Issue #6, opened 2016-07-12
Compute Acceptance Test Issues


jtopjian:
I tried running the `TestListFlavor` acceptance test and had to fix a lot of errors in the `acceptance/openstack/compute/v2` directory to just get the test to run. The test passed, but incorrectly -- no flavors were printed.

Here's the diff of what needed to be changed just to get the test to run:

https://github.com/jtopjian/gophercloud/commit/74cd982df7d38472c66955b8061165763e332c34

It looks like there have been a good amount of changes that have happened to the code base - are these the types of things that should be documented in the MIGRATING doc?

I'd be happy to submit PRs to fix the tests, but having no flavors printed is throwing me off. It looks like there might be some auth issues and I have no doubt that someone else can figure that one out quicker than me.

Nice work on this new repo, by the way!

jrperritt:
>> I tried running the TestListFlavor acceptance test and had to fix a lot of errors

You're right; the acceptance tests haven't been touched in a while. See my comment [here](https://github.com/rackspace/gophercloud/issues/592#issuecomment-231936468).

>> It looks like there have been a good amount of changes that have happened to the code base - are these the types of things that should be documented in the MIGRATING doc?

Yes

>> I'd be happy to submit PRs to fix the tests

Great. Please read the [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md) (which I'm continuing to add to) first.

jtopjian:
Sounds good.

If it helps, I have [this stuff](https://github.com/jtopjian/terraform-devstack) which I have been using to run the Terraform acceptance tests. I'm hoping to keep the number of configurations down to two:

* standard: current api versions (currently sorting out issues with lbaas v2 and keystone v3)
* legacy: lbaas v1 and keystone v2

It doesn't look like there is an official way of restarting devstack-based services any more, which means it's not possible to snapshot these configurations into an image and launch on-demand... I usually build them once every week or two and manually fire off the Terraform acceptances whenever I review a PR.

jrperritt:
closed by #8

Issue #7, opened 2016-07-13
compute/v2/flavors: error parsing empty swap


jtopjian:
A flavor being returned by OpenStack looks like:

```json
    {
      "OS-FLV-DISABLED:disabled": false,
      "OS-FLV-EXT-DATA:ephemeral": 0,
      "disk": 160,
      "id": "5",
      "links": [
        {
          "href": "http://10.1.11.212:8774/v2.1/80839b1b20aa41a3a4bf32718bae67cb/flavors/5",
          "rel": "self"
        },
        {
          "href": "http://10.1.11.212:8774/80839b1b20aa41a3a4bf32718bae67cb/flavors/5",
          "rel": "bookmark"
        }
      ],
      "name": "m1.xlarge",
      "os-flavor-access:is_public": true,
      "ram": 16384,
      "rxtx_factor": 1,
      "swap": "",
      "vcpus": 8
    },
```

When no swap is set, the returned value is `""` which triggers the error:

```
Unable to retrieve flavors: json: cannot unmarshal string into Go value of type int
```

`Swap` is defined in Gophercloud as an [int](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/flavors/results.go#L44), as it should, because Nova defines it as an [int](https://github.com/openstack/nova/blob/e8cdef91f82f3a46d6f195d96cbd450934119022/nova/compute/flavors.py#L52-L63) as well. 

Was [mapstructure](https://github.com/rackspace/gophercloud/blob/master/openstack/compute/v2/flavors/results.go) doing more lenient casting of values?

jrperritt:
Just another misguided OpenStack design...

>> Was mapstructure doing more lenient casting of values?

Optionally, yes. There was a `WeakDecode` method that would handle some conversions like `int` -> `string` for you.

I guess the way to do this is to define an `UnmarshalJSON` method on `Flavor` (so that it implements `json.Unmarshaler` and handle (probably two-step it) it there.

Hopefully this won't be a recurring theme.

jrperritt:
Closed by #12

PR #8, opened 2016-07-13
Compute Acceptance Test Fixes

jtopjian:
This PR enables individual compute acceptance tests to be run. Only a few acceptance tests are passing at this time, however, these changes are required for Go to validate the `acceptance/openstack/compute/v2` package.

For #6


jtopjian:
@jrperritt No doubt I'm jumping the gun on getting fixes/changes into this new repo, so please feel free to nit (or nix) all of this.

I began going through the compute acceptance tests and thought about compiling them all into one large PR, but it looks like that would be one _very_ large PR. In addition, some of the test failures might require changes to the corresponding packages (see #7). So I thought I'd start with this which at least allows individual tests to be run.

jtopjian:
@jrperritt Curiosity is getting the best of me and I've started going through each of the Compute acceptance tests to see what fixes they need or how I can practice cleaning them up. I have each of the above tests also in separate branches if that would be easier, but some require a few common fixes. I can look at making each branch completely standalone, though.

acceptance/openstack/compute/v2/servers_test.go
```diff
@@ -11,170 +10,91 @@ import (
...
-
-	return server, err
-}
+	if err = waitForStatus(client, server, "ACTIVE"); err != nil {
+		t.Fatalf("Unable to wait for server: %v", err)
+	}
+	defer func() {
+		servers.Delete(client, server.ID)
+		t.Logf("Server deleted.")
+	}()
```
jtopjian: I've moved the delete defer to be triggered after the instance has been confirmed ACTIVE. This is so if instance creation fails, the error'd instance still remains so the tester can investigate why the instance failed to launch. When defer was in its original location, the instance would immediately be deleted, the UUID never printed (though this can be fixed), and troubleshooting was much harder.

acceptance/openstack/compute/v2/servers_test.go
```diff
@@ -11,170 +10,91 @@ import (
...
+	}
 }
 
-func createServer(t *testing.T, client *gophercloud.ServiceClient, choices *ComputeChoices) (*servers.Server, error) {
-	if testing.Short() {
-		t.Skip("Skipping test that requires server creation in short mode.")
+func TestServersCreateDestroy(t *testing.T) {
+	client, err := newClient()
+	if err != nil {
+		t.Fatalf("Unable to create a compute client: %v", err)
```
jtopjian: I see occurrences of both `%v` and `%s` for printing errors. What's the best format in this case?

acceptance/openstack/compute/v2/servers_test.go
```diff
@@ -276,24 +261,23 @@ func TestActionReboot(t *testing.T) {
...
+		t.Logf("Server deleted.")
+	}()
 
 	rebootOpts := &servers.RebootOpts{
 		Type: servers.SoftReboot,
 	}
-	res := servers.Reboot(client, server.ID, rebootOpts)
-	if res.Err == nil {
-		t.Fatal("Expected the SDK to provide an ArgumentError here")
-	}
```
jtopjian: I wasn't sure why this was in the test. Rather than commenting it out, I figured I would yank it out, comment here, and add it back if necessary. :)

acceptance/openstack/compute/v2/servers_test.go
jrperritt: Short answer: for errors, I think the output is the same for both, so I'm indifferent.

Longer answer: in general, `%s` is for formatting strings, so using it with an `error` will call the error's `Error() string` method (the only method required to satisfy the `error` interface). `%v` (notably `%+v`) is useful for collection types (`map`, `struct`), where the '+' also prints the field names (if any).

jrperritt:
As far as I'm concerned, you can't make the acceptance tests any worse than they were before this PR. Because of that, As long as you're only touching the acceptance tests (or docs), I'm not concerned about reviewing this (and therefore not concerned about the size of it). Once you're happy with it (and it builds), let me know and I'll merge it.

acceptance/openstack/compute/v2/servers_test.go
jtopjian: :+1: 

Sorry for commenting and then clobbering the commits.

I've been going with `%v` to support the case where the error would contain extra info that `%s` would possibly strip. I figure if it's an acceptance test, the tester will want to see all information possible.

jtopjian:
> As far as I'm concerned, you can't make the acceptance tests any worse than they were before this PR. 

I'm gonna take that as a compliment :D

Sounds good. This has been pretty fun and a good way for me to determine what needs to be (or can be, for the better) changed in Terraform.

jrperritt:
>> I'm gonna take that as a compliment :D

It was, and a slight to myself.

jtopjian:
@jrperritt I think this is ready to go!

There are probably a few nits here and there and I wasn't able to successfully run the `defsecrules` tests due to that extension requiring a custom environment, but other than that, everything runs great!

script/acceptancetest_environments/keystonev2-lbaasv1.sh
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
```
jrperritt: Did you mean for this to be committed, and, if so, can you comment on what it does? Something like `# Installs keylogger` would suffice.

jrperritt:
Can you merge or rebase with `master`? After that and that ^, this looks good to merge.

script/acceptancetest_environments/keystonev2-lbaasv1.sh
jtopjian: Yup, meant to commit it, but oops on not having any useful comments.

jtopjian:
@jrperritt Done and done.

jrperritt:
Thank you for all the work on this. +2

jtopjian:
> Thank you for all the work on this.

No problem at all. I'll probably go through a few of the others that I'm familiar with (block storage, networking).

PR #13, opened 2016-07-20
Fix v3 auth with project scope

carolynvs:
When I use the AuthenticatedClient to authenticate using identity v3 and a project scope is specified (i.e. TenantName), I receive the following error

```go
auth := gophercloud.AuthOptions{
		IdentityEndpoint: "http://example.com/v3",
		Username:  "carolyn",
		Password: "secret",
		TenantName: "admin",
		DomainName: "default",
	}
	identityService, authErr := openstack.AuthenticatedClient(auth)
```

```
The base Identity V3 API does not accept authentication by TenantName
```

I have updated the client to use the modified auth options, so that the Tenant ID/Name is cleared out to pass validation. I've also updated the v3 auth test so that without this fix, the test fails with the error above.

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -174,16 +174,16 @@ func v3auth(client *gophercloud.ProviderClient, endpoint string, options gopherc
 	}
 
 	v3Opts := tokens3.AuthOptions{
-		IdentityEndpoint: options.IdentityEndpoint,
-		Username:         options.Username,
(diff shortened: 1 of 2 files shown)


carolynvs:
@jrperritt This one is ready for review too.

jrperritt:
Ah, very good. +2

PR #14, opened 2016-07-20
Fix example commands to setup a new fork

carolynvs:
I've updated the contributing guide with the commands I used to setup a new fork.


jrperritt:
+2

Issue #22, opened 2016-07-26
Unable to parse Block Storage v1 Volume.CreatedAt


jtopjian:
When trying to create a volume, the following error is returned:

```
parsing time "2016-07-26T01:21:26.911126" as "2006-01-02T15:04:05.999999Z": cannot parse "" as "Z"
```

It looks like the [CreatedAt](https://github.com/gophercloud/gophercloud/blob/9b7b9e6a417dcbe0495f02a1eb381f3bcd193230/openstack/blockstorage/v1/volumes/results.go#L21) type needs changed to `gophercloud.JSONRFC3339MilliNoZ`.

jrperritt:
closed by #23

PR #23, opened 2016-07-26
Change Block Storage v1 Volume.CreatedAt Type

jtopjian:
This commit changes the Volume.CreatedAt type from
gophercloud.JSONRFC3339Milli to gophercloud.JSONRFC3339MilliNoZ.

For #22

### DIFF
--- a/openstack/blockstorage/v1/volumes/results.go
+++ b/openstack/blockstorage/v1/volumes/results.go
@@ -18,7 +18,7 @@ type Volume struct {
 	// Indicates whether this is a bootable volume.
 	Bootable string `json:"bootable"`
 	// The date when this volume was created.
-	CreatedAt gophercloud.JSONRFC3339Milli `json:"created_at"`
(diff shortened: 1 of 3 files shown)


jtopjian:
Although the API samples in the Cinder repository show timestamps with Z, as best as I can tell, timestamps are really without Z.

From looking at the unit tests [here](https://github.com/openstack/cinder/blob/6cfecade9e2ee86bbb7d95c3c401c9e4c70f6a96/cinder/tests/unit/api/v1/test_volumes.py), a small Python snippet can be run that shows a timestamp without Z:

```python
import datetime
import iso8601

print datetime.datetime(1900, 1, 1, 1, 1, 1,tzinfo=iso8601.iso8601.Utc())
```

1900-01-01 01:01:01+00:00

jrperritt:
I can confirm this. I hit this while migrating `rackspace/gophercloud` over here. I thought I'd fixed it, but it turns out I fixed it in a vendored repository. +2

jtopjian:
Thanks!

Issue #32, opened 2016-08-03
Acceptance Test Ideas


jtopjian:
I started working on the Block Storage acceptance tests and wanted to reuse some of the work done in the Compute acceptance tests. This ranged from something as simple as initializing a client to creating a server to attach a volume.

Would it be worthwhile to make further changes to the tests (and test file structure) to do this?

jrperritt:
closed by #33

PR #33, opened 2016-08-03
Compute Acceptance Test Changes

jtopjian:
This commit makes more changes to the compute acceptance tests:

* Makes all reusable functions exportable so other APIs can use them.
* Centralizes client initialization and environment variable checks.

For #32


jtopjian:
This first run was just rearranging everything. I can also add some godoc comments to the functions in `compute.go`. An idea I had was to have a file similar to `compute.go` in each of the API tests that could be used as references, samples, and examples of Gophercloud.

jtopjian:
@jrperritt I used this same structure for the Block Storage acceptance tests and worked very well. In addition, being able to call the `CreateServer` function from Compute made writing tests for volume attachments easy.

Any thoughts on this? I'm happy to restructure or rename anything. What do you think of the general idea of sharing test components between OpenStack components, though?

jrperritt:
> Any thoughts on this?

I like it.

> What do you think of the general idea of sharing test components between OpenStack components, though?

For acceptance tests, approved and encouraged.

jtopjian:
:tada: 

Let me know if you'd like me to include godoc comments or would prefer to not have them. I can definitely add them, but don't want to write them all out if you don't want them here. :)

If not, this is ready to go from my end.

jrperritt:
I'd like godoc to be the official (and only) source of documentation, so any and all comments are welcome and encouraged.

jtopjian:
@jrperritt I added some godoc comments - let me know if they need work. I ended up amending the commit since I re-ordered the functions to be alphabetical and figured it would just cause a bunch of diff noise.

jrperritt:
Thank you for all the work on this. +2

Issue #10, opened 2016-07-18
New Branch to Support Magnum Bays


carolynvs:
The magnum API isn't solidified yet. I will have access to an environment this week and will update the ticket with additional details on each action.
- POST /bays - create a cluster
- GET /bays/[id] - get cluster info
- PUT /bays/[id] - update a cluster, currently that only means increasing the node count
- DELETE /bays - delete a cluster

carolynvs:
@jrperritt Would you please assign this to me? 😄

jrperritt:
>> The magnum API isn't solidified yet.

To be clear, we're waiting until the API is finalized to add support for this, correct?

>> Would you please assign this to me?

I think only members of the org (and maybe collaborators?) can be assigned issues. If it's important for accounting purposes, I can try to figure out how to get you added. Otherwise, it's known the issue is yours.

jrperritt:
Because the Magnum API isn't finalized yet, I think the best we can do here is create a `magnum` branch that can hold the code for the unfinalized API. [If and] when the API becomes solidified, we can review and merge into `master`.

jrperritt:
Branch is here: https://github.com/gophercloud/gophercloud/tree/magnum

carolynvs:
Thanks!

jrperritt:
Changed issue title to be more practical for now. If and when the API stabilizes we can look at supporting this.

PR #38, opened 2016-08-09
Support deleting Magnum bays

carolynvs:
For #10.

### DIFF
--- a/openstack/containerorchestration/v1/bays/requests.go
+++ b/openstack/containerorchestration/v1/bays/requests.go
@@ -88,7 +88,7 @@ func Create(c *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResul
 
 // Delete accepts a unique ID and deletes the bay associated with it.
 func Delete(c *gophercloud.ServiceClient, bayID string) (r DeleteResult) {
-	opts := &gophercloud.RequestOpts{ErrorContext: common.ErrDeleteFailed{}}
-	_, r.Err = c.Delete(deleteURL(c, bayID), opts)
+	ro := &gophercloud.RequestOpts{ErrorContext: &common.ErrorResponse{}}
+	_, r.Err = c.Delete(deleteURL(c, bayID), ro)
 	return
 }
--- a/openstack/containerorchestration/v1/bays/requests.go
+++ b/openstack/containerorchestration/v1/bays/requests.go
@@ -1,7 +1,6 @@
 package bays
 
 import (
-	"encoding/json"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/containerorchestration/v1/common"
 	"github.com/gophercloud/gophercloud/pagination"
@@ -87,26 +86,9 @@ func Create(c *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResul
 	return
 }
 
-type ErrDeleteFailed struct { }
-
-// Error400 extracts the actual error message from the body of the response
-func (d ErrDeleteFailed) Error400(r gophercloud.ErrUnexpectedResponseCode) error {
-	var s *common.ErrorResponse
-	err := json.Unmarshal(r.Body, &s)
-	if err != nil {
-		return gophercloud.ErrDefault400{ErrUnexpectedResponseCode: r}
-	}
-
-	return s
-}
-
-func (d ErrDeleteFailed) Error() string {
-	return "Unable to delete bay"
-}
-
 // Delete accepts a unique ID and deletes the bay associated with it.
 func Delete(c *gophercloud.ServiceClient, bayID string) (r DeleteResult) {
-	opts := &gophercloud.RequestOpts{ErrorContext: ErrDeleteFailed{}}
+	opts := &gophercloud.RequestOpts{ErrorContext: common.ErrDeleteFailed{}}
 	_, r.Err = c.Delete(deleteURL(c, bayID), opts)
 	return
 }
--- a/openstack/containerorchestration/v1/bays/requests.go
+++ b/openstack/containerorchestration/v1/bays/requests.go
@@ -48,7 +48,8 @@ func List(c *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pager {
 
 // Get retrieves a specific bay based on its unique ID.
 func Get(c *gophercloud.ServiceClient, id string) (r GetResult) {
-	_, r.Err = c.Get(getURL(c, id), &r.Body, nil)
+	ro := &gophercloud.RequestOpts{ErrorContext: &common.ErrorResponse{}}
+	_, r.Err = c.Get(getURL(c, id), &r.Body, ro)
 	return
 }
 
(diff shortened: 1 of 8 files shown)


carolynvs:
@jrperritt This is now ready for review.

openstack/containerorchestration/v1/bays/requests.go
```diff
@@ -84,3 +86,27 @@ func Create(c *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResul
...
+		return gophercloud.ErrDefault400{ErrUnexpectedResponseCode: r}
+	}
+
+	return s
+}
+
+func (d ErrDeleteFailed) Error() string {
+	return "Unable to delete bay"
+}
+
```
jrperritt: Lets move L89-L106 to `errors.go`.

jrperritt:
Ping me after that ^ one, and I'll merge.

carolynvs:
@jrperritt Alrighty! I've moved those lines into `errors.go`.

carolynvs:
@jrperritt Oops, actually let's hold off I just found one more response code (409) that needs special handling. My kingdom for a documented API! 😄

carolynvs:
@jrperritt Okay I've updated my Create, Get and Delete methods to use a custom error handler to extract the real error message from the body. If it can't for any reason, it falls back to the default error message for the status code.

carolynvs:
@jrperritt Sorry for the trickle of changes! This is ready to review.

openstack/containerorchestration/v1/bays/results.go
```diff
@@ -86,3 +86,18 @@ func ExtractBays(r pagination.Page) ([]Bay, error) {
 	err := (r.(BayPage)).ExtractInto(&s)
 	return s.Bays, err
 }
+
+// DeleteHeader represents the headers returned in the response from a Delete request.
+type DeleteHeader struct {
```
jrperritt: test for this?

openstack/containerorchestration/v1/common/errors.go
```diff
@@ -0,0 +1,118 @@
...
+	e.ErrUnexpectedResponseCode = r
+	err := json.Unmarshal(r.Body, &e)
+	return err == nil
+}
+
+// Error400 extracts the actual error message from the body of the response
+func (e *ErrorResponse) Error400(r gophercloud.ErrUnexpectedResponseCode) error {
+	if e.unwrapError(r) {
+		return e
+	} else {
```
jrperritt: Because of the `return` in the `if` clause, we can get rid of all these `else`s.

openstack/containerorchestration/v1/common/errors.go
```diff
@@ -0,0 +1,118 @@
...
+// Error returns the error message details from the server
+func (e *ErrorResponse) Error() string {
+	errors := []string{}
+	for _, e := range e.Errors {
+		errors = append(errors, e.Detail)
+	}
+	return strings.Join(errors, "\n")
+}
+
+func (e *ErrorResponse) unwrapError(r gophercloud.ErrUnexpectedResponseCode) bool {
```
jrperritt: Let's return an `error` and check for `nil` in the calling function:
```go
func (e *ErrorResponse) unwrapError(r gophercloud.ErrUnexpectedResponseCode) error {
	e.ErrUnexpectedResponseCode = r
	return json.Unmarshal(r.Body, &e)
}

if e.unwrapError(r) == nil {
...
}
```

jrperritt: A bit more explanation of reasoning for that ^: when returning `bool`, we'd like to name the function something that suggests a binary response (e.g. `canUnwrapError`), which we could do here, but in addition to checking if the error can be unwrapped, it also unwraps it, which wouldn't be obvious by the function name `canUnwrapError`.

Open for discussion.

carolynvs: I wasn't returning the error because the error handling doesn't rely upon the content of the error at all. Basically I was following the cast pattern, which returns the cast value and a boolean representing if the cast was successful.

How about naming it `tryUnwrapError` and the calling it like this: `if e.tryUnwrapError(r) {...}`

openstack/containerorchestration/v1/bays/results.go
carolynvs: Oops! That's used and should be removed.

carolynvs:
@jrperritt  I have made 2 out of 3 changes. Let me know if you are okay with my suggestion for `unwrapError`.

openstack/containerorchestration/v1/common/errors.go
jrperritt: I've been considering this off and on, and I have a suggestion. First, though, a comment that this whole thing reminds me needs to go in the README:
The way errors are handled currently is more of an experiment than a certain library feature; basically, I'm not sure how I feel about the design of it. It exists for 2 reasons:
1. When we get an authentication error, we'd like to be able to retry the operation (or run a user-defined function).
2. To allow overriding the error message returned by OpenStack, since many times the error messages returned are vague and/or unhelpful
The reason I mention that is because I'm not sure we gain anything by implementing all the `Error***` interfaces here.

Now the suggestion: it looks like the error logic here is to unmarshal the `error` into an `Error` if possible, otherwise return original error. Because of that, could we just create an `ExtractErr` method on `DeleteResult` with that logic? Something like the following (untested):

```go
func (h *DeleteResult) ExtractErr() error {
  if h.Err != nil {
    if e, ok := unwrapError(h.Err); ok {
      return e
    }
    return h.Err
  }
  return nil
}
``

Comment (carolynvs) [inline openstack/containerorchestration/v1/common/errors.go]: That would need to be for every result type because Magnum returns that same error structure for all failed operations (create/delete/update/get). I can go back to returning the generic http error and providing a function to unmashall the real error. This isn't my library so I'll do the error handling however you need it. Just let me know what you'd like to see and I'll update the PR. 😄

Comment (carolynvs) [inline openstack/containerorchestration/v1/common/errors.go]: That would need to be for every result type because Magnum returns that same error structure for all failed operations (create/delete/update/get). I can go back to returning the generic http error and providing a function to unmashall the real error. This isn't my library so I'll do the error handling however you need it. Just let me know what you'd like to see and I'll update the PR. 😄

Comment (carolynvs) [inline openstack/containerorchestration/v1/common/errors.go]: I'll update the PR with your suggestion but for every result type, not just Delete so you can look it over.

Comment (jrperritt) [inline openstack/containerorchestration/v1/common/errors.go]: > That would need to be for every result type because Magnum returns that same error structure for all failed operations (create/delete/update/get).

Ah, OK. That would be tedious.

> I'll update the PR with your suggestion but for every result type, not just Delete so you can look it over.

No, that's OK. Instead, what about having that `ExtractErr` method on something like `ErrResult` and having all the results embed that type?

```go
type DeleteResult struct {
  ErrResult
}

type ErrResult struct {
  gophercloud.ErrResult
}

func (h *ErrResult) ExtractErr() error {
  if h.Err != nil {
    if e, ok := unwrapError(h.Err); ok {
      return e
    }
    return h.Err
  }
  return nil
}
```

carolynvs: Sure thing, thanks! I'll get that updated today.

jrperritt:
Upon review, my suggestion for handling the embedded error may be unfeasible at the moment. Let's just leave this as-is and fix later if need-be. +2

Issue #39, opened 2016-08-10
identity/v2: Error when adding role to user


jtopjian:
When adding a role to a user, the following error is seen:

```
Unable to grant role to user: Expected HTTP response code [201 202] when accessing [PUT http://10.1.7.61:35357/v2.0/tenants/20e0168198504ef6bb77844b23df18aa/users/11d01713fd874ae4bf20
55b2a8023147/roles/OS-KSADM/017257e31a1e421b8b533afc27882521], but got 200 instead
```

AFAICT, which isn't saying much, unless an explicit status is given, and the request is successful, a 200 response is returned:

https://github.com/openstack/keystone/blob/94e83aff172feee3874604ab1a92d4038be4965f/keystone/common/wsgi.py#L732-L733

And I don't see an explicit status set here: 

https://github.com/openstack/keystone/blob/94e83aff172feee3874604ab1a92d4038be4965f/keystone/assignment/controllers.py#L150

jrperritt:
closed by #40

PR #40, opened 2016-08-10
Allow return code 200 when adding role to user

jtopjian:
For #39


jtopjian:
@jrperritt ping for review.

If I was more confident in reading the Keystone code, I would have only used 200 and not 200/201.

jrperritt:
Yeah, I think that's fine. +2

PR #44, opened 2016-08-10
Add openstack/auth_yaml.go

raphink:
This aims to fix https://github.com/rackspace/gophercloud/issues/487.

/cc @emonty @mcanevet

for #104

### DIFF
--- a/openstack/auth_yaml.go
+++ b/openstack/auth_yaml.go
@@ -1,0 +1,131 @@
+package openstack
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+
+	"github.com/gophercloud/gophercloud"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	ErrNoCloudYaml = fmt.Errorf("Clouds.yaml file not found.")
+)
+
+type Clouds struct {
+	Clouds map[string]struct {
+		Profile       string   `yaml:"profile"`
+		RegionName    string   `yaml:"region_name"`
+		Regions       []string `yaml:"regions"`
+		DNSAPIVersion string   `yaml:"dns_api_version"`
+		Auth          struct {
+			AuthURL     string `yaml:"auth_url"`
+			Username    string `yaml:"username"`
+			Password    string `yaml:"password"`
+			ProjectName string `yaml:"project_name"`
+			ProjectID   string `yaml:"project_id"`
+		} `yaml:"auth,omitempty"`
+	}
+}
+
+// AuthOptionsFromYaml fills out an identity.AuthOptions structure with the settings found in a specific
+// cloud entry of a clouds.yaml file (as per http://docs.openstack.org/developer/os-client-config)
+// If `env` is set to true, it will also override these values with the various
+// OS_* environment variables.  The following variables provide sources of truth: OS_AUTH_URL, OS_USERNAME,
+// OS_PASSWORD, OS_TENANT_ID, and OS_TENANT_NAME.  Of these, OS_USERNAME, OS_PASSWORD, and OS_AUTH_URL must
+// have settings, or an error will result.  OS_TENANT_ID and OS_TENANT_NAME are optional.
+func AuthOptionsFromYaml(cloud string, env bool) (gophercloud.AuthOptions, error) {
+	clouds := &Clouds{}
+	cloudsContent, err := cloudsYaml()
+	if err != nil {
+		return nilOptions, err
+	}
+	err = yaml.Unmarshal(cloudsContent, clouds)
+	if err != nil {
+		return nilOptions, fmt.Errorf("Failed to unmarshal YAML: %v", err)
+	}
+
+	auth := clouds.Clouds[cloud].Auth
+
+	authURL := auth.AuthURL
+	username := auth.Username
+	password := auth.Password
+	tenantID := auth.ProjectID
+	tenantName := auth.ProjectName
+
+	if env {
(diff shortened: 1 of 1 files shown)


raphink:
@jtopjian @jrperritt is that good for you?

jtopjian:
Hi @raphink! I think just as a formality, there should be a corresponding issue opened for this. Check the [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md) for more information.

As for code review, I'll leave that one for @jrperritt. :)

raphink:
Hi @jtopjian . There was an issue on the rackspace project: https://github.com/rackspace/gophercloud/issues/487

Should this issue be copied to the gophercloud project?

raphink:
There https://github.com/gophercloud/gophercloud/issues/104

openstack/auth_yaml.go
```diff
@@ -0,0 +1,131 @@
...
+		Username:         username,
+		Password:         password,
+		TenantID:         tenantID,
+		TenantName:       tenantName,
+	}
+
+	return ao, nil
+}
+
+func cloudsYaml() ([]byte, error) {
```
jrperritt: I understand that `os-client-config` has logic that looks in all these different places and decides what to do for the user, but I think we should just have the user tell us the file location:
```go
func AuthOptionsFromYaml(yamlPath, cloud string) (gophercloud.AuthOptions, error) {
...
}
```

openstack/auth_yaml.go
```diff
@@ -0,0 +1,131 @@
...
+
+	auth := clouds.Clouds[cloud].Auth
+
+	authURL := auth.AuthURL
+	username := auth.Username
+	password := auth.Password
+	tenantID := auth.ProjectID
+	tenantName := auth.ProjectName
+
+	if env {
```
raphink: This provides overrides through the environment, which is the default behavior usually. If this isn't done here, it makes it much harder to implement, as you'll have to call both `AuthOptionsFromYaml` and `AuthOptionsFromEnv` and merge the keys one by one.

openstack/auth_yaml.go
raphink: This means every client will need to reimplement the default locations for this file, even though they are listed in the spec.

jrperritt: Presumably a user has only one config file. It seems like letting them explicitly specify the location and name of the file is what we'd want.

openstack/auth_yaml.go
jrperritt: I don't see this as a common situation. Moreover, a user could just:
```go
ao, err := openstack.AuthOptionsFromYaml("config.yaml", "myCloud")
ao.Username = os.Getenv("OS_USERNAME")
```

raphink:
OK, I guess I'll move the getenv + file loop logic to terraform then...

openstack/auth_yaml.go
raphink: Another reason why I'd done this is the checks, e.g.:

```go
	if authURL == "" {
		err := gophercloud.ErrMissingInput{Argument: "authURL"}
		return nilOptions, err
	}
```

It's OK to not have variables set in the YAML files, so long as they are overridden in the environment. If I remove the environment option, then I'll have to remove the checks, too.

openstack/auth_yaml.go
raphink: So you'd want another parameter to the method for specifying the file to parse.

mcanevet:
@jtopjian @jrperritt is there something preventing this PR to be merged?

jtopjian:
I haven't been following this one, but it looks like the hold is due to more discussion being needed on some parts.

From reading the comments, I can see both sides. I'm hesitant to give an opinion because I have no clue how to design software correctly, but:

Gophercloud isn't really a full-fledged client. It's a core library. The packages for communicating with the various APIs are thin. There's some convenience here and there (detecting environment variables for auth, the various `IDFromName` functions), but it's mostly the bare minimum needed to communicate with an API. I've grown to appreciate this very much.

IMO, this feature feels slightly more convenient than core, so I can understand the pushback in some areas. For example, instead of checking a bunch of different directories for the config file, it's much simpler to specify the file path. Then Gophercloud never has to worry about updating default paths. 

Yes, that means that everything that consumes Gophercloud will have to re-implement those same types of checks. For now. I suspect this is why https://github.com/openstack/os-client-config is actually a separate project rather than part of either OSC or Keystone, right?

Maybe this function is better suited for something discussed [here](https://github.com/gophercloud/gophercloud/issues/237) than in Gophercloud directly? In the end, you'd still end up passing values to the core AuthOptions. And tools like Terraform could implement the "utility" package rather than rewriting the implementation.

I could be totally wrong, though.

jtopjian:
Hi all,

A new [utils](https://github.com/gophercloud/utils) repo was created a few weeks ago. I think this feature might be better suited there. Creating a draft of this in the utils repo is on my to-do list, which I will _eventually_ get to, but I wanted to leave a note here in case someone else wanted to submit it first. I would be happy to review.

jtopjian:
Hi all,

https://github.com/gophercloud/utils/pull/2 has just been merged which adds initial support for `os-client-config`-like functionality, including `clouds.yaml`. I don't have an in-depth amount of knowledge of `os-client-config`, so I anticipate it requiring some follow-up patches. Please don't hesitate to patch or submit an issue.

I'm going to go ahead and close this.

Issue #45, opened 2016-08-11
Committed vim swp file


jtopjian:
Looks like I accidentally committed a `.swp` file in #43:

https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/blockstorage/v1/.blockstorage.go.swp

jrperritt:
closed by #46

PR #46, opened 2016-08-11
vim swp file

jtopjian:
This commit removes an swp file that snuck in and adds a
.gitignore file to prevent further ones.

For #45


jtopjian:
@jrperritt ping for review. and sorry about that :/

jrperritt:
No worries.

Issue #48, opened 2016-08-13
Identity Acceptance Tests


jtopjian:
Continuing along with acceptance tests, this issue is for Identity acceptance tests.

jrperritt:
closed by #49

PR #49, opened 2016-08-13
Identity Acceptance Test Cleanup

jtopjian:
For #48 

The v2 and v3 tests are mutually exclusive. Each have to be run on separate environments. I've tested both as such and both pass.


jtopjian:
@jrperritt ping for review

jrperritt:
+2

PR #60, opened 2016-08-16
Suport creating Magnum certificates

carolynvs:
For #10.

- [x] Import an existing private key into a bay
- [x] Generate a new private key and load it into a bay
- [x] Unit Tests
- [x] Acceptance Tests
- [x] Rebase after #38 is merged

PR #64, opened 2016-08-18
Generate shell scripts in the Magnum credentials bundle

carolynvs:
For #10.

In order for the credentials bundle to be useful, not only does it need to include TLS certificates but shell scripts to correctly configure docker or kubectl to use them, e.g.  `docker.env`.

Here's how someone would use them:

1. Create a bay
2. Wait for it to become active
3. Generate the credentials bundle and save it to disk
4. Source the appropriate shell script based on the type of cluster (kubernetes/swarm) and the user shell (powershell/bash/cmd/fish). For example `source docker.env` or `. kubectl.ps1`.
5. Use the appropriate client (docker/kubectl) to interact with the bay.

- [x] Generate swarm compatible scripts for swarm clusters
- [x] Generate kubernetes compatible scripts kubernetes clusters
- [x] Unit Tests
- [x] Rebase when #60 is merged.


carolynvs:
@jrperritt This is now ready for review.

Issue #34, opened 2016-08-06
Keystoneless Client


jtopjian:
If it's not already possible, what's the general outline for making a client that bypasses Keystone? In particular, this would be for Swift clients that authenticate directly with Swift. It would resemble this: 

https://github.com/ncw/swift/blob/master/auth.go#L85-L94

Is this something that would make sense to have in Gophercloud or is there a requirement to have all clients go through Keystone?

jrperritt:
`gophercloud.ProviderClient`'s `Request` method in pretty generic. It'll make any request based on the URL and headers you give it. Would this be something that calls `ProviderClient.Request` with pre-populated fields?

Can you provide a code example (Go or pseudo) of what you have in mind and how you'd use it? (I looked at the link, but still don't know in what context it'd be used).

jtopjian:
> Would this be something that calls ProviderClient.Request with pre-populated fields?

After writing the below out, I came to the realization of "yes" :smile: 

The specific goal that I'm trying to accomplish is using Gophercloud in Swift environments that use Swift's native (non-Keystone) authentication, which is done by passing the headers found in the `ncw/swift` link above.

I _think_ this is along the right lines(?):

```golang
providerClient, _ := openstack.NewClient("http://example.com:8080/v1")
swiftClient, _ := openstack.NewObjectStorageV1(providerClient, gophercloud.EndpointOpts{})
```

And at this point, I should have a basic `gophercloud.ServiceClient` that never hit Keystone... but I still need to add those headers, which I think I can do [here](https://github.com/gophercloud/gophercloud/blob/master/provider_client.go#L84), but I'm not sure how to do that. I think I'd have to write my own `openstack.NewClient` in order to create the `ProviderClient` the way I want. 

Or would I create a new type of ProviderClient that has its own [AuthenticatedHeaders](https://github.com/gophercloud/gophercloud/blob/master/provider_client.go#L76) function that sets the `ApiKey` and `UserName`?

jtopjian:
Oh, also: so it looks like this can be done because how flexible you've implemented all of the pieces (along the lines of the Debug question I had the other week, which the link you gave me might also be the example implementation I'm looking for... will look into that now).

Let's say I roll this into something like `openstack.NewSwiftAuthClient()`, would that be something you'd be interested in merging or would that stay as an app-specific implementation?

jrperritt:
[...]

PR #65, opened 2016-08-20
Swauth Support

jtopjian:
Swauth is an internal authentication system for Swift. This commit adds
support for Swauth in a way that still enables the resulting
*gophercloud.ServiceClient to still be able to use the existing ObjectStorage
API calls.

For #34

### DIFF
--- a/openstack/objectstorage/v1/swauth/requests.go
+++ b/openstack/objectstorage/v1/swauth/requests.go
@@ -21,8 +21,8 @@ func (opts AuthOpts) ToAuthOptsMap() (map[string]string, error) {
 	return gophercloud.BuildHeaders(opts)
 }
 
-// Authenticate performs an authentication request for a Swauth-based user.
-func Authenticate(c *gophercloud.ProviderClient, opts AuthOptsBuilder) (r GetAuthResult) {
+// Auth performs an authentication request for a Swauth-based user.
+func Auth(c *gophercloud.ProviderClient, opts AuthOptsBuilder) (r GetAuthResult) {
 	h := make(map[string]string)
 
 	if opts != nil {
@@ -54,7 +54,7 @@ func Authenticate(c *gophercloud.ProviderClient, opts AuthOptsBuilder) (r GetAut
 // NewObjectStorageV1 creates a Swauth-authenticated *gophercloud.ServiceClient
 // client that can issue ObjectStorage-based API calls.
 func NewObjectStorageV1(pc *gophercloud.ProviderClient, authOpts AuthOpts) (*gophercloud.ServiceClient, error) {
-	auth, err := Authenticate(pc, authOpts).Extract()
+	auth, err := Auth(pc, authOpts).Extract()
 	if err != nil {
 		return nil, err
 	}
--- a/openstack/objectstorage/v1/swauth/requests.go
+++ b/openstack/objectstorage/v1/swauth/requests.go
@@ -1,0 +1,70 @@
+package swauth
+
+import "github.com/gophercloud/gophercloud"
+
+// AuthOptsBuilder describes struct types that can be accepted by the Auth call.
+// The AuthOpts struct in this package does.
+type AuthOptsBuilder interface {
+	ToAuthOptsMap() (map[string]string, error)
+}
+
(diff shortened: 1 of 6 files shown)


jtopjian:
@jrperritt ping for review

openstack/objectstorage/v1/swauth/requests.go
```diff
@@ -0,0 +1,70 @@
...
+import "github.com/gophercloud/gophercloud"
+
+// AuthOptsBuilder describes struct types that can be accepted by the Auth call.
+// The AuthOpts struct in this package does.
+type AuthOptsBuilder interface {
+	ToAuthOptsMap() (map[string]string, error)
+}
+
+// AuthOpts specifies an authentication request.
+type AuthOpts struct {
```
jrperritt: Are either and/or both these fields required? If so, we can add 
a `required: "true"` struct tag to each of them.

openstack/objectstorage/v1/swauth/requests.go
```diff
@@ -0,0 +1,70 @@
...
+	Key string `h:"X-Auth-Key"`
+}
+
+// ToAuthOptsMap formats an AuthOpts structure into a request body.
+func (opts AuthOpts) ToAuthOptsMap() (map[string]string, error) {
+	return gophercloud.BuildHeaders(opts)
+}
+
+// Authenticate performs an authentication request for a Swauth-based user.
+func Authenticate(c *gophercloud.ProviderClient, opts AuthOptsBuilder) (r GetAuthResult) {
```
jrperritt: For consistency with the rest of the library, some options:
1. Rename function to `Auth` and result to `AuthResult`.
2. Rename result to `AuthenticateResult` and opts to `AuthenticateOpts`
3. Other suggestions

openstack/objectstorage/v1/swauth/requests.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	r.Err = err
+
+	return r
+}
+
+// NewObjectStorageV1 creates a Swauth-authenticated *gophercloud.ServiceClient
+// client that can issue ObjectStorage-based API calls.
+func NewObjectStorageV1(pc *gophercloud.ProviderClient, authOpts AuthOpts) (*gophercloud.ServiceClient, error) {
```
jrperritt: :+1:

jrperritt:
Just a couple of comments above ^

openstack/objectstorage/v1/swauth/requests.go
jtopjian: Quick thought: Is it worthwhile for this to go in `client.go` or `clients.go`?

jrperritt: I was thinking something similar. It's up to you; I'm not worried about about it because we can move around functions/types in a backwards-compatible way as long as it's intra-package.

jtopjian:
@jrperritt Thanks! Ping for re-review. :)

openstack/objectstorage/v1/swauth/requests.go
jrperritt: @jtopjian this one?

jtopjian: `:facepalm:`

I had that, then started renaming things to `Authenticate`, realized I liked `Auth` better, did `git checkout .` etc etc... hang on :)

openstack/objectstorage/v1/swauth/requests.go
```diff
@@ -0,0 +1,70 @@
...
+type AuthOptsBuilder interface {
+	ToAuthOptsMap() (map[string]string, error)
+}
+
+// AuthOpts specifies an authentication request.
+type AuthOpts struct {
+	// User is an Swauth-based username in username:tenant format.
+	User string `h:"X-Auth-User" required:"true"`
+	// Key is a secret/password to authenticate the User with.
+	Key string `h:"X-Auth-Key" required:"true"`
```
jrperritt: :+1:

jrperritt:
+2

jtopjian:
Thank you for your help with this!

jrperritt:
And you for yours

Issue #66, opened 2016-08-24
[wip] Add support for "Extend volume size" action to block storage client interface.


marioluan:
Openstack doc:
http://developer.openstack.org/api-ref-blockstorage-v2.html#extendVolume

marioluan:
I've already submited a PR: https://github.com/gophercloud/gophercloud/pull/67
Is there any change to merge that PR into master branch and make a new release?
cc @jamiehannaford @jrperritt @smashwilson

marioluan:
@jrperritt Thanks!

PR #67, opened 2016-08-24
Adds block storage extend's action support

marioluan:
This aims to fix https://github.com/gophercloud/gophercloud/issues/66
Openstack's source code: https://github.com/openstack/cinder/blob/8.1.0/cinder/volume/drivers/disco/disco.py#L379


jrperritt:
You'll need to remove your Godeps and .gitignore files

marioluan:
@jrperritt done!

openstack/blockstorage/v2/extensions/volumeactions/requests.go
```diff
@@ -172,3 +172,35 @@ func TerminateConnection(client *gophercloud.ServiceClient, id string, opts Term
...
+// ExtendSizeOptsBuilder allows extensions to add additional parameters to the
+// ExtendSize request.
+type ExtendSizeOptsBuilder interface {
+	ToVolumeExtendSizeMap() (map[string]interface{}, error)
+}
+
+// ExtendSizeOpts contain options for extending the size of an existing Volume. This object is passed
+// to the volumes.ExtendSize function.
+type ExtendSizeOpts struct {
+	NewSize int `json:"new_size,omitempty"`
```
jrperritt: What units is `NewSize`? GBs?

Should this be required? If so:
```
NewSize int `json:"new_size" required:"true"`
```

What happens if a value less than the current size is provided?

marioluan: @jrperritt
I've added the required tag to the field and also added a documentation to it, regarding the format (GB).

Openstack returns an error if the value is less than the current size. Should I do something about it?

jrperritt: >  Should I do something about it?

No, I was just curious if OpenStack would downsize as well. Nothing we can do without separately querying the current size.

jrperritt:
+2

PR #68, opened 2016-08-26
Add os-volume_upload_image action to volumeactions

sample:
This action makes volume-backed image (cli equivalent cinder upload-to-image)


jtopjian:
@jrperritt This can be closed, too.

Issue #69, opened 2016-08-27
Network Provider SegmentationID type error


jtopjian:
Even though `segmentation_id` is [defined](http://developer.openstack.org/api-ref/networking/v2/?expanded=list-networks-provider-network-detail) as a string, it's actually being returned as a number.

Examples can be seen [here](http://specs.openstack.org/openstack/neutron-specs/specs/api/networking_general_api_information.html) as well as throughout the Neutron code where it's defined as Int.

So I think [SegmentationID](https://github.com/gophercloud/gophercloud/blob/e1c6ceb3abcbec6c99ada70e81346c8e1f98a490/openstack/networking/v2/extensions/provider/results.go#L48) needs to be updated as an `int`. VXLAN allows for 16 million vxlans, and thus 16 million VXLAN IDs, which `int` can handle, but maybe there's another network type that goes past an `int`'s max number and this should be `int64`? Or deal with that if it arises?

jrperritt:
Can we just `string(id)` in an `Unmarshal` method? I think IDs should be strings, but I'm open to discussion.

jtopjian:
That makes sense.

However, browsing through the [neutron code](https://github.com/openstack/neutron/search?utf8=%E2%9C%93&q=segmentation_id) it looks like `segmentation_id` is always defined and treated as a number. 

On one hand I like the comfort of a string, but on the other it feels weird to treat this field differently than how it's defined in OpenStack... which is ironic given the amount of type smoothing that is done in other places. :)

jtopjian:
Oh, actually, another argument for a string is in the case of zero value.

Neutron is returning "null" rather than 0. Having a `segmentation_id` of 0 might cause [unexpected results](https://www.google.ca/?ion=1&espv=2#q=vlan%200).

I'll go with a string and rework this.

jrperritt:
closed by #70

PR #70, opened 2016-08-27
Ensure returned segmentation_id is a string

jtopjian:
For #69

### DIFF
--- a/openstack/networking/v2/extensions/provider/results.go
+++ b/openstack/networking/v2/extensions/provider/results.go
@@ -1,6 +1,9 @@
 package provider
 
 import (
+	"encoding/json"
+	"strconv"
+
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	"github.com/gophercloud/gophercloud/pagination"
 )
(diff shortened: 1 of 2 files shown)


jtopjian:
@jrperritt ping for review

openstack/networking/v2/extensions/provider/results.go
```diff
@@ -48,6 +51,29 @@ type NetworkExtAttrs struct {
...
+		return err
+	}
+
+	*n = NetworkExtAttrs(networkExtAttrs.tmp)
+
+	switch t := networkExtAttrs.SegmentationID.(type) {
+	case float64:
+		n.SegmentationID = strconv.FormatFloat(t, 'f', -1, 64)
+	case string:
+		n.SegmentationID = string(t)
```
jrperritt: Nothing wrong with this, but because we know `t` is of type `string` here (because we're in `case string`), we should be able to just:
```go
n.SegmentationID = t
``` 
(Note: no change required, more of a for-future-reference comment)

jrperritt:
+2

PR #73, opened 2016-08-29
Add VolumeType option to bootfromvolume extension

sample:
Useful when need to choose volume-type for boot device.

### DIFF
--- a/acceptance/openstack/compute/v2/bootfromvolume_test.go
+++ b/acceptance/openstack/compute/v2/bootfromvolume_test.go
@@ -29,6 +29,8 @@ func TestBootFromVolumeSingleVolume(t *testing.T) {
 			UUID:       choices.ImageID,
 			SourceType: bootfromvolume.Image,
 			VolumeSize: 10,
+			VolumeType: "fast",
+			DestinationType: "volume",
 		},
 	}
 
--- a/openstack/compute/v2/extensions/bootfromvolume/requests.go
+++ b/openstack/compute/v2/extensions/bootfromvolume/requests.go
@@ -39,6 +39,8 @@ type BlockDevice struct {
 	GuestFormat string `json:"guest_format,omitempty"`
 	// VolumeSize is the size of the volume to create (in gigabytes).
 	VolumeSize int `json:"volume_size"`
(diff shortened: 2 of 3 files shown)


jtopjian:
Hi there,

Just a comment about the acceptance test:

The acceptance test will fail unless the person running the test has created a volume type of "fast" beforehand. Perhaps a good way to handle this is to create a new test that is only run if the environment `OS_VOLUME_TYPE` is set.

You can see how environment variables are obtained in the tests [here](https://github.com/sample/gophercloud/blob/1f19437317ab8345c4c307ddf9c2306fa8be9afa/acceptance/openstack/compute/v2/bootfromvolume_test.go#L22-L25).

You can add new environment variables [here](https://github.com/sample/gophercloud/blob/1f19437317ab8345c4c307ddf9c2306fa8be9afa/acceptance/clients/clients.go). 

If you make `OS_VOLUME_TYPE` a required variable, it would be nice to have the volume type created and defined in the [reference devstack script](https://github.com/sample/gophercloud/blob/1f19437317ab8345c4c307ddf9c2306fa8be9afa/script/acceptancetest_environments/keystonev2-lbaasv1.sh).

Let me know if you have any questions or thoughts :)

jrperritt:
@sample in addition, before this gets reviewed/merged, it needs to adhere to the [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md).

jtopjian:
I was taking another look at this and I'm unable to see where the Block Device Mapping extension supports a Block Storage volume type. BDM supports a `device_type`, though that's a different kind of "type".

It's certainly possible to specify a Block Device type in a roundabout way: First create the block device volume with a specified [type](https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v2/volumes/requests.go#L40) and then configure the instance's device mapping to use that existing volume.

Other than that, and unless I'm terribly mistaken, I don't think this is valid.

jtopjian:
I'm going to go ahead and close this PR. As mentioned above, I wasn't able to see where this is implemented in OpenStack. If I'm incorrect about this, do let me know and we'll get it sorted out.

Issue #50, opened 2016-08-14
Unmarshal error when copying object


jtopjian:
I'm seeing the following error when trying to extract the results of a copy:

```
json: cannot unmarshal string into Go value of type int64
```

Here's the code I'm using:

```go
copyResult := objects.Copy(client, containerName, objectName, copyOpts)
copyHeader, err := copyResult.Extract()
```

Here's what `copyResult` looks like:

```
objects.CopyResult{HeaderResult:gophercloud.HeaderResult{Result:gophercloud.Result{Body:interface {}(nil), Header:http.Header{"Last-Modified":[]string{"Sun, 14 Aug 2016 23:11:50 GMT"}, "Etag":[]string{"4cb9bc8724551e6c7415f1080df84869"}, "X-Copied-From-Account":[]string{"AUTH_a54442e26ad8409a905659a268197eae"}, "Content-Type":[]string{"text/html; charset=UTF-8"}, "X-Trans-Id":[]string{"tx76234fbd166a462ca6e84-0057b0fab5"}, "Date":[]string{"Sun, 14 Aug 2016 23:11:49 GMT"}, "Content-Length":[]string{"0"}, "X-Copied-From":[]string{"test-container-Jy8w56q9/test-object-chsLSwuc"}, "X-Copied-From-Last-Modified":[]string{"Sun, 14 Aug 2016 23:11:50 GMT"}}, Err:error(nil)}}}
```

[ContentLength](https://github.com/gophercloud/gophercloud/blob/6dea3f17463923a8ea359d85ce2acaeb3ebb1c23/openstack/objectstorage/v1/objects/results.go#L262) is triggering the error

This is probably the same as the other unmarshal errors, but I'm not sure what to do since `CopyResult` is of type `gophercloud.HeaderResult`. Doing it incorrectly is probably going to break more things down the chain.

As well, `ContentLength` is defined as a string for the other structs. So maybe changing `int64` to `string` is the solution?

jrperritt:
Same logic as [this](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/results.go#L169) should work, no?

jrperritt:
I'd prefer to have types as they should be e.g. `ContentLength` as a number, times as `time.Time`s, ...

jtopjian:
> Same logic as this should work, no?

Ah, yes. I see how this works now. I got side tracked into `HeaderResult`.

[...]

PR #74, opened 2016-08-29
More specific types for ObjectStorage response object fields

jrperritt:
for #50 

- [x] Accounts
- [x] Containers
- [x] Objects
- [x] Accounts unit tests
- [x] Containers unit tests
- [x] Objects unit tests


jrperritt:
Travis is struggling for some reason, but the tests pass.

Issue #78, opened 2016-08-31
Network Acceptance test updates


jtopjian:
Following suit, it would be nice to update and standardize the network acceptance tests.

jrperritt:
closed by #79

PR #79, opened 2016-08-31
Networking Acceptance Test Cleanup

jtopjian:
for #78


jtopjian:
@jrperritt ping for review

Also included is a second devstack script to bring up a keystone v3 and lbaas v2 environment. The existing devstack script can be used for keystone v2 and lbaas v1. If there's an opportunity to run automated acceptance tests, I'm not sure how best to run the tests separately in an automated fashion -- maybe an extra build tag?

jrperritt:
> I'm not sure how best to run the tests separately in an automated fashion

Which tests specifically? 
`go test github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2/...` should run just the networking v2 acceptance tests.

jtopjian:
Right, but, specifically for networking, lbaas and lbaasv2 can't be run on the same OpenStack environment. Given that lbaas is deprecated, it's probably ignorable altogether.

In that case, the only other combination would be keystone v2 and v3.

jrperritt:
Ah, I see. Off the top of my head, one alternative to leaving it as is and letting users run the tests the want to run: 
In the relevant test packages (maybe in `init` functions), check the user's catalog for the service version, set a global flag, and, in the `Test*` functions, run or skip based on the flag.

jtopjian:
Good ideas :)

jrperritt:
Did you want to make any more changes to this or is it still ready to merge?

jtopjian:
Oh, sorry - I meant to follow up this morning to say that the automated test ideas discussed are probably best suited for another PR.

IMO, this is ready to go. Let me know if you'd like to see anything fixed up, though.

jrperritt:
+2

PR #83, opened 2016-09-01
Refactor WaitFor

stgleb:
Refactor WaitFor by using select instead of
checking timeout all the time. Add negative unit
tests for WaitFor and ability to skip long tests
on timeout. Check condition right before first blocking
in case if it completes fast.

### DIFF
--- a/testing/util_test.go
+++ b/testing/util_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"testing"
 
+	"errors"
 	"github.com/gophercloud/gophercloud"
 	th "github.com/gophercloud/gophercloud/testhelper"
 )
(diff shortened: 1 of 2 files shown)


jtopjian:
@stgleb I'd like to discuss this a bit more, so can you open a corresponding issue for this per the [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md)? 

Namely, this isn't triggering a timeout for me. I modified the function to add some logging, and [here's](https://gist.github.com/jtopjian/2fff393a232a76000dabc732932ef27d) what I'm seeing. You can see in the output that more than 2 seconds have elapsed, yet a timeout didn't occur. Maybe I'm doing something wrong, though.

jrperritt:
closed by #221

PR #89, opened 2016-09-07
Use new endpoint type for magnum

carolynvs:
Magnum has changed the endpoint type in the service catalog from "container" to "container-infra".


carolynvs:

PR #90, opened 2016-09-07
Add BeginDetaching call for Cinder v2

eyakubovich:
Per Cinder API, os-begin_detaching action needs to
be called prior to os-terminate_connection. For #117 

Corresponding Cinder code: https://github.com/openstack/cinder/blob/stable/newton/cinder/api/contrib/volume_actions.py#L138


philips:
LGTM

jrperritt:
will review tomorrow

jrperritt:
+2

PR #92, opened 2016-09-09
Check that Create token request succeeded

eyakubovich:
If there was a network error, this needs to be
handled to avoid nil pointer derferences later.

For #109


philips:
LGTM

jrperritt:
+2

Issue #86, opened 2016-09-04
Issue creating server with metadata


jtopjian:
When creating a server with metadata in the Create request, the metadata is not being recognized.

Upon digging into why, I've found a few metadata inconsistencies but am not sure what the best overall fix would be:

1. `Metadata` is defined in the [request](https://github.com/gophercloud/gophercloud/blob/a4d4930ed40c27a2c51486d2de6880f36fe74bd4/openstack/compute/v2/servers/requests.go#L159) as a `map[string]string` but is a `map[string]interface{}` in the [result](https://github.com/gophercloud/gophercloud/blob/a4d4930ed40c27a2c51486d2de6880f36fe74bd4/openstack/compute/v2/servers/results.go#L157). This has not changed from the Rackspace code, but should either one or the other be used in both places? 

2. The JSON tag in the request is declaring to omit the `metadata` field. This is used in some other areas of the request, but I believe that is so those fields can be given extra attention in the [ToServerCreateMap](https://github.com/gophercloud/gophercloud/blob/a4d4930ed40c27a2c51486d2de6880f36fe74bd4/openstack/compute/v2/servers/requests.go#L184).

I believe it is the addition of this JSON tag that is ultimately causing the metadata to not be recognized during the creation. But should the `CreateOpts` be changed to:

```go
Metadata map[string]string `json:"metadata,omitempty"`
```

Or should something like the following be added to `ToServerCreateMap`:

```go
if len(opts.Metadata) > 0 {
  metadata := make(map[string]string, len(opts.Metadata))
  for k, v := range opts.Metadata {
    metadata[k] = v
  }
  b["metadata"] = metadata
}
```

I'm inclined to go with just the JSON tag, but wanted to ask in case there's a reason for the latter.

jrperritt:
> I believe it is the addition of this JSON tag that is ultimately causing the metadata to not be recognized during the creation

Yep, I missed that one

> ```Metadata map[string]string `json:"metadata,omitempty"````

That should do it

jrperritt:
The `metadata` in the result should be changed to `map[string]string` as well.

jtopjian:
Sounds good. I can get a PR for this in the next day or two. Thanks!

jrperritt:
I wonder if I left it out because `gophercloud.BuildRequestBody` doesn't handle maps. If it does handle them, the JSON tag should work fine. If not, option 2.

jtopjian:
[...]

PR #93, opened 2016-09-10
Server Metadata Update

jtopjian:
This commit updates the Server Metadata so it can be parsed and
translated using JSON tags.

The Server Metadata result has been changed to map[string]string to
match the request.

A unit test was added to verify the above and a piece of metadata has
been added to the server creation acceptance test.

For #86


jtopjian:
@jrperritt ping for review

jrperritt:
+2

Issue #94, opened 2016-09-12
Sometimes No ImageRef is OK


jtopjian:
Currently, server creation will [fail](https://github.com/gophercloud/gophercloud/blob/37b56e8781b116e76a7818251674c63a30d5dd63/openstack/compute/v2/servers/requests.go#L222) if neither `ImageRef` and `ImageName` are provided, but sometimes it's OK for the `ImageRef` to be empty. Notably when [booting from a volume](https://github.com/openstack/nova/blob/9d9212c2a696950045301328a58e06d029926c3b/nova/api/openstack/compute/servers.py#L936).

What's the best way to modify the `ToServerCreateMap` to account for this? Does `opts` have access to `BlockDevice` (or `block_device_mapping`) inside `servers.ToServerCreateMap`?

If it doesn't, is the next option to modify the check for `ImageRef` and `ImageName` to not fail when neither are provided?

jrperritt:
This is a good example of why errors should be exported but internal-only.

> Does opts have access to BlockDevice (or block_device_mapping) inside servers.ToServerCreateMap?

No, but it's in `bfv.ToServerCreateMap` where this should be handled. It would entail type-asserting a non-nil error returned from `servers.ToServerCreateMap` to see if it was the "NoImageNameOrID" error and, if so, ignoring it.

jtopjian:
> No, but it's in bfv.ToServerCreateMap where this should be handled. It would entail type-asserting a non-nil error returned from servers.ToServerCreateMap to see if it was the "NoImageNameOrID" error and, if so, ignoring it.

Ahh - I see. Clever.  I'll see if I can get this in place and submit a PR.

Thanks!

jrperritt:
closed by #97

Issue #76, opened 2016-08-30
use interfaces for errors


jrperritt:
Instead of exporting custom error types, I think it'd be better to not export them, and instead have them implement an interface (e.g. `InvalidInput`, `MissingInput`).

Needs more thought.

chaosaffe:
Interfaces are used to ensure common functionality. Unless you intend to implement specific functionality on `InvalidInput` and `MissingInput` then they would match the `error` interface as well as any other interface that has the same functionality. There is no way to declare that a struct belongs to a specific interface, you can only implement the functionality required by the interface. So your options are to add frivolous functionality to match the interface, or use the `error` interface and have it match `error`, `InvalidInput` and `MissingInput` interfaces, negating the idea of separating them. 

In golang the common practice (see the core and `x` packages for reference) to export custom error types to allow users to do appropriate error checking and handling.

pierreprinetti:
Error handling has slightly improved in v2. I don't think we're there yet though, and I would invite you to formulate a use-case you'd want Gophercloud to handle more ergonomically. I'll go ahead and close this, please open a new issue with a case if you are still interested.

PR #97, opened 2016-09-13
Allow ImageRef to be empty when block device is present

jtopjian:
For #94

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -219,23 +219,21 @@ func (opts CreateOpts) ToServerCreateMap() (map[string]interface{}, error) {
 		b["networks"] = networks
 	}
 
-	// If ImageRef isn't provided, use ImageName to ascertain the image ID.
+	// If ImageRef isn't provided, check if ImageName was provided to ascertain
+	// the image ID.
 	if opts.ImageRef == "" {
-		if opts.ImageName == "" {
-			err := ErrNeitherImageIDNorImageNameProvided{}
-			err.Argument = "ImageRef/ImageName"
-			return nil, err
+		if opts.ImageName != "" {
+			if sc == nil {
+				err := ErrNoClientProvidedForIDByName{}
+				err.Argument = "ServiceClient"
+				return nil, err
+			}
+			imageID, err := images.IDFromName(sc, opts.ImageName)
+			if err != nil {
+				return nil, err
+			}
+			b["imageRef"] = imageID
 		}
-		if sc == nil {
-			err := ErrNoClientProvidedForIDByName{}
-			err.Argument = "ServiceClient"
-			return nil, err
-		}
-		imageID, err := images.IDFromName(sc, opts.ImageName)
-		if err != nil {
-			return nil, err
-		}
-		b["imageRef"] = imageID
 	}
 
 	// If FlavorRef isn't provided, use FlavorName to ascertain the flavor ID.
@@ -424,23 +422,21 @@ func (opts RebuildOpts) ToServerRebuildMap() (map[string]interface{}, error) {
 		return nil, err
 	}
 
-	// If ImageRef isn't provided, use ImageName to ascertain the image ID.
+	// If ImageRef isn't provided, check if ImageName was provided to ascertain
+	// the image ID.
 	if opts.ImageID == "" {
-		if opts.ImageName == "" {
-			err := ErrNeitherImageIDNorImageNameProvided{}
-			err.Argument = "ImageRef/ImageName"
-			return nil, err
+		if opts.ImageName != "" {
+			if opts.ServiceClient == nil {
+				err := ErrNoClientProvidedForIDByName{}
+				err.Argument = "ServiceClient"
+				return nil, err
+			}
+			imageID, err := images.IDFromName(opts.ServiceClient, opts.ImageName)
+			if err != nil {
+				return nil, err
+			}
+			b["imageRef"] = imageID
 		}
-		if opts.ServiceClient == nil {
-			err := ErrNoClientProvidedForIDByName{}
-			err.Argument = "ServiceClient"
-			return nil, err
-		}
-		imageID, err := images.IDFromName(opts.ServiceClient, opts.ImageName)
-		if err != nil {
-			return nil, err
-		}
-		b["imageRef"] = imageID
 	}
 
 	return map[string]interface{}{"rebuild": b}, nil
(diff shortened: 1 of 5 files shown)


jtopjian:
@jrperritt thoughts on this?

openstack/compute/v2/servers/requests.go
```diff
@@ -182,6 +182,7 @@ type CreateOpts struct {
 
 // ToServerCreateMap assembles a request body based on the contents of a CreateOpts.
 func (opts CreateOpts) ToServerCreateMap() (map[string]interface{}, error) {
+	var serverCreateMapError error
```
jrperritt: Hmm. I see the issue now: I was thinking we shouldn't have to modify this at all, but we do because the request body hasn't been fully populated when the error occurs. Options include:
1. Removing the logic that checks for an image ID/name and let OpenStack handle it
2. Fake it by providing an `ImageID` if it and `ImageName` don't exist, call `servers.ToServerCreateMap`, then delete the `image_ref` map entry.
3. Move the `ImageID`/`ImageName` checking logic down to the bottom of `servers.ToServerCreateMap` such that the request body is completely populated and the `error` returned will be that one or `nil`.

What do you think of those options (or others)?

jtopjian: Out of the three, I think #1 is the best way to go. It resumes the "return early" format and enables booting from volume without any modifications. The error returned by OpenStack should be sufficient for the dev/user to figure out what's going on and change their gophercloud-based app accordingly.

I like it better than the change I made - I prefer sticking to the return early form.

#2 introduces a fix unique to gophercloud.

#3 feels like it's literally pushing the problem away until later.

That's just my $0.02 - I'm no dev.

jrperritt: > 2. introduces a fix unique to gophercloud.

I don't follow: In what way could we create a fix that would be non-unique to Gophercloud?

> 3. feels like it's literally pushing the problem away until later.

I agree that 3. is weak, but it would eliminate the need for `serverCreateMapError`.

Of course the most straightforward solution would be to not call `servers.ToServerCreateMap` and instead put the server creation logic in `bfv.ToServersCreateMap` (yes, resulting in code duplication...).

jtopjian: > I don't follow: In what way could we create a fix that would be non-unique to Gophercloud?

If the code were assign a value of "THIS-IMAGE-DOES-NOT-EXIST" to `ImageRef`, its purpose would strictly be as a dummy placeholder to fix how Gophercloud implemented bfv support.

IMO, this kind of fix is more of a meta-fix, building gophercloud's logic on top of itself, rather than being a clean abstraction to the API.

tl;dr: it feels hacky.

> Of course the most straightforward solution would be to not call servers.ToServerCreateMap and instead put the server creation logic in bfv.ToServersCreateMap (yes, resulting in code duplication...).

Indeed. I would support and be happy to implement either 1. or this option. I'm also more tha happy to leave the decision of which to you ;)

jrperritt: > tl;dr: it feels hacky.

No doubt.

> I would support and be happy to implement either 1. or this option

I agree with your original analysis. Let's go with option 1.

jtopjian:
Removing the ImageRef and ImageName checks was easy enough. Then I wanted to see if I could write an acceptance test that could easily catch the incorrect creation of a non-bfv server that didn't have either set.

I was able to grab the error [here](https://github.com/jtopjian/gophercloud/blob/4b0245f556da994b22e5124abf1cf4f82f634bd8/acceptance/openstack/compute/v2/servers_test.go#L104), which is:

```
{"badRequest": {"message": "Missing imageRef attribute", "code": 400}}
```

And then I went down a long rabbit hole of seeing if I could get the `message` part of that JSON string as the actual error message, so one could do:

```go
if error.Error() == "Missing imageRef attribute" {
...
```

And beyond that, maybe raising the existing `servers.ErrNoImageIDProvided`. But I got totally tangled up and lost. I started implementing a copy of the custom 404 errors found [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/errors.go), but then it looked like I had to implement some `RequestOpts` [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/requests.go#L270) so I could set [ErrorContext](https://github.com/jtopjian/gophercloud/blob/4b0245f556da994b22e5124abf1cf4f82f634bd8/provider_client.go#L104).

Any idea how that could easily be done?

jrperritt:
> but then it looked like I had to implement some RequestOpts here so I could set ErrorContext.

`ErrorContext` is/was more of an experiment that's almost certainly going away (purpose of #76 ), so definitely avoid that if you can.

> Any idea how that could easily be done?

To test the exact structure of the error response, create a type with that structure and `json.Unmarshal` into it. To test the error message is what you expect, you could use a regular expression on `err.Error()`.

This issue is a good example of why defining errors in a flexible (for the library) and public (for users) way is hard. That message coming back from OpenStack could be anything really, now or in the future, so building an application that depends on the exact error message would be very brittle. `ErrorContext` only somewhat helps, because an application would still be broken if any exported, Gophercloud-defined error type changed in any way. I think errors as interfaces will help with the issue.

jrperritt:
Alternatively, if you know the exact error message, you can use `strings.Contains`:
```go
strings.Contains(`{"badRequest": {"message": "Missing imageRef attribute", "code": 400}}`, "Missing imageRef"))
```

jtopjian:
> ErrorContext is/was more of an experiment that's almost certainly going away (purpose of #76 ), so definitely avoid that if you can.

👍 

> To test the exact structure of the error response, create a type with that structure and json.Unmarshal into it. 

Yup. I started to get tangled up trying to get the error to become the new default error when that exact message was returned. It became more of a puzzle I wanted to solve rather than something good to have in place because...

> That message coming back from OpenStack could be anything really, now or in the future, so building an application that depends on the exact error message would be very brittle. 

Yes, exactly.

> I think errors as interfaces will help with the issue.

OK. In this case, perhaps continuing with the removal of the ImageRef and letting it flow through to OpenStack is still a good idea. A new error system could be added later rather than implementing one form and changing it out later.

Thanks for the input! I'll clean up this PR and let you know when it's ready.

jtopjian:
@jrperritt ping for review!

Along with the related acceptance test changes, I found a few minor issues - let me know if I should pull these out for another PR.

I also found a few other places in the bfv extension that needs worked on -- I'll open another issue/pr for those since they're not related to ImageRef.

Issue #98, opened 2016-09-15
Boot From Volume Updates


jtopjian:
I'm pretty sure `DestinationType` is required because I'm getting errors returned whenever it's omitted. 

Additionally, there are only two `DestinationType`s possible, volume and local, so it might be worthwhile to have them as defined types. However, I'm not sure what to do since there's an existing `SourceType` of `Volume` -- have a combined type of `SourceDestinationType` and then check for the allowable ones in the `ToServersCreateMap`?

Also, `VolumeSize` can be omitted in certain situations.

The different combinations of allowed block device options are described [here](http://docs.openstack.org/developer/nova/block_device_mapping.html).

jtopjian:
I've done some sketching of these changes [here](https://github.com/jtopjian/gophercloud/commit/30eaaca8450d4aef545cbdb54b77205f6dc7bf07). I don't want to open a PR on it until #94 is taken care of first.

jrperritt:
> I'm pretty sure DestinationType is required because I'm getting errors returned whenever it's omitted

Per the style guide, we really need source code proof of the issue and resolution. I don't doubt that you're getting errors, but we can't know the cause for sure without reading the source (e.g. it may be an invalid combination of parameters, an old version of the nova API, etc).

> Also, VolumeSize can be omitted in certain situations.

Same as above.

> The different combinations of allowed block device options are described here.

I know it's tempting to rely on the documentation, but because it isn't auto-generated from the source, it should be considered unreliable.

jtopjian:
> Per the style guide, we really need source code proof of the issue and resolution.

I appreciate this enforcement.

AFAICT, Nova is only making `source_type` required [here](https://github.com/openstack/nova/blob/841fc2ecde608a012660828011cf1f5dea322026/nova/block_device.py#L76).

> (e.g. it may be an invalid combination of parameters, an old version of the nova API, etc).

Yes, in this case the error looks to be coming from libvirt. While the error is being caught by Nova, it is not being caught in a way that should make `destination_type` required. This is probably a bug and I'd imagine that this is hypervisor-specific. Therefore, keeping `destination_type` as a non-required field makes sense.

> Also, VolumeSize can be omitted in certain situations.

Reading through the BlockDevice and general Nova API code, there are a number of places where `volume_size` is not checked. 

[...]

PR #100, opened 2016-09-17
bfv updates

jtopjian:
For #98


jtopjian:
@jrperritt Thoughts on the changes to `SourceType`? Maybe some validation like:

```go
if opts.SourceType == SourceDestinationType.Blank {
  // SourceType cannot be Blank
}
```

In some ways it feels awkward, but having `SourceType` as a defined type and `DestinationType` as a string feels a little awkward, too.

Or is there some way to have `Volume` exist as both that is just outside of my Go knowledge?

jtopjian:

Please see the comments in #98 for API compatibility and let me know if there's any more research you'd like me to do.

With regard to the code specifically, the only thing I'm unsure about is the shared Type between source and destination.

Re my other comment about validity, it seems there are so many different combinations that can be done here (including letting `SourceType` be "blank", contrary to what I originally thought above), that it's probably best to just let the Compute service sort it out and return an error ?

jrperritt:
Will review tomorrow

openstack/compute/v2/extensions/bootfromvolume/requests.go
```diff
@@ -5,26 +5,29 @@ import (
...
-	Snapshot SourceType = "snapshot"
+	Snapshot SourceDestinationType = "snapshot"
 	// Image SourceType
-	Image SourceType = "image"
+	Image SourceDestinationType = "image"
 	// Blank SourceType
-	Blank SourceType = "blank"
+	Blank SourceDestinationType = "blank"
+	// Local DestinationType
+	Local SourceDestinationType = "local"
```
jrperritt: How about we leave `SourceType` as-is, prefix the current constants with 'Source', and define a new `DestinationType`:
```go
type (
  DestinationType string
  SourceType string
)

const (
  SourceVolume SourceType = "volume"
  DestinationVolume DestinationType = "volume"
  ...
)
```

openstack/compute/v2/extensions/bootfromvolume/requests.go
```diff
@@ -34,11 +37,11 @@ type BlockDevice struct {
 	DeleteOnTermination bool `json:"delete_on_termination"`
 	// DestinationType is the type that gets created. Possible values are "volume"
 	// and "local".
-	DestinationType string `json:"destination_type,omitempty"`
+	DestinationType SourceDestinationType `json:"destination_type,omitempty"`
 	// GuestFormat specifies the format of the block device.
 	GuestFormat string `json:"guest_format,omitempty"`
 	// VolumeSize is the size of the volume to create (in gigabytes).
-	VolumeSize int `json:"volume_size"`
+	VolumeSize int `json:"volume_size,omitempty"`
```
jrperritt: :+1:

jtopjian:
@jrperritt ping for review

jrperritt:
will review tomorrow

openstack/compute/v2/extensions/bootfromvolume/requests.go
```diff
@@ -5,40 +5,65 @@ import (
...
-	// Image SourceType
-	Image SourceType = "image"
 	// Blank SourceType
 	Blank SourceType = "blank"
+
+	// Volume DestinationType
+	DestinationVolume DestinationType = "volume"
+
+	// Image SourceType
+	Image SourceType = "image"
```
jrperritt: These should all have either a 'Destination' or 'Source' prefix.
Also, let's group them by type, no alphabetically.

jrperritt:
Also, please do **not** rebase after a review has started

jtopjian:
@jrperritt Yeah, that's my bad - I need to break that habit.

Please see 599720ce8db8f65581b9a4072ebc1f15972ca8ef -- does that look OK for the rename / reorder?

jrperritt:
This looks good. +2

jtopjian:
Thank you again for your help and patience :)

Issue #101, opened 2016-09-19
Subnet Gateway Behavior


jtopjian:
When creating a subnet, there are three different actions that can happen:

1. When `gateway_ip` is omitted, Neutron will provision a default gateway.
2. When `gateway_ip` is specified with a value of nil/None, no gateway is provisioned.
3. When `gateway_ip` is specified and is a valid IP, that IP is used for the gateway.

Currently, gophercloud can do 1 and 3, but not 2.

[reference](https://github.com/openstack/neutron/blob/ba788da398b31d5433a91bdc72ff2695b475fa41/neutron/api/v2/attributes.py#L119-L127)

jrperritt:
closed by #102

PR #102, opened 2016-09-19
Update Subnet Gateway Behavior

jtopjian:
This commit enables all three behaviors of a gateway during subnet creation and
updating.

If a GatewayIP is omitted, Neutron will provision a default gateway.
If a GatewayIP is set to an empty string, no gateway will be provisioned.
If a GatewayIP is specified, it will be used as the gateway IP.

A convenience function of `StringToPointer` was added to help pass empty
strings as pointers.

For #101

### DIFF
--- a/acceptance/openstack/networking/v2/networking.go
+++ b/acceptance/openstack/networking/v2/networking.go
@@ -63,14 +63,75 @@ func CreateSubnet(t *testing.T, client *gophercloud.ServiceClient, networkID str
 	subnetOctet := tools.RandomInt(1, 250)
 	subnetCIDR := fmt.Sprintf("192.168.%d.0/24", subnetOctet)
 	subnetGateway := fmt.Sprintf("192.168.%d.1", subnetOctet)
-	iFalse := false
 	createOpts := subnets.CreateOpts{
 		NetworkID:  networkID,
 		CIDR:       subnetCIDR,
 		IPVersion:  4,
 		Name:       subnetName,
-		EnableDHCP: &iFalse,
+		EnableDHCP: gophercloud.Disabled,
 		GatewayIP:  &subnetGateway,
+	}
+
+	t.Logf("Attempting to create subnet: %s", subnetName)
+
+	subnet, err := subnets.Create(client, createOpts).Extract()
+	if err != nil {
+		return subnet, err
+	}
+
+	t.Logf("Successfully created subnet.")
+	return subnet, nil
+}
+
+// CreateSubnetWithDefaultGateway will create a subnet on the specified Network
+// ID and have Neutron set the gateway by default An error will be returned if
+// the subnet could not be created.
+func CreateSubnetWithDefaultGateway(t *testing.T, client *gophercloud.ServiceClient, networkID string) (*subnets.Subnet, error) {
+	subnetName := tools.RandomString("TESTACC-", 8)
+	subnetOctet := tools.RandomInt(1, 250)
+	subnetCIDR := fmt.Sprintf("192.168.%d.0/24", subnetOctet)
+	createOpts := subnets.CreateOpts{
+		NetworkID:  networkID,
+		CIDR:       subnetCIDR,
+		IPVersion:  4,
+		Name:       subnetName,
+		EnableDHCP: gophercloud.Disabled,
+	}
+
+	t.Logf("Attempting to create subnet: %s", subnetName)
+
+	subnet, err := subnets.Create(client, createOpts).Extract()
+	if err != nil {
+		return subnet, err
+	}
+
+	t.Logf("Successfully created subnet.")
+	return subnet, nil
+}
+
+// CreateSubnetWithNoGateway will create a subnet with no gateway on the
+// specified Network ID.  An error will be returned if the subnet could not be
+// created.
+func CreateSubnetWithNoGateway(t *testing.T, client *gophercloud.ServiceClient, networkID string) (*subnets.Subnet, error) {
+	var noGateway = ""
+	subnetName := tools.RandomString("TESTACC-", 8)
+	subnetOctet := tools.RandomInt(1, 250)
+	subnetCIDR := fmt.Sprintf("192.168.%d.0/24", subnetOctet)
+	dhcpStart := fmt.Sprintf("192.168.%d.10", subnetOctet)
+	dhcpEnd := fmt.Sprintf("192.168.%d.200", subnetOctet)
+	createOpts := subnets.CreateOpts{
+		NetworkID:  networkID,
+		CIDR:       subnetCIDR,
+		IPVersion:  4,
+		Name:       subnetName,
+		EnableDHCP: gophercloud.Disabled,
+		GatewayIP:  &noGateway,
+		AllocationPools: []subnets.AllocationPool{
+			{
+				Start: dhcpStart,
+				End:   dhcpEnd,
+			},
+		},
 	}
 
 	t.Logf("Attempting to create subnet: %s", subnetName)
--- a/openstack/networking/v2/subnets/requests.go
+++ b/openstack/networking/v2/subnets/requests.go
@@ -79,7 +79,7 @@ type CreateOpts struct {
 	Name            string                `json:"name,omitempty"`
 	TenantID        string                `json:"tenant_id,omitempty"`
 	AllocationPools []AllocationPool      `json:"allocation_pools,omitempty"`
-	GatewayIP       *string               `json:"gateway_ip"`
+	GatewayIP       *string               `json:"gateway_ip,omitempty"`
 	IPVersion       gophercloud.IPVersion `json:"ip_version,omitempty"`
 	EnableDHCP      *bool                 `json:"enable_dhcp,omitempty"`
 	DNSNameservers  []string              `json:"dns_nameservers,omitempty"`
@@ -88,7 +88,16 @@ type CreateOpts struct {
 
 // ToSubnetCreateMap casts a CreateOpts struct to a map.
 func (opts CreateOpts) ToSubnetCreateMap() (map[string]interface{}, error) {
-	return gophercloud.BuildRequestBody(opts, "subnet")
+	b, err := gophercloud.BuildRequestBody(opts, "subnet")
+	if err != nil {
+		return nil, err
+	}
+
+	if m := b["subnet"].(map[string]interface{}); m["gateway_ip"] == "" {
+		m["gateway_ip"] = nil
+	}
+
+	return b, nil
 }
 
 // Create accepts a CreateOpts struct and creates a new subnet using the values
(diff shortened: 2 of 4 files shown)


jtopjian:
@jrperritt ping for review/thoughts?

The [last time](https://github.com/rackspace/gophercloud/commit/f92ae6c762ef68e7fbe914f793d14356805a0a50) I came across this, I added the `NoGateway` field, so I'm trying something different this time.

I think this is just an area where any solution in Go might feel awkward. Some might think that omitting the GatewayIP means they don't want a gateway at all rather than to have Neutron make one.

If this isn't a good solution, maybe the `NoGateway` field should be revisited? `NoGateway` was emulating the `neutron` cli tool's `--no-gateway` flag, but maybe that's something too app-specific for Gophercloud?

params.go
```diff
@@ -181,6 +181,12 @@ func IntToPointer(i int) *int {
 	return &i
 }
 
+// StringToPointer is a function for converting strings into string pointers.
+// This is useful when passing in options to operations.
+func StringToPointer(s string) *string {
```
jrperritt: How is this different from `MaybeString` (the function below this one)?

jtopjian: It's different because it's called `StringToPointer`. 

Yes, good catch. IIRC, I saw `MaybeString` and without reading it, disregarded it as something that would check if something was a string, bool, int, etc.

openstack/networking/v2/subnets/requests.go
```diff
@@ -79,7 +79,7 @@ type CreateOpts struct {
 	Name            string                `json:"name,omitempty"`
 	TenantID        string                `json:"tenant_id,omitempty"`
 	AllocationPools []AllocationPool      `json:"allocation_pools,omitempty"`
-	GatewayIP       *string               `json:"gateway_ip"`
+	GatewayIP       *string               `json:"-"`
```
jtopjian: The map should have three different forms:

1. User explicitly sets a gateway: 

`map[network_id:12345 cidr:1.2.3.0/24 gateway_ip:1.2.3.1]`

2. User does not want a gateway to exist: 

`map[network_id:12345 cidr:1.2.3.0/24 gateway_ip:]`

3. User wants Neutron to create a gateway, which will end up being `1.2.3.1`: 

`map[network_id:12345 cidr:1.2.3.0/24]`

jrperritt: OK, adding `omitempty` to the struct tag:

```
requests_test.go:22: m: map[subnet:map[network_id:12345 cidr:1.2.3.4.]]
requests_test.go:32: m: map[subnet:map[network_id:12345 cidr:1.2.3.4. gateway_ip:]]
```

jtopjian: I swear that was not working for me! I'll be cleaning up that other PR so I'll double-check this.

jtopjian: It's not working for me with `omitempty`. 

To confirm: the only change you made to master was the addition of `omitempty` to the `GatewayIP` tag?

Are you printing the map after `omitempty` would have stripped away the nil value? Isn't that the point of `omitempty`?

I'm printing `b` right after it was created in `Create` and either it's omitted (because it was nil) or it has a value of `""` (which Neutron does not want).

params.go
jtopjian: Unless I'm mistaken, one difference is that `StringToPointer` will return `&""`, where `MaybeString` will not.

openstack/networking/v2/subnets/requests.go
jrperritt: OK, I've done some testing, and I think I see where the discrepancy lies:

> 2) User does not want a gateway to exist:

> map[network_id:12345 cidr:1.2.3.0/24 gateway_ip:]

Not exactly. What we want is `gateway_ip:null` (or equivalently `gateway_ip:<nil>`). What's above represents `gateway_ip:""`, which OpenStack indeed does not accept.

With that in mind, I agree that this cannot be done as-is. Adding `omitempty` to the struct tag and modifying `ToSubnetCreateMap` to the following should fix it:

```go
func (opts CreateOpts) ToSubnetCreateMap() (map[string]interface{}, error) {
	b, err := gophercloud.BuildRequestBody(opts, "subnet")
	if err != nil {
		return nil, err
	}

	if m := b["subnet"].(map[string]interface{}); m["gateway_ip"] == "" {
		m["gateway_ip"] = nil
	}

	return b, nil
}
```

With those 2 changes:
- not providing a value for `CreateOpts.GatewayIP` will give the user the OpenStack default gateway
- providing the [a pointer to] the empty string will result in the subnet not having a gateway.

Now, is that reasonable or should it be flipped, where the user has to explicitly ask for the default gateway and otherwise gets no gateway?

params.go
jrperritt: > Unless I'm mistaken, one difference is that StringToPointer will return &"", where MaybeString will not.

Yes, I know :) I don't think we need this function. Creating a pointer to a `string` (as evidenced by this function) is trivial, and I suspect creating one for a `string` constant is uncommon. If you'd like something for the tests, you can define a exported variable in one of the common testing packages:
```go
var(
  Empty = ""
  EmptyPtr = &empty
)
```

openstack/networking/v2/subnets/requests.go
jtopjian: Cool - thank you for the attention on this. :)

I think the proposed workflow is good. This mirrors the default behavior in the Neutron CLI tool.

jtopjian:
@jrperritt `ToSubnetCreateMap` has been updated and `StringToPointer` has been removed.

openstack/networking/v2/subnets/requests.go
```diff
@@ -112,15 +121,30 @@ type UpdateOptsBuilder interface {
 // UpdateOpts represents the attributes used when updating an existing subnet.
 type UpdateOpts struct {
 	Name           string      `json:"name,omitempty"`
-	GatewayIP      string      `json:"gateway_ip,omitempty"`
+	GatewayIP      *string     `json:"-"`
```
jtopjian: Yes, totally forgot about Update. Coming up.

jtopjian:
@jrperritt Update updated, acceptance and unit tests added for gateway update.

openstack/networking/v2/subnets/requests.go
```diff
@@ -112,15 +121,24 @@ type UpdateOptsBuilder interface {
 // UpdateOpts represents the attributes used when updating an existing subnet.
 type UpdateOpts struct {
 	Name           string      `json:"name,omitempty"`
-	GatewayIP      string      `json:"gateway_ip,omitempty"`
+	GatewayIP      *string     `json:"gateway_ip,omitempty"`
```
jtopjian: In this case, I think reverting `UpdateOpts` is the best choice. While it makes for being inconsistent with `CreateOpts`, it more naturally enforces allowed values. I think that's the style more favored here.

I'll update the code, but won't push anything until confirmation.

jrperritt: > So is the only option left simply to ignore the case in the code and document the behavior?

Yes, this `Update` should work like all the others.

jtopjian: OK, so no changes needed to the Update part then?

If not, I have an extra unit test to push and then I think it's ready for another full review.

openstack/networking/v2/subnets/testing/requests_test.go
```diff
@@ -276,6 +277,7 @@ func TestCreate(t *testing.T) {
 		NetworkID: "d32019d3-bc6e-4319-9c1d-6722fc136a22",
 		IPVersion: 4,
 		CIDR:      "192.168.199.0/24",
+		GatewayIP: gophercloud.MaybeString("192.168.199.1"),
```
jrperritt: `MaybeString` is intended for internal use (but created before an `internal` package was a thing). Because users refer to tests for direction, let's define an IP address as a constant (or variable, if it'll get changed) and take its address here.

openstack/networking/v2/subnets/testing/requests_test.go
```diff
@@ -469,6 +556,64 @@ func TestUpdate(t *testing.T) {
...
+        "id": "08eae331-0402-425a-923c-34f7cfe39c1b"
+    }
+}
+	`)
+	})
+
+	var noGateway = ""
+	opts := subnets.UpdateOpts{
+		Name:      "my_new_subnet",
+		GatewayIP: &noGateway,
```
jrperritt: :+1:

openstack/networking/v2/subnets/requests.go
jrperritt: I think what you have there is sufficient.

jtopjian:
@jrperritt I've updated what should be the final commit (let me know if these should have been appended rather than squashed - I know you've asked for that before and I have probably slipped in that habit)

The two changes were: removal of `MaybeString` and I added an additional unit test I had not committed before.

acceptance/openstack/networking/v2/networking.go
```diff
@@ -63,13 +63,12 @@ func CreateSubnet(t *testing.T, client *gophercloud.ServiceClient, networkID str
...
 	subnetCIDR := fmt.Sprintf("192.168.%d.0/24", subnetOctet)
 	subnetGateway := fmt.Sprintf("192.168.%d.1", subnetOctet)
-	iFalse := false
 	createOpts := subnets.CreateOpts{
 		NetworkID:  networkID,
 		CIDR:       subnetCIDR,
 		IPVersion:  4,
 		Name:       subnetName,
-		EnableDHCP: &iFalse,
+		EnableDHCP: gophercloud.Disabled,
```
jtopjian: @jrperritt Is this along the same lines of `MaybeString`? I found this to be a helpful both in terms of being a shortcut as well as readability ("DHCP is Disabled").

jrperritt:
> let me know if these should have been appended rather than squashed

A few things:
- firstly, it's OK; no changes needed here,
- secondly, the merger can (and should, for targeted PRs) now squash all commits in the PR upon merging,  
- thirdly, last item under `On Pull Requests` section of style guide: "Unless explicitly asked, do not squash commits in the middle of a review; only append. It makes it difficult for the reviewer to see what's changed from one review to the next."

jrperritt:
+2

jtopjian:
Thank you for your help on this one!

openstack/networking/v2/subnets/requests.go
```diff
@@ -88,7 +88,16 @@ type CreateOpts struct {
...
 // ToSubnetCreateMap casts a CreateOpts struct to a map.
 func (opts CreateOpts) ToSubnetCreateMap() (map[string]interface{}, error) {
-	return gophercloud.BuildRequestBody(opts, "subnet")
+	b, err := gophercloud.BuildRequestBody(opts, "subnet")
+	if err != nil {
+		return nil, err
+	}
+
+	if m := b["subnet"].(map[string]interface{}); m["gateway_ip"] == "" {
+		m["gateway_ip"] = nil
```
fatmcgav: @jtopjian / @jrperritt Shouldn't this line have been `b["gateway_ip"] = nil`?

Else how does `b["gateway_ip"]` get set to `nil`?

Issue #99, opened 2016-09-16
Error 400 while disassociating floating IP


jccomputing:
When disassociating a floating IP from a Neutron port, I receive a 400 error: "Invalid request due to incorrect syntax or missing required parameters."

``` go
import "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips"
floatingips.Update(client, floatingIP.ID, floatingips.UpdateOpts{})
```

I dumped the request and I saw that the `PortID` attribute of `UpdateOpts` is serialized as an empty string. The Neutron API reference states that it should be a null value.

I'm not sure what could be a proper fix for this, but when I convert `PortID` to a pointer, it works fine:

``` go
type UpdateOpts struct {
     PortID *string `json:"port_id"`
}
```

jrperritt:
closed by #103

PR #103, opened 2016-09-20
Fix Floating IP Disassociation

jtopjian:
This commit fixes floating IP disassociation by changing the PortID to a
string pointer rather than a string. This allows a value of "null" to be
passed which is what the Networking API is looking for.

For #99


jtopjian:
@jrperritt ping for review

jrperritt:
+2

PR #105, opened 2016-09-26
Add create multiple servers

stgleb:
Add min_count and max_count fields to request,
for creating multiple servers.

Closes https://github.com/rackspace/gophercloud/issues/597

### DIFF
--- a/openstack/compute/v2/servers/errors.go
+++ b/openstack/compute/v2/servers/errors.go
@@ -69,3 +69,20 @@ type ErrServerNotFound struct {
 func (e ErrServerNotFound) Error() string {
 	return fmt.Sprintf("I couldn't find server [%s]", e.ID)
 }
+
+type ErrMinGreaterThanMax struct {
+	Min int
+	Max int
+}
+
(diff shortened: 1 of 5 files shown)


jtopjian:
@stgleb Sorry for the delay on this one. A few notes:

While there is an existing Rackspace issue for this, it's best to open a new Gophercloud issue. 

Additionally, this is actually a Nova extension defined here:

https://github.com/openstack/nova/blob/b9e94252e290d09d6c3e8e5ec794163cd702f0c9/nova/api/openstack/compute/multiple_create.py

Because of that, the implementation in Gophercloud should be similar to the other existing server creation extensions. For example:

* https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/bootfromvolume/requests.go
* https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/keypairs/requests.go

Let me know if you need help with this. If you're no longer interested in getting this implemented, that's no problem at all, but maybe let me know so I can go ahead and close this.

jrperritt:
closing due to inactivity

PR #110, opened 2016-09-30
support skipping the validity check for the server's certificate

dixudx:
Sometimes the certificate is self-signed, mainly used for internal development, testing and etc.

> x509: certificate signed by unknown authority

This patch supports such scenario.

### DIFF
--- a/auth_options.go
+++ b/auth_options.go
@@ -52,6 +52,8 @@ type AuthOptions struct {
 	// TokenID allows users to authenticate (possibly as another user) with an
 	// authentication token ID.
 	TokenID string `json:"-"`
+
+	InsecureSkipTlsVerify bool
 }
 
 // ToTokenV2CreateMap allows AuthOptions to satisfy the AuthOptionsBuilder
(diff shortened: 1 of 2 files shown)


jtopjian:
@dixudx Per the [Style Guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md), there needs to be a corresponding issue opened that describes the reason for the Pull Request, even if the Pull Request sufficiently details its own case.

Ignoring that for the moment, I think this feature is better handled outside of Gophercloud. This enables the application to more uniquely handle app-specific configuration and leaves Gophercloud to focus just on API interaction.

An example of how to configure the http client to accept "insecure" certificates can be seen [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/config.go#L43-L92) in Terraform.

Note that the Terraform implementation also handles additional features such as Proxy settings, custom intermediate certificates and even client certificate authentication. It's really quite brilliant that Gophercloud can accept outside customizations like this :smile:

Issue #62, opened 2016-08-17
reorganize unit tests


jrperritt:
The coverage stats are getting thrown off because `go cover` assumes the tests are testing the package they're in. Though I'd prefer to keep all the test files together, the easiest thing to do for now is move the `requests_test.go` files back into the package they test.

jrperritt:
closed by #115

jrperritt:
Instead of reorganizing the tests, I changed how the coverage was calculated.

PR #115, opened 2016-10-07
aggregate coverage metrics of unit tests across packages

jrperritt:
for #62


jrperritt:
The unit test coverage number [above] is now more representative of reality.

Issue #114, opened 2016-10-06
Support for shared file systems API V2 (manila)


mvalkon:
Gophercloud does not support [the shared file systems API](http://developer.openstack.org/api-ref-share-v2.html). We are using this API in our organization and will implement a library for go to use in our product. If possible, only the latest microversion of the API will be implemented.

We'd like to contribute our implementation to the gophercloud library. This issue is opened to track the progress for the PR's regarding this feature.

jrperritt:
Great. The [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md) contains useful information when making contributions.

mvalkon:
Any pointers on ways to orchestrate the PR's would be valued. We're aiming to produce small PR's implementing one API operation each, but this leads to PR's having duplicated content until merged.

jrperritt:
> We're aiming to produce small PR's implementing one API operation each, but this leads to PR's having duplicated content until merged.

That's expected. If the PRs are small, they can be reviewed (and merged) more quickly, allowing you to go to the next one. The PRs will be reviewed in order.

Aside from that, can you either create individual issues or checkboxes in this issue's description for each operation/PR to aid in record-keeping?

mvalkon:
@jrperritt Sure, do you have a preferred approach? I'd probably go for the checkbox one, but have no strong opinion.

mvalkon:
Opted to go with separate issues in order to maintain sanity in the record-keeping.

jrperritt:
>> Opted to go with separate issues

Great, no preference here.

I will review all of the currently-open Manila PRs tomorrow.

mvalkon:
@jrperritt thanks a lot!

pospispa:
I'd like to use Gophercloud in Kubernetes to communicate with the shared file systems API (Manila).
It's great that a part of the work was successfully completed. However, a part of the work is in PRs that are waiting for review and merge.

From my point of view it seems that the Gophercloud project got stuck as there are no new commits in the master branch since mid-November.

Please, are there any plans to revive this project and have reviews and merges to master start happening again?

mvalkon:
@pospispa I'd be willing to continue contributing on my own time if this project is revived. 

@jrperritt @jtopjian can you update on the situation?

jtopjian:
[...]

PR #118, opened 2016-10-12
Feature/filestorage sharenetworks create

ehdou:
Add support for Share Network Create operation
For  #130

### DIFF
--- a/openstack/sharedfilesystems/v2/sharenetworks/requests.go
+++ b/openstack/sharedfilesystems/v2/sharenetworks/requests.go
@@ -1,0 +1,46 @@
+package sharenetworks
+
+import "github.com/gophercloud/gophercloud"
+
+// CreateOptsBuilder allows extensions to add additional parameters to the
+// Create request.
+type CreateOptsBuilder interface {
+	ToShareNetworkCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts contains options for creating a ShareNetwork. This object is
+// passed to the sharenetworks.Create function. For more information about
+// these parameters, see the ShareNetwork object.
+type CreateOpts struct {
+	// The UUID of the Neutron network to set up for share servers
+	NeutronNetID string `json:"neutron_net_id,omitempty"`
+	// The UUID of the Neutron subnet to set up for share servers
+	NeutronSubnetID string `json:"neutron_subnet_id,omitempty"`
+	// The UUID of the nova network to set up for share servers
+	NovaNetID string `json:"nova_net_id,omitempty"`
+	// The share network name
+	Name string `json:"name"`
+	// The share network description
+	Description string `json:"description"`
+}
+
+// ToShareNetworkCreateMap assembles a request body based on the contents of a
+// CreateOpts.
+func (opts CreateOpts) ToShareNetworkCreateMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "share_network")
+}
+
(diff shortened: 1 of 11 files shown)


ehdou:
Manila share network Create : https://github.com/openstack/manila/blob/master/api-ref/source/share-networks.inc#L162

ehdou:
@jrperritt feel free to review if you have a moment

jrperritt:
Thank you for following the style guide as closely as you could. One issue: this link (https://github.com/gophercloud/gophercloud/pull/118#issuecomment-253145324) looks like it's to documentation, not source code. In order to adequately review this PR (and indeed the others), we'll want a link to the function/schema/view from the source code to verify fields, types, return codes, etc. For example, many OpenStack services have a `views` directory for response objects, and some have a `schemas` directory for request body information. Still others have neither of those and just have logic defined in functions. As an example, have a look at the link in the 3rd bullet point of the [`On Pull Requests`](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md#on-pull-requests) section of the style guide.

ehdou:
Thanks for your comment. I tried to find links in the code but it's not so simple. I found the [view](https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/api/views/share_networks.py) and the [DB](https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/db/sqlalchemy/models.py#L745). The [create](https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/api/v1/share_networks.py#L240) doesn't explicitly mention the return code, I'll try to find it. I also noticed that `name` and `description` are not required, I'll update my code in a moment.

ehdou:
@jrperritt If I remove the test case for failing to build the request body (because `name` and `description` are not required parameters) then the coverage decreases -0.004%. How can I overcome this problem ? I don't see anything else I can test in my PR, except for `NewSharedFileSystemV2` that gets the Manila client, but then I would be the only one to test the client. Any suggestion ?

ehdou:
It's a bit difficult to follow but response objects are assigned [here](https://github.com/openstack/manila/blob/master/manila/api/openstack/wsgi.py#L828). The default return code for response object is [200](https://github.com/openstack/manila/blob/master/manila/api/openstack/wsgi.py#L369).

mvalkon:
@ehdou please replace #114 with #130 in the PR description.

openstack/sharedfilesystems/v2/sharenetworks/requests.go
```diff
@@ -0,0 +1,46 @@
...
+// these parameters, see the ShareNetwork object.
+type CreateOpts struct {
+	// The UUID of the Neutron network to set up for share servers
+	NeutronNetID string `json:"neutron_net_id,omitempty"`
+	// The UUID of the Neutron subnet to set up for share servers
+	NeutronSubnetID string `json:"neutron_subnet_id,omitempty"`
+	// The UUID of the nova network to set up for share servers
+	NovaNetID string `json:"nova_net_id,omitempty"`
+	// The share network name
+	Name string `json:"name"`
```
jrperritt: `,omitempty`? Also for `Description` (below).

jrperritt: Actually, hold off on that. Let's merge this one and fix that later.

jrperritt:
+2

PR #122, opened 2016-10-13
Feature/filestorage sharenetworks delete

ehdou:
Add support for Share Network Delete operation
For #130

PR #124, opened 2016-10-13
Feature/filestorage sharenetworks list

ehdou:
Add support for Share Network List
For #130

### DIFF
--- a/acceptance/openstack/sharedfilesystems/v2/sharenetworks_test.go
+++ b/acceptance/openstack/sharedfilesystems/v2/sharenetworks_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/sharenetworks"
 )
 
 func TestShareNetworkCreateDestroy(t *testing.T) {
@@ -20,4 +21,67 @@ func TestShareNetworkCreateDestroy(t *testing.T) {
 	PrintShareNetwork(t, shareNetwork)
 
 	defer DeleteShareNetwork(t, client, shareNetwork)
+}
+
+func TestShareNetworkList(t *testing.T) {
+	client, err := clients.NewSharedFileSystemV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create a shared file system client: %v", err)
+	}
+
+	allPages, err := sharenetworks.List(client, sharenetworks.ListOpts{}).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to retrieve share networks: %v", err)
+	}
+
+	allShareNetworks, err := sharenetworks.ExtractShareNetworks(allPages)
+	if err != nil {
+		t.Fatalf("Unable to extract share networks: %v", err)
+	}
+
+	for _, shareNetwork := range allShareNetworks {
+		PrintShareNetwork(t, &shareNetwork)
+	}
+}
+
+// The test creates 2 shared networks and verifies that only the one(s) with
+// a particular name are being listed
+func TestShareNetworkListFiltering(t *testing.T) {
+	client, err := clients.NewSharedFileSystemV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create a shared file system client: %v", err)
+	}
+
+	shareNetwork, err := CreateShareNetwork(t, client)
+	if err != nil {
+		t.Fatalf("Unable to create share network: %v", err)
+	}
+	defer DeleteShareNetwork(t, client, shareNetwork)
+
+	shareNetwork, err = CreateShareNetwork(t, client)
+	if err != nil {
+		t.Fatalf("Unable to create share network: %v", err)
+	}
+	defer DeleteShareNetwork(t, client, shareNetwork)
+
+	options := sharenetworks.ListOpts{
+		Name: shareNetwork.Name,
+	}
+
+	allPages, err := sharenetworks.List(client, options).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to retrieve share networks: %v", err)
+	}
+
+	allShareNetworks, err := sharenetworks.ExtractShareNetworks(allPages)
+	if err != nil {
+		t.Fatalf("Unable to extract share networks: %v", err)
+	}
+
+	for _, listedShareNetwork := range allShareNetworks {
+		if listedShareNetwork.Name != shareNetwork.Name {
+			t.Fatalf("The name of the share network was expected to be %s", shareNetwork.Name)
+		}
+		PrintShareNetwork(t, &listedShareNetwork)
+	}
 }
--- a/acceptance/openstack/sharedfilesystems/v2/sharenetworks_test.go
+++ b/acceptance/openstack/sharedfilesystems/v2/sharenetworks_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/sharenetworks"
+	"github.com/gophercloud/gophercloud/pagination"
 )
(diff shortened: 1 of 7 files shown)


ehdou:
[List function](https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/api/v1/share_networks.py#L108) in Manila code.

ehdou:
@jrperritt This PR is ready for review. It's based on [PR#122](https://github.com/gophercloud/gophercloud/pull/122) and add support for list

mvalkon:
@ehdou please replace #114 with #130 in the PR description.

openstack/sharedfilesystems/v2/sharenetworks/requests.go
```diff
@@ -50,3 +53,60 @@ func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
...
+	// The nova network ID
+	NovaNetID string `q:"nova_net_id"`
+	// The network type. A valid value is VLAN, VXLAN, GRE or flat
+	NetworkType string `q:"network_type"`
+	// The Share Network name
+	Name string `q:"name"`
+	// The Share Network description
+	Description string `q:"description"`
+	// The Share Network IP version
+	IPVersion int `q:"ip_version"`
```
jrperritt: s/int/gophercloud.IPVersion

openstack/sharedfilesystems/v2/sharenetworks/requests.go
```diff
@@ -50,3 +53,60 @@ func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
...
+	// The Share Network segmentation ID
+	SegmentationID int `q:"segmentation_id"`
+	// List all share networks created after the given date
+	CreatedSince string `q:"created_since"`
```
jrperritt: These should probably be date objects commiserate with the required format.

openstack/sharedfilesystems/v2/sharenetworks/results.go
```diff
@@ -28,15 +31,36 @@ type ShareNetwork struct {
...
+	UpdatedAt gophercloud.JSONRFC3339MilliNoZ `json:"updated_at"`
 }
 
 type commonResult struct {
 	gophercloud.Result
 }
 
+// ShareNetworkPage is a pagination.pager that is returned from a call to the List function.
+type ShareNetworkPage struct {
+	pagination.SinglePageBase
```
jrperritt: I see limit and offset options: https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/api/v1/share_networks.py#L172

Are you sure there is no pagination allowed for this type?

openstack/sharedfilesystems/v2/sharenetworks/urls.go
```diff
@@ -9,3 +9,7 @@ func createURL(c *gophercloud.ServiceClient) string {
 func deleteURL(c *gophercloud.ServiceClient, id string) string {
 	return c.ServiceURL("share-networks", id)
 }
+
+func listURL(c *gophercloud.ServiceClient) string {
+	return c.ServiceURL("share-networks", "detail")
```
jrperritt: It looks like there are separate `list` and `list detail` operations?

openstack/sharedfilesystems/v2/sharenetworks/results.go
```diff
@@ -28,15 +31,36 @@ type ShareNetwork struct {
 	// The Share Network description
 	Description string `json:"description"`
 	// The date and time stamp when the Share Network was created
-	CreatedAt string `json:"created_at"`
+	CreatedAt gophercloud.JSONRFC3339MilliNoZ `json:"created_at"`
```
jrperritt: :+1:

openstack/sharedfilesystems/v2/sharenetworks/requests.go
ehdou: Do you mean using gophercloud.JSONRFC3339MilliNoZ instead of string for the dates ? If so, that doesn't seem to work. BuildQueryString crashes as it is calling isZero() on the value.

openstack/sharedfilesystems/v2/sharenetworks/urls.go
ehdou: There is both operations but I chose to implement only one as we are planning to only use this one. Would you like me to implement the other one, too ?

openstack/sharedfilesystems/v2/sharenetworks/requests.go
ehdou: fixed

ehdou:
@jrperritt Thanks for your comments, I fixed them and rebased this branch

openstack/sharedfilesystems/v2/sharenetworks/urls.go
jrperritt: Not necessarily, but this isn't the `List` URL, right? It's the `List Detail` URL. We should name the operation `ListDetail` (and this function `listDetailURL`) if it's returning the full list response.

openstack/sharedfilesystems/v2/sharenetworks/results.go
```diff
@@ -28,15 +31,36 @@ type ShareNetwork struct {
...
+	UpdatedAt gophercloud.JSONRFC3339MilliNoZ `json:"updated_at"`
 }
 
 type commonResult struct {
 	gophercloud.Result
 }
 
+// ShareNetworkPage is a pagination.pager that is returned from a call to the List function.
+type ShareNetworkPage struct {
+	pagination.LinkedPageBase
```
jrperritt: Why is this a `LinkedPageBase` (i.e. where are the links in the page responses)?

acceptance/openstack/sharedfilesystems/v2/sharenetworks_test.go
```diff
@@ -21,3 +22,66 @@ func TestShareNetworkCreateDestroy(t *testing.T) {
 
 	defer DeleteShareNetwork(t, client, shareNetwork)
 }
+
+func TestShareNetworkList(t *testing.T) {
```
jrperritt: I suspect if you had a pagination acceptance test (instead of calling `AllPages`) with more than 1 page, you'd see if the `LinkedPageBase` is correct (I suspect it isn't).

openstack/sharedfilesystems/v2/sharenetworks/results.go
ehdou: Sorry I probably misunderstood the pagination types. Which type should I use when I have both `limit` and `offset` as query parameters ? I don't see any that would work and I can't find an example in the other services.

jrperritt: It's probably this: https://github.com/gophercloud/gophercloud/blob/master/pagination/marker.go#L19

For that, you'll need to create a `LastMarker` method and a `NextPageURL` that sets `offset` in the query string instead of `marker`.

acceptance/openstack/sharedfilesystems/v2/sharenetworks_test.go
ehdou: I added one now

openstack/sharedfilesystems/v2/sharenetworks/results.go
ehdou: Done

openstack/sharedfilesystems/v2/sharenetworks/urls.go
ehdou: Done

ehdou:
@jrperritt I updated the list function to MarkerPageBase. LastMarker should now return the correct `offset` depending on the values of `offset` and `limit` in the previous query. If pagination is not needed (because `limit` is not present for example), then `offset` is set to maxInt value. I also added an acceptance test for paginaton

jrperritt:
+2

PR #127, opened 2016-10-13
feature/shared file systems: show share details

mvalkon:
This pull request implements [`showing`][1] a [`share`][2]. 

Manila implementation related links:
https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/share/api.py#L1225

For #129  

[1]: http://developer.openstack.org/api-ref/shared-file-systems/?expanded=#show-share-details
[2]: http://docs.openstack.org/admin-guide/shared-file-systems-key-concepts.html#share

PR #120, opened 2016-10-12
feature/shared file systems: delete a share

mvalkon:
This PR implements [`deleting`][1] a Manila [`share`][2].

Manila implementation of delete can be found:
https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/share/api.py#L675

For #129 

[1]: http://developer.openstack.org/api-ref/shared-file-systems/?expanded=delete-share-detail#delete-share
[2]: http://docs.openstack.org/admin-guide/shared-file-systems-key-concepts.html#share

PR #119, opened 2016-10-12
feature/shared file systems: create share

mvalkon:
Adds support for creating a [`share`][1]. 

For #129  

[1]: http://developer.openstack.org/api-ref/shared-file-systems/?expanded=create-share-detail

PR #128, opened 2016-10-13
feature/shared file systems: share acceptance tests

mvalkon:
This PR adds acceptance tests for creating, deleting and getting a [`share`](http://docs.openstack.org/admin-guide/shared-file-systems-key-concepts.html#share).

For #129


mvalkon:
This PR will require a bit more work. A share is dependent (by default, I suppose this is configurable) of a share network. On devstack default installation of Manila with the default nfs plugin, I noticed that a share can only be successfully created if the share network is connected to the _private_ neutron network. I'll have to think about how to handle this dependency in the acceptance tests. Maybe the creation of this share network can be forced in the environment setup script, and then the name or ID of the network can be passed to the test in an environment variable.

jtopjian:
This is just my $0.02, but I think that's a great approach. Besides the standard `OS_USERNAME` env variables and such, there are a few (`OS_EXTGW_ID`, `OS_FLAVOR_ID_RESIZE`, etc) that serve as gateways from the test environment to the acceptance test.

Somewhere toward the end of `keystonev2-lbaasv1.sh`, just add whatever shell logic is required to set up the sfs environment and store the core information in some env variables. Then create corresponding fields in the `AcceptanceTestChoices` struct.

jrperritt:
SGTM

mvalkon:
@jtopjian @jrperritt please take a look when you have time, thanks!

acceptance/openstack/sharedfilesystems/v2/shares.go
```diff
@@ -0,0 +1,63 @@
...
+	if err != nil {
+		t.Fatalf("Unable to delete share %s: %v", share.ID, err)
+	}
+
+	t.Logf("Deleted share: %s", share.ID)
+}
+
+// PrintShare prints some information of the share
+func PrintShare(t *testing.T, share *shares.Share) {
+	asJSON, err := json.MarshalIndent(share, "", " ")
```
jtopjian: I thought about doing this as well. The benefit is that there are no external dependencies, but the downside is that the fields will be converted to their JSON form (ex: `TenantID` to `tenant_id`). Maybe there's a way around this?

mvalkon: @jtopjian I checked out your branch where you're using `kr/pretty`, but did not try it myself. Does the pretty-printing work around this issue? I think printing the responses the way they are currently printed in many acceptance tests is a little cumbersome.

jtopjian: > Does the pretty-printing work around this issue?

It does, yes. I initially tried the same method you're using (`json.MarshalIndent`), but noticed it changes the names to JSON format. That's when I went hunting for an external library. It'd be great if printing a nicely formatted struct was possible without an external dependency.

> I think printing the responses the way they are currently printed in many acceptance tests is a little cumbersome.

I agree 100%. When I initially added all of the `Print<Resource>` functions, it was easier for me to just manually inventory all of the fields. I figured I'd defer a better solution for later.

jtopjian: > Does the pretty-printing work around this issue?

It does, yes. I initially tried the same method you're using (`json.MarshalIndent`), but noticed it changes the names to JSON format. That's when I went hunting for an external library. It'd be great if printing a nicely formatted struct was possible without an external dependency.

> I think printing the responses the way they are currently printed in many acceptance tests is a little cumbersome.

I agree 100%. When I initially added all of the `Print<Resource>` functions, it was easier for me to just manually inventory all of the fields. I figured I'd defer a better solution for later.

Edit: Additionally, if an external dependency is the only way to go, `kr/pretty` might not be the best choice. I also looked at `go-spew`, but that seemed to be _very_ verbose to the point where too much meta information got in the way.

I don't have a strong opinion about a solution, only that I agree with you that it'd be more maintainable to not manually inventory each field of the resource. :)

mvalkon: @jtopjian @jrperritt I propose we decide on this in a wider scope, and refactor the acceptance tests once we have a solution. If possible, I'd like to merge this PR as it is, unless there more comments. Once we have an idea on what to do with the printing, I can commit to opening a PR which will address that. I can also take a look at some options wrt. an external lib. Thanks a lot for comments so far!

jtopjian: Sounds good - absolutely works for me.

jrperritt:
@mvalkon This PR should only have changes to acceptance test files and scripts, right? Can you update it accordingly?

mvalkon:
@jrperritt this is based on #127 in order for the tests to actually pass. The commits from 6855ac38 onwards are the ones belonging to this PR, and should only contain acceptance test related changes. Or should I rebase this on top of master? I did not want to do this as I wasn't sure if the acceptance tests are run in a ci-system. Sorry for the confusion

jrperritt:
> Or should I rebase this on top of master? ...Sorry for the confusion

The confusion is my fault; let me try to explain: Basing one PR on another is perfectly fine, but (to me) it isn't actually ready for review until it's been rebased or merged with whatever PR it depends upon. That way the reviewer isn't going through the same files on every PR. So when you're asking for a review, I'm assuming that the PR can be merged as-is. Again, my fault for not clarifying.

> I wasn't sure if the acceptance tests are run in a ci-system

Currently, they are not.

mvalkon:
@jrperritt I'm still a little confused, but I think we can move forwards :)

> Basing one PR on another is perfectly fine, but (to me) it isn't actually ready for review until it's been rebased

This PR is rebased on top of #127. So from the commit history, you should be able to see the commits related only for this PR. 

> or merged with whatever PR it depends upon. 

In this case we would have to merge PRs #119 #120 #127 and this one into a massive PR, which I don't think is a good idea. How about I will tag each PR belonging to #129 as PENDING, and you can comment & merge in order? I can then rebase each PR after merged.

mvalkon:
@jrperritt this should be good for review / merge now.

jrperritt:
@jtopjian for review

mvalkon:
@jtopjian @jrperritt bump

mvalkon:
@jrperritt @jtopjian pinging for review again

jtopjian:
@mvalkon I say definitely go forward with whatever works for the acceptance tests. I've been meaning to revisit the "pretty print" stuff, but have just procrastinated on it. :(

mvalkon:
@jtopjian i've also been out of time for this, hopefully getting around to it during december. @jrperritt could this be merged?

openstack/sharedfilesystems/v2/shares/util.go
```diff
@@ -0,0 +1,22 @@
+package shares
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// WaitForStatus will continually poll the resource, checking for a particular
+// status. It will do this for the amount of seconds defined.
+func WaitForStatus(c *gophercloud.ServiceClient, id, status string, secs int) error {
```
jrperritt: This is an example of what I was referring to in https://github.com/gophercloud/gophercloud/pull/128#issuecomment-254581988. I'm not sure I want to have a `WaitForStatus` function in every service. Putting this as an unexported function somewhere in under the `acceptance` directory would be fine though.

mvalkon: Ah I see, I was following someones lead here. I'll move this.

mvalkon: @jrperritt I decided to move this under the acceptance tests of shared file systems, what do you think? It's not optimal but generalizing requires more work which would probably warrant it's own PR.

jrperritt: That is fine

Issue #131, opened 2016-10-14
Shared file system: Security Services


ehdou:
This issue is raised to track the development of the security services in the shared file systems api v2
- [x] show security services with details (#135)
- [x] list security services with details (#134)
- [x] create a security service (#132)
- [x] delete a security service (#133)
- [x] update a security service (#136)

This issue is part of [#114](https://github.com/gophercloud/gophercloud/issues/114)

jtopjian:
🎉

PR #126, opened 2016-10-13
Feature/filestorage sharenetworks update

ehdou:
Add support for Share Network Update
For #130

PR #132, opened 2016-10-14
Feature/filestorage securityservices create

ehdou:
Add support for Security Service Create
For #131

### DIFF
--- a/openstack/sharedfilesystems/v2/securityservices/requests.go
+++ b/openstack/sharedfilesystems/v2/securityservices/requests.go
@@ -1,0 +1,52 @@
+package securityservices
+
+import "github.com/gophercloud/gophercloud"
+
+// CreateOptsBuilder allows extensions to add additional parameters to the
+// Create request.
+type CreateOptsBuilder interface {
+	ToSecurityServiceCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts contains options for creating a SecurityService. This object is
+// passed to the securityservices.Create function. For more information about
+// these parameters, see the SecurityService object.
+type CreateOpts struct {
+	// The security service name
+	Name string `json:"name"`
+	// The security service description
+	Description string `json:"description,omitempty"`
+	// The security service type. A valid value is ldap, kerberos, or active_directory
+	Type string `json:"type" required:"true"`
+	// The DNS IP address that is used inside the tenant network
+	DNSIP string `json:"dns_ip,omitempty"`
+	// The security service user or group name that is used by the tenant
+	User string `json:"user,omitempty"`
+	// The user password, if you specify a user
+	Password string `json:"password,omitempty"`
+	// The security service domain
+	Domain string `json:"domain,omitempty"`
+	// The security service host name or IP address
+	Server string `json:"server,omitempty"`
+}
+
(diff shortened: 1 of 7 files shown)


ehdou:
Here is the [DB](https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/db/sqlalchemy/models.py#L728) representation of security services, the [view](https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/api/views/security_service.py) and the [create function](https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/api/v1/security_service.py#L191)

ehdou:
@jrperritt This PR is ready for review. It's based on [PR#126](https://github.com/gophercloud/gophercloud/pull/126) and adds support for security service Create

ehdou:
@jrperritt This PR is ready for review

openstack/sharedfilesystems/v2/securityservices/requests.go
```diff
@@ -0,0 +1,52 @@
...
+
+// CreateOpts contains options for creating a SecurityService. This object is
+// passed to the securityservices.Create function. For more information about
+// these parameters, see the SecurityService object.
+type CreateOpts struct {
+	// The security service name
+	Name string `json:"name"`
+	// The security service description
+	Description string `json:"description,omitempty"`
+	// The security service type. A valid value is ldap, kerberos, or active_directory
```
jrperritt: Are these the only [3] types allowed? If so, we should probably create a type for this that has those three as `const`s of that type.

openstack/sharedfilesystems/v2/securityservices/requests.go
```diff
@@ -0,0 +1,52 @@
...
+type CreateOptsBuilder interface {
+	ToSecurityServiceCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts contains options for creating a SecurityService. This object is
+// passed to the securityservices.Create function. For more information about
+// these parameters, see the SecurityService object.
+type CreateOpts struct {
+	// The security service name
+	Name string `json:"name"`
```
jrperritt: Can the name by empty or should this have an `omitempty` or `required:true` tag?

openstack/sharedfilesystems/v2/securityservices/requests.go
```diff
@@ -0,0 +1,52 @@
...
+	// The security service description
+	Description string `json:"description,omitempty"`
+	// The security service type. A valid value is ldap, kerberos, or active_directory
+	Type string `json:"type" required:"true"`
```
jrperritt: Let's move required fields to the top of the struct.

openstack/sharedfilesystems/v2/securityservices/requests.go
ehdou: The name is optional, I added `omitempty`

openstack/sharedfilesystems/v2/securityservices/requests.go
ehdou: good point, done

ehdou:
@jrperritt This PR is ready for re-review

Issue #129, opened 2016-10-13
Shared file systems: Shares


mvalkon:
This issue is raised to track the development of the [`shares`](http://developer.openstack.org/api-ref/shared-file-systems/?expanded=#shares) -actions in the shared file systems api v2
- [x] Create a share (#119)
- [x] Delete a share (#120)
- [x] List shares (#1095)
- [x] Show a share (#127)
- [ ] Manage share
- [x] Update share

This issue is a part of #114

kayrus:
Reference for the `Update share`:
API ref: https://developer.openstack.org/api-ref/shared-file-system/?expanded=update-share-detail#update-share
Manila code: v2 ShareController inherits v1 https://github.com/openstack/manila/blob/4f22d0b19e782800b06deba74e6ec15708e5db33/manila/api/v1/shares.py#L210..L214 -> https://github.com/openstack/manila/blob/4f22d0b19e782800b06deba74e6ec15708e5db33/manila/api/v2/shares.py#L26 -> https://github.com/openstack/manila/blob/4f22d0b19e782800b06deba74e6ec15708e5db33/manila/api/v2/shares.py#L40

pierreprinetti:
I am closing this issue, as the implementation (thanks @kayrus) seems to have mostly covered the current needs. Please open a new issue if something is missing and you need it!

pierreprinetti:
and thanks @mvalkon !

PR #134, opened 2016-10-14
Feature/filestorage securityservices list

ehdou:
Add support for Security Service List
For #131

PR #145, opened 2016-10-17
feature/shared file systems: list shares

mvalkon:
This pull request implements [`listing of shares`](http://developer.openstack.org/api-ref/shared-file-systems/?expanded=list-shares-detail#list-shares) with and without details

Manila implementation details:
https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/api/v1/shares.py#L110
https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/api/v1/shares.py#L153

For #129

### DIFF
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -88,6 +88,10 @@ type ListOptsBuilder interface {
 
 // ListOpts is used to filter the List() call
 type ListOpts struct {
+	// Limit can be used to limit the query results
+	Limit int `q:"limit"`
+	// Offset can be used to define the starting point of share listing
+	Offset int `q:"offset"`
 	// Admin-only option to list shares from all tenants
 	AllTenants bool `q:"tenant_id"`
 	// Name of the share
@@ -97,7 +101,7 @@ type ListOpts struct {
 	// ShareServerID is the UUID of the share server
 	ShareServerID string `q:"share_server_id"`
 	// MetaData is key-value-pairs of custom metadata
-	MetaData map[string]string `q:"metadata"`
+	Metadata map[string]string `q:"metadata"`
 	// ShareTypeID is the UUID of the share type
 	ShareTypeID string `q:"share_type_id"`
 	// SortKey defines the sorting key
@@ -111,7 +115,7 @@ type ListOpts struct {
 	// ProjectID is the UUID of the project in which the share belongs to
 	ProjectID string `q:"project_id"`
 	// IsPublic determines the level of visibility of the share
-	IsPublic bool `q:"is_public"`
+	IsPublic *bool `q:"is_public"`
 	// ConsistencyGroupID is the UUID of the consistency group to which the share belongs to
 	ConsistencyGroupID string `q:"consistency_group_id"`
 }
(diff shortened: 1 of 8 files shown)


mvalkon:
The [api docs][1] are describing both `query` and `body` parameters for listing shares, however I could only find the `query` implementation in manila source. If anyone knows more about the details, feel free to comment.

[1]: http://developer.openstack.org/api-ref/shared-file-systems/?expanded=list-shares-detail#list-shares

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -2,12 +2,13 @@ package shares
...
 import (
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
 )
 
 // Share contains all information associated with an OpenStack Share
 type Share struct {
 	// The availability zone of the share
-	AvailabilityZone string `json:"availability_zone"`
+	AvailabilityZone string `json:"availability_zone,omitempty"`
```
mvalkon: Not sure if this is a good idea to have each of the fields as `omitempty`. It allows us to re-use the type for short and long listing though. Open to suggestions on better alternatives

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -76,6 +77,20 @@ func (r commonResult) Extract() (*Share, error) {
 	return s.Share, err
 }
 
+// SharePage is a pagination.pager that is returned from a call to the List func
+type SharePage struct {
+	pagination.SinglePageBase
```
mvalkon: Based on the discussion in #134 I'm starting to question if `SingleBasePage` should be a `MarkerBasePage`. Given that in the Manila API we support `limit` and `offset`, maybe it should be? 

I have a naive acceptance test for pagination calling `EachPage()` for shares, which seems to work with `SinglePageBase`, but I'm not confident that the test would reveal a problem.

mvalkon:
@jrperritt do you have any comments regarding this one while #128 is waiting for a review?

mvalkon:
@jrperritt this one should be ready for review.

jrperritt:
Maybe I'm missing something, but I only see 2 files changed, and they're acceptance tests. Nothing wrong with that, but are the title and description correct?

mvalkon:
@jrperritt you're correct. I've apparently pushed the wrong local branch. I'll have to rebase the correct one, re-run the tests and push again. I'll tag this as WIP until I'm sure it works.

mvalkon:
@jrperritt finally managed to setup acceptance test environment and re-ran tests. Everything passes, please review.

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -79,3 +80,67 @@ func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
...
+}
+
+// ListOpts is used to filter the List() call
+type ListOpts struct {
+	// Limit can be used to limit the query results
+	Limit int `q:"limit"`
+	// Offset can be used to define the starting point of share listing
+	Offset int `q:"offset"`
+	// Admin-only option to list shares from all tenants
+	AllTenants bool `q:"tenant_id"`
```
jtopjian: Should this be `q:"all_tenants"` ?

mvalkon: Most likely should, I will re-check this PR asap.

jtopjian:
This was implemented in #1095

Issue #140, opened 2016-10-17
Shared file system: Share Types


ehdou:
This issue is raised to track the development of the share types in the shared file systems api v2
- [x] List share types (#143)
- [x] List default share types (#144)
- [x] List extra specs for a share type (#146)
- [x] Create a share type (#141)
- [x] Show share type access with details (#149)
- [x] Set an extra spec for a share type (#147)
- [x] Unset an extra spec for a share type (#148)
- [x] Add access for a share type (#150)
- [x] Remove access for a share type (#151)
- [x] Delete a share type (#142)

This issue is part of #114

jrperritt:
To make it easier to follow, can you put the PR next to its corresponding check box in the description like I've started doing?

ehdou:
Of course, I'll do it now

ehdou:
@mvalkon Can you do the same for the issues you have opened ?

jtopjian:
🎉

PR #144, opened 2016-10-17
Feature/filestorage sharetype getdefault

ehdou:
Add support for Share Type Get Default
For #140

PR #146, opened 2016-10-18
Feature/filestorage sharetype getextraspecs

ehdou:
Add support for Share Type Get Extra Specs
For #140

### DIFF
--- a/openstack/sharedfilesystems/v2/sharetypes/results.go
+++ b/openstack/sharedfilesystems/v2/sharetypes/results.go
@@ -68,3 +68,28 @@ func ExtractShareTypes(r pagination.Page) ([]ShareType, error) {
 type GetDefaultResult struct {
 	commonResult
 }
+
+// ExtraSpecs contains all the information associated with extra specifications
+// for an Openstack ShareType.
+type ExtraSpecs struct {
+	Specs map[string]interface{}
+}
+
+type extraSpecsResult struct {
+	gophercloud.Result
+}
+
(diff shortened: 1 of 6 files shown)


ehdou:
[Get extra specs function](https://github.com/openstack/manila/blob/16c522bf5e5c3d37018caef11dd7852462193327/manila/api/v1/share_types_extra_specs.py#L33) for share type

ehdou:
@jrperritt This PR is ready for review. It's based on [PR#144](https://github.com/gophercloud/gophercloud/pull/144) and adds support for share type Get Extra Specs

ehdou:
@jrperritt This PR is ready for review

openstack/sharedfilesystems/v2/sharetypes/results.go
```diff
@@ -68,3 +68,28 @@ func ExtractShareTypes(r pagination.Page) ([]ShareType, error) {
 type GetDefaultResult struct {
 	commonResult
 }
+
+// ExtraSpecs contains all the information associated with extra specifications
+// for an Openstack ShareType.
+type ExtraSpecs struct {
```
jrperritt: If `ExtraSpecs` is just a map with [varying] key-value pairs, we don't need to wrap it in a `struct`:
```go
type ExtraSpecs map[string]interface{}
```

ehdou:
@jrperritt This PR is ready for re-review

jrperritt:
+2

PR #155, opened 2016-10-20
Bugfix: use gophercloud.JSONRFC3339NoZ in db/v1 APIs

fmy:
db/v1 APIs return time values with the "2006-01-02T15:04:05" format.
https://github.com/openstack/trove/blob/129fac7d5374e18a428afa1b5c0259743677222e/api-ref/source/samples/db-instance-status-detail-response.json

But trove api documentation is different.
http://developer.openstack.org/api-ref-database-v1.html#Database_Instances

### DIFF
--- a/openstack/db/v1/configurations/results.go
+++ b/openstack/db/v1/configurations/results.go
@@ -1,19 +1,17 @@
 package configurations
 
 import (
-	"time"
-
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
 
 // Config represents a configuration group API resource.
 type Config struct {
-	Created              time.Time `json:"created"`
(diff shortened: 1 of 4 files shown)


jtopjian:
@fmy Ugh, time formats :)

I'm inclined to agree with this change since there have been other occurrences, but doing a general search of "created" on the Trove repo shows both the Z and non-Z format being used. It _seems_ like the Z-format is being used in Rackspace (for example, [here](https://github.com/openstack/trove/blob/4bedc8b6e0fd462a7811baa7efcc9f9a0cb5859b/integration/apidocs/src/resources/samples/db-gs-create-database-instance-response.json)), though maybe that's just a coincidence. Ultimately we'll want to find the piece of actual Python code that determines the format.

Can you open an Issue for further discussion on this?

PR #158, opened 2016-10-21
Use json.Number for flavorId

fmy:
flavorId is defined as "xsd:string".
But it must be an integer value.
json.Number is suitable for this situation.
http://developer.openstack.org/api-ref-database-v1.html#flavors

https://bugs.launchpad.net/trove/+bug/1165077

### DIFF
--- a/openstack/db/v1/flavors/results.go
+++ b/openstack/db/v1/flavors/results.go
@@ -1,6 +1,7 @@
 package flavors
 
 import (
+	"encoding/json"
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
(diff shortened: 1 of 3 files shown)


jtopjian:
@fmy Per the [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md), there needs to be a corresponding issue opened to go with the PR. This can separate the discussion of the PR vs the code review.

I'm not too familiar with Trove, but I was able to find [this](https://github.com/openstack/trove/commit/21fee8c3c554ba158c2bbe977131522a204c22cc) commit which converts Flavor ID to a string. This commit happened after the bug report you posted and it's the latest commit in the `trove/flavor` and `trove/db` directories, so I'm inclined to think that it's the current design.

Additionally, if a Trove Flavor ID is a reference to a Nova/Compute Flavor ID, it has to be a string.

If this is incorrect, can you open an issue for further discussion?

Issue #161, opened 2016-10-23
FWaaS Rule - Allow Any Protocol


jtopjian:
@jrperritt quick side note: I just switched Terraform over to gophercloud/gophercloud. Thank you for your help over the past few months! I'm now going through the backlog of Terraform issues, some of which require upstream fixes.

FWaaS Rules should allow a protocol of "any", which is achieved by setting `protocol` to `null`.

Here's a list of acceptable protocols, the first being `None`: https://github.com/openstack/neutron-fwaas/blob/905c798db15e0a02bffe2014013be99f9943f56f/neutron_fwaas/extensions/firewall.py#L170-L172

Here's the output of the Neutron CLI client setting `--protocol any`:

``` json
{
    "firewall_rule": {
        "action": "allow",
        "destination_ip_address": "192.168.1.2",
        "enabled": "true",
        "ip_version": 4,
        "name": "foo",
        "protocol": null,
        "source_ip_address": "192.168.1.1"
    }
}
```

jrperritt:
closed by #162

PR #162, opened 2016-10-23
Allow Any Protocol in FWaaS Rules

jtopjian:
This commit enables FWaaS Rules to allow any protocols. It does this
by creating constants for each valid type of protocol, and upon the
type of ProtocolAny being used, the request's protocol parameter will
be null.

For #161


jtopjian:
@jrperritt ping for review

Rather than the pointer-to-string method used in Subnets, I thought I'd try keeping everything more simple. Since there's a finite number of protocols available, they can all be defined as constants. This enables `ProtocolAny` to have any value, and upon triggering that value, set the request value as `null`. Let me know if this sounds OK.

Looks like coverage is failing due to not checking [this](https://github.com/jtopjian/gophercloud/blob/2aaaca3496bdaddc0b25d8b64593da3be67ee4b4/openstack/networking/v2/extensions/fwaas/rules/requests.go#L115-L117)

jrperritt:
> Let me know if this sounds OK.

I think so.

+2

PR #165, opened 2016-10-23
Identity v3 Projects Create

jtopjian:
For #157 
Depends on #164

Issue #157, opened 2016-10-20
missing support for keystone project CRUD


kaccardi:
gophercloud lacks support for keystone projects.

- [x] list projects (#163)
- [x] get project (#164)
- [x] create project (#165)
- [x] delete project (#166)
- [x] update project (#167)

jtopjian:
For listing projects, I think these are the relevant areas in code:
* https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L255-L265
* https://github.com/openstack/keystone/blob/5ed77daf0a36b448ccf30ec97a25e4d1e31fd86b/keystone/common/controller.py#L606

#163 does not contain support for the comparatives that the second link describes. I think this is probably acceptable for now, though.

jrperritt:
@jtopjian  Thank you for taking the time to research and read the source code for these operations.

jtopjian:
No problem at all! Thanks for the help with reviews.

PR #166, opened 2016-10-23
Identity v3 Projects Delete

jtopjian:
For #157 
Depends on #165


jtopjian:
@jrperritt This should be ready for review.

PR #167, opened 2016-10-24
Identity v3 Projects Update

jtopjian:
For #157 
Depends on #166


jtopjian:
@jrperritt This is ready for review

jrperritt:
Looks good

PR #169, opened 2016-10-24
Expose server type on baymodels

carolynvs:
Server type specifies if a bay (cluster) is vm or lxc.


carolynvs:
@jrperritt This is ready for review

PR #170, opened 2016-10-31
computer/v2/servers: Check if opts.UserData is already Base64 Encoded

fatmcgav:
Currently, the `UserData` field is always Base64 encoded, even if it could already be Base64 encoded. 

This improves the behaviour by checking if the current value is already Base64 encoded, and skips the encoding if it is. 
Otherwise, it will Base64 encode before submitting the request.

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -145,7 +145,7 @@ type CreateOpts struct {
 	SecurityGroups []string `json:"-"`
 
 	// UserData contains configuration information or scripts to use upon launch.
-	// Create will base64-encode it for you.
+	// Create will base64-encode it for you, if it isn't already.
 	UserData []byte `json:"-"`
 
 	// AvailabilityZone in which to launch the server.
@@ -190,8 +190,16 @@ func (opts CreateOpts) ToServerCreateMap() (map[string]interface{}, error) {
 	}
 
 	if opts.UserData != nil {
-		encoded := base64.StdEncoding.EncodeToString(opts.UserData)
-		b["user_data"] = &encoded
+		// Check if we've already got a Base64 encoded string; dont double-encode
+		_, base64DecodeError := base64.StdEncoding.DecodeString(string(opts.UserData))
+
+		if base64DecodeError == nil {
+			preEncoded := string(opts.UserData)
+			b["user_data"] = &preEncoded
+		} else {
+			encoded := base64.StdEncoding.EncodeToString(opts.UserData)
+			b["user_data"] = &encoded
+		}
 	}
 
 	if len(opts.SecurityGroups) > 0 {
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -190,16 +190,13 @@ func (opts CreateOpts) ToServerCreateMap() (map[string]interface{}, error) {
 	}
 
 	if opts.UserData != nil {
-		// Check if we've already got a Base64 encoded string; dont double-encode
-		_, base64DecodeError := base64.StdEncoding.DecodeString(string(opts.UserData))
-
-		if base64DecodeError == nil {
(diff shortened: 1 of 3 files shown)


fatmcgav:
@jtopjian I spotted this one whilst playing with Terraform's [template_cloudinit_config](https://www.terraform.io/docs/providers/template/d/cloudinit_config.html) resources, which Base64 encoded `user_data` by default.

fatmcgav:
@jrperritt Ping for a review on this one please.

jtopjian:
This makes sense to me. I have seen this topic come up in other (non-OpenStack) areas where assumptions are made as to where the encoding is actually done.

jtopjian:
@fatmcgav Oh, sorry, can you open a corresponding issue for this to match the PR? Check the [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md) for more details.

fatmcgav:
@jtopjian Done :)

fatmcgav:
@jtopjian / @jrperritt Any chance of getting this one merged in?

Can the update the vendored version in TF :)

jtopjian:
@fatmcgav If you're running into this issue specifically with the `template_cloudinit_config` data source, a temporary workaround is to set `base64_encode` to false. This is what I've been doing.

openstack/compute/v2/servers/requests.go
```diff
@@ -190,8 +190,16 @@ func (opts CreateOpts) ToServerCreateMap() (map[string]interface{}, error) {
...
+		// Check if we've already got a Base64 encoded string; dont double-encode
+		_, base64DecodeError := base64.StdEncoding.DecodeString(string(opts.UserData))
+
+		if base64DecodeError == nil {
+			preEncoded := string(opts.UserData)
+			b["user_data"] = &preEncoded
+		} else {
+			encoded := base64.StdEncoding.EncodeToString(opts.UserData)
+			b["user_data"] = &encoded
+		}
```
jrperritt:
```go
if opts.UserData != nil {
  var userData string
  if _, err := base64.StdEncoding.DecodeString(string(opts.UserData)); err != nil {
    userData = base64.StdEncoding.EncodeToString(opts.UserData)
  } else {
    userData = string(opts.UserData)
  }
  b["user_data"] = &userData
}
```

fatmcgav: @jrperritt Looks like a nice optimization... Have made the changes.

jrperritt:
Also, I think we should have a unit test for this. You can just add the `UserData` field on to one of the existing unit tests in `requests_test.go` and check that it's what it should be with `TestJSONRequest` in `fixtures.go`. [This test](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/testing/fixtures.go#L479) would be fine.

fatmcgav:
Sure, I'll give it a go... 🙃

fatmcgav:
@jrperritt One quick q, if I may... 

How would you suggest I handle the difference in request/response field names? 
That is, the request field is `user_data` [1], whereas the response field is `OS-EXT-SERV-ATTR:user_data` [2].

Cheers
Gav

[1] http://developer.openstack.org/api-ref/compute/?expanded=create-server-detail
[2] http://developer.openstack.org/api-ref/compute/?expanded=show-server-details-detail

jrperritt:
We aren't concerned with the response field here, only the request field. Populate the `UserData` field of `servers.CreateOpts` and check in `TestJSONRequest` that it is what it should be.

fatmcgav:
@jrperritt Cheers for the pointer, I was overthinking the test... 

Have added 21a44e132331d6d46ab4187411c85ced6b1a9963 - Let me know your thoughts. 

Cheers
Gav

jrperritt:
Looks great. Thank you.

fatmcgav:
:) Cheers

Issue #172, opened 2016-11-01
Designate - Zone Read


grahamhayes:
Add support for OpenStack DNS - Designate Zone Reads (List + Get)

jtopjian:
@grahamhayes I wanted to touch base about the Designate work. If you're pressed for time, I would be happy to (attempt to) take over. This is something I'd really like to see in Gophercloud. :)

jrperritt:
@jtopjian if you're still interested in implementing this, please go ahead

jtopjian:
On it.

jrperritt:
closed by #271

Issue #75, opened 2016-08-30
making all times `time.Time`


jrperritt:
Currently some operation results use `string` while others use Gophercloud custom types (e.g. `gophercloud.JSONUnix`). Creating an `Unmarshal` method on each result and doing the conversion there. One benefit would be not having to export all the custom time types.

jrperritt:
I'm going to start this one tomorrow.

jrperritt:
closed by #190

PR #173, opened 2016-11-01
Add Designate Zone List / Get Support

grahamhayes:
For #172

https://github.com/openstack/designate/blob/stable/newton/designate/objects/adapters/api_v2/zone.py#L23-L55

### DIFF
--- a/openstack/dns/v2/zones/results.go
+++ b/openstack/dns/v2/zones/results.go
@@ -1,0 +1,63 @@
+package zones
+
+import (
+	"encoding/json"
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// Extract interprets a GetResult, CreateResult or UpdateResult as a concrete Zone.
+// An error is returned if the original call or the extraction failed.
+func (r commonResult) Extract() (Zone, error) {
+	var s Zone
+	err := r.ExtractInto(&s)
+	return s, err
+}
+
+// GetResult is the deferred result of a Get call.
+type GetResult struct {
+	commonResult
+}
+
+// ZonePage is a single page of Zonje results.
+type ZonePage struct {
+	pagination.LinkedPageBase
+}
+
+// IsEmpty returns true if the page contains no results.
+func (p ZonePage) IsEmpty() (bool, error) {
+	services, err := ExtractZones(p)
+	return len(services) == 0, err
+}
+
+// ExtractZones extracts a slice of Services from a Collection acquired from List.
+func ExtractZones(r pagination.Page) ([]Zone, error) {
+	var s struct {
+		Zones []Zone `json:"zones"`
+	}
+	err := (r.(ZonePage)).ExtractInto(&s)
+	return s.Zones, err
+}
+
+type Zone struct {
+	ID            string                          `json:"id,omitempty"`
+	PoolID        string                          `json:"pool_id,omitempty"`
+	ProjectID     string                          `json:"project_id,omitempty"`
+	Name          string                          `json:"name,omitempty"`
+	Email         string                          `json:"email,omitempty"`
+	TTL           int                             `json:"ttl,omitempty"`
+	Serial        json.Number                     `json:"serial,omitempty"`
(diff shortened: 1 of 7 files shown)


grahamhayes:
Not sure what I can do for coveralls above - none of the service clients have any tests ....

jrperritt:
Akk. Sorry, the merge conflict is my fault.

> Not sure what I can do for coveralls above - none of the service clients have any tests ....

I'm not concerned about the service client tests.

openstack/dns/v2/zones/results.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// Extract interprets a GetResult, CreateResult or UpdateResult as a concrete Zone.
+// An error is returned if the original call or the extraction failed.
+func (r commonResult) Extract() (Zone, error) {
```
jrperritt: `*Zone`

openstack/dns/v2/zones/results.go
```diff
@@ -0,0 +1,63 @@
...
+	var s Zone
```
jrperritt: `*Zone`

openstack/dns/v2/zones/results.go
```diff
@@ -0,0 +1,63 @@
...
+// ExtractZones extracts a slice of Services from a Collection acquired from List.
+func ExtractZones(r pagination.Page) ([]Zone, error) {
+	var s struct {
+		Zones []Zone `json:"zones"`
+	}
+	err := (r.(ZonePage)).ExtractInto(&s)
+	return s.Zones, err
+}
+
+type Zone struct {
```
jrperritt: +`Description` field

openstack/dns/v2/zones/results.go
```diff
@@ -0,0 +1,63 @@
...
+	ID            string                          `json:"id,omitempty"`
+	PoolID        string                          `json:"pool_id,omitempty"`
+	ProjectID     string                          `json:"project_id,omitempty"`
+	Name          string                          `json:"name,omitempty"`
+	Email         string                          `json:"email,omitempty"`
+	TTL           int                             `json:"ttl,omitempty"`
+	Serial        json.Number                     `json:"serial,omitempty"`
```
jrperritt: Because `json.Number` and the different time formats (like `JSONRFC3339MilliNoZ`) are kind of awkward to use, we're trying to move toward only having built-in data types in response fields (see #75). To that end, perhaps we could implement `json.Unmarshaler` (see [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/imageservice/v2/images/results.go#L82) for an example for both `json.Number` and `time.Time`) for `Zone`?

openstack/dns/v2/zones/results.go
```diff
@@ -0,0 +1,63 @@
...
+	Status        string                          `json:"status,omitempty"`
+	Action        string                          `json:"action,omitempty"`
+	Version       int                             `json:"version,omitempty"`
+	Attributes    map[string]string               `json:"attributes,omitempty"`
+	Type          string                          `json:"type,omitempty"`
+	Masters       []string                        `json:"masters,omitempty"`
+	CreatedAt     gophercloud.JSONRFC3339MilliNoZ `json:"created_at,omitempty"`
+	UpdatedAt     gophercloud.JSONRFC3339MilliNoZ `json:"updated_at,omitempty"`
+	TransferredAt gophercloud.JSONRFC3339MilliNoZ `json:"transferred_at,omitempty"`
+}
```
jrperritt: Is there a link that describes the uses of these fields and their types?

grahamhayes: http://developer.openstack.org/api-ref/dns/?expanded=create-zone-detail#create-zone would be the best resource for the time fields

openstack/dns/v2/zones/results.go
grahamhayes: sure, I can do that.

I was having some issues with the serial number, as gophercloud seems to marshal and unmarshal - which causes the serial number to swap to scientific notation, which is then parsed as a string. `json.Number` was the only thing that would work for me.

jrperritt:
closed in favor of #271

Issue #57, opened 2016-08-15
Designate - Recordset Get / List


grahamhayes:

jtopjian:
Resolved by #279

PR #175, opened 2016-11-01
[wip] Designate Recordset Get / List Support

grahamhayes:
For #57 

https://github.com/openstack/designate/blob/stable/newton/designate/objects/adapters/api_v2/recordset.py#L24-L54

WIP as it depends on #173 (commit 2c31a49) for base client setup


jrperritt:
closing in favor of #279

Issue #168, opened 2016-10-24
BuildQueryString is not allowing parameters with a value of false


jtopjian:
This is due to the `isZero` function doing a [recursive check](https://github.com/gophercloud/gophercloud/blob/master/params.go#L235) when the field is a pointer.

Is this expected behavior? If so, how can something like `projects?enabled=false` be accomplished?

jrperritt:
`*bool`

jrperritt:
~~*bool~~

Ah, hmm. Let me think about it.

jrperritt:
I suspect `bool` is a rare type for a query string value. Maybe we just handle those individually in the `ToListQuery` function.

jrperritt:
or maybe `isZero` is changed to use the value of the pointer.

jtopjian:
```shell
$ grep -R \*bool ~/go/src/github.com/gophercloud/gophercloud | grep q:
```

It looks like there's a good amount being used, but definitely less than other types.

Whichever solution you think is best. 😄

jrperritt:
I think `isZero` is wrong here. For any pointer, `nil` should return `true`, anything else `false`.

jtopjian:
So just change:

```go
return isZero(v.Elem())
```

to

```go
return false
```

?

If so, I can make a PR for this as well as confirm I see the expected actions in my Keystone Project branches.

jrperritt:
Yes (and appropriate additions to the already-existing unit test for `isZero`). I suspect I wrote it recursively for some reason, but I can't remember why (let this be a documentation lesson for me). If any tests break because of the change, we can consider changing the types causing the issues.

jrperritt:
closed by #181

PR #181, opened 2016-11-04
Update how isZero handles Pointers

jtopjian:
This commit modifies the isZero function so it checks to see if
a pointer is actually set rather than if it points to a zero value.

For #168

### DIFF
--- a/params.go
+++ b/params.go
@@ -307,6 +307,7 @@ func BuildQueryString(opts interface{}) (*url.URL, error) {
 
 				// if the field is set, add it to the slice of query pieces
 				if !isZero(v) {
+				loop:
 					switch v.Kind() {
 					case reflect.String:
 						params.Add(tags[0], v.String())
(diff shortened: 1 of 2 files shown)


jtopjian:
@jrperritt ping for review.

I only accounted for Pointers to Bool in the switch statement. I think that should be sufficient?

params.go
```diff
@@ -314,6 +314,12 @@ func BuildQueryString(opts interface{}) (*url.URL, error) {
 						params.Add(tags[0], strconv.FormatInt(v.Int(), 10))
 					case reflect.Bool:
 						params.Add(tags[0], strconv.FormatBool(v.Bool()))
+					case reflect.Ptr:
+						pv := v.Elem()
+						switch pv.Kind() {
```
jrperritt: This is good, but to get all the `Kind`s without duplicating code, we should be able to move this to the top of the `switch` and add a label:
```go
loop:
  switch v.Kind() {
    case reflect.Ptr:
      v = v.Elem()
      goto loop
    case reflect.String:
    ...
}
```

jtopjian: Oh cool -- thanks!

jtopjian:
@jrperritt ping for review

params.go
```diff
@@ -307,13 +307,17 @@ func BuildQueryString(opts interface{}) (*url.URL, error) {
...
 				if !isZero(v) {
+				loop:
 					switch v.Kind() {
 					case reflect.String:
 						params.Add(tags[0], v.String())
 					case reflect.Int:
 						params.Add(tags[0], strconv.FormatInt(v.Int(), 10))
 					case reflect.Bool:
 						params.Add(tags[0], strconv.FormatBool(v.Bool()))
+					case reflect.Ptr:
```
jtopjian: Ah. Understood. Sorry about that.

jtopjian:
@jrperritt ping for review

jrperritt:
thank you. this looks good.

Issue #182, opened 2016-11-07
Swift Authentication: Ensure StorageURL ends in a Slash


jtopjian:
I ran into an issue where a container was attempted to be created via the URL:

```
http://localhost:8080/v1/AUTH_testtf-test-container
```

When it should have been:

```
http://localhost:8080/v1/AUTH_test/tf-test-container
```

jrperritt:
closed by #183

PR #183, opened 2016-11-07
Ensure StorageURL ends in a slash

jtopjian:
For #182 

Not sure if this is the best way of doing this.

### DIFF
--- a/openstack/objectstorage/v1/swauth/requests.go
+++ b/openstack/objectstorage/v1/swauth/requests.go
@@ -1,8 +1,6 @@
 package swauth
 
-import (
-	"github.com/gophercloud/gophercloud"
-)
+import "github.com/gophercloud/gophercloud"
 
 // AuthOptsBuilder describes struct types that can be accepted by the Auth call.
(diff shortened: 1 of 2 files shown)


jtopjian:
@jrperritt ping for review. Let me know if there's a better way to handle this (some existing means in Gophercloud) as well as if there's a test I can add.

jrperritt:
> Not sure if this is the best way of doing this.

use [`gophercloud.NormalizeURL`](https://github.com/gophercloud/gophercloud/blob/master/util.go#L41)

jtopjian:
@jrperritt Does this look OK?

```go
	swiftClient := &gophercloud.ServiceClient{
		ProviderClient: pc,
		Endpoint:       gophercloud.NormalizeURL(auth.StorageURL),
	}
```

jrperritt:
yeah, that looks fine

jtopjian:
@jrperritt Thank you for the help. Ping for review.

jrperritt:
+2

Issue #184, opened 2016-11-07
Volumes V2 Should Expose Extensions


spjmurray:
Volume lost the ability to parse `os-vol-tenant-attr:tenant_id` in the past.  It'd be nice to have access to this again for metric collection (e.g. we can map volume usage back to specific projects/domains and watch trends).  I'm willing to do the work just need some pointers on the best way to implement this extension.

jrperritt:
Ah, right. Because there can be arbitrarily many extension fields, I thought it would be best to let users add them as-needed. In this case, it really just amounts to a new package for that extension with a `results.go` file with something like:
```go
package tenantattr

type VolumeExt struct {
  TenantID `json:"os-vol-tenant-attr:tenant_id"`
}
```

Then in your application code you'd have something like (untested):
```go
type MyVolume struct {
  volumes.Volume
  tenantattr.VolumeExt
}

var v struct {
  MyVolume `json:"volume"`
}

err := volumes.Get(client, volID).ExtractInto(&v)
fmt.Println(v.TenantID)
```

I should mention this is not in stone; I'm open to suggestions for a better way of implementing extensions [to response objects].

spjmurray:
Got you. Perhaps a nice API to allow extraction of arbitrary extensions is a decent measure. Zero changes to library code and access to all the things from the client side

jrperritt:
It's hard to say without seeing the coded implementation, but if you'd like to make a PR or provide your idea for the implementation here, maybe we can flush out the details.

spjmurray:
Inbound tomorrow

spjmurray:
In the end I just bodged it in client code https://github.com/spjmurray/telegraf/blob/openstack_stats/plugins/inputs/openstack/openstack.go#L246.  Is this the right/clean way to do it?  I guess if so this issue can be closed

jtopjian:
Since this is an [official extension](https://github.com/openstack/cinder/blob/a15459a27777e98599d70392191b0a6f0288ff34/cinder/api/contrib/volume_tenant_attribute.py), I'd like to see this included and I could do the work if your current solution is working.

[...]

PR #186, opened 2016-11-08
tenantattr extension for cinder

jrperritt:
For #184 

code in this PR:

https://github.com/openstack/cinder/blob/a15459a27777e98599d70392191b0a6f0288ff34/cinder/api/contrib/volume_tenant_attribute.py#L27


jrperritt:
Let's see how well this works.

closes #184

Issue #191, opened 2016-11-16
Unmarshal with nova-network Numeric IDs


jtopjian:
By some weird chance, I never tested security groups with nova-network until today. I'm seeing unmarshal errors with numeric secgroup IDs:

```
=== RUN   TestSecGroupsCreate
--- FAIL: TestSecGroupsCreate (2.18s)
	secgroup_test.go:42: Unable to create security group: json: cannot unmarshal number into Go value of type string
```

I see that support for numeric and string formatted IDs for both secgroups and rules [should be supported](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/secgroups/results.go#L10-L12) though, but I'm not seeing where the casting is done. I apologize if I'm overlooking it.

jrperritt:
That comment was probably valid for the old repo when we used a third-party package to handle it. In this repo, that object will need an `UnmarshalJSON` method to handle multiple possible types.

I'd definitely accept a PR for this, but do note that Nova networks and security groups have both been deprecated and will likely soon (perhaps in January 2017, but not sooner) move out of this repo. See #189 for the issue.

jtopjian:
OK, Cool. I was pretty sure this needed the Unmarshal stuff like the others, but the comment threw me off. 

I'll get a PR for this together. I've been meaning to comment on #189 when I got a chance. I'll add some quick thoughts now.

jtopjian:
I also just noticed that nova-network-based floating IPs are also affected.

jrperritt:
closed by #192

jtopjian:
@jrperritt Thanks for merging #192. I also submitted #217 and linked it to this issue since both are related to nova-network support (#192 is for the security groups and #217 is for floating IPs). I can open another Issue for #217 or if you want to re-open this or just leave everything as-is.

jtopjian:
Thank you! :tada:

PR #192, opened 2016-11-17
Handle Unmarshaling Compute Security Group IDs

jtopjian:
For #191


jtopjian:
@jrperritt ping for review

jtopjian:
@jrperritt I think this is better.

jrperritt:
+2

Issue #159, opened 2016-10-22
Z and non-Z time formats in trove api.


fmy:
Create an issue corresponds to PR #155 .

I think some solutions below.
- suffix Z values if not exists
- fix trove api's python code

esalipe:
Created an alternative fix for this that keeps using time.Time in #196. As described in the PR, trove stores the times in `datetime` objects and then generates time values using `isoformat()` which means the returned values are in non-Z format.

fmy:
@esalipe Thanks! I'll close this PR.

jrperritt:
closed by #196

PR #196, opened 2016-11-21
db/instances: expect non-Z time format

esalipe:
Changes parsing of time fields for database instances to match the non-Z format returned by trove.

Trove stores the times as `datetime` objects:
(created) https://github.com/openstack/trove/blob/stable/newton/trove/db/models.py#L35
(updated) https://github.com/openstack/trove/blob/stable/newton/trove/db/models.py#L60

The `datetime` is converted to a string using `isoformat()`, meaning no Z:
https://github.com/openstack/trove/blob/stable/newton/trove/common/base_wsgi.py#L453

This PR solves the same issue as #155 but keeps using `time.Time` in the `Instance` struct.

For #159

### DIFF
--- a/openstack/db/v1/instances/results.go
+++ b/openstack/db/v1/instances/results.go
@@ -57,31 +57,23 @@ type Instance struct {
 	Datastore datastores.DatastorePartial
 }
 
-func (s *Instance) UnmarshalJSON(b []byte) error {
+func (r *Instance) UnmarshalJSON(b []byte) error {
 	type tmp Instance
-	var p *struct {
+	var s struct {
 		tmp
-		Created string `json:"created"`
-		Updated string `json:"updated"`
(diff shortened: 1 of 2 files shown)


openstack/db/v1/instances/results.go
```diff
@@ -56,6 +57,33 @@ type Instance struct {
 	Datastore datastores.DatastorePartial
 }
 
+func (s *Instance) UnmarshalJSON(b []byte) error {
```
jrperritt: I think the logic here is sound, but let's make it conform to the other `UnmarshalJSON` methods. See [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/orchestration/v1/stacks/results.go#L62) for an example.

esalipe: Done.

jrperritt:
+2

Issue #160, opened 2016-10-22
flavorId should be string in trove api.


fmy:
It's an issue corresponds to PR #158 .

> I'm inclined to think that it's the current design.
> Additionally, if a Trove Flavor ID is a reference to a Nova/Compute Flavor ID, it has to be a string.
> If this is incorrect, can you open an issue for further discussion?

Agree 😄 
Flavor ID should be string.
I will change `json.Number` to `string`. It's OK ?

esalipe:
Created an alternative PR #197 to fix this. As described in the PR the types used for returning the flavor ID are different depending on whether the response is for `instances` or for `flavors`. Links to the trove source code are also provided in the PR.

fmy:
Thanks.

jrperritt:
closed by #197

PR #197, opened 2016-11-21
db: fix flavor id parsing

esalipe:
* Trove flavor IDs returned with instances are always strings
https://github.com/openstack/trove/blob/stable/newton/trove/instance/models.py#L212
* Trove flavor IDs returned with flavors are integers or null, there is a
  str_id parameter that contains the ID in string format
https://github.com/openstack/trove/blob/stable/newton/trove/flavor/views.py#L33

Because of the differing return types this PR solves #160 by not referring to `flavor.Flavors` from instance. Instead instance uses its own struct with the ID as type string.

For #160


jrperritt:
LGTM. +2

Issue #198, opened 2016-11-21
DB Instances: IP parameter missing in result


esalipe:
In case a Trove database instance has no hostname, gophercloud returns no information on how to connect to it.

In such a case Trove will return the IP address(es) of the database instance instead:
https://github.com/openstack/trove/blob/stable/newton/trove/instance/views.py#L46

This information could be parsed and returned to the user by gophercloud.

jrperritt:
closed by #199

PR #199, opened 2016-11-21
db/instance: add support for IPs in response body

esalipe:
For #198 

Parses IP addresses returned by Trove when no hostname is available:
https://github.com/openstack/trove/blob/stable/newton/trove/instance/views.py#L46


jrperritt:
+2

Issue #202, opened 2016-11-23
Unable to authenticate the request due to an error: Expected HTTP response code [204 404] when accessing [HEAD https://xxx.xxx.com:35357/v3/auth/tokens], but got 200 instead


dixudx:
There is a bug in current v3 token validation.

> Unable to authenticate the request due to an error: Expected HTTP response code [204 404] when accessing [HEAD https://xxx.xxx.com:35357/v3/auth/tokens], but got 200 instead

Actually the valid return codes are,

```
Normal response codes: 200

Error response codes: 413,405,404,403,401,400,503
```

For more detailed information, refer to [keystone v3 token validate doc](http://developer.openstack.org/api-ref/identity/v3/index.html?expanded=validate-and-show-information-for-token-detail)

jtopjian:
I can confirm this behavior, though I think the correct codes for this should be: 200, 400, 401, 403, 404. While the others are valid codes in Keystone, I'm not sure if they make sense in this context.

Here's a list of the Keystone exception codes: https://github.com/openstack/keystone/blob/master/keystone/exception.py

And here's prior research on the 200 response: https://github.com/gophercloud/gophercloud/issues/39

jtopjian:
Fixed by #203

PR #203, opened 2016-11-23
fix v3 token validate

dixudx:
For #202

### DIFF
--- a/openstack/identity/v3/tokens/requests.go
+++ b/openstack/identity/v3/tokens/requests.go
@@ -182,13 +182,13 @@ func Get(c *gophercloud.ServiceClient, token string) (r GetResult) {
 func Validate(c *gophercloud.ServiceClient, token string) (bool, error) {
 	resp, err := c.Request("HEAD", tokenURL(c), &gophercloud.RequestOpts{
 		MoreHeaders: subjectTokenHeaders(c, token),
-		OkCodes:     []int{204, 404},
+		OkCodes:     []int{200, 204, 400, 401, 403, 404},
 	})
 	if err != nil {
 		return false, err
 	}
 
-	return resp.StatusCode == 204, nil
+	return resp.StatusCode == 200 || resp.StatusCode == 204, nil
 }
 
 // Revoke immediately makes specified token invalid.
--- a/openstack/identity/v3/tokens/testing/requests_test.go
+++ b/openstack/identity/v3/tokens/testing/requests_test.go
@@ -482,7 +482,7 @@ func TestValidateRequestFailure(t *testing.T) {
 func TestValidateRequestError(t *testing.T) {
 	testhelper.SetupHTTP()
 	defer testhelper.TeardownHTTP()
-	client := prepareAuthTokenHandler(t, "HEAD", http.StatusUnauthorized)
+	client := prepareAuthTokenHandler(t, "HEAD", http.StatusMethodNotAllowed)
 
 	_, err := tokens.Validate(&client, "abcdef12345")
 	if err == nil {


openstack/identity/v3/tokens/requests.go
```diff
@@ -182,13 +182,13 @@ func Get(c *gophercloud.ServiceClient, token string) (r GetResult) {
 func Validate(c *gophercloud.ServiceClient, token string) (bool, error) {
 	resp, err := c.Request("HEAD", tokenURL(c), &gophercloud.RequestOpts{
 		MoreHeaders: subjectTokenHeaders(c, token),
-		OkCodes:     []int{204, 404},
+		OkCodes:     []int{200, 400, 401, 403, 404, 405, 413, 503},
```
jrperritt: Is there a compelling reason not to go with @jtopjian 's suggestion of just using 200, 400, 401, 403, and 404?

jtopjian:
@dixudx Just looping back around to this one. From doing previous research on Keystone error codes, I'm pretty sure `200, 400, 401, 403, 404` are all of the valid ones. Although the docs might say otherwise, the ones I ilsted were what I was able to find in the actual implementing code.

Specifically with 503, I'm not sure if that should be considered an acceptable code. That usually means an error on the server side.

Thoughts?

dixudx:
@jtopjian 

Below are some useful links for your reference:

* [keystone v3 api token validate doc](https://developer.openstack.org/api-ref/identity/v3/index.html?expanded=validate-and-show-information-for-token-detail#validate-and-show-information-for-token)
* Related Keystone source codes:
     * [keystone/auth/routers.py#L29](https://github.com/openstack/keystone/blob/master/keystone/auth/routers.py#L29)
     * [keystone/auth/controllers.py#L298-L307](https://github.com/openstack/keystone/blob/master/keystone/auth/controllers.py#L298-L307)
     * [keystone/common/controller.py#L84-L165](https://github.com/openstack/keystone/blob/master/keystone/common/controller.py#L84-L165)
 
Yeah, `503` may be not a quite appropriate ok code. I think `200, 400, 401, 403, 404, 405` would be ok.

jtopjian:
Thanks. Just to give some details for other discussions you and I have had, plus for future work: 

I understand that this PR is very simple and it might seem like nit-picking with regard to exact status codes, but making sure Gophercloud conforms exactly to what the OpenStack API services report is really important as it prevents a _lot_ of future problems. 

The API docs are great resources, but unfortunately they sometimes have errors and sometimes are missing information. A good example of this is [here](https://github.com/gophercloud/gophercloud/pull/288#issuecomment-285588009). The only reference to the `extra` field was in an example code output for using `curl` with Keystone -- it's not mentioned in the API reference, but it's an important piece for supporting Keystone users.

So reasons like this are why there is such a strong requirement to use code as a reference point. And don't get me wrong, digging into the code is _not_ fun. It sucks. But it's well worth the effort to get things right. 

(of course, the best thing to do would be to file bugs about the documentation. Maybe one day I will collect the areas I know of and submit bugs for all of them.)

Back to this specific PR, I think the best area to look for the return codes are in these two files:

* https://github.com/openstack/keystone/blob/master/keystone/exception.py
* https://github.com/openstack/keystone/blob/master/keystone/common/wsgi.py

For example: https://github.com/openstack/keystone/blob/master/keystone/exception.py#L88

I can't find any reference to `413`, `REQUEST_ENTITY_TOO_LARGE`, `405`, `METHOD_NOT_ALLOWED` in all of the Keystone code. Only in documentation... that makes me not want to trust those as actual codes. It's possible I'm wrong, but maybe the best way to go is to not include 405 for now and add it later if it turns out it's actually used. If 405 was added _now_ and it really isn't used, I doubt anyone would take the time to remove it from Gophercloud.

Make sense?

dixudx:
Sounds good. I will modify ok codes in this PR to `200, 400, 401, 403, 404`.

Thanks

jtopjian:
@dixudx The Travis test failed on validating the return code. The unit test is expecting a 204. I did a search for 204 in Keystone, and sure enough, it's a valid code:

https://github.com/openstack/keystone/blob/da472c5f3b9c9d80f54284300a818acc2741c7e9/keystone/common/wsgi.py#L255

I hate to do this, but can you add 204? That should be it, and we can get this merged :)

dixudx:
@jtopjian Already added. Pls get merged.

jtopjian:
👍 

Thanks! Yes, you're right. It was just the unit test that was failing. Thank you for your patience with this one.

Issue #194, opened 2016-11-18
Cinder Scheduler API


spjmurray:
Same as #179 it's be nice to get out an overall statistic for storage back-ends for admins to perform trend analysis and capacity planning.

Specific API to be targeted is http://developer.openstack.org/api-ref/block-storage/v2/index.html?expanded=list-back-end-storage-pools-detail

Will get the boilerplate written and tested while the discussion ensues.

jtopjian:
Given there's a timestamp in the response, it might make sense to wait until #190 is merged before working on this.

spjmurray:
Thanks for the heads up @jtopjian I'll ensure it meets expectation!

spjmurray:
Turns out worrying about timestamps isn't a big issue.  After some digging it seems that the API documentation and reality are about as remote as the Andromeda galaxy.  At a top level you have only got 2 fixed fields:

https://github.com/openstack/cinder/blob/master/cinder/api/views/scheduler_stats.py

Then it is up to the driver to respond with evidently what it likes (e.g. Pure storage vs Ceph RBD - it's all context specific.  However the scheduler relies on a bunch of fields to work, which gives us common ground to go off and importantly extract the usage stats out of:

https://github.com/openstack/cinder/blob/master/doc/source/devref/drivers.rst#volume-stats

jtopjian:
I spent some time reviewing this. The API implementation is dizzying. Like @spjmurray mentioned, it's up to each volume driver to build the dataset to return. Some have a standard definition, but there's enough variance between all to cause concern.

In general, I think [this](https://github.com/openstack/cinder/blob/c6a940b885da369e746601a4a84cfacdbdf76a75/cinder/interface/volume_driver.py#L56) is the best list to go off of. Scanning the various drivers, most implement this set. However, there are a handful that specify even more fields ([example](https://github.com/openstack/cinder/blob/f2c1066f710feb9259be2669e35a57cd102bc33f/cinder/volume/drivers/solidfire.py). I don't have access to these drivers to see if they're somehow filtered. I doubt they are. 

So I think a good implementation would be to define a Result as the set of required and optional fields _and then_ implement an extra field called something like `extra_fields`, which would be a `map[string]interface{}` to catch the remaining. I think this last part could probably wait, both for ease of getting this in place as well as waiting on a decision/agreement of a standard way of implementing this as there are a few other PRs needing to do that.

[...]

PR #205, opened 2016-11-25
Cinder Storage Pools API

spjmurray:
Addresses issue #194.  The documentation is a little hazy links to code in the issue discussion thread.


openstack/blockstorage/storagepools/results.go
```diff
@@ -0,0 +1,108 @@
...
+			FreeCapacityGB interface{} `json:"free_capacity_gb"`
+			TotalCapacityGB interface{} `json:"total_capacity_gb"`
+		} `json:"capabilities"`
+	}
+	err = json.Unmarshal(b, &q)
+	if err != nil {
+		return err
+	}
+
+	// Should be a numeric, "unknown", "infinite"
```
jtopjian: Maybe add unit tests for the unknown and infinite cases?

jtopjian:
One inline comment. In addition, can you run the code through `gofmt`? If you're a vi user, the `vim-go` plugin will take care of that :)

openstack/blockstorage/storagepools/results.go
```diff
@@ -0,0 +1,108 @@
...
+	VendorName string `json:"vendor_name"`
+	VolumeBackendName string `json:"volume_backend_name"`
+}
+
+type StoragePool struct {
+	Name string `json:"name"`
+	Capabilities `json:"capabilities"`
+}
+
+func (s *StoragePool) UnmarshalJSON(b []byte) error {
```
jtopjian: Style nit: this should be `(r *StoragePool)`

openstack/blockstorage/storagepools/results.go
```diff
@@ -0,0 +1,108 @@
...
+	// Unmarshal the generic stuff
+	type tmp StoragePool
+	var p *struct {
```
jtopjian: Style nit: this should be `var s`

spjmurray:
@jtopjian it's almost like you knew adding those unit tests would throw up an error ... good catch :)

jrperritt:
@spjmurray still correctly titled `[wip]` or ready for review?

openstack/blockstorage/storagepools/results.go
```diff
@@ -0,0 +1,122 @@
...
+			r.Capabilities.FreeCapacityGB = q.Capabilities.FreeCapacityGB.(float64)
+		case string:
+			keyword := q.Capabilities.FreeCapacityGB.(string)
+			switch keyword {
+			case "unknown":
+				r.Capabilities.FreeCapacityGB = 0.0
+			case "infinite":
+				r.Capabilities.FreeCapacityGB = math.Inf(1)
+			default:
+				return fmt.Errorf("capabilities.free_capacity_gb: unexpected string %v", keyword)
```
jtopjian: @jrperritt Would it be better to set the value to 0.0 or return an error?

openstack/blockstorage/storagepools/requests.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// Formats a ListOpts into a query string.
+func (opts ListOpts) ToStoragePoolsListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
+// List makes a request against the API to list hypervisors.
+func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) (r ListResult) {
```
jrperritt: We'll want to paginate this, even it's a single-page type.

openstack/blockstorage/storagepools/results.go
```diff
@@ -0,0 +1,122 @@
...
+	DriverVersion     string  `json:"driver_version"`
+	FreeCapacityGB    float64 `json:"-"`
+	StorageProtocol   string  `json:"storage_protocol"`
+	TotalCapacityGB   float64 `json:"-"`
+	VendorName        string  `json:"vendor_name"`
+	VolumeBackendName string  `json:"volume_backend_name"`
+	// Optional Fields
+	ReservedPercentage       int64   `json:"reserved_percentage"`
+	LocationInfo             string  `json:"location_info"`
+	QoSSupprt                bool    `json:"QoS_support"`
```
jrperritt: `QoSSupport`

openstack/blockstorage/storagepools/results.go
```diff
@@ -0,0 +1,122 @@
...
+	ProvisonedCapacityGB     float64 `json:"provisoned_capacity_gb"`
```
jrperritt: Haven't looked at the schema, but I suspect these should be "Provisioned" (and "provisioned")

openstack/blockstorage/storagepools/results.go
```diff
@@ -0,0 +1,122 @@
...
+	TotalVolumes             int64   `json:"total_volumes"`
+	FilterFunction           string  `json:"filter_function"`
+	GoodnessFuction          string  `json:"goodness_function"`
+	Mutliattach              bool    `json:"multiattach"`
+	SparseCopyVolume         bool    `json:"sparse_copy_volume"`
+}
+
+type StoragePool struct {
+	Name         string `json:"name"`
+	Capabilities `json:"capabilities"`
```
jrperritt:
```go
Capabilities Capabilities `json:"capabilities"`
```

openstack/blockstorage/storagepools/results.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+
+	// Should be a numeric, "unknown", "infinite"
+	if q.Capabilities.FreeCapacityGB != nil {
+		switch t := q.Capabilities.FreeCapacityGB.(type) {
+		case float64:
+			r.Capabilities.FreeCapacityGB = q.Capabilities.FreeCapacityGB.(float64)
+		case int:
+			r.Capabilities.FreeCapacityGB = q.Capabilities.FreeCapacityGB.(float64)
```
jrperritt: That will panic. Type-assert to `int`, then cast to `float64`. Better would be writing a unit test for it.

openstack/blockstorage/storagepools/results.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+	}
+
+	// Should be a numeric, "unknown", "infinite"
+	if q.Capabilities.TotalCapacityGB != nil {
+		switch t := q.Capabilities.TotalCapacityGB.(type) {
+		case float64:
+			r.Capabilities.TotalCapacityGB = q.Capabilities.TotalCapacityGB.(float64)
+		case int:
+			r.Capabilities.TotalCapacityGB = q.Capabilities.TotalCapacityGB.(float64)
```
jrperritt: Same here as above (type-assert to `int`, then cast to `float64`).

spjmurray:
@jtopjian @jrperritt Should be much better now...

openstack/blockstorage/v2/extensions/schedulerstats/results.go
```diff
@@ -0,0 +1,107 @@
...
+		case string:
+			keyword := s.FreeCapacityGB.(string)
+			switch keyword {
+			case "infinite":
+				r.FreeCapacityGB = math.Inf(1)
+			default:
+				r.FreeCapacityGB = 0.0
+			}
+		default:
+			return fmt.Errorf("capabilities.free_capacity_gb: unexpected type %v", t)
```
jtopjian: I think all default actions should be to just set to `0.0` rather than return an error.

jtopjian:
Thanks, Simon :)

openstack/blockstorage/extensions/schedulerstats/results.go
```diff
@@ -37,6 +36,25 @@ type StoragePool struct {
 	Capabilities Capabilities `json:"capabilities"`
 }
 
+// Generic function to parse a capacity value which may be a numeric
+// value, "unknown", or "infinite"
+func ParseCapacity(capacity interface{}) float64 {
```
jrperritt: I'd prefer to just put this logic in `UnmarshalJSON`, but if you want to use a function, it should to be unexported (lowercase).

Issue #208, opened 2016-12-01
Bug: identity/v2/user tenant_id is wrong


dbaumgarten:
The field TenantID in gophercloud/openstack/identity/v2/requests.go  CommonOpts struct should have the struct tag `json:"tenantId,omitempty"` instead of `json:"tenant_id,omitempty"` as according to the api doc, (http://developer.openstack.org/api-ref/identity/v2-admin/?expanded=create-user-admin-endpoint-detail), the Keystone source (https://github.com/openstack/keystone/blob/stable/liberty/keystone/identity/schema.py) and my own experiments, keystone expects the field to be called tenantId.

Additionally the TenantID field in the gophercloud/openstack/identity/v2/results.go User-struct should be removed, as neither List nor Get requests will ever return a TenantID.

If thats ok, I would happily fix this an create a PR.

jtopjian:
It looks like the user schema was only recently added: 

https://github.com/openstack/keystone/blob/fab399e26cdbe7cffba895f99d7247896ec6cb82/keystone/identity/schema.py

But browsing some other files such as the [controller](https://github.com/openstack/keystone/blob/master/keystone/identity/controllers.py) (and its history) shows mention of `tenantId` in the user actions ([example](https://github.com/openstack/keystone/blob/master/keystone/identity/controllers.py#L62)), so I also think `tenantId` is correct.

dbaumgarten:
What about the TenantID field in the user result struct? As far as I can see, there is no request, that returns a tenantid inside a user-object/structure, so that field in the result struct will always be empty.
While an unused field in a struct is not a real Problem, I think it should be removed to prevent confusion.

jrperritt:
closed by #211

PR #211, opened 2016-12-02
Fixed TenantID struct-tag in user/CommonOpts

dbaumgarten:
For #208

(Links to source and documentation inside issue)


jrperritt:
+2

Issue #212, opened 2016-12-03
Personality isn't working 


brucemartins:
Data is not being sent properly, I already have a fix for this and have tested it with terraform and would like to make a PR with the fix

jtopjian:
Some prior discussion about this can be found here: https://github.com/hashicorp/terraform/pull/10517

Some of the acceptance tests implement "personality", but there isn't verification done within the instance to confirm it was successful.

jrperritt:
closed by #213

PR #213, opened 2016-12-03
fixed issue where personality wasn't working in request w/ config_drive

brucemartins:

For #212

code in this PR:

[PUT URLS HERE]


jrperritt:
+2

Issue #207, opened 2016-11-29
PUT/DELETE for nova quota-sets


dbaumgarten:
At the moment there is only a function for getting the current nova quotas for a tenant/project and no possibility to update (PUT) or reset (DELETE) them. 

I would be happy to contribute this functionality to gophercloud. ( In fact, as I needed this for work I have already written the needed code)

If you are ok with that, I would open a PR.

jrperritt:
closed by #214

PR #214, opened 2016-12-05
Added functionality for updating and resetting compute quotas.

dbaumgarten:
For #207

code in this PR:

https://github.com/openstack/nova/blob/stable/liberty/nova/api/openstack/compute/schemas/quota_sets.py#L28

Also while coding this I found and fixed two small errors in the Get-Method.
   - The ssh-key-quota should be "key_pairs" instead of "keypairs"
   - The quotas "server_groups" and "server_group_members" where missing

Unit and acceptance tests added and succeeding.

### DIFF
--- a/openstack/compute/v2/extensions/quotasets/requests.go
+++ b/openstack/compute/v2/extensions/quotasets/requests.go
@@ -10,3 +10,88 @@ func Get(client *gophercloud.ServiceClient, tenantID string) GetResult {
 	_, res.Err = client.Get(getURL(client, tenantID), &res.Body, nil)
 	return res
 }
+
+//Updates the quotas for the given tenantID and returns the new quota-set
+func Update(client *gophercloud.ServiceClient, tenantID string, opts UpdateOptsBuilder) (res UpdateResult) {
+	reqBody, err := opts.ToComputeQuotaUpdateMap()
+	if err != nil {
+		res.Err = err
+		return
+	}
+
+	_, res.Err = client.Put(updateURL(client, tenantID), reqBody, &res.Body, &gophercloud.RequestOpts{OkCodes: []int{200}})
+	return res
+}
+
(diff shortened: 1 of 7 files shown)


openstack/compute/v2/extensions/quotasets/requests.go
```diff
@@ -10,3 +10,87 @@ func Get(client *gophercloud.ServiceClient, tenantID string) GetResult {
...
+}
+
+type UpdateOptsBuilder interface {
+	//Extra specific name to prevent collisions with interfaces for other quotas (e.g. neutron)
+	ToComputeQuotaUpdateMap() (map[string]interface{}, error)
+}
+
+func (opts UpdateOpts) ToComputeQuotaUpdateMap() (map[string]interface{}, error) {
+
+	quotas, err := gophercloud.BuildRequestBody(opts, "quota_set")
```
jrperritt: Let's reduce this to 
```go
return gophercloud.BuildRequestBody(opts, "quota_set")
```

openstack/compute/v2/extensions/quotasets/requests.go
```diff
@@ -10,3 +10,87 @@ func Get(client *gophercloud.ServiceClient, tenantID string) GetResult {
...
+	return quotas, err
+}
+
+//Convenience method to fill an UpdateOpts-struct from a QuotaSet-struct
+func (opts *UpdateOpts) FillFromQuotaSet(set *QuotaSet) {
```
jrperritt: I can understand to appeal of a method like this (perhaps retrieving one user's quota and setting it as another's), but I'm not sure I want to commit to it; similar methods could be justified for every resource. You can put it in ` acceptance/openstack/compute/v2/compute.go` though if you like.

jrperritt:
+2

Issue #116, opened 2016-10-11
servers.WaitForStatus timeout does not work as expected


sb10:
I would expect the purpose of the timeout in WaitForStatus() is that after the timeout period ends, WaitForStatus returns regardless of the status.

However, servers.WaitForStatus() is implemented with gophercloud.WaitFor(), but the latter does not actually interrupt its predicate function after the timeout period. Instead it waits for the predicate to return a value.

Unfortunately it seems that servers.WaitForStatus makes calls that can hang indefinitely in the case of a problem bringing up the server.

jtopjian:
It looks like this is what #83 is implementing, however, it removes the ability for the entire `WaitFor` function to stop after a set amount of time.

I think both forms of a timeout are applicable here, since there are two main cases to account for:

1. The predicate function never returns (network issue, server issue)
2. The satisfied value is never returned.

jrperritt:
closed by #221

PR #221, opened 2016-12-28
Refactor WaitFor

jtopjian:
This commit modifies WaitFor in a few ways:

It replaces time.Now().Second() with time.Now().Unix() in order to
account for rolling minutes (for example, when WaitFor starts at 59
seconds, every timeout check will result in a negative number).

A "retry" timer has also been added. This will cause the predicate
to be retried every n seconds if there hasn't been a response. This
is to account for server or network issues that would cause the
predicate to be lost or hang indefinitely.

A combination of using both timeout and retry can be effective in
handling faulty requests as well as a master kill switch to stop.

For #116 
Should supersede #83  

`ticks` and `tocks` have been left in for review purposes. They can be removed once confirmed that this implementation looks good.

### DIFF
--- a/util.go
+++ b/util.go
@@ -2,6 +2,7 @@ package gophercloud
 
 import (
 	"errors"
+	"fmt"
 	"net/url"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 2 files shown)


jtopjian:
Travis errors will be fixed upon first review, too.

jtopjian:
Marking as WIP, but I'd like a quick review of the implementation to see if this is worthwhile or not.

util.go
```diff
@@ -10,26 +11,58 @@ import (
...
+// To handle situations when the predicate response might hang indefinitely,
+// the predicate will be retried every "retry" seconds.
 // Resource packages will wrap this in a more convenient function that's
 // specific to a certain resource, but it can also be useful on its own.
-func WaitFor(timeout int, predicate func() (bool, error)) error {
-	start := time.Now().Second()
-	for {
-		// Force a 1s sleep
-		time.Sleep(1 * time.Second)
+func WaitFor(timeout, retry int, predicate func() (bool, error)) error {
```
jrperritt: Let's just fix `WaitFor` without `retry` for now.

util.go
```diff
@@ -10,26 +11,58 @@ import (
...
+		ch := make(chan WaitForResult, 1)
+		go func() {
+			defer close(ch)
+			satisfied, err := predicate()
+			result.Success = satisfied
+			result.Error = err
+		}()
+
+		if retry > 0 {
+			select {
```
jrperritt: This is good but take it out of the `if` block.

util.go
```diff
@@ -10,26 +11,58 @@ import (
...
+			case <-ch:
+				fmt.Println("tick")
+				if result.Error != nil {
+					return result.Error
+				}
+				if result.Success {
+					return nil
+				}
+			case <-time.After(time.Duration(retry) * time.Second):
```
jrperritt: s/retry/timeout

util.go
```diff
@@ -10,26 +11,58 @@ import (
...
+				fmt.Println("tock")
+				continue
```
jrperritt: `return fmt.Errorf("a timeout occurred")`

util.go
jrperritt: At the moment, this doesn't work on a timeout (e.g. 
```go
WaitFor(1, 0, func()(bool, error) error {
  time.Sleep(10000000 * time.Second)
  return true, nil
}
```

I think the changes mentioned below should fix it though.

util.go
```diff
@@ -10,26 +11,58 @@ import (
...
-		satisfied, err := predicate()
-		if err != nil {
-			return err
-		}
-		if satisfied {
-			return nil
+		time.Sleep(1 * time.Second)
+
+		var result WaitForResult
+		ch := make(chan WaitForResult, 1)
```
jrperritt: Though there's nothing wrong with this, note that you're doing anything with this value when you receive it below; it could just as well be a `chan bool`.

util.go
```diff
@@ -10,26 +11,58 @@ import (
...
+		go func() {
+			defer close(ch)
+			satisfied, err := predicate()
+			result.Success = satisfied
+			result.Error = err
+		}()
+
+		if retry > 0 {
+			select {
+			case <-ch:
```
jrperritt: If you wanted to send the `WaitForResult`, you'd receive it here:
```go
case wfr := <-ch:
//...
```

util.go
jrperritt: Ah, you're also not sending it on the channel.

jtopjian:
@jrperritt Thanks! That all makes total sense and makes the implementation much more simple.

jrperritt:
+2 Thank you for fixing this

Issue #222, opened 2017-01-03
Can't fetch all private flavors in a single request


mfranc213:
Gophercloud's Flavor struct is missing an `IsPublic` field.  Compare:
	[Nova's Flavor object](https://github.com/openstack/nova/blob/stable/newton/nova/objects/flavor.py#L221) with [Gophercloud's Flavor struct](https://github.com/gophercloud/gophercloud/blob/magnum/openstack/compute/v2/flavors/requests.go#L17). Furthermore, Gophercloud's *flavors* requests don't allow an `is_public` to be specified.
	
In *flavors* requests made to Nova's API, the `is_public` argument is optional. When the argument is not supplied, Nova uses [the value](https://github.com/openstack/nova/blob/stable/newton/nova/api/openstack/compute/flavors.py#L64) **True**. The argument is also [ternary](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/v2/flavors.py#L112).  When the argument is supplied, the following values are allowed:

* **True** returns private flavors associated with that project as well as (a bit confusingly) public flavors
* **False** (admin only) returns only private flavors, across all projects
* **None** (admin only) is interpreted by Nova as a request to [not filter](https://github.com/openstack/nova/blob/stable/newton/nova/api/openstack/compute/flavors.py#L67) by `is_public`, so it returns all public and private flavors across all projects.

All *four* options are tested in novaclient: [no is_public argument](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/tests/unit/v2/test_flavors.py#L35), [None](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/tests/unit/v2/test_flavors.py#L60), [False](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/tests/unit/v2/test_flavors.py#L67), and [True](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/tests/unit/v2/test_flavors.py#L74).  **True** is the default value in novaclient's [method signature](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/v2/flavors.py#L95). 

Because Gophercloud's flavor requests don't allow `is_public` to be specified, Nova uses the value **True**.  So there is no way, currently, to obtain the list of all private flavors other than by fetching flavors project by project, but even then, because Gophercloud's Flavor struct doesn't contain the IsPublic field, there is no way to distinguish public from private flavors through Gophercloud.

[...]

PR #223, opened 2017-01-03
Allow all private flavors to be fetched (#222)

mfranc213:
Because Gophercloud's flavor requests don't allow is_public
to be specified, Nova uses the value True. So there is no way,
currently, to obtain the list of all private flavors other than
by fetching flavors project by project, but even then, because
Gophercloud's Flavor struct doesn't contain the IsPublic field,
there is no way to distinguish public from private flavors through
Gophercloud.

For #222

Gophercloud's Flavor struct is missing an `IsPublic` field.  Compare:
	[Nova's Flavor object](https://github.com/openstack/nova/blob/stable/newton/nova/objects/flavor.py#L221) with [Gophercloud's Flavor struct](https://github.com/gophercloud/gophercloud/blob/magnum/openstack/compute/v2/flavors/requests.go#L17). Furthermore, Gophercloud's *flavors* requests don't allow an `is_public` to be specified.
	
In *flavors* requests made to Nova's API, the `is_public` argument is optional. When the argument is not supplied, Nova uses [the value](https://github.com/openstack/nova/blob/stable/newton/nova/api/openstack/compute/flavors.py#L64) **True**. The argument is also [ternary](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/v2/flavors.py#L112).  When the argument is supplied, the following values are allowed:

* **True** returns private flavors associated with that project as well as (a bit confusingly) public flavors
* **False** (admin only) returns only private flavors, across all projects
* **None** (admin only) is interpreted by Nova as a request to [not filter](https://github.com/openstack/nova/blob/stable/newton/nova/api/openstack/compute/flavors.py#L67) by `is_public`, so it returns all public and private flavors across all projects.

All *four* options are tested in novaclient: [no is_public argument](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/tests/unit/v2/test_flavors.py#L35), [None](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/tests/unit/v2/test_flavors.py#L60), [False](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/tests/unit/v2/test_flavors.py#L67), and [True](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/tests/unit/v2/test_flavors.py#L74).  **True** is the default value in novaclient's [method signature](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/v2/flavors.py#L95). 

Because Gophercloud's flavor requests don't allow `is_public` to be specified, Nova uses the value **True**.  So there is no way, currently, to obtain the list of all private flavors other than by fetching flavors project by project, but even then, because Gophercloud's Flavor struct doesn't contain the IsPublic field, there is no way to distinguish public from private flavors through Gophercloud.


mfranc213:
New commit for this PR ready for review, responding to https://github.com/gophercloud/gophercloud/issues/222#issuecomment-270299091 (thanks Joe!).

jtopjian:
@mfranc213 Sorry for the delay on this. 

It looks like there are some conflicts that need resolved due to other PRs being merged. Once that's fixed, there are a few small fixes to make, which I can point out then. 

Since this has been opened for so long, if you're no longer able to work on this, let me know and I can take over.

mfranc213:
Hi jtopjian.  Great!  Yes, let me know what the changes are and I'll take care of them.  Thanks so much.

Margaret

openstack/compute/v2/flavors/requests.go
```diff
@@ -11,6 +11,33 @@ type ListOptsBuilder interface {
...
+// request options are ternary, which is why FlavorType is a string. The following values are
+// allowed:
+//
+//      Project (the default): Returns only public flavors and private flavors associated with that project.
+//      Private (admin only):  Returns only private flavors, across all projects.
+//      All (admin only):      Returns all public and private flavors across all projects.
+//
+// If no IsPublic argument is supplied in the request, OpenStack treats the FlavorType as Project.
+type FlavorType string
+const (
```
jtopjian: format nit: add a newline between `type` and `const`

openstack/compute/v2/flavors/requests.go
```diff
@@ -11,6 +11,33 @@ type ListOptsBuilder interface {
...
+    switch flavorType {
+    case Project:
+        return "Project"
+    case Private:
+        return "Private"
+    case All:
+        return "All"
+    }
+    return "Unknown"
+}
```
mfranc213: Done.

openstack/compute/v2/flavors/requests.go
```diff
@@ -21,22 +21,12 @@ type ListOptsBuilder interface {
 //
 // If no IsPublic argument is supplied in the request, OpenStack treats the FlavorType as Project.
 type FlavorType string
+
 const (
-        Project FlavorType =  "true"
-        Private FlavorType = "false"
-        All FlavorType = "None"
+	Project FlavorType = "true"
```
jrperritt: Let's name these `FlavorType`s something else. `flavors.Project` doesn't convey a flavor type. `TypeProject` would be clearer.

jtopjian: You making that comment made me realize you're correct, but I also think `Project` muddies things, too.

What about `FlavorScope` or `FlavorAccessScope`?

jtopjian: Or just `FlavorAccess`, too.

mfranc213: FlavorAccess it is.  Much better than FlavorType.

jrperritt: Perhaps I was misunderstood: The constant names should be changed to something more meaningful.

jtopjian: Ah, I see what you are saying now. One would never actually refer to the type name in practice, only the constant name and that's what will end up being used in code. I'm sorry.

I think @mfranc213's change of `FlavorType` to `FlavorAccess` works well with your original request:

```go
listOpts := flavors.listOpts{
  FlavorAccess: flavors.TypePublic,
}
```

It seems to fit well with how it's described in the [documentation](https://github.com/openstack/python-novaclient/blob/stable/newton/novaclient/v2/flavors.py#L101-L103). 

yes/no?

mfranc213: Oh yes, I did misunderstand you.  Are you suggesting that the constant named `Project` should instead be named, for example, `FlavorAccess`?  That didn't make sense to me, which is why I interpreted you the way I did.

I'm having a hard time thinking of more accurate and descriptive constant names than what I have.  That's a comment about my limitations rather than the virtues of the names I suggested.  I wish the  OS community had stuck with the `Tenant` vernacular.  I think `flavors.Tenant` is more immediately understood than `flavors.Project`.

Let me know how you would like me to proceed (including if it's "Margaret: propose something better".)

mfranc213: Woops my last comment was written without my having seen your comment from 4 mins ago.  Reading your comment now.

mfranc213: Okay, I think at this point I'm just waiting for @jtopjian to vote on the matter.

jrperritt: This is my fault; I wasn't clear enough. My concern is that a name like `Public` is a bit vague; `PublicAccess` would be better. In addition, `FlavorType` and `FlavorAccess` are OK, but we're in the `flavors` package, so the functions and types are assumed to relate to flavors. Here's a suggestion to make the code a bit more readable:
```go
type AccessType string

const (
  PublicAccess AccessType = "true"
  PrivateAccess AccessType = "false"
  AllAccess AccessType = "None"
)
```
And it would be used like:
```go
listOpts := flavors.listOpts{
  FlavorAccess: flavors.PublicAccess,
}
```

Note that the "Type" part of "AccessType" is not because it's a Go type (which it is), but because it is the type of access granted.

@jtopjian, @mfranc213 questions or thoughts about that?

jtopjian: Yep, all understood, including the overloaded `Type`. :)

mfranc213: I like `AccessType` and I the idea of appending "Access" to the names of all three types.  But `PublicAccess` can't be one of these types, because there is no API call that allows you to obtain all and only those flavors that are public.  

It really is the case that the Nova API gives you three options in your fetch: 

* all flavors, public or private, across all projects,
* private flavors across all projects, or 
* flavors available within a specific project (either because they are private to that project or because they are public).

mfranc213: I need to reverse what I said above.  Constructing replacement reply now ...

mfranc213: I've reconsidered my earlier stance on the term "Access" in this context, because we're not directly talking about access.  The `All` type, for instance, means that all flavors are returned, not that, as I think it would imply, anyone can access that API call.  In fact only admins can use the `All` option.

I wonder if we should use the vernacular of the Nova API itself. I was trying to avoid it because I think it's so confusing--it's a ternary masquerading as a binary, and beyond that the request-response relationships are bizarre.  Reluctantly, I'm thinking of this:

    type IsPublic string

    const (
        Project IsPublic = "true"
        Private IsPublic = "false"
        All     IsPublic = "None"
    )

And used like:

    listOpts := flavors.listOpts{
       IsPublic: flavors.All,
    }

mfranc213: Or maybe:

    const (
        ProjectOnly IsPublic = "true"
        AllPrivate  IsPublic = "false"
        All         IsPublic = "None"
    )

jrperritt: @mfranc213 This is directly from the source code (that @jtopjian linked to in an above comment):
```
:param is_public: Filter flavors with provided access type (optional).
                          None means give all flavors and only admin has query
                          access to all flavor types.
```
I think it should look something like what I suggested in my last comment.

mfranc213: Okay, I'm persuaded.  Will push next commit shortly.  Thanks for your comments and help.

openstack/compute/v2/flavors/results.go
```diff
@@ -46,9 +46,11 @@ type Flavor struct {
 	Swap int `json:"swap"`
 	// VCPUs indicates how many (virtual) CPUs are available for this flavor.
 	VCPUs int `json:"vcpus"`
+	// IsPublic indicates whether the flavor is public.
+	IsPublic bool `json:"is_public"`
 }
 
-func (r *Flavor) UnmarshalJSON(b []byte) error {
+func (f *Flavor) UnmarshalJSON(b []byte) error {
```
jtopjian: @mfranc213 I'm so sorry. I don't know how I missed this before. `r` is actually correct here. See the Naming section in the [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md) for more info.

mfranc213: That's my fault.  I forgot about that instruction in the style guide!

jrperritt:
@mfranc213 Thank you for your patience and persistence :)

Issue #224, opened 2017-01-05
DB Instances: Fault information missing in result


esalipe:
When a DB instance has errored, Trove provides fault information in instance detail responses:
https://github.com/openstack/trove/blob/stable/newton/trove/instance/models.py#L395

The fault information class is defined here:
https://github.com/openstack/trove/blob/stable/newton/trove/instance/models.py#L1479
Only part of the fault information is returned in the response:
https://github.com/openstack/trove/blob/stable/newton/trove/instance/views.py#L128

Support for parsing the fault information would be useful as the information makes understanding why a DB instance is in ERROR state much simpler.

jtopjian:
@esalipe I've fixed up your original branch and merged it as #1014. Let me know if you have any questions. :)

PR #225, opened 2017-01-05
db/instances: add support for fault information in response body

esalipe:
For #224 

Parses fault information returned by Trove when a DB instance has faulted:
https://github.com/openstack/trove/blob/stable/newton/trove/instance/views.py#L96

### DIFF
--- a/openstack/db/v1/instances/results.go
+++ b/openstack/db/v1/instances/results.go
@@ -1,6 +1,7 @@
 package instances
 
 import (
+	"encoding/json"
 	"time"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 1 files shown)


jtopjian:
@esalipe Sorry for the delay. It looks like there's a merge conflict from another PR. Would it be possible for you to rebase this off of master?

In addition, could you add a unit test to verify the fault implementation? I suspect that there _might_ be an issue with the Instance's `UnmarshalJSON` taking precedence, but I'm hopefully wrong about that.

openstack/db/v1/instances/results.go
```diff
@@ -49,13 +61,33 @@ type Instance struct {
 	// The build status of the instance.
 	Status string
 
+	// Fault information (only available when the instance has errored)
+	Fault Fault
```
jrperritt: Let's change the type to `*Fault`. That way it can be checked like
```go
if instance.Fault != nil {
...
}
```
Is that how it's expected to be used? (Specifically, I suspect it's the case that `Fault` may be populated when the error returned by Gophercloud from a call to OpenStack is `nil`.)

jrperritt:
closing due to inactivity

Issue #228, opened 2017-01-12
Remove createErr Function from Identity Packages


jtopjian:
Per discussion in #165, there are a few `createErr` functions in the identity packages that don't seem to be used.

jrperritt:
closed by #229

PR #229, opened 2017-01-12
Removing unused createErr functions from Identity packages

jtopjian:
For #228


jtopjian:
@jrperritt ping for review

jrperritt:
Thank you :+1:

Issue #232, opened 2017-01-18
Endless recursive requests to server when server response error


armstrongli:
Here's the output from client when server is not stable. Then client sends more requests until connection reset.

```
...
ng request: Successfully re-authenticated, but got error ex
ecuting request: Successfully re-authenticated, but got error executing request: Successfully re-authenticated, but got error executi
ng request: Successfully re-authenticated, but got error executing request: Successfully re-authenticated, but got error executing re
quest: Successfully re-authenticated, but got error executing request: Successfully re-authenticated, but got error executing request
: Successfully re-authenticated, but got error executing request: Put https://os-r-object.vip.*.*.com/v1/KEY_3e4c92efc6444df19
f7df401a2de4a29/backup: read tcp 10.147.160.18:36116->10.14.41.20:443: read: connection reset by peer
```

armstrongli:
Fixed in PR: https://github.com/gophercloud/gophercloud/pull/233

jrperritt:
See https://github.com/gophercloud/gophercloud/pull/233#issuecomment-274197054

armstrongli:
@jrperritt Please reopen  this issue as the root cause is found and fixed in this PR: https://github.com/gophercloud/gophercloud/pull/604

jrperritt:
Looking now

jrperritt:
closed by #604

Issue #230, opened 2017-01-16
Investigate sensible defaults for common HTTP functionality (e.g. re-auth limit, logging)


jrperritt:
These have caused confusion for enough people to consider adding defaults. In the case of implementing a default re-auth limit, it would likely be a backwards-incompatible change (for anyone counting on infinitely trying to re-auth).

kayrus:
Logging was added in https://github.com/gophercloud/utils/tree/master/client
@jtopjian I believe the issue can be closed.

PR #233, opened 2017-01-18
add recursive flag to avoid endless requests to server for unstable a…

armstrongli:
…uthentication & authorization

For https://github.com/gophercloud/gophercloud/issues/232


jrperritt:
To limit re-auth attempts, you need to create a custom `HTTPClient`. See [here](https://github.com/gophercloud/gophercloud/blob/master/FAQ.md#implementing-default-logging-and-re-authentication-attempts).

Default re-auth limits may be implemented in the future (see #230), but it won't look like this PR.

Issue #226, opened 2017-01-06
Ports: Removing Security Groups and Allowed Address Pairs


jtopjian:
The Neutron command-line client provides port update options of `--no-security-groups` and `--no-allowed-address-pairs` which passes empty arrays to the API server:

```
DEBUG: keystoneauth.session REQ: curl -g -i 
-X PUT http://10.1.15.127:9696/v2.0/ports/c9f02a7a-4aac-44ed-a2bc-947db82621d2.json 
-H "User-Agent: python-neutronclient" 
-H "Content-Type: application/json" 
-H "Accept: application/json" 
-H "X-Auth-Token: {SHA1}861547906c8c47c61d33008eae9c7877b09a6a4b" 
-d '{"port": {"security_groups": []}}'

DEBUG: keystoneauth.session REQ: curl -g -i 
-X PUT http://10.1.15.127:9696/v2.0/ports/e937b274-e3f8-4b64-b056-0885875620b1.json 
-H "User-Agent: python-neutronclient" 
-H "Content-Type: application/json" 
-H "Accept: application/json" 
-H "X-Auth-Token: {SHA1}c2478d044e2294608a20ac440eed08f42bc1d500" 
-d '{"port": {"allowed_address_pairs": []}}'
```

In order to emulate this behavior, the `omitempty` tags for [these fields](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/ports/requests.go#L122-L123) needs to be removed.

As for the server-side, I'm pretty sure this is where the empty array is being validated:

* [port](https://github.com/openstack/neutron/blob/7b3d7ba98e04286b14950cc9ab5c699a418dbb45/neutron/db/securitygroups_db.py#L723)
* [allowed_address_pairs](https://github.com/openstack/neutron/blob/0b9935a1411e58005370c762b5a5895e385e45b2/neutron/db/allowedaddresspairs_db.py#L104)

jrperritt:
closed by #236

PR #236, opened 2017-01-19
Ports: Remove Security Groups and Address Pairs

jtopjian:
This commit enables security groups and address pairs to be removed
from a port. This is done by allowing an empty value to be passed
for either attribute in the port update request.

For #226


jtopjian:
@jrperritt ping for review

Issue #245, opened 2017-02-08
BlockStorage v1: Availability Should be AvailabilityZone


jtopjian:
The Block Storage v1 volume creation request is using [`Availability`](https://github.com/gophercloud/gophercloud/blob/c0df5212759612eb99d1178816ad620952c549f1/openstack/blockstorage/v1/volumes/requests.go#L19) when I believe it should be `AvailabilityZone` (and subsequently `availability_zone`).

I think [this](https://github.com/openstack/cinder/blob/master/cinder/api/v1/volumes.py#L274) is where `availability_zone` is being checked. As well, the following can be seen using `python-cinderclient`:

```
DEBUG:keystoneclient.session:REQ: curl -g -i -X POST https://example.com:8776/v1/1234/volumes -H "User-Agent: python-cinderclient" -H "Content-Type: application/json" -H "Accept: application/json" -H "X-Auth-Token: {SHA1}abcd" -d '{"volume": {"status": "creating", "availability_zone": "foo", "source_volid": null, "display_description": null, "snapshot_id": null, "user_id": null, "size": 1, "display_name": "jttest", "imageRef": null, "attach_status": "detached", "volume_type": null, "project_id": null, "metadata": {}}}'
```

jrperritt:
closed by #246

PR #246, opened 2017-02-08
Use AvailabilityZone instead of Availability

jtopjian:
This commit updates the Block Storage v1 create request to use
AvailabilityZone instead of Availability.

For #245


jtopjian:
@jrperritt ping for review

jrperritt:
+2

Issue #248, opened 2017-02-09
Time values for volumes.Volume are incorrectly handled


markdryan:
The volume.Volumes struct has two time fields, CreatedAt and UpdatedAt.  Both of these fields are defined to be of type time.Time.  However, a separate type, gophercloud.JSONRFC3339MilliNoZ, is used to unmarshall them.  This creates two problems:

1. It's not possible to unmarshall a json representation of a volumes.Volume structure that contains a CreatedAt or UpdatedAt field with a timezone, e.g., "created_at": "2015-09-17T03:35:03.000000Z".  According to the [spec](https://developer.openstack.org/api-ref/block-storage/v3/index.html?expanded=list-accessible-volumes-detail,list-accessbile-volumes-with-details-detail#list-accessbile-volumes-with-details) this should be possible.
2. Timestamps without time zone information are processed incorrectly.  They are parsed using time.Parse which treats them as being UTC times which is not right.  Again according to the [spec](https://developer.openstack.org/api-ref/block-storage/v3/index.html?expanded=list-accessible-volumes-detail,list-accessbile-volumes-with-details-detail#list-accessbile-volumes-with-details) the timestamps are ISO8601 timestamps and ISO8601 timestamps without time zone information are in local and not UTC time.

The result of all this is, is that volume timestamps fields will almost always be incorrect, unless both client and server are in the UK and it's winter.

markdryan:
I've created a PR that illustrates the issue and proposes a fix:

https://github.com/gophercloud/gophercloud/pull/249

The fix works by a) changing the type used to unmarshall volumes.Volume timestamps to gophercloud.JSONRFC3339Milli and b) fixing the UnmarshallJSON method for this type to handle timestamps that have timezone information and those that do not.  I chose this type as it was currently unused.  However, a new type could easily be introduced.

jtopjian:
@markdryan I'm a little confused about this issue. Have you run into a deployment where the time parsing here is causing an issue?

From browsing the Cinder code, Cinder, the database schema, and Oslo are storing and retrieving everything as UTC. Any volumes I create are stored and retrieved as UTC. If I set my server's time zone to my local zone (MDT), all dates are still stored as UTC.

Now, I totally get that not having the `Z` at the end is incorrect ISO8601 form. Googling around, I see this is not uncommon, especially with MySQL. But I get that others might have different environments configured, which is why I'm curious if you've seen this issue first hand.

[...]

PR #249, opened 2017-02-09
Fix time handling in volumes.Volume

markdryan:
The code that parsed time values for the volumes.Volume CreatedAt and UpdatedAt fields
failed if time zone information was included in such values.  It also incorrectly interpreted
time values with no time zone information as being UTC time whereas the spec indicates
that they are in fact in local time.  This commit fixes the issue by permitting
time zone information in the volume.Volumes time fields and also correctly handling
time values that do no contain any timezone information.

For #248

code in this PR:

https://github.com/openstack/cinder/blob/master/cinder/objects/base.py#L245

I'm afraid I'm not a python programmer but as far as I can work out, this is where the created_at and updated_at fields are defined.


markdryan:
Abandoning PR.  The error was seen when using a custom volume service (ciao) that returned timezone information in its date fields.  While this is not incorrect to do this it was agreed that the issue should be fixed in ciao rather than gophercloud.

Issue #251, opened 2017-02-11
ImageService v2: Image Paging


jtopjian:
I think there's an issue with paging of images.

Currently a "next page" will be built regardless of if the `next` attribute is set or not:

https://github.com/gophercloud/gophercloud/blob/c0df5212759612eb99d1178816ad620952c549f1/openstack/imageservice/v2/images/results.go#L151

https://github.com/gophercloud/gophercloud/blob/c0df5212759612eb99d1178816ad620952c549f1/openstack/imageservice/v2/images/urls.go#L41

The result of the above will be something like: https://example.com:9292/v2. This gets passed to the pager and because of the base portion of the URL being set, [this](https://github.com/gophercloud/gophercloud/blob/c0df5212759612eb99d1178816ad620952c549f1/pagination/pager.go#L118) part is never hit.

The base URL will then be retrieved, triggering a 401 error.

I think this is a bug with Gophercloud rather than Glance, but [here's](https://github.com/openstack/glance/blob/b7c627ef68c303d4c3182c1a41b36185447039a3/glance/api/v2/images.py#L837) where the `next` attribute is being built.

I think the solution is something as simple as:

```go
func (r ImagePage) NextPageURL() (string, error) {
	var s struct {
		Next string `json:"next"`
	}
	err := r.ExtractInto(&s)
	if err != nil {
		return "", err
	}

        if s.Next == "" {
                return "", nil
        }

	return nextPageURL(r.URL.String(), s.Next), nil
}
```

If so, I'm happy to submit a PR.

jrperritt:
I don't think I've used the Glance API yet (hence part of the reason this may have gone unnoticed). Feel free to submit something that works for you.

jrperritt:
closed by #254

PR #254, opened 2017-02-14
ImageService v2: Fix paging when next URL is blank

jtopjian:
For #251 

If I get some time, I'll write up some acceptance tests for the ImageService package. As an interim, [Terraform image data source](https://github.com/jtopjian/terraform/blob/e077dd1e7be407a41cb8cab664db64d9b7fe2969/builtin/providers/openstack/data_source_openstack_images_image.go) I'm working on was able to succeed with this patch, so it kind of acts like an acceptance test.

side note: I'll probably be writing more data sources in the future. They map very well with ListOpts, so they might prove useful for finding other bugs in the List areas.

### DIFF
--- a/openstack/imageservice/v2/images/results.go
+++ b/openstack/imageservice/v2/images/results.go
@@ -156,6 +156,11 @@ func (r ImagePage) NextPageURL() (string, error) {
 	if err != nil {
 		return "", err
 	}
+
+	if s.Next == "" {
+		return "", nil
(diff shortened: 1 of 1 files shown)


jrperritt:
SGTM

jrperritt:
This is ready to merge, yes?

jtopjian:
Oops - yes, it is.

Issue #259, opened 2017-02-17
ImageService v2: Error Creating Properties


jtopjian:
The combination of current [JSON tags](https://github.com/gophercloud/gophercloud/blob/c5c590a7ee87267c978586fdd1655db0dd4b6e05/openstack/imageservice/v2/images/requests.go#L102) is causing the JSON body to not be rendered properly to the API.

The following isn't JSON, but a Python dict I dumped from the request on the server:

```python
{
  'name': 'Rancher TerraformAccTest', 
  'container_format': 'bare', 
  'visibility': 'private', 
  'protected': False, 
  'disk_format': 'qcow2',
    '-': {'architecture': 'x86_64'}, 
  'architecture': 'x86_64'
}
```

Note the last two entries.

I think `omitempty` can be removed since the `ToImageCreateMap` function checks for nil anyway? If so, I have a PR ready to go for that along with the start of acceptance tests.

jrperritt:
Confirmed. Please go ahead.

jrperritt:
closed by #264

PR #264, opened 2017-02-17
ImageService v2: Fixing Create Properties

jtopjian:
For #259

### DIFF
--- a/acceptance/openstack/imageservice/v2/images_test.go
+++ b/acceptance/openstack/imageservice/v2/images_test.go
@@ -1,0 +1,26 @@
+// +build acceptance imageservice images
+
+package v2
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+)
+
+func TestImagesCreateDestroyEmptyImage(t *testing.T) {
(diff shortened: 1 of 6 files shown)


acceptance/openstack/imageservice/v2/images_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+)
+
+func TestImagesCreateDestroyEmptyImage(t *testing.T) {
+	client, err := clients.NewImageServiceV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create an image service client: %v", err)
```
jtopjian: as an aside: `%v` here was always a mistake. When I did the initial cleanup of the tests, I had meant this to be `%#v`, but without understanding the structure of the error. I had thought there would be good bit of info in the error structure to print out. Turns out `%#v` would have been a bad choice. `%v` could easily just be `%s`. Maybe I'll fix that globally one day.

jtopjian:
@jrperritt ping for review

jrperritt:
Is there a unit test for this?

jtopjian:
Yes :)

jtopjian:
Thank you!

Issue #209, opened 2016-12-01
quotasets.Get().Extract() fails with JSON parser error on large values


sb10:
I am trying to use quotasets.Get().Extract() on a tenant that has 9,216,000 MB of memory quota. It fails with this error:

> json: cannot unmarshal number 9.216e+06 into Go value of type int

Taking a look at what Get() returns before the Extract():

> {quotaResult:{Result:{Body:map[quota_set:map[security_group_rules:20 security_groups:10 key_pairs:100 instances:1
0 ram:9.216e+06 fixed_ips:-1 metadata_items:128 server_groups:10 injected_file_content_bytes:10240 cores:20 id:083cd4d7080f44988ff7d6
423e958d5e injected_files:5 injected_file_path_bytes:255 server_group_members:10 floating_ips:10]] Header:map[] Err:<nil>}}}

dbaumgarten:
This seems to be en issue with encoding/json. If the json decoder encounters a number in scientific notation (e.g. 9.216e+06 ) it assumes the number to be of type float64. The Ram field in the QuotaSet-Struct is of type int. So the decoder doesn't try to convert the float to an int and throws an error instead. (Here is an example: https://play.golang.org/p/gG4neHJmMN )
However, this does not seem to be a bug in encoding/json. Is is the expected behavior ( https://github.com/golang/go/issues/5562 )

Do you know if the OpenStack API returns the number in scientific notation, or does gophercloud somehow accidentally do the conversion?

sb10:
I'm not sure what the OpenStack API returns, but with my debugging (as per my previous comment) I can see that the struct that ExtractInto() (and subsequently json.Marshal) operates on contains a float64 (after conversion from interface{}).

sb10:
Changing the definition of type QuotaSet Ram field from int to float64 does fix the problem, but would no doubt break lots of existing code.

dbaumgarten:
Knowing exactly how the response from OpenStack looks would probably help to find a better fix for this. Can you just throw a Print() into gophercloud/provider_client.go Request() to get the raw response Body?

dbaumgarten:
I think I found the root of the problem.
gophercloud/provider_client.go Request() decodes the response body into an interface{}. Because of the lack of type information in an interface{} is decides to use int64 as type for Ram.
[...]

PR #271, opened 2017-02-26
DNS Zones: List / Get

jtopjian:
For #172 

Supersedes #173 

Here's some more info on the DNS methods: https://github.com/openstack/designate/blob/stable/newton/designate/api/v2/controllers/zones/__init__.py

And on the schema: 

https://github.com/openstack/designate/blob/stable/newton/designate/objects/zone.py
https://github.com/openstack/designate/blob/6ddf20aba5c361629168ce2f3103c1e8052528d5/designate/objects/adapters/base.py

### DIFF
--- a/acceptance/openstack/dns/v2/zones.go
+++ b/acceptance/openstack/dns/v2/zones.go
@@ -1,0 +1,1 @@
+package v2
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -310,6 +310,16 @@ func NewDBV1(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*
 	return &gophercloud.ServiceClient{ProviderClient: client, Endpoint: url}, nil
 }
 
+// NewDNSV2 creates a ServiceClient that may be used to access the v2 DNS service.
+func NewDNSV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
+	eo.ApplyDefaults("dns")
+	url, err := client.EndpointLocator(eo)
+	if err != nil {
+		return nil, err
+	}
+	return &gophercloud.ServiceClient{ProviderClient: client, Endpoint: url}, nil
+}
+
 // NewImageServiceV2 creates a ServiceClient that may be used to access the v2 image service.
 func NewImageServiceV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
 	eo.ApplyDefaults("image")
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -317,7 +317,10 @@ func NewDNSV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (
 	if err != nil {
 		return nil, err
 	}
-	return &gophercloud.ServiceClient{ProviderClient: client, Endpoint: url}, nil
+	return &gophercloud.ServiceClient{
+		ProviderClient: client,
+		Endpoint:       url,
+		ResourceBase:   url + "v2/"}, nil
 }
 
 // NewImageServiceV2 creates a ServiceClient that may be used to access the v2 image service.
--- a/openstack/dns/v2/zones/results.go
+++ b/openstack/dns/v2/zones/results.go
@@ -2,6 +2,7 @@ package zones
 
 import (
 	"encoding/json"
+	"strconv"
 	"time"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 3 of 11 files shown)


jtopjian:
I'm still working on this. Unit tests pass, but I want to get some basic acceptance tests first just to make sure things work correctly against a running Designate service.

The first commit is just a rebase to get things up to date with master. The other commits fix things up to get the unit tests to pass.

I'm going to add some inline comments with a few questions.

openstack/client.go
```diff
@@ -317,7 +317,10 @@ func NewDNSV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (
 	if err != nil {
 		return nil, err
 	}
-	return &gophercloud.ServiceClient{ProviderClient: client, Endpoint: url}, nil
+	return &gophercloud.ServiceClient{
+		ProviderClient: client,
+		Endpoint:       url,
+		ResourceBase:   url + "v2/"}, nil
```
jtopjian: @jrperritt is this the correct thing to do? I noticed this was done for some services, but not all. I haven't dug into why (been a little pressed for time) but will do later just so I see why.

openstack/dns/v2/zones/results.go
```diff
@@ -90,6 +93,7 @@ func (r *Zone) UnmarshalJSON(b []byte) error {
 		CreatedAt     gophercloud.JSONRFC3339MilliNoZ `json:"created_at"`
 		UpdatedAt     gophercloud.JSONRFC3339MilliNoZ `json:"updated_at"`
 		TransferredAt gophercloud.JSONRFC3339MilliNoZ `json:"transferred_at"`
+		Serial        interface{}                     `json:"serial"`
```
jtopjian: Without this, I was getting scientific notations as described in #209. I saw this was done for `compute/v2/flavors` for `Swap`, so implemented it here.

Is this the best way to go? Or should this wait until a solution in #209 is implemented?

jtopjian:
@jrperritt There looks to be an issue with `AllPages`. In the unit tests, it's returning a count of 3 consistently. In the acceptance test, it's randomly returning 0 or 3. If 3, one of the zones is entirely blank.

Other than that, and with the other comments in mind, this is ready for a first review. :)

jrperritt:
@jtopjian Please rebase with master to pull in #276 and see if that fixes the `AllPages` issue.

jtopjian:
Working great :)

jtopjian:
@jrperritt ping for review

jtopjian:
Just noticed ListOpts was not implemented. _Now_ it should be ready for a first review.

acceptance/openstack/dns/v2/zones.go
```diff
@@ -0,0 +1 @@
+package v2
```
jrperritt: This file is by convention just named `pkg.go` or `doc.go`.

openstack/client.go
```diff
@@ -310,6 +310,19 @@ func NewDBV1(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*
...
+func NewDNSV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
+	eo.ApplyDefaults("dns")
+	url, err := client.EndpointLocator(eo)
+	if err != nil {
+		return nil, err
+	}
+	return &gophercloud.ServiceClient{
+		ProviderClient: client,
+		Endpoint:       url,
+		ResourceBase:   url + "v2/"}, nil
```
jrperritt: This is for when the catalog's URL for this service doesn't include the version; it occurred with Networking v2 I think. Is that also the case here?

acceptance/openstack/dns/v2/zones.go
jtopjian: Ah - OK, thanks.

If it's OK with you, this file gets renamed to `dns.go` in the next PR which follows the pattern of the other acceptance tests I've done. It also has actual contents.

No issues renaming it to `pkg.go` in this PR if you prefer, though.

openstack/client.go
jtopjian: Got it. And that's correct: the catalog I'm testing against does not have a version in the URL.

acceptance/openstack/dns/v2/zones.go
jrperritt: Ah, no problem. The name of the file isn't really important.

openstack/client.go
jrperritt: :+1:

jrperritt:
Looks good

Issue #272, opened 2017-03-01
Floating IP acceptance test failing


dbaumgarten:
Hi,

I just noticed that the acceptance-tests for floating-ips aren't working.

Running: go test -v floatingips_test.go layer3.go  (in the appropriate folder)
Results in:
` networking.go:83: Attempting to create subnet: TESTACC-RXoeMDT0
        networking.go:90: Successfully created subnet.
        layer3.go:46: Attempting to create external router: TESTACC-QXGreeZZ
        floatingips_test.go:55: Unable to create router: Invalid request due to incorrect syntax or missing required parameters.
        networking.go:188: Attempting to delete subnet: 080b375b-b3aa-413f-b356-c3beb692483a
        networking.go:195: Deleted subnet: 080b375b-b3aa-413f-b356-c3beb692483a`

And so the Test fails.

I am testing on OpenStack Liberty.

jtopjian:
@dbaumgarten The floating IP test was indeed broke. #274 fixes it.

The README at https://github.com/gophercloud/gophercloud/tree/master/acceptance mentions how to run acceptance tests. For the floating IP one, I did:

```shell
$ gophercloudtest TestLayer3FloatingIPsCreateDelete networking/v2/extensions/layer3
```

I also had to be an admin user since it's creating a subnet on the default Neutron network in my environment. That might vary with your environment.

Let me know if you're still having issues. :)

dbaumgarten:
Ok. I looked at this at little further.
The tests failed, because the tried to add a subnet to my floating-ip-network. That didn't work, because the used account doesnot have admin privileges (thank god).

Why are the tests doing this? Why aren't they creating their subnets on a "normal" network?
Adding a subnet to the floating-ip network is nothing I would like a test to do on my productive OpenStack environment. And I also would not want the test to run with admin privileges. (If not absolutely necessary)

Can't we just make the tests use a normal network? What about using the network provided via OS_NETWORK_NAME ? We could just look up its id and use this for the subnets.

jtopjian:
That indeed looks a bit strange. I want to say that's a mistake, but something is making me think there was a reason why I made the test that way. It might have been because the ability to create a router with interfaces on the non-admin subnet and the public subnet was not possible at the time.

If you're able to, feel free to rewrite the test to account for a more normal workflow. I can also take a look when I get a moment, too, but it might take me a few days.

jrperritt:
closed by #277

PR #277, opened 2017-03-03
Layer3 acceptance tests dont use floating-network for subnets

dbaumgarten:
For #272 

Layer3 acceptance tests now use the network provided via OS_NETWORK_NAME to create their subnets instead of OS_EXTGW_ID.

No source/documentation to link to.


acceptance/openstack/networking/v2/extensions/layer3/layer3.go
```diff
@@ -246,3 +248,24 @@ func WaitForRouterInterfaceToDetach(client *gophercloud.ServiceClient, routerInt
 		return false, nil
 	})
 }
+
+func GetNetworkIdByName(client *gophercloud.ServiceClient, name string)(string,error){
```
jtopjian: Isn't is possible to use the existing `IDFromName` function in the `networks` package?

jrperritt:
@dbaumgarten You've committed a file that has 16,000+ lines

dbaumgarten:
Using IDFromName from the networks package is definetly the right thing to do. I just wasn't aware of its existence.
I also just noticed this strange (and huge) file. It seems like i messed  up when using git...
I will fix this right now.

jtopjian:
@dbaumgarten Short of the above, the test looks good to me. Thanks for fixing it up.

As an aside, there are some existing scripts [here](https://github.com/gophercloud/gophercloud/tree/master/script/acceptancetest_environments) to build out devstack environments for acceptance tests. They handle setting the required environment variables like [`OS_NETWORK_ID`](https://github.com/gophercloud/gophercloud/blob/master/script/acceptancetest_environments/keystonev3-lbaasv2.sh#L198) which would be the equivalent of what one would set `OS_NETWORK_NAME` to, so it's possible to just nix the whole `OS_NETWORK_NAME` part.

As a second aside, [these](https://github.com/jtopjian/terraform-devstack) are more up-to-date scripts that I've been using. I should either replace the existing scripts in this repo to add a link pointing to them. [packstack-standard](https://github.com/jtopjian/terraform-devstack/tree/master/packstack-standard) has been my daily build environment for the past few months. Devstack started to get too hairy to maintain.

dbaumgarten:
@jtopjian Thanks for the hint. Using a standadised testing environment seems like a good idea.

Desprite the test-environment-scripts setting OS_NETWORK_ID, there is no trace of this variable in AcceptanceTestChoicesFromEnv() and it isn't documented in the acceptance readme.

Should we add OS_NETWORK_ID to AcceptanceTestChoicesFromEnv() and/or the readme?

jtopjian:
whoops! Looks like I got Terraform and Gophercloud mixed up. `OS_NETWORK_ID` is used in Terraform.

In that case, `OS_NETWORK_NAME` is totally fine. For my own sanity I might update Gophercloud to use `OS_NETWORK_ID` in the future. If you'd like to do that, go for it :)

Issue #54, opened 2016-08-15
Designate - Zone Create / Update / Delete


grahamhayes:
Add support for OpenStack DNS - Designate

jtopjian:
Resolved by #278

PR #278, opened 2017-03-05
DNS v2: Zone Create, Update, Delete

jtopjian:
For #54 
Supersedes #174 

The following URLs detail how this API works:

https://github.com/openstack/designate/blob/stable/newton/designate/objects/zone.py
https://github.com/openstack/designate/blob/stable/newton/designate/objects/adapters/base.py
https://github.com/openstack/designate/blob/stable/newton/designate/objects/adapters/api_v2/zone.py
https://github.com/openstack/designate/blob/stable/newton/designate/api/v2/controllers/zones/__init__.py
https://github.com/openstack/designate/tree/stable/newton/api-ref/source/samples/zones

@jrperritt This contains more than one method. I don't know how I can break it up while still retaining @grahamhayes's original work. I could chop it up, but then you'll be looking at a lot of red removed code.


jtopjian:
One odd bit about this one: the Delete method returns a response with a body. This might be to monitor the phases of the delete process. I don't think the built-in `client.Delete` method doesn't allow this, so a zone-specific one would have to be written. 

Using the standard `client.Delete` works and that's what's being used here. I figured adding the zone-specific Delete can be done later. Unless that's not a good idea and it should be implemented off the bat?

Adding the zone-specific Delete would also mean breaking the pattern of other Delete calls in Gophercloud. Instead of doing `ExtractErr()`, deletes would have to do `Extract()`. That's probably OK, and perhaps that's done elsewhere in OpenStack services that haven't been implemented, but something to just be aware of.

jtopjian:
@jrperritt Ready for review. The only thing I'm not sure about is the Delete stuff in the above comment.

jtopjian:
> The only thing I'm not sure about is the Delete stuff in the above comment.

Ah, nevermind. I see how I can do this. I'll work on it and re-notify.

jtopjian:
okie dokie. I think that should do it.

@jrperritt ready for review.

jrperritt:
Will review today.

openstack/dns/v2/zones/requests.go
```diff
@@ -55,3 +55,108 @@ func Get(client *gophercloud.ServiceClient, zoneID string) (r GetResult) {
...
+	Description string `json:"description,omitempty"`
+
+	// Name of the zone.
+	Name string `json:"name,required"`
+
+	// Masters specifies zone masters if this is a secondary zone.
+	Masters []string `json:"masters,omitempty"`
+
+	// TTL is the time to live of the zone.
+	TTL int `json:"-"`
```
jrperritt: No change necessary here, just a comment: `json:"ttl,omitempty"` should be fine here since 0 isn't allowed. That would clean up `ToZoneCreate/UpdateMap` a bit.

PR #279, opened 2017-03-05
DNS v2: RecordSet List, Get

jtopjian:
For #57 
Supersedes #175 

URLs:

https://github.com/openstack/designate/blob/master/designate/api/v2/controllers/recordsets.py
https://github.com/openstack/designate/blob/master/designate/api/v2/controllers/common.py
https://github.com/openstack/designate/blob/master/designate/objects/recordset.py
https://github.com/openstack/designate/blob/master/designate/objects/adapters/api_v2/recordset.py
https://github.com/openstack/designate/tree/stable/newton/api-ref/source/samples/recordsets

### DIFF
--- a/openstack/dns/v2/recordsets/results.go
+++ b/openstack/dns/v2/recordsets/results.go
@@ -1,6 +1,9 @@
 package recordsets
 
 import (
+	"encoding/json"
+	"time"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
@@ -9,10 +12,10 @@ type commonResult struct {
 	gophercloud.Result
 }
 
-// Extract interprets a GetResult, CreateResult or UpdateResult as a concrete Zone.
+// Extract interprets a GetResult, CreateResult or UpdateResult as a concrete RecordSet.
 // An error is returned if the original call or the extraction failed.
-func (r commonResult) Extract() (RecordSet, error) {
-	var s RecordSet
+func (r commonResult) Extract() (*RecordSet, error) {
+	var s *RecordSet
 	err := r.ExtractInto(&s)
 	return s, err
 }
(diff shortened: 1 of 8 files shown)


jrperritt:
@jtopjian Your `zones.Delete` function in this PR and the one it depends on differ in signature. After you rebase and fix merge conflicts, I'll review this one.

jtopjian:
@jrperritt Thanks! This has been fixed up.

openstack/dns/v2/recordsets/results.go
```diff
@@ -0,0 +1,113 @@
...
+	commonResult
+}
+
+// RecordSetPage is a single page of RecordSet results.
+type RecordSetPage struct {
+	pagination.LinkedPageBase
+}
+
+// IsEmpty returns true if the page contains no results.
+func (r RecordSetPage) IsEmpty() (bool, error) {
```
jrperritt: I don't _think_ we need these anymore. At least, that was my intention when I wrote [this](https://github.com/gophercloud/gophercloud/blob/master/pagination/linked.go#L78). Let me know if something breaks if this is deleted.

openstack/dns/v2/recordsets/results.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// CreatedAt is the date when the recordset was created.
+	CreatedAt time.Time `json:"-"`
+
+	// UpdatedAt is the date when the recordset was updated.
+	UpdatedAt time.Time `json:"-"`
+
+	// Links includes HTTP references to the itself,
+	// useful for passing along to other APIs that might want a recordset reference.
+	Links map[string]interface{} `json:"links"`
```
jrperritt: `[]gophercloud.Link`

openstack/dns/v2/recordsets/testing/fixtures.go
```diff
@@ -0,0 +1,168 @@
...
+            "zone_name": "example.com.",
+            "created_at": "2014-10-24T19:59:44.000000",
+            "version": 1,
+            "type": "A",
+            "status": "PENDING",
+            "action": "CREATE"
+        }
+    ],
+    "links": {
+        "self": "http://127.0.0.1:9001/v2/zones/2150b1bf-dee2-4221-9d85-11f7886fb15f/recordsets?limit=1"
```
jrperritt: We really want to be careful about page types. In this case, it does appear to be a `LinkPageBase`, but these must have `next` fields so that Gophercloud knows how to get to the next page. If the service uses a `LinkPageBase` that has a field named something other than `next`, we have to define a [`NextPageURL`](https://github.com/gophercloud/gophercloud/blob/master/pagination/linked.go#L24) method on the page type.

jtopjian: 👍 

I'll spend some time digging into the paging system. It's probably the area I'm least familiar with in Gophercloud.

jrperritt:
@jtopjian If you don't think you'll be able to get to the changes before the end of the weekend, just let me know, and I'll take over this PR.

jtopjian:
@jrperritt I _think_ I'm OK. I spent some time earlier today and will continue tomorrow. Just working on the Link stuff. If I'm still stuck tomorrow, I'll let you know.

openstack/dns/v2/recordsets/results.go
jtopjian: I'm getting errors when I remove this. I think it's required because the DNS API does not bundle links into an array.

jtopjian:
@jrperritt OK, I think I've got this sorted out.

All requested changes have been made, broken up into three commits:

1. General requested updates: I think these are all standard.
2. Pagination changes: 
  a. I think `IsEmpty` is required because `LinkedPageBase`'s `IsEmpty` is expecting an array?  
  b. I added a few unit tests to make sure paging is working correctly. An acceptance test shows it's working correctly with an actual API deployment.
3. Individual RecordSet links: the links here are, again, not an array, so I implemented a custom unmarshal. Is this correct?

If the above changes to paging and links is correct, I can amend the Zone stuff as it probably works the same way.

jrperritt:
> a. I think IsEmpty is required because LinkedPageBase's IsEmpty is expecting an array?

Correct, I missed that. The default `IsEmpty` should work for JavaScript arrays. It could also be modified to work with maps like this, but that's for another issue.

> Individual RecordSet links: the links here are, again, not an array

Of course they're not. How silly of me to expect consistency across services.

> , so I implemented a custom unmarshal. Is this correct?

Yeah, I think that's the way to go.

Thank you for finishing this up so quickly.

PR #176, opened 2016-11-01
[wip] Add Designate Recordset Create / Update / Delete

grahamhayes:
For #56 

https://github.com/openstack/designate/blob/stable/newton/designate/objects/adapters/api_v2/recordset.py#L24-L54

WIP as it depends on #173 and #175 (commit 2c31a49 and commit 02a787f) for base client setup

Issue #56, opened 2016-08-15
Designate - Recordset Create / Update / Delete


grahamhayes:

jtopjian:
Resolved by #280

PR #280, opened 2017-03-05
[Pending #279] DNS v2: RecordSet Create, Update, Delete

jtopjian:
For #56 
Supersedes #176 

URLs:

https://github.com/openstack/designate/blob/master/designate/api/v2/controllers/recordsets.py
https://github.com/openstack/designate/blob/master/designate/api/v2/controllers/common.py
https://github.com/openstack/designate/blob/master/designate/objects/recordset.py
https://github.com/openstack/designate/blob/master/designate/objects/adapters/api_v2/recordset.py
https://github.com/openstack/designate/tree/stable/newton/api-ref/source/samples/recordsets

### DIFF
--- a/openstack/dns/v2/recordsets/requests.go
+++ b/openstack/dns/v2/recordsets/requests.go
@@ -61,66 +61,103 @@ func Get(client *gophercloud.ServiceClient, zoneID string, rrsetID string) (r Ge
 	return
 }
 
-// UpdateOptsBuilder allows extensions to add additional attributes to the Update request.
+// CreateOptsBuilder allows extensions to add additional attributes to the Create request.
 type CreateOptsBuilder interface {
-	ToZoneCreateMap() (map[string]interface{}, error)
+	ToRecordSetCreateMap() (map[string]interface{}, error)
 }
 
-// UpdateOpts specifies the base attributes that may be updated on an existing server.
+// CreateOpts specifies the base attributes that may be used to create a RecordSet.
 type CreateOpts struct {
-	ZoneID  string   `json:"zone_id,omitempty"`
-	Name    string   `json:"name,omitempty"`
-	Type    string   `json:"type,omitempty"`
+	// Name is the name of the RecordSet.
+	Name string `json:"name,required"`
+
+	// Description is a description of the RecordSet.
+	Description string `json:"description,omitempty"`
+
+	// Records are the DNS records of the RecordSet.
 	Records []string `json:"records,omitempty"`
-	TTL     int      `json:"ttl,omitempty"`
+
+	// TTL is the time to live of the RecordSet.
+	TTL int `json:"-"`
+
+	// Type is the RRTYPE of the RecordSet.
+	Type string `json:"type,omitempty"`
 }
 
-// ToServerUpdateMap formats an UpdateOpts structure into a request body.
-func (opts CreateOpts) ToZoneCreateMap() (map[string]interface{}, error) {
-	return gophercloud.BuildRequestBody(opts, "")
+// ToRecordSetCreateMap formats an CreateOpts structure into a request body.
+func (opts CreateOpts) ToRecordSetCreateMap() (map[string]interface{}, error) {
+	b, err := gophercloud.BuildRequestBody(opts, "")
+	if err != nil {
+		return nil, err
+	}
+
+	if opts.TTL > 0 {
+		b["ttl"] = opts.TTL
+	} else {
+		b["ttl"] = nil
+	}
+
+	return b, nil
 }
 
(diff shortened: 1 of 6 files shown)


jrperritt:
@jtopjian If you'd like to rebase this one, I'll review it.

jtopjian:
@jrperritt rebased, but I haven't done a second review on it, so it might be a little rough. Feel free to review, or I can double-check it in the next day or so.

openstack/dns/v2/recordsets/requests.go
```diff
@@ -60,3 +60,104 @@ func Get(client *gophercloud.ServiceClient, zoneID string, rrsetID string) (r Ge
...
+	}
+	_, r.Err = client.Post(baseURL(client, zoneID), &b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{201, 202},
+	})
+	return
+}
+
+// UpdateOptsBuilder allows extensions to add additional attributes to the Update request.
+type UpdateOptsBuilder interface {
+	ToRecordSetSetUpdateMap() (map[string]interface{}, error)
```
jrperritt: `ToRecordSetUpdateMap`

openstack/dns/v2/recordsets/requests.go
```diff
@@ -60,3 +60,104 @@ func Get(client *gophercloud.ServiceClient, zoneID string, rrsetID string) (r Ge
...
+	if opts.TTL > 0 {
+		b["ttl"] = opts.TTL
+	} else {
+		b["ttl"] = nil
+	}
+
+	return b, nil
+}
+
+// Create changes the service type of an existing service.
```
jrperritt: emend

openstack/dns/v2/recordsets/requests.go
```diff
@@ -60,3 +60,104 @@ func Get(client *gophercloud.ServiceClient, zoneID string, rrsetID string) (r Ge
...
+}
+
+// ToRecordSetCreateMap formats an CreateOpts structure into a request body.
+func (opts CreateOpts) ToRecordSetCreateMap() (map[string]interface{}, error) {
+	b, err := gophercloud.BuildRequestBody(opts, "")
+	if err != nil {
+		return nil, err
+	}
+
+	if opts.TTL > 0 {
```
jrperritt: I think this block can be removed by adding `ttl,omitempty` to the `TTL` field in `CreateOpts`.

openstack/dns/v2/recordsets/requests.go
```diff
@@ -60,3 +60,104 @@ func Get(client *gophercloud.ServiceClient, zoneID string, rrsetID string) (r Ge
...
+
+// CreateOptsBuilder allows extensions to add additional attributes to the Create request.
+type CreateOptsBuilder interface {
+	ToRecordSetCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts specifies the base attributes that may be used to create a RecordSet.
+type CreateOpts struct {
+	// Name is the name of the RecordSet.
+	Name string `json:"name,required"`
```
jrperritt: `json:"name" required:"true"`

jtopjian:
@jrperritt wrt `TTL`, it's one of those fields that should be added to the request body with a `null` value sometimes. This is for when the RR should inherit the TTL of the zone. It can happen during creation as well as update (removing its TTL and using the zone's).

I had sudden flashbacks to spending ~2 hours observing how the TTL behaves when I was originally working on this PR.

jrperritt:
@jtopjian What if I don't want the zone to have a TTL? Can I leave it empty? If so, it should probably be a `*int`, where `nil` means no TTL, 0 means inherited TTL, and > 0 means provided TTL. I'm remembering this happening somewhere else in this library before. I think that's how we handled it then, but I not sure.

jrperritt:
Actually, I think I got that backwards: it should probably be 0 means no TTL, and nil means inherit from parent.

jtopjian:
@jrperritt OK, ran some tests again.

`TTL` can be omitted during creation. If omitted, the RR uses the Zone's TTL. Note that this is not reflected in the API. The TTL will be returned with a `null` value, which the OpenStack client renders as `None` and Gophercloud as 0.

But actual DNS requests will return the RR with the TTL of the Zone.

If the TTL is >0, then the it overrules the Zone's TTL.

If TTL is set to 0 (`*int`), the server throws an error.

During updates, a `null` field can clear a previously set TTL. So I think the way TTL is handled in update is correct.

AFAICT, it's not possible to not set a TTL. Inheriting and not setting are really the same thing.

jrperritt:
OK, SGTM

jtopjian:
Thanks!

Issue #273, opened 2017-03-01
RouterID for floatingips


dbaumgarten:
Hi,

the field router_id is missing in the ListOpts and the FloatingIP structs.

The documentation states it should be there and my test indicate the same.

I have the required commit ready (~4 lines), but did not yet create a PR, because I can't pass the acceptance tests because of #272 .

jtopjian:
Just to add some info to this, I think [this](https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/l3.py) is where the floating IP fields are specified. The fields match up with the [API docs](https://developer.openstack.org/api-ref/networking/v2/index.html?expanded=list-floating-ips-detail).

jrperritt:
closed by #283

PR #283, opened 2017-03-06
Added RouterID field to floatingip

dbaumgarten:
For #273

Source: https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/l3.py
Docs: https://developer.openstack.org/api-ref/networking/v2/index.html?expanded=list-floating-ips-detail


jrperritt:
Looks good

Issue #216, opened 2016-12-18
Image Metadata/Properties Type


jtopjian:
I'm still researching this one, but it looks as though image metadata and properties can hold more than `map[string]string`.

This was originally discovered [here](https://github.com/hashicorp/terraform/issues/10767) and per the output [here](https://github.com/hashicorp/terraform/issues/10767#issuecomment-267507276),  the properties of the image contains a nested array.

I was able to reproduce the reported issue in question by setting similar metadata/properties on an existing image:

```shell
$ openstack --debug image set --property block_device_mapping='[{"device_name": "/dev/vdb", "boot_index": -1, "source_type": "blank", "destination_type": "local", "volume_size": 1, "delete_on_termination": true}]' jttest
```

Similarly, I was able to add arbitrary arrays and hashes/maps as properties:

```shell
$ openstack --debug image set --property foo='[{"a": "b"}]' jttest
$ openstack --debug image set --property bar='["a","b","c"]' jttest
$ openstack --debug image set --property baz='{"a": "b"}' jttest
```

So it doesn't seem like the API is enforcing a schema for properties like this (meaning: `block_device_mapping` isn't an exception to the otherwise `map[string]string`).

When Gophercloud hits a property like this, the following error is triggered:

```
Unexpected error from Get: json: cannot unmarshal array into Go value of type string
```

This is reproducible both in unit and acceptance tests.

Changing `map[string]string` to `map[string]interface{}` quickly resolves the issue, but I'm unsure if that's the best solution or if that's _all_ that is required to fix this.

Regardless of the fix, I believe both the `compute/v2/images` and `images/v2/images` packages will need updated.

TODO (for either myself or if anyone else wants to do this):

* Dig into the Glance code and find where it's setting metadata/properties
* Determine the difference between "metadata" and "properties". The `images/v2/images` defines both, but the `openstack` command-line client only allows updates to `properties`. `properties` are displayed as `metadata` through the Compute API (which I understand will be deprecated soon).

dvianello:
Hi @jtopjian, any news on this? We're currently hit pretty badly by this bug at the moment!

jtopjian:
I'll admit to total procrastination on this one. I'll make some time for this soon. :)

jtopjian:
Good news: This is, in fact, a special case for `block_device_mapping` and only in Compute Image API:

https://github.com/openstack/nova/blob/master/nova/image/glance.py#L672-L721

[...]

PR #285, opened 2017-03-07
Compute v2: Handle JSON Image Metadata

jtopjian:
The Compute API will attempt to parse JSON values for the image metadata values of `block_device_mapping` and `mappings`.

For #216 

@jrperritt I can't seem to get this to work. Do you mind taking a look when you have time? The unit tests are set up with the current error I'm seeing.

### DIFF
--- a/openstack/compute/v2/images/results.go
+++ b/openstack/compute/v2/images/results.go
@@ -46,7 +46,7 @@ type Image struct {
 
 	Updated string
 
-	Metadata map[string]string
+	Metadata map[string]interface{}
 }
 
 // ImagePage contains a single page of results from a List operation.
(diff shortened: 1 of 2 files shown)


jrperritt:
You shouldn't need the `UnmarshalJSON` method; changing to `map[string]interface{}` should suffice. I don't have time to delve into the issue, but the values look correct; the problem is likely with the test: you're using `reflect.DeepEqual`, which recursively checks types as well. I'd guess the `boot_index` fields have different number types (something like int vs float). That's not this library's concern though; after successfully unmarshaling the `Metadata` into a `map[string]interface{}`, it's up to the user to handle it. You can confirm by converting both `boot_index` values to `int` (or `float{32,64}`) in the test.

jtopjian:
>  the problem is likely with the test: you're using reflect.DeepEqual, which recursively checks types as well.

Yeah, that's exactly where the problem is being hit. The order of the keys in the map should not matter with `reflect.DeepEqual`, right? 

I agree about the Unmarshal. That was used for two other attempts I was playing around with: first, omitting `block_device_mapping` entirely and also trying to re-marshal the JSON back to a string.

I'll take another run at this. Thanks!

jtopjian:
@jrperritt You were right -- it was `boot_index` causing the problem. Thank you for the pointer :)

This is ready for review.

Issue #252, opened 2017-02-14
support for v3 auth is part broken, part incomplete


majewsky:
Our corporate OpenStack installations support Identity v3 exclusively, and they use the whole feature set, down to users from one domain having roles in projects in different domains. When I look at [`func AuthOptionsFromEnv`](https://github.com/gophercloud/gophercloud/blob/50cdddf51c54e8df178af065055cbc5b3cf442e1/openstack/auth_env.go#L15), about half of the environment variables that I need are missing, and for some, the implementation in Gophercloud is wrong.

I'll recap the terminology of the OpenStack Identity API here in case some readers are not familiar with it: In the v2 API, users are limited to *tenants*, and users can only do work in their respective tenant, so the sufficient auth options are:

```go
UserName   string // from $OS_USERNAME
UserID     string // from $OS_USERID
TenantName string // from $OS_TENANT_NAME
TenantID   string // from $OS_TENANT_ID
```

That's also what's in `type AuthOptions` (plus endpoint URL plus password). In the v3 API, users reside in *domains*, and tenants (now called *projects*) do so, too. A token can be scoped to a domain, to a project in a domain (not necessarily the user's domain), or to nothing. The sufficient auth options in this case are: [(reference)](https://github.com/openstack/python-openstackclient/blob/83cf09df28e200ba40f403504db3b11e623978fd/doc/source/man/openstack.rst#environment-variables)

```go
UserName          string // from $OS_USERNAME
UserID            string // from $OS_USERID
UserDomainName    string // from $OS_USER_DOMAIN_NAME
UserDomainID      string // from $OS_USER_DOMAIN_ID
//for domain scope:
DomainName        string // from $OS_DOMAIN_NAME
DomainID          string // from $OS_DOMAIN_ID
//for project scope:
ProjectName       string // from $OS_PROJECT_NAME
ProjectID         string // from $OS_PROJECT_ID
ProjectDomainName string // from $OS_PROJECT_DOMAIN_NAME
ProjectDomainID   string // from $OS_PROJECT_DOMAIN_ID
```

[...]

PR #286, opened 2017-03-07
Update some docs for some nits found

emonty:
For #252

I was reading through some of the docs and just noticed a few things. Most of them are trivial, but I figured I'd make a PR anyway. It's possible that some of the words in this might help with the situation described in 252 (cross-domain scoping should work fine if projects are specified by ID. specifying domain information is only useful if the user wants to specify both user and project by name - but since both can be specified by id, then it's really not an actual issue)

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ import (
 
 // Option 1: Pass in the values yourself
 opts := gophercloud.AuthOptions{
-  IdentityEndpoint: "https://my-openstack.com:5000/v2.0",
+  IdentityEndpoint: "https://openstack.example.com:5000/v2.0",
   Username: "{username}",
   Password: "{password}",
 }
--- a/auth_options.go
+++ b/auth_options.go
@@ -1,7 +1,7 @@
 package gophercloud
 
 /*
-AuthOptions stores information needed to authenticate to an OpenStack cluster.
+AuthOptions stores information needed to authenticate to an OpenStack Cloud.
 You can populate one manually, or use a provider's AuthOptionsFromEnv() function
 to read relevant information from the standard environment variables. Pass one
 to a provider's AuthenticatedClient function to authenticate and obtain a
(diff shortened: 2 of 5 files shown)


cdent:
I heartily approve and would like to subscribe to your newsletter.

👍 

(I reviewed this and like it.)

auth_options.go
```diff
@@ -31,9 +31,16 @@ type AuthOptions struct {
 	DomainName string `json:"name,omitempty"`
 
 	// The TenantID and TenantName fields are optional for the Identity V2 API.
+	// The same fields are knows as project id and project name in the Identity
```
jrperritt: s/knows/known
s/project name/project name, respectfully,/

README.md
```diff
@@ -74,7 +74,7 @@ import (
 
 // Option 1: Pass in the values yourself
 opts := gophercloud.AuthOptions{
-  IdentityEndpoint: "https://my-openstack.com:5000/v2.0",
+  IdentityEndpoint: "https://openstack.example.com:5000/v2.0",
```
jrperritt: Why change all these URLs? [curious]

doc.go
```diff
@@ -4,11 +4,16 @@ clouds. The library has a three-level hierarchy: providers, services, and
 resources.
 
 Provider structs represent the service providers that offer and manage a
-collection of services. Examples of providers include: OpenStack, Rackspace,
-HP. These are defined like so:
+collection of services. Examples of public OpenStack cloud providers include:
```
jrperritt: Instead of listing a bunch of providers, I'd rather just scrap this sentence.

auth_options.go
emonty: K. Fixing

README.md
emonty: Bah. I explained in the individual commit, but the PR doesn't really show that ...

"RFC 2606 defines example.com as a reserved domain to be used in tests
and examples. It is guaranteed to not be actually resolvable. Update
the uses of my-openstack.com to use example.com instead."

Essentially, there's a domain that's built in to the RFCs for this.

doc.go
emonty: WFM!

README.md
jrperritt: Interesting :+1:

jrperritt:
@emonty Thank you for the doc updates

Issue #287, opened 2017-03-09
Identity v3: Users CRUD


jtopjian:
* [x] list / get (#288)
* [x] create (#392)
* [x] delete (#393)
* [x] update (#394)

General API reference: https://developer.openstack.org/api-ref/identity/v3/index.html#users

API processing: 
* https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/controllers.py#L193
* https://github.com/openstack/keystone/blob/master/keystone/identity/backends/sql.py

Schema:
* https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/schema.py#L65-L74
* https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/backends/sql_model.py#L28

Options:
* https://github.com/openstack/keystone/blob/master/keystone/identity/backends/resource_options.py
* https://github.com/openstack/keystone/blob/4c83fb0f13d0a09db7a4ca66c3b0dc2745c34c5f/releasenotes/notes/bp-per-user-auth-plugin-reqs-feb95fd907be4b40.yaml
* https://github.com/openstack/keystone/blob/1ec73f3b97337bc3d98f34734abded8639fcffc9/releasenotes/notes/bug-1659995-f3e716de743b7291.yaml

List:
* https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/controllers.py#L227

Get:
* https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/controllers.py#L242

Create:
* https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/controllers.py#L214
* https://github.com/openstack/keystone/blob/master/keystone/identity/backends/sql.py#L127
* https://github.com/openstack/keystone/blob/master/keystone/identity/schema.py#L80

Delete:
* https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/controllers.py#L275
https://github.com/openstack/keystone/blob/master/keystone/identity/backends/sql.py#L330

Update:
[...]

Issue #260, opened 2017-02-17
ImageService v2: Unable to Read Properties


jtopjian:
The current `Image` struct looks for a JSON [`properties`](https://github.com/gophercloud/gophercloud/blob/ff77917b4d967827a1a375c97214f06b030e519a/openstack/imageservice/v2/images/results.go#L66) object, but one isn't actually returned by the server. Instead, properties are mixed in with all of the other top-level key/values.

I haven't dug into the Glance API / code enough to see what the best way of dealing with this is yet. I see the `python-glanceclient` making requests for a schema, so it's possible a schema is used as the basis of what is considered a "base" property and what is considered extra. 

The easier solution might be to just compare what else has been sent down by the server and if it's not defined in the `Image` struct, add it to the `Properties` map.

jrperritt:
There is a Go feature (issue #15314) planned to be added soon (Go1.9early) that I think would solve this for us (a `[]string` would be returned in the `Unmarshal` `error` that would contain any unused fields). Until then, I think we should your easier solution.

jrperritt:
@jtopjian I think `RemainingKeys` could be used for this, no?

jtopjian:
Yep :)

jtopjian:
Fixed by #376

PR #288, opened 2017-03-09
Identity v3: Users Get / List

jtopjian:
For #287

### DIFF
--- a/openstack/identity/v3/users/requests.go
+++ b/openstack/identity/v3/users/requests.go
@@ -20,7 +20,7 @@ type ListOpts struct {
 	Enabled *bool `q:"enabled"`
 
 	// IdpID filters the response by an Identity Provider ID.
-	IdPID string `q:"enabled"`
+	IdPID string `q:"idp_id"`
 
 	// Name filters the response by username.
 	Name string `q:"name"`
@@ -56,35 +56,8 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 	})
 }
 
-// GetOptsBuilder allows extensions to add additional parameters to
-// the Get request.
-type GetOptsBuilder interface {
-	ToUserGetQuery() (string, error)
-}
-
-// GetOpts allows you to modify the details included in the Get request.
-type GetOpts struct{}
-
-// ToUserGetQuery formats a GetOpts into a query string.
-func (opts GetOpts) ToUserGetQuery() (string, error) {
-	q, err := gophercloud.BuildQueryString(opts)
-	return q.String(), err
-}
-
 // Get retrieves details on a single user, by ID.
-func Get(client *gophercloud.ServiceClient, id string, opts GetOptsBuilder) (r GetResult) {
-	url := getURL(client, id)
-	if opts != nil {
-		query, err := opts.ToUserGetQuery()
-		if err != nil {
-			r.Err = err
-			return
-		}
-		url += query
-	}
-
-	_, r.Err = client.Get(url, &r.Body, &gophercloud.RequestOpts{
-		OkCodes: []int{200},
-	})
+func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
+	_, r.Err = client.Get(getURL(client, id), &r.Body, nil)
 	return
 }
--- a/openstack/identity/v3/users/requests.go
+++ b/openstack/identity/v3/users/requests.go
@@ -1,0 +1,90 @@
+package users
+
+import (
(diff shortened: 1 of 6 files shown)


jtopjian:
@jrperritt Ready for review

jtopjian:
Found a missing field while working on Create. That's all of them now.

jtopjian:
Nope. Something is strange here. Making WIP for now.

jtopjian:
@jrperritt Looks like Keystone also supports an arbitrary number of properties like Glance does in #260. There's little documentation on this, but search for "email" [here](https://github.com/openstack/keystone/blob/464ec0b3da48d0ca294dbf58e7409d013a2c4528/doc/source/devref/api_curl_examples.rst) and note this [`to_dict`](https://github.com/openstack/keystone/blob/a926700cf49895313f553b37a9186c625fc9ce4d/keystone/common/sql/core.py#L138) function. 

The second link is noteworthy because it seems Keystone switches between adding the extra properties to an `extra` dict _or_ just adds them to the general top-level JSON structure.

So while I was trying to avoid #260, I ended up having to solve it anyway.

No doubt the solution I came up with in the latest commit isn't the cleanest way of doing this. Can you take a look when you have time?

jtopjian:
@jrperritt The `extra` field here is what I was talking about for #194. In [this](https://github.com/gophercloud/gophercloud/pull/288/commits/8c6fd3a301f410fe58fcfa5939e509b40928454b) commit, I'm using `reflect` to bundle the extra fields into the `extra` field. I'm all thumbs when it comes to `reflect`, so let me know if there's a better way of doing this.

In the case of this PR, `extra` is the name of the actual field. #316 added `AdditionalProperties` to the request body because that was also the name in the API. #260 will be `Properties`. But when there are cases when we add our own "extra" field, we should probably standardize on a name. `Extra`?

jrperritt:
>  let me know if there's a better way of doing this.

I think that way is correct. That functionality would seem appropriate to put in a utility file in the `internal` package for other `Extract` function to use.

> added AdditionalProperties to the request body because that was also the name in the API

Ah, I didn't realize that.

> we should probably standardize on a name. Extra?

Other ideas: `Overflow`, `More`, `Other`.

jrperritt:
@jtopjian Was this just waiting on `RemainingKeys`?

jtopjian:
@jrperritt Yeah, I'll be working on this this week.

jtopjian:

I'm not sure what happened, but somewhere along the way my branch got mangled. Since I had to rebase, I figure I'd squash the commits.

I reviewed everything today and even found a new user attribute of `Options`. The links in #287 have been updated, too.

openstack/identity/v3/users/requests.go
```diff
@@ -0,0 +1,90 @@
...
+// ListOpts allows you to query the List method.
+type ListOpts struct {
+	// DomainID filters the response by a domain ID.
+	DomainID string `q:"domain_id"`
+
+	// Enabled filters the response by enabled users.
+	Enabled *bool `q:"enabled"`
+
+	// IdpID filters the response by an Identity Provider ID.
+	IdPID string `q:"enabled"`
```
jrperritt: `q:"?"`

openstack/identity/v3/users/requests.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// GetOptsBuilder allows extensions to add additional parameters to
+// the Get request.
+type GetOptsBuilder interface {
+	ToUserGetQuery() (string, error)
+}
+
+// GetOpts allows you to modify the details included in the Get request.
+type GetOpts struct{}
```
jrperritt: What are the options here?

jtopjian: There aren't any. This is only to satisfy the `ToUserGetQuery` method for possible extension. Should this be removed altogether?

jrperritt: I don't think any of the other services offer this. I'm trying to think of what options could even by provided considering the ID uniquely identifies the user.

jtopjian: I agree. I did the same for Projects, but in hindsight, I have no idea why. I _must_ have had a reason or known of some customization in the wild.

I'm OK with removing this and adding it later if it comes up.

jrperritt:
This one looks good to me. We'll probably want to remove the `GetOptionsBuilder` from `v3/projects` in a separate PR.

jtopjian:
Thanks!

I'll submit a PR for `v3/projects` now while it's on my mind.

PR #295, opened 2017-03-15
use $GOPATH in the readme of acceptance tests

mvalkon:
Enforce `$GOPATH` instead of `~/go` in `gophercloudest()` bash function. Everyone should have `$GOPATH` set, but it may not point to ~/go.

For #296


mvalkon:
@jtopjian @jrperritt please check this. Did not create an issue for such a small change but can do so if required.

jtopjian:
LGTM, but might want to make a quick issue anyway for consistency. :)

Issue #298, opened 2017-03-17
Network v2: Port Security Extension


jtopjian:
The Port Security extension defines a single attribute (`port_security_enabled`) which will toggle Port Security on individual ports as well as all ports on a specific network.

https://github.com/openstack/neutron-lib/blob/65fc1a49af00ba660d99b3f187602ceb35e739fa/neutron_lib/api/definitions/port_security.py

* [x] Update External Network extension
* [x] Update Port Binding extension
* [x] Climb out of rabbit hole.
* [x] Get / List Extensions (#299)
* [x] Create Extensions
* [x] Update Extensions

jrperritt:
@jtopjian all done here?

jtopjian:
Yes! 3 days to the 1 year mark :smile: 

I remember starting this one thinking it'd be done in an afternoon.

PR #299, opened 2017-03-17
Network v2: Port Security Get / List

jtopjian:
For #298


jtopjian:
@jrperritt March was a few weeks ago, right? :)

We use port security at work and are able to get around the lack of support in Gophercloud by using Terraform's `value_specs` field, but I just noticed this PR sitting in queue and wanted to revisit adding proper support.

Ping for review.

`openstack/networking/v2/extensions/portsecurity/results.go` is the only actual implemented feature. Everything else is tests and docs.

@dklyle "extensions" are probably the most difficult implementation in Gophercloud (both in terms of initially understanding how to use them and how they're implemented in the core of Gophercloud). This PR is a pretty clean example of how to implement it. Let me know if you have questions.

Issue #301, opened 2017-03-19
Netowrk v2: multiple provider extention


takaishi:
We can't create multiple provider network now.
So, add attribute "segments" to bind multiple physical network for logical network.

### api docuemnt

"segments" is array of segment object, segment has provider:network_type and provider:physical_network, provider:segmentation_id.

- https://developer.openstack.org/api-ref/networking/v2/?expanded=create-network-detail,update-segment-detail#create-network

### neutron source code
- https://github.com/openstack/neutron/blob/stable/mitaka/neutron/extensions/multiprovidernet.py#L24
- https://github.com/openstack/neutron/blob/stable/mitaka/neutron/extensions/providernet.py#L22

jtopjian:
Resolved by #302

Issue #300, opened 2017-03-18
Problem with Unmarshalling Extensions


jtopjian:
I wanted to add the `port_security_enabled` extension and ran into some problems...

The extension adds support for extra fields in the request and extra fields in the results. I wanted to tackle the latter first, thinking it would be as simple as the other extensions (volume tenants, server availability zones) that were recently added. Reconfiguring the `networks` package to account for the modified `Extract` functions/methods caused other network extension packages to break.

Things were going well with fixing the broken extensions until I ran into the `provider` extension which requires `UnmarshalJSON` to smooth over the `SegmentationID` field.

It looks like there's difficulties triggering `UnmarshalJSON` on anonymous/nested structs:

* golang github issue 17877
* [this](http://www.howtobuildsoftware.com/index.php/how-do/HX6/go-why-embedded-structs-are-not-decoded-when-there-is-an-unmarshal-implementation) might also be relevant.

The result is that the following pattern no longer works:

```go
type NetworkWithExt struct {
  networks.Network
  provider.NetworkProviderExt
}
```

`UnmarshalJSON` is triggered, but it strips out the `Network` struct. This can be worked around by embedding `networks.Network` in the `provider.NetworkProviderExt` struct, but that breaks the above pattern and might possibly break when someone wants to add another extension (`portsbinding`) to the mix.

The latest 4 commits of [this](https://github.com/jtopjian/gophercloud/commits/networkv2-extensions-refactor) branch are my attempt to cleanly show the problem. There are some notes in `provider`'s unit tests with some additional information.

I'm not sure what the best solution would be here. I like the ability to compose structures with extended fields, but if it only works on extensions that don't require type modifications, that makes things a bit complicated.

I think the magical best solution would be if there was a way to retrieve a network or server result and it automatically included all of the extensions, but I have no idea how that would work. Each extension somehow register itself as an extension to its parent. When the parent is unmarshalled, it applies an extract on each registered extension. Since extract would only grab the fields configured in its struct, it would ignore all other fields, which is kind of what's done [here](https://github.com/jtopjian/gophercloud/blob/78ad5ba5e226e1895fa22002c08be870127a29cd/openstack/networking/v2/extensions/provider/testing/results_test.go#L49-L62). Of course, there's been the discovery of two resources that place arbitrary fields in the JSON body.

In any case, I'm stuck. It's entirely possible I'm overlooking something really simple, too.

jrperritt:
[...]

PR #302, opened 2017-03-19
network v2: support to create multiple provider

takaishi:
For https://github.com/gophercloud/gophercloud/issues/301

- https://github.com/openstack/neutron/blob/stable/mitaka/neutron/extensions/multiprovidernet.py#L24
- https://github.com/openstack/neutron/blob/stable/mitaka/neutron/extensions/providernet.py#L22

refs: https://github.com/hashicorp/terraform/pull/10265


jtopjian:
@takaishi Thanks for this. 

Just a note that there's already a [`provider`](https://github.com/gophercloud/gophercloud/tree/master/openstack/networking/v2/extensions/provider) package. Right now it only supports retrieval of the provider attributes. The modifications you made to the `networking/v2/networks` package should be moved there.

To do that, check out how the [`keypairs`](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/keypairs/requests.go#L9) compute extension is configured with `CreateOptsExt`.

And another note: the `provider` package will be modified soon so it works similar to how other extensions are starting to be structured (see [here](https://github.com/gophercloud/gophercloud/tree/master/openstack/blockstorage/extensions/volumetenants) and [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/availabilityzones/results.go) for examples). See #298 and #300 for more details on the status of this. Working on the `CreateOptsExt` and `UpdateOptsExt` stuff shouldn't be affected, though.

takaishi:
@jtopjian Thank you for comment.
I moved segments to `provider` from `network`.
So, if `provider` package will refactor soon, should I wait it?

jtopjian:
@takaishi I think working on the create and update portions is safe to do. The only piece that will change is the "read". There hasn't been a PR opened for that yet, so it's still yet to be done, too.

Thanks for making those changes. Another existing example to look at is the FWaaS [`routerinsertion`](https://github.com/gophercloud/gophercloud/tree/master/openstack/networking/v2/extensions/fwaas/routerinsertion) extension.

Note how it doesn't need a `Create` function. This might have been the confusing part about the `keypairs` extension because it included both the ability to extend the Server create (to inject a keypair into the server) as well as the ability to create keypairs.

The `routerinsertion` extension, on the other hand, simply extends the FWaaS create, which is what you want to do here with the provider stuff. You can see how this works [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/fwaas/routerinsertion/testing/requests_test.go#L56-L68).

Ideally, the `provider` extension will work like this:

```go
networkCreateOpts := neworks.CreateOpts{
        Name: "mynetwork",
}

providerCreateOpts := provider.CreateOptsExt{
        CreateOptsBuilder: networkCreateOpts,
        SegmentationID: "1234567890",
}

network, err := networks.Create(client, providerCreateOpts).Extract()
```

See how it's extending the original Network options and still using the original `networks.Create` function? Because of this, `CreateOptsExt` in `requests.go` only needs to contain the _new_ fields.

I'm assuming you're adding this functionality for Terraform, so to use Terraform as an example, see [here](https://github.com/hashicorp/terraform/blob/master/builtin/providers/openstack/resource_openstack_compute_instance_v2.go#L394-L446) for how the Server `createOpts` is continually extended depending on the features that were requested.

Let me know if that makes sense - I'd be happy to explain in more detail.

takaishi:
@jtopjian 

OK, I will create PRs.
I understand to write extention code and fixed, Thank you!

jtopjian:
@takaishi If I understand correctly, the Neutron extension `provider` enables a single provider network to be added into the `network` body:

```json
{
    "network": {
        "status": "ACTIVE",
        "subnets": [
            "54d6f61d-db07-451c-9ab3-b9609b6b6f0b"
        ],
        "name": "private-network",
        "admin_state_up": true,
        "tenant_id": "4fd44f30292945e481c7b8a0c8908869",
        "shared": true,
        "id": "d32019d3-bc6e-4319-9c1d-6722fc136a22",

        "provider:physical_network": null,
        "provider:network_type": "local",
        "provider:segmentation_id": null
    }
}
```

and the `multiprovider` extension adds support for a `segments` list that contains one or more `provider` fields:

```json
{
	"network": {
			"name": "sample_network",
			"admin_state_up": true,
			"shared": true,
			"tenant_id": "12345",
			"segments": [
				{
					"provider:segmentation_id": 666,
					"provider:physical_network": "br-ex",
					"provider:network_type": "vxlan"
				},
				{
					"provider:segmentation_id": 615,
					"provider:physical_network": "br-ex",
					"provider:network_type": "vxlan"
				}
			]
	}
}
```

Is it safe to say that if someone would create a `multiprovider` request with only 1 `segment` element, it is the same as a standard request for the `provider` extension? If this is true, it would make sense to always just use `segments` in apps like Terraform to make things simple.

takaishi:
@jtopjian Yes, when create `multipleprovider` with segments has 1 provider, network response is same `provider`.
I agree to use `segments` always.

takaishi:
@jrperritt This PR is ready to review.

takaishi:
@jtopjian This PR is ready to review.

jtopjian:
👍 

I'm at the OpenStack Summit at the moment so it might take me a few days to get to this, but I'll definitely review it soon :)

openstack/networking/v2/extensions/provider/requests.go
```diff
@@ -0,0 +1,30 @@
+package provider
+
+import (
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+)
+
+type CreateOptsExtBuilder interface {
+	ToProviderCreateMap() (map[string]interface{}, error)
+}
```
jtopjian: This type can be removed.

openstack/networking/v2/extensions/provider/results.go
```diff
@@ -49,6 +49,16 @@ type NetworkExtAttrs struct {
 	// instance, if network_type is vlan, then this is a vlan identifier;
 	// otherwise, if network_type is gre, then this will be a gre key.
 	SegmentationID string `json:"provider:segmentation_id"`
+
+	// Array og Segment for define multiple physical binfig for logical network.
```
jtopjian:
```go
// Segments is an array of Segment which defines multiple physical bindings to logical networks.
```

openstack/networking/v2/extensions/provider/requests.go
```diff
@@ -0,0 +1,30 @@
...
+
```
jtopjian:
```go
// CreateOptsExt adds a Segments option to the base Network CreateOpts.
```

openstack/networking/v2/extensions/provider/requests.go
```diff
@@ -0,0 +1,30 @@
...
+type CreateOptsExt struct {
+	networks.CreateOptsBuilder
+	Segments        []Segment `json:"segments,omitempty"`
+}
+
+func (opts CreateOptsExt) ToNetworkCreateMap() (map[string]interface{}, error) {
```
jtopjian:
```go
// ToNetworkCreateMap adds segments to the base network creation options.
```

openstack/networking/v2/extensions/provider/results.go
```diff
@@ -49,6 +49,16 @@ type NetworkExtAttrs struct {
...
+	Segments []Segment `json:"segments"`
+}
+
+// Define physical binding for logical network.
```
jtopjian:
```go
// Segment defines a physical binding to a logical network.
```

openstack/networking/v2/extensions/provider/urls.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "github.com/gophercloud/gophercloud"
+
+func rootURL(c *gophercloud.ServiceClient) string {
+	return c.ServiceURL("networks")
+}
+
+func createURL(c *gophercloud.ServiceClient) string {
+	return rootURL(c)
+}
```
jtopjian: This file can be removed

takaishi:
@jtopjian Thank you for review! I fixed docs and removed type & file.

jtopjian:
@takaishi This looks good to me - thanks!

Issue #243, opened 2017-01-29
Display error message returned from OpenStack


takaishi:
Hello,
When OpenStack return following error message:

`{"badRequest": {"message": "Invalid IP protocol 112.", "code": 400}}`

but `ErrorDefault400.Error()` is `Invalid request due to incorrect syntax or missing required parameters.` (https://github.com/gophercloud/gophercloud/blob/master/errors.go#L106)
If response has message, Error() should return response message?
For example: 

```go
type BadRequestBody struct {
	BadRequest struct {
		Message string `json:"message"`
		Code    int    `json:"code"`
	} `json:"badRequest"`
}

func (e ErrDefault400) Error() string {
	if string(e.Body) != "" {
		body := new(BadRequestBody)
		err := json.Unmarshal(e.Body, body)
		if err != nil {
			return fmt.Sprintf("%s", err)
		}
		return body.BadRequest.Message
	} else {
		return "Invalid request due to incorrect syntax or missing required parameters."
	}
}

```

khdegraaf:
I am working on a fix (although in my private fork) for this problem.  Hopefully will have time to do PR in official repository in a couple weeks.

Xiol:
Any word on when this is likely to be resolved?

jtopjian:
I've just merged #703 which should resolve this issue. Apologies for the delay.

PR #303, opened 2017-03-19
Compute V2: Display error message when 400 Bad Request.

takaishi:
For #243

- https://github.com/openstack/nova/blob/015870d05d7e0848a6d0d9d275370890ee3b9829/nova/api/openstack/wsgi.py#L1010
- https://github.com/openstack/nova/blob/015870d05d7e0848a6d0d9d275370890ee3b9829/nova/api/openstack/wsgi.py#L1035-L1045

### DIFF
--- a/openstack/compute/v2/servers/errors.go
+++ b/openstack/compute/v2/servers/errors.go
@@ -3,6 +3,7 @@ package servers
 import (
 	"fmt"
 
+	"encoding/json"
 	"github.com/gophercloud/gophercloud"
 )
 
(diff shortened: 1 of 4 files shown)


takaishi:
@jtopjian This PR is ready to review.

jtopjian:
I'll defer this one to @jrperritt for when he has time.

jrperritt:
After skimming the code, it looks good, but this what you're trying to achieve here is what I intended the default behaviour to be. That is, you shouldn't need create many custom errors. If it's not already doing that, the problem is further up in the error handling. I'll take a look at it within the next week.

databus23:
@jrperritt Any chance we can proceed on this? not sure what you mean with solving this *further up* in the error handling. getting the actual error message from a server create call is very helpful because otherwise without access to the nova logs there is no way of telling what was rejected by the service.

databus23:
@jrperritt Any chance to get some feedback on this PR? How can we proceed on this?

jtopjian:
This was resolved in #703

PR #304, opened 2017-03-21
Call UnmarshalJSON for embedded types that implement json.Unmarshaler; Update Network v2 Provider Extension to be consistent with other extensions

jrperritt:
For #300 and part of #298

### DIFF
--- a/openstack/networking/v2/extensions/external/results.go
+++ b/openstack/networking/v2/extensions/external/results.go
@@ -1,76 +1,8 @@
 package external
 
-import (
-	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
-	"github.com/gophercloud/gophercloud/pagination"
-)
-
-// NetworkExternal represents a decorated form of a Network with based on the
+// NetworkExt represents a decorated form of a Network with based on the
 // "external-net" extension.
-type NetworkExternal struct {
-	// UUID for the network
-	ID string `json:"id"`
-
-	// Human-readable name for the network. Might not be unique.
-	Name string `json:"name"`
-
-	// The administrative state of network. If false (down), the network does not forward packets.
-	AdminStateUp bool `json:"admin_state_up"`
-
-	// Indicates whether network is currently operational. Possible values include
-	// `ACTIVE', `DOWN', `BUILD', or `ERROR'. Plug-ins might define additional values.
-	Status string `json:"status"`
-
-	// Subnets associated with this network.
-	Subnets []string `json:"subnets"`
-
-	// Owner of network. Only admin users can specify a tenant_id other than its own.
-	TenantID string `json:"tenant_id"`
-
-	// Specifies whether the network resource can be accessed by any tenant or not.
-	Shared bool `json:"shared"`
-
+type NetworkExt struct {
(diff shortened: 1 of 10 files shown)


jrperritt:
cc @jtopjian

openstack/networking/v2/extensions/external/results.go
```diff
@@ -1,76 +1,8 @@
...
-	// Subnets associated with this network.
-	Subnets []string `json:"subnets"`
-
-	// Owner of network. Only admin users can specify a tenant_id other than its own.
-	TenantID string `json:"tenant_id"`
-
-	// Specifies whether the network resource can be accessed by any tenant or not.
-	Shared bool `json:"shared"`
-
+type NetworkExt struct {
```
jtopjian: I think extension names should be renamed to something more unique. Out of curiosity, I tried to mix two extensions in like so:

```go
type NetworkWithExt {
        networks.Network
        provider.NetworkExt
        external.NetworkExt
}
```

And got a Go error of:

```
duplicate field NetworkExt
```

results.go
```diff
@@ -100,12 +79,60 @@ func (r Result) ExtractIntoStructPtr(to interface{}, label string) error {
...
+	b, err := json.Marshal(rawitem)
+	if err != nil {
+		return err
+	}
+
+	err = json.Unmarshal(b, to)
+	if err != nil {
+		return err
+	}
+
```
jtopjian: I think I'm able to follow along what's going on here. The only thing I'm unsure about is why a Marshal/Unmarshal is being performed here and again on each anonymous struct?

openstack/networking/v2/extensions/external/results.go
jrperritt: Good point. Will change.

results.go
jrperritt: I'll add some comments and unit tests. Basically, the first `json.Unmarshal` will handle any non response objects the don't have embedded structs. The other is specifically for embedded structs.

When a struct is embedded, the parent will inherit a method defined on the embedded struct unless that same method is defined on the parent.

jrperritt:
I'm going to think about this some more and do some more testing: what's there was kind of thrown together in haste (read: I don't like it). In my head, if each embedded struct implements `json.Unmarshaler`, it should just work.

jrperritt:
OK, I'm convinced that it doesn't work like I thought it did. It looks like the following rules are applied for `json.Unmarshal(b, object)`:
1. If `object` implements `json.Unmarshaler`, `object`'s `UnmarshalJSON` method is called,
2. otherwise, if `object` embeds any number of subojects, and at most one of those subobjects implements `json.Unmarshaler`, the subobject's `UnmarshalJSON` method is called,
3. otherwise, a default `json.Unmarshal` is run on `object`. For an `object` with multiple embedded subobjects, each implementing `json.Unmarshaler`, this means none of those embedded subobjects' `UnmarshalJSON` methods will be called.

#3 above is the problem we're seeing.

jtopjian:
@jrperritt Any reason to keep this one open?

jrperritt:
I don't think so. Closing in favor of #443

Issue #306, opened 2017-03-23
Network v2: Security Group update


jtopjian:
Security Groups are able to be updated. The fields with `allow_put` are allowed to be updated:

https://github.com/openstack/neutron/blob/master/neutron/extensions/securitygroup.py#L220-L240

jrperritt:
closed by #307

PR #307, opened 2017-03-23
Network v2: Security Group update

jtopjian:
For #306


jtopjian:
@jrperritt ready for review

jrperritt:
LGTM

Issue #312, opened 2017-03-28
ObjectStorage v1: accounts.GetResult should parse quota


majewsky:
The `X-Account-Meta-Quota-Bytes` header is recognized [here](https://github.com/openstack/swift/blob/2aa5811c7fdf86df302f5ce8b0413bf51e86e7ab/swift/common/middleware/account_quotas.py#L84-L85) for PUT requests and then stored with a Swift account just like any `X-Account-Meta-` type header.

This issue is specifically about the GET request, where the header is returned and should be parsed in the same way that e.g. the [`X-Account-Bytes-Used` header is already handled](https://github.com/gophercloud/gophercloud/blob/9a5595b8ffe429439c63781cdc312c254e4ad990/openstack/objectstorage/v1/accounts/results.go#L79) by GetResult's UnmarshalJSON.

jtopjian:
Implemented in #313 and #1031

PR #313, opened 2017-03-28
ObjectStorage v1: recognize X-Account-Meta-Quota-Bytes header in accounts.GetHeader

majewsky:
For #312

code in this PR:

https://github.com/openstack/swift/blob/stable/newton/swift/common/middleware/account_quotas.py

The only part that might be controversial is the usage of `-1` to encode "no quota". I'm being consistent with other OpenStack services' quota handling here. Swift is a special snowflake because it does not report JSONs, but stuffs everything in headers. If you don't like `quota == -1`, I can also change the type of `GetHeader.Quota` to `*int64`, so that `nil` would indicate absence of a quota.

### DIFF
--- a/openstack/objectstorage/v1/accounts/results.go
+++ b/openstack/objectstorage/v1/accounts/results.go
@@ -61,7 +61,9 @@ func (r UpdateResult) Extract() (*UpdateHeader, error) {
 
 // GetHeader represents the headers returned in the response from a Get request.
 type GetHeader struct {
-	BytesUsed      int64     `json:"-"`
+	BytesUsed int64 `json:"-"`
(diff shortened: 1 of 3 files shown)


majewsky:
Any update on this?

jtopjian:
@majewsky Sorry for the delay on this.

I think `*int` is better here. I agree that `-1` would be more consistent with most other OpenStack quotas, but in Swift, since a missing Quota header indicates no/unlimited quota, I think `*int` fits this more appropriately.

jrperritt:
> I think *int64 is better here.

I agree

majewsky:
@auhlig Can you please take care of this while I'm away?

auhlig:
Sure @majewsky.

Issue #255, opened 2017-02-15
Re-authentication fails due to missing project id in context


xrl:
I am using a branch of kubernetes which moves away from the rackspace gophercloud and over to this more actively maintained gophercloud lib. The kubernetes health checks (a master scans nova for all running instances, making sure they're still present. same idea for cinder vols.) would fail after about 12 hours, owing to a re-auth-related failure.

I added some extra info to the reauth wrapping error:

```
 func (e ErrDefault400) Error() string {
-       return "Invalid request due to incorrect syntax or missing required parameters."
+       return fmt.Sprintf("Invalid request due to incorrect syntax or missing required parameters. URL: %v, Body: %v", e.ErrUnexpectedResponseCode.URL, string(e.ErrUnexpectedResponseCode.Body))
 }
```

and rebuilt kubernetes. Now I see these errors for cinder/nova:

Nova:
```
Successfully re-authenticated, but got error executing request: Invalid request due to incorrect syntax or missing required parameters. URL: https://my.cloud:8774/v2.1/cfef9feef46441a29ab31117690d352a/servers/detail?name=%5Ekube-node-03%24&status=ACTIVE, Body: {"badRequest": {"message": "Malformed request URL: URL's project_id 'cfef9feef46441a29ab31117690d352a' doesn't match Context's project_id 'None'", "code": 400}}
```

and

Cinder:
```
Attach volume "3f4d29d9-8407-4c2a-a435-645757d5d25c" to instance "8c97ee08-b345-462b-80eb-2a0033d9bd5e" failed with Successfully re-authenticated, but got error executing request: Invalid request due to incorrect syntax or missing required parameters. URL: https://my.cloud:8776/v2/cfef9feef46441a29ab31117690d352a/volumes/detail, Body: {"badRequest": {"message": "Malformed request url", "code": 400}}
```

The nova error is more useful and points out that a context (is that the same thing as a token?) doesn't specify the `project_id`.

Sounds like there is a bug in the re-auth code if it is generating bad tokens. The code generally works fine for the first 12 hours.

jrperritt:
> Sounds like there is a bug in the re-auth code if it is generating bad tokens

Firstly, that would indeed be strange: the re-auth code is exactly the auth code. I've never had an issue re-authenticating, but I also don't use project ID.
Secondly, so far there's no proof that this library is the cause; this could just as well be your OpenStack environment or Kubernetes. Have you taken measures to rule out those possibilities?

[...]

PR #314, opened 2017-03-28
Remove checks for TenantId

xrl:
For #255 

code in this PR: Um, not sure.

I am unable to reauth in my project. The issue is that the scope is not generated correctly after the first auth. All subsequent auth lose access to the TenantId because it is mutated and set to `""`.

```golang
                scope.ProjectID = opts.TenantID
-               opts.TenantID = ""
-               opts.TenantName = ""
```

to support that change I also comment out a sanity check. I'm not sure if that sanity check is necessary for other environments but my standalone script works fine without it.

### DIFF
--- a/auth_options.go
+++ b/auth_options.go
@@ -139,14 +139,6 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
 	// if insufficient or incompatible information is present.
 	var req request
 
-	// Test first for unrecognized arguments.
-	if opts.TenantID != "" {
-		return nil, ErrTenantIDProvided{}
-	}
-	if opts.TenantName != "" {
-		return nil, ErrTenantNameProvided{}
-	}
-
 	if opts.Password == "" {
 		if opts.TokenID != "" {
 			// Because we aren't using password authentication, it's an error to also provide any of the user-based authentication
@@ -259,8 +251,6 @@ func (opts *AuthOptions) ToTokenV3ScopeMap() (map[string]interface{}, error) {
 
 	if opts.TenantID != "" {
 		scope.ProjectID = opts.TenantID
-		opts.TenantID = ""
-		opts.TenantName = ""
 	} else {
 		if opts.TenantName != "" {
 			scope.ProjectName = opts.TenantName


xrl:
To faciliate all this library hacking I made a standalone project which would create a gophercloud client, fetch instance data, reauth, fetch instance data again (essentially reproducing the failing kubernetes action).

Standalone project: https://github.com/xrl/openstack-token-test
Output of running it in my OS environment:

```
starting openstack-token test
creating NewClient
authenticating Provider
Starting loop at Tue Mar 28 23:31:03 +0000 UTC 2017
doLoop called
creating compute handle
starting pagination
pagination complete
Found 131 instances matching : [whoa, big, list]
Loop completed at Tue Mar 28 23:31:05 +0000 UTC 2017
Doing a reauth to see if it triggers bad token bug
Done with reauth
Starting loop at Tue Mar 28 23:31:11 +0000 UTC 2017
doLoop called
creating compute handle
starting pagination
pagination complete
Found 131 instances matching : [whoa, big, list]
Loop completed at Tue Mar 28 23:31:13 +0000 UTC 2017
now sleeping for an hour...
```

Without my changes it looked like:

```
starting openstack-token test
creating NewClient
authenticating Provider
Starting loop at Tue Mar 28 23:35:42 +0000 UTC 2017
doLoop called
creating compute handle
starting pagination
pagination complete
Found 131 instances matching : [whoa, big, list]
Loop completed at Tue Mar 28 23:35:44 +0000 UTC 2017
Doing a reauth to see if it triggers bad token bug
Done with reauth
Starting loop at Tue Mar 28 23:35:50 +0000 UTC 2017
doLoop called
creating compute handle
failed on loop: No suitable endpoint could be found in the service catalog.
```

xrl:
@jrperritt you may be interested to see this fix. You mentioned you didn't use this scoping method so that gave me a good start for what could be wrong.

This fixes it for me so that's nice. I will try putting it in to my custom kube build and verify stability.

xrl:
@jrperritt can I get a review on this? It's a small PR and it actually just removes errant code and it's had a few users who say it works in practice.

jtopjian:
Thanks for catching this.

I'm going to follow up with a PR to include an acceptance test + remove one other piece in `auth_options.go`.

PR #316, opened 2017-04-03
Add arbitrary key/value scheduler hints ( was add reservation in scheduler hints)

msimonin:
Some OpenStack installation[[1](https://www.chameleoncloud.org/
)] use a lease system.  Leases are usually
provided by blazar[[2](https://wiki.openstack.org/wiki/Blazar
)] and each lease is composed of a list of
reservations. When creating servers, one must pass a valid reservation
id in order to indicate to the system which reservation will be used by
the future resource.

For #317

### DIFF
--- a/openstack/compute/v2/extensions/schedulerhints/requests.go
+++ b/openstack/compute/v2/extensions/schedulerhints/requests.go
@@ -27,8 +27,8 @@ type SchedulerHints struct {
 	TargetCell string `json:"target_cell,omitempty"`
 	// BuildNearHostIP specifies a subnet of compute nodes to host the instance.
 	BuildNearHostIP string
-	// Reservation specifies the id of a reservation obtained in blazar
-	Reservation string
+	// AdditionalProperies are arbitrary key/values that are not validated by nova.
+	AdditionalProperties map[string]interface{}
 }
 
 // CreateOptsBuilder builds the scheduler hints into a serializable format.
@@ -118,15 +118,10 @@ func (opts SchedulerHints) ToServerSchedulerHintsCreateMap() (map[string]interfa
 		sh["cidr"] = "/" + ipParts[1]
 	}
 
-	if opts.Reservation != "" {
-		if !uuidRegex.MatchString(opts.Reservation) {
-			err := gophercloud.ErrInvalidInput{}
-			err.Argument = "schedulerhints.SchedulerHints.Reservation"
-			err.Value = opts.Reservation
-			err.Info = "The reservation_id must be in UUID format."
-			return nil, err
+	if opts.AdditionalProperties != nil {
+		for k, v := range opts.AdditionalProperties {
+			sh[k] = v
 		}
-		sh["reservation"] = opts.Reservation
 	}
 
 	return sh, nil
--- a/openstack/compute/v2/extensions/schedulerhints/requests.go
+++ b/openstack/compute/v2/extensions/schedulerhints/requests.go
@@ -27,6 +27,8 @@ type SchedulerHints struct {
 	TargetCell string `json:"target_cell,omitempty"`
 	// BuildNearHostIP specifies a subnet of compute nodes to host the instance.
 	BuildNearHostIP string
+	// Reservation specifies the id of a reservation obtained in blazar
+	Reservation string
 }
 
 // CreateOptsBuilder builds the scheduler hints into a serializable format.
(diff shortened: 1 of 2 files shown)


jtopjian:
@msimonin Thanks for this! I'm familiar with Chameleon and would love to see features added that go toward helping it.

At first glance, I don't see `reservation` defined in the [scheduler_hints schema](https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/schemas/scheduler_hints.py). Is it defined elsewhere in Nova? Or is `scheduler_hints` able to accept arbitrary key/value pairs?

If it's defined elsewhere in Nova, your approach might still be good. But if `reservation` is defined outside of the core Nova `scheduler_hints` extension, it would be better to create a Go package that defines the extension (even if it's for one additional field). If `reservation` is defined in Blazar, then creating something under `openstack/blazar/v2` might make sense.

But if `reservation` works because `scheduler_hints` is able to accept any arbitrary key/value pairs, then rather than explicitly defining `reservation`, support for accepting any key should be added. Otherwise, what is preventing 10 other arbitrary additions to `scheduler_hints`? :)

This is is why the PR template asks for specific lines of code -- to help Gophercloud match what is defined in the actual API.

I hope that makes sense.

msimonin:
Thanks @jtopjian, absolutely, that makes sense.
I had a look on how  `scheduler_hints`  are handled. 
The [scheduler_hints schema](https://github.com/openstack/nova/blob/48f9c365befdb6ab583bfa2773e0969b32d69ca7/nova/api/openstack/compute/schemas/scheduler_hints.py) sets the `additionalProperties:True` allowing extra `key/value` to be added under the `scheduler_hints` object. This won't be stripped from the request by the [validator](https://github.com/openstack/nova/blob/48f9c365befdb6ab583bfa2773e0969b32d69ca7/nova/api/validation/validators.py#L166).

So in my understanding (I may be wrong), `scheduler_hints` allows extra  key/value to be specified and thus is partially validated.
In the blazar case `reservation` are used to [filter hosts](https://github.com/openstack/blazar-nova/blob/c3919458d289e113372ee819b33cef5e0efca990/blazarnova/scheduler/filters/blazar_filter.py#L78) and make pass only them in the reservation.

So in your opinion, what could be the best approach ? 
- allowing arbitrary key/value ?
- create a dedicated go package to handle this key ? 

Hope we can find a clean way to have this ! :)

jtopjian:
Nice - that validator link is the key. [This](https://github.com/openstack/nova/blob/d639b7502c90a5fbdea7fb03b5a944f60f0db335/nova/api/validation/__init__.py#L26) might also be relevant in terms of information, not implementation.

IMO, go for the arbitrary k/v in `openstack/compute/v2/extensions/schedulerhints/requests.go`. I swear this has been handled before, but it's too early in the morning for me to recall where. Might want to browse the various `requests.go` files and see.

In general, I think adding a field called `AdditionalProperties` of type `map[string]string` or `map[string]interface{}` is the best way to go. Then in `ToServerSchedulerHintsCreateMap`, iterate over `AdditionalProperties`, place each key in `sh`, then make sure `AdditionalProperties` is not placed in `sh` and sent as part of the request.

With regard to validating the UUID for `reservation`, I would just nix that. Validation will be moved to whatever piece of code declares `reservation` in the scheduler hint.

I think, anyway. @jrperritt ?

msimonin:
@jtopjian, @jrperritt I've updated the merge request to allow arbitrary key/value to be passed in the scheduler hints under an `AdditionnalProperties` fields. 

The topic of the PR and associated is less general than that, let me know what can I can do next.

jtopjian:
@msimonin This looks good to me - nice work! Thank you for your help with digging into the core way this works, too.

@jrperritt does all of the final reviews. Hang tight until then :)

priteau:
I confirm that, as far as I know, Nova can accept any string as a scheduler hint name. If the hint is not supported by any enabled filter, the launch will proceed as if the hint had not been provided.

The default list of filters can be extended by modules external to Nova, which is what Blazar is doing. Other projects can do the same and create new scheduler hints. See for example this [Volume Affinity Filter](https://github.com/Mirantis/nova-scheduler-volume-affinity-filter/blob/master/README.rst) from Mirantis. So I agree with the decision to allow arbitrary key/value pairs.

msimonin:
I renamed the PR title. 
@jrperritt  Is there any chance to get this merged soon ?

jrperritt:
@jtopjian 
> In general, I think adding a field called AdditionalProperties of type map[string]string or map[string]interface{} is the best way to go. Then in ToServerSchedulerHintsCreateMap, iterate over AdditionalProperties, place each key in sh, then make sure AdditionalProperties is not placed in sh and sent as part of the request.

Agree

jtopjian:
This looks good to me. Thanks, @msimonin and @priteau!

Issue #320, opened 2017-04-22
listing region availability zones?


zetaab:
Is there method for listing availability zones? https://developer.openstack.org/api-ref/compute/?expanded=get-availability-zone-information-detail#get-availability-zone-information

if no, could it be implemented?

dixudx:
@zetaab It has already been implemented. Please refer to [availabilityzones/requests.go](https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/v2/availabilityzones/requests.go#L9).

dixudx:
It is missing in [gophercloud godoc](https://godoc.org/github.com/gophercloud/gophercloud). Will add it later.

dixudx:
@zetaab  [availabilityzones/requests.go](https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/v2/availabilityzones/requests.go#L9) is shared file system availabilityzones, not compute availabilityzones.

I'll submit a new PR on this new feature. Sorry for misleading you.

yongfengdu:
Should this issue being closed since this is tracked in #327?

jtopjian:
@yongfengdu No, this needs to stay open for two reasons:

1. All Gophercloud Pull Requests must have a corresponding issue.
2. #327 seems to have gone stale. Likely someone else could/should follow-up with a new PR.

dstdfx:
@jtopjian I'd like to finish this

jtopjian:
@dstdfx Sure, go for it :)

jtopjian:
Implemented in #704 

@dstdfx Thank you again for your work on this.

PR #322, opened 2017-04-25
add missing doc files to support autogenerated on godoc

dixudx:
* add missing doc files;
* some new apis are not shown/autogenerated on [GoDoc](https://godoc.org/github.com/gophercloud/gophercloud), e.g. *availability zones* (#320)

### DIFF
--- a/openstack/compute/v2/extensions/bootfromvolume/doc.go
+++ b/openstack/compute/v2/extensions/bootfromvolume/doc.go
@@ -1,0 +1,1 @@
+package bootfromvolume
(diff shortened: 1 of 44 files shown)


jtopjian:
@dixudx I agree that the Gophercloud packages should follow a consistent format, which includes a standard `doc.go` file across everything. But I don't believe the lack of `doc.go` files is preventing documentation from being generated. I see all packages listed at the bottom of the main doc page. Unless I am missing something?

But if this is true, I would prefer if more time went into the missing `doc.go` files to add relevant documentation (an exception being for the `testing` directories). I'm not implying that _you_ should do all of the work. It's something multiple people can do in pieces.

Perhaps a good format could be:

```go
/*
1-3 sentence summary of the package.

Example of how to do something with the package.

  foo, err := package.Bar(client, createOpts)

Example 2 of how to do something.
*/
```

Thoughts?

jrperritt:
I agree with @jtopjian. Adding accurate documentation is always welcome, but I don't think this qualifies as that.

PR #153, opened 2016-10-18
Feature/filestorage availabilityzones list

ehdou:
Add support for Availability Zones List
For #152

Issue #152, opened 2016-10-18
Shared file system: Availability Zones


ehdou:
This issue is raised to track the development of the availability zones in the shared file systems api v2
- [x] List Availability zones (#153)

This issue is part of #114

PR #323, opened 2017-04-25
mv availabilityzones to compute extensions

dixudx:
For #152 (Shared file system: Availability Zones), implementation #153 is not the right one.

#153 implements [Compute Availability zones (os-availability-zone)](https://developer.openstack.org/api-ref/compute/?expanded=get-availability-zone-information-detail,get-detailed-availability-zone-information-detail#get-availability-zone-information), which is quite different from [Shared File System Availability zones](https://developer.openstack.org/api-ref/shared-file-systems/index.html?expanded=list-security-services-detail).

So I prefer to move this implementation to folder  [openstack/compute/v2/extensions/availabilityzones](https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions/availabilityzones).

When this pr is merged, I'll try to implement #152.


jtopjian:
@dixudx This doesn't seem correct to me. 

The _compute_ availability zone extension defines attributes such as [`zoneState`](https://github.com/openstack/nova/blob/b9e94252e290d09d6c3e8e5ec794163cd702f0c9/nova/api/openstack/compute/availability_zone.py#L43-L45) (which can also be seen in the API reference docs).

I believe the sharedfilesystem availability zone package is correct and the compute extension is missing.

dixudx:
@jtopjian I don't agree except `zoneState`.

The url of [Compute Availability zone](https://developer.openstack.org/api-ref/compute/?expanded=get-availability-zone-information-detail,get-detailed-availability-zone-information-detail#get-availability-zone-information) is `/os-availability-zone`, which is exactly implemented by [sharedfilesystems/v2/availabilityzones/urls.go](https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/v2/availabilityzones/urls.go#L6).

jtopjian:
@dixudx ... there's also `zoneName` and `hosts` that are missing from your PR. 

With regard to the URL, one piece of the URL is the same, but that's just a coincidence. One URL is going to be hitting the Compute API and the other is hitting the Shared Filesystem API, so the full URLs for both are actually different.

Can you find references in the OpenStack code (which is one requirement to have a PR accepted) that the compute extension returns id, name, created_at, and updated_at like is happening with the shared filesystem extension?

dixudx:
For *compute Availability zone*, the corresponding openstack code can be found at [compute/availability_zone.py#L60](https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/availability_zone.py#L60).

*compute Availability zone* and *share file system availability zone* is quite different including the apis and return value.

@jtopjian You're right. So sorry that I just noticed the url and ignore the actual return value.

But we still need to update the url in [sharedfilesystems/v2/availabilityzones/urls.go#L6](https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/v2/availabilityzones/urls.go#L6).

jtopjian:
@dixudx No problem at all :)

jtopjian:
I'm going to close this PR, but let me know if you have any questions.

PR #324, opened 2017-04-26
fix availability zones url in shared file system

dixudx:
@jtopjian fix the url as we discussed in #323

Reference: [shared-file-systems list-availability-zones](https://developer.openstack.org/api-ref/shared-file-systems/index.html?expanded=list-security-services-detail,list-all-major-versions-detail,show-details-of-specific-api-version-detail#list-availability-zones)

### DIFF
--- a/openstack/sharedfilesystems/v2/availabilityzones/testing/fixtures.go
+++ b/openstack/sharedfilesystems/v2/availabilityzones/testing/fixtures.go
@@ -10,14 +10,7 @@ import (
 )
 
 func MockListResponse(t *testing.T) {
-	th.Mux.HandleFunc("/os-availability-zone", func(w http.ResponseWriter, r *http.Request) {
-		th.TestMethod(t, r, "GET")
-		th.TestHeader(t, r, "X-Auth-Token", fake.TokenID)
-
-		w.Header().Add("Content-Type", "application/json")
-		w.WriteHeader(http.StatusOK)
-
-		fmt.Fprintf(w, `
+	response := `
         {
             "availability_zones": [
                 {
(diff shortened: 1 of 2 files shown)


jtopjian:
@dixudx Yep, looks like the URL was indeed renamed. Possibly to avoid confusion with the compute extension :smile: 

The actual routing code can be seen [here](https://github.com/openstack/manila/blob/eeeb648cc380a383fc74cd107517be9b0678441b/manila/api/v2/router.py#L71-L84).

This is interesting, though, because this deprecation is within the v2 API. I would think that both URLs should be supported. I'm interested to hear @jrperritt's thoughts when he has time.

Also, just for consistency with submission rules, can you open a corresponding issue for this PR?

jtopjian:
I'm going to ping @mvalkon and @ehdou to get their thoughts on the URL rename. They might be able to shed some light about if the new URL should be used exclusively or if both URLs should be supported within the `v2` package namespace.

mvalkon:
I suppose that it depends on how we want gophercloud to support the microversioning of the shared file systems API. Can't say I remember this perfectly but I think when me and @ehdou were working on the shared file systems implementation, we decided that we'll only support the latest microversion. Obviously wrt availability zones we made a mistake.

Now if someone is using the < 2.7 version of shared file systems API with gophercloud, and we change this, it'll break their code. We have to decide whether this is acceptable or then find a way to support the previous microversions of the API.

jtopjian:
@mvalkon Thanks for the input :)

> Now if someone is using the < 2.7 version of shared file systems API with gophercloud, and we change this, it'll break their code. We have to decide whether this is acceptable or then find a way to support the previous microversions of the API.

If I understand correctly, the current URL works for 2.7 and below, right? It's just deprecated from 2.7+. 

I propose keeping the current URL until there's support to determine what microversion is returned from the catalog. This way, the current solution works with all v2 deployments and no one is left out in the cold.

If Gophercloud used the new URL right from the start, I'd be in favour of adding support for the deprecated URL, but the same caveat applies: once there was support for determining the microversion.

So if anyone wants to try to tackle microversion discovery, perhaps just for the Shared Filesystem API initially, go for it. :)

jtopjian:
Per the discussion above, I'm going to close this PR. Let me know if anyone has any questions, though.

Issue #319, opened 2017-04-12
Get/set instance ssh host key?


sb10:
Now that golang.org/x/crypto/ssh.ClientConfig requires a HostKeyCallback, I realise I haven't been doing any host key checking when ssh'ing to the instances I create via Gophercloud. Is there any way I can find out what the public key of those instances are? Or generate host keys myself and inject them when I create the instances?

`nova console-log [instance name]` shows the host key since it gets printed to the console during boot up. Is that, or could that, be captured and provided by gophercloud?

https://www.ibm.com/developerworks/cloud/library/cl-openstack-pythonapis/ (search for "ssh host key") shows an example of parsing this output. Is there at least a gophercloud method of getting the raw console output?

jtopjian:
The ability to show the console output has been merged in #329 and #1032. The ability to parse the SSH key is out of scope of Gophercloud. An example function to do this can be seen in #329, however it was removed in #1032.

PR #329, opened 2017-05-03
get the raw console output from the server instance

dixudx:
For #319 

code in this PR:

* [openstack compute api doc on `os-getConsoleOutput`](https://developer.openstack.org/api-ref/compute/?expanded=show-console-output-os-getconsoleoutput-action-detail#show-console-output-os-getconsoleoutput-action)
* [novaclient/v2/servers.py `get_console_output`](https://github.com/openstack/python-novaclient/blob/master/novaclient/v2/servers.py#L1680-L1693)

### DIFF
--- a/openstack/compute/v2/servers/results.go
+++ b/openstack/compute/v2/servers/results.go
@@ -4,9 +4,11 @@ import (
 	"crypto/rsa"
 	"encoding/base64"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/url"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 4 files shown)


openstack/compute/v2/servers/results.go
```diff
@@ -92,6 +99,41 @@ func (r GetPasswordResult) ExtractPassword(privateKey *rsa.PrivateKey) (string,
...
+			searchFlag = false
+		}
+
+		if searchFlag && strings.HasPrefix(value, "ssh-rsa") {
+			return value, nil
+		}
+	}
+
+	return "", errors.New("ssh host key not found")
+}
```
jtopjian: I appreciate what this function is doing, but I think it's too niche to be included in Gophercloud as a whole. This function could easily be added to the application which needs to parse the console output. Let's remove this.

jtopjian:
For completeness, here's the server implementation:

https://github.com/openstack/nova/blob/77b11a9e31e265c338857326418bfe8cd135c1f8/nova/api/openstack/compute/console_output.py

And here's the return code:

https://github.com/openstack/nova/blob/master/nova/api/openstack/wsgi.py#L316

PR #334, opened 2017-05-09
gofmt on the whole project

dixudx:
Run `gofmt` on the whole project to fix the format.


jtopjian:
👍 

Thanks for this :)

Issue #335, opened 2017-05-09
add codes formating to travis CI


dixudx:
Currently there is no CI running against code formatting.

jtopjian:
Closed by #334, #336, and #339

PR #336, opened 2017-05-09
add code formatting check for travis ci

dixudx:
For #335

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ install:
 - go get -v -tags 'fixtures acceptance' ./...
 - go get github.com/wadey/gocovmerge
 - go get github.com/mattn/goveralls
+- go get golang.org/x/tools/cmd/goimports
 go:
 - 1.7
 - tip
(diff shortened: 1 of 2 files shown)


dixudx:
@jtopjian This CI failed due to pr #334. It just proves that this PR is quite important to guarantee the code format for every commit.

When #334 is merged, this CI will pass.

jtopjian:
This seems reasonable to me.  

@jrperritt (since this involves modifying Travis): any objections?

jrperritt:
Agree. It's something I've mentioned before, but never got around to doing. I think running `go fmt` as part of the CI is a good idea.

jtopjian:
@jrperritt `go fmt` to actually rewrite the incoming files?

jrperritt:
Well, a `go fmt` check is fine, too.

jtopjian:
Ah, sorry. This tripped me up, too. `goimports` does everything `go fmt` does plus handles import formatting. That's why I assumed you were looking for _more_ than just a format report :)

jtopjian:
@dixudx Thanks for putting this together :)

Issue #337, opened 2017-05-09
No vendor is provided


dixudx:
Add `vendor` folder to avoid breaking down when their dependent packages changed or went away.

jtopjian:
Copying comment from #338:

I don't think vendoring is a big requirement here. There are only two packages required and I would be very surprised if an update to them broke Gophercloud. The one benefit of vendoring is that the dev/user doesn't have to do a go get to get started. But again, there's only two required packages.

In addition, if Gophercloud were to begin vendoring, I think it might be best to use the upcoming official Go dep tool, too.

I do appreciate the suggestion, though. :)

PR #338, opened 2017-05-09
add vendor packages

dixudx:
For #337


jtopjian:
I don't think vendoring is a big requirement here. There are only two packages required and I would be very surprised if an update to them broke Gophercloud. The one benefit of vendoring is that the dev/user doesn't have to do a `go get` to get started. But again, there's only two required packages.

In addition, if Gophercloud were to begin vendoring, I think it might be best to use the upcoming official Go dep tool, too.

I do appreciate the suggestion, though. :)

Issue #53, opened 2016-08-15
fix microversion header


jrperritt:
Currently all `ServiceClient` methods erroneously add the Nova microversion header: https://github.com/gophercloud/gophercloud/blob/master/service_client.go#L52

Currently, I don't think there's a good way to know what type of `ServiceClient` the user has. One option would be to add a `Type` field to `ServiceClient` and switch on it in the (not yet existing) `ServiceClient` `Request` method.

pospispa:
Created PR https://github.com/gophercloud/gophercloud/pull/341 to fix this issue.

jtopjian:
Some non-coded thoughts on implementing this:

This might be relevant:

http://docs-draft.openstack.org/15/462815/4/check/gate-api-wg-docs-ubuntu-xenial/b8479be//doc/build/html/guidelines/consuming-catalog.html

As might this:

https://github.com/openstack/api-wg/blob/master/guidelines/microversion_specification.rst

> Currently, I don't think there's a good way to know what type of ServiceClient the user has. One option would be to add a Type field to ServiceClient and switch on it in the (not yet existing) ServiceClient Request method.

I think what's currently in [client.go](https://github.com/gophercloud/gophercloud/blob/master/openstack/client.go) is a good start. It has the standard service type names and allows a user to override a service type name. I don't know of an application currently taking advantage of that, _but_ the mechanism is there.

I think the `EndpointLocator` type is a good place to implement the first referenced doc without too much trouble. If the `EndpointLocator` functions implemented the whole spec, there would be a good amount of logic added to them, including reading a possibly bundled external spec sheet of service types and their known aliases. But one step at a time.

Now, an interesting thing mentioned in the second doc is how the version header is specified:

```
OpenStack-API-Version: [SERVICE_TYPE] [VERSION_STRING]
```

This is different than what's currently implemented in Gophercloud and what has been proposed for Manila. We should see if the above format is valid or if it's just a future spec.

If it is a valid format, this makes things a little easy: the `type` is already being declared when creating the service client, so as mentioned initially, add `Type` to the service client and set it, then use it to form the standardized URL.

But if URLs like `X-OpenStack-Nova-API-Version` and `X-Openstack-Manila-Api-Version` are the only accepted forms at this time, then Gophercloud should probably do the same pattern as it's doing with type: apply a sane default, but allow it to be overridden. 

Ideally the header _and_ requested version should be easily configurable to the user, so maybe extend `EndpointOpts` with these?

[...]

PR #341, opened 2017-05-16
Added Type into ServiceClient to Distinguish Different Services

pospispa:
All ServiceClient methods insert the OpenStack Nova HTTP Microversion header ("X-OpenStack-Nova-API-Version") into HTTP requests even though the request may be to a different service than Nova. This is incorrect.

That's why a mapping of serviceType to HTTPMicroversionHeader is added. Default value (empty string) is mapped to OpenStack Nova HTTP Microversion header because of backward compatibility.

The serviceType constants will be made public as soon as it will be needed.

For #53 

code in this PR: N/A

### DIFF
--- a/service_client.go
+++ b/service_client.go
@@ -21,7 +21,31 @@ type ServiceClient struct {
 	// as-is, instead.
 	ResourceBase string
 
+	// Type of service, e.g. Nova, Manila, etc.
+	Type serviceType
+
 	Microversion string
+}
+
+type serviceType string
+
+const (
+	novaServiceType   serviceType = "Nova"
+	manilaServiceType serviceType = "Manila"
+)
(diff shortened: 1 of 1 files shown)


pospispa:
@jtopjian @jrperritt please review

jrperritt:
I'm not sure this is the best way to implement this. I'll want to read more about it before we merge anything.

pospispa:
@jrperritt OK, will wait and looking forward to your comments.

jrperritt:
See #355 for my take on how to handle this.

pospispa:
Solution in https://github.com/gophercloud/gophercloud/pull/355 was chosen that's why this PR is closed without merge.

Issue #340, opened 2017-05-16
Shared file systems: Get Microversion


pospispa:
Shared file systems v2 API uses microversions. Some services like [share instance export locations](https://developer.openstack.org/api-ref/shared-file-systems/?expanded=id234-detail#share-instance-export-locations-since-api-v2-9) are supported from a specific microversion.

That's why getting the shared file systems API microversion is necessary.

Related issue: [fix microversion header](https://github.com/gophercloud/gophercloud/issues/53).

jtopjian:
@pospispa @jrperritt Unless I am totally misunderstanding how this all works, isn't this an appropriate solution?

https://github.com/jtopjian/gophercloud/commit/5bf58ab1beb18b3220bd0b63f74104a56e1f273e
https://github.com/jtopjian/gophercloud/commit/7f48e23cca311be0fb1796e61bbded54cfefb7d6

Note that the above was done in a rush and might be sloppy, but if it looks like a good foundation, I can submit a proper PR.

One piece of missing functionality that I see in the above is for the user to be able to specify a microversion (or to opt out at all).

pospispa:
@jtopjian your implementation is really a good foundation.
I've tried it out [here](https://github.com/pospispa/gophercloud/commit/73d2fbf333e31055439b18cacf4c5330af99a6e9) and I haven't found any problems in it.

I will add a few comments directly to your code.

pospispa:
This issue is fixed by https://github.com/gophercloud/gophercloud/pull/409

PR #342, opened 2017-05-17
Get manila api microversion

pospispa:
For https://github.com/gophercloud/gophercloud/issues/340

code in this PR:

https://github.com/openstack/manila/blob/stable/newton/manila/api/versions.py#L88

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -8,12 +8,19 @@ import (
 	"github.com/gophercloud/gophercloud"
 	tokens2 "github.com/gophercloud/gophercloud/openstack/identity/v2/tokens"
 	tokens3 "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
+	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/shares"
 	"github.com/gophercloud/gophercloud/openstack/utils"
 )
 
 const (
 	v20 = "v2.0"
 	v30 = "v3.0"
+)
+
+const (
+	apiVersionStatusCurrent    = "CURRENT"
+	apiVersionStatusSupported  = "SUPPORTED"
+	apiVersionStatusDeprecated = "DEPRECATED"
 )
 
 // NewClient prepares an unauthenticated ProviderClient instance.
@@ -261,7 +268,20 @@ func NewBlockStorageV2(client *gophercloud.ProviderClient, eo gophercloud.Endpoi
 
 // NewSharedFileSystemV2 creates a ServiceClient that may be used to access the v2 shared file system service.
 func NewSharedFileSystemV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
-	return initClientOpts(client, eo, "sharev2")
+	serviceClient, err := initClientOpts(client, eo, "sharev2")
+	if err != nil {
+		return serviceClient, err
+	}
+	respMicroversions := shares.GetMicroversion(serviceClient)
+	// keep ServiceClient.Microversion empty in case an attempt to get a Microversion fails
+	if extractedMicroversionsReqResp, err := respMicroversions.ExtractMicroversion(); err == nil {
+		for _, extractedMicroversion := range *extractedMicroversionsReqResp {
+			if extractedMicroversion.Status == apiVersionStatusCurrent && len(extractedMicroversion.Version) > 0 {
+				serviceClient.Microversion = extractedMicroversion.Version
+			}
+		}
+	}
+	return serviceClient, nil
 }
 
 // NewCDNV1 creates a ServiceClient that may be used to access the OpenStack v1
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -1,8 +1,6 @@
 package shares
 
-import (
-	"github.com/gophercloud/gophercloud"
-)
+import "github.com/gophercloud/gophercloud"
 
 // CreateOptsBuilder allows extensions to add additional parameters to the
 // Create request.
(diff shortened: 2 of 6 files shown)


pospispa:
@jtopjian @jrperritt please review

jtopjian:
@pospispa Thanks for working on this! I'll be able to dedicate some time to this over the next few days.

pospispa:
@jtopjian thank you for letting me know.
I'm looking forward to your comments.

jrperritt:
On hold for a decision on #341

pospispa:
@jtopjian @jrperritt this PR was waiting for decision on https://github.com/gophercloud/gophercloud/pull/341
As https://github.com/gophercloud/gophercloud/pull/341 is closed without merge and the https://github.com/gophercloud/gophercloud/pull/355 is merged I updated this PR according to current master.
I also updated https://github.com/gophercloud/gophercloud/pull/346 that is based on this PR, however, is still WIP.

To sum it up, this PR is ready for review. Please, take a look.

pospispa:
@jtopjian please, do you plan to take a look on this PR.

pospispa:
@jtopjian I'm sorry, I don't precisely understand what you mean. Please, do you mean that sharedfilesystems package should use the [func ListVersions](https://github.com/gophercloud/gophercloud/blob/master/openstack/orchestration/v1/apiversions/requests.go#L9) to get the current sharedfilesystems Microversion?

jtopjian:
Yes, kind of. It should use a `ListVersions` function, but that function is probably better suited in a dedicated `sharedfilesystems/apiversions` package (note how it's not `sharedfilesystems/v2/apiversions` since, in theory, any version could/should be able to use this functionality. This differs from how the existing `apiversion` packages are today, but that will most likely change to them being placed into a non-versioned directory).

The reason for the dedicated package is because this functionality has a high probability of re-use. You've implemented microversion detection for `shares`, but what if `sharetypes` then also needs to work with microversions?

Let me know if that clarifies things :)

pospispa:
@jtopjian yes it clarifies things. Now, I understand your original comment better.

pospispa:
@jtopjian please, let me try to explain my use case. A new service client for a particular shared filesystem endpoint is created in [func NewSharedFileSystemV2](https://github.com/gophercloud/gophercloud/blob/master/openstack/client.go#L263). The `func NewSharedFileSystemV2` finds an endpoint that is major version specific, e.g. `/v2/`. I need to extend the `func NewSharedFileSystemV2` in such a way that the func except finding endpoint the [Microversion](https://github.com/gophercloud/gophercloud/blob/master/service_client.go#L30) is also initialized to the [Microversion](https://docs.openstack.org/developer/manila/devref/api_microversion_dev.html) value of the found endpoint.

In previous comments we touched orchestration service API [list versions](https://developer.openstack.org/api-ref/orchestration/v1/index.html?expanded=list-versions-detail) that lists all **MAJOR** API versions of orchestration service and is at [`/`](https://developer.openstack.org/api-ref/orchestration/v1/index.html?expanded=list-versions-detail).
I agree that functions for listing all **MAJOR** API versions of an OpenStack service shall be in `openstack/servicename/apiversions` package.

However, my use case is not listing all major versions of shared filesystems service at [`/`](https://developer.openstack.org/api-ref/shared-file-systems/?expanded=list-all-major-versions-detail), however, I need to get Microversion of a particular endpoint that already has a particular **MAJOR** version in its path, e.g. [`/v2/`](https://developer.openstack.org/api-ref/shared-file-systems/?expanded=show-details-of-specific-api-version-detail). That's why in my opinion the package should be `sharedfilesystems/v2/microversion` because it's **MAJOR** version specific.

Please, do you agree?

jtopjian:
@pospispa The reason I suggested placing the `apiversions` package in a non-versioned directory is because it is possible to get all versions (both major and minor) by querying the root URL of the service.

As an exercise:

```bash
#!/bin/bash

oscurl() {
  token=$(openstack token issue -c id -f value)
  curl -s -H "X-Auth-Token:$token" $1
}

echo "Cinder API versions"
oscurl http://localhost:8776/ | python -mjson.tool

echo "Nova API versions"
oscurl http://localhost:8774/ | python -mjson.tool

echo "Glance API versions"
oscurl http://localhost:9292/ | python -mjson.tool

echo "Manila API versions"
oscurl http://localhost:8786/ | python -mjson.tool

echo "Manila v2 API versions"
oscurl http://localhost:8786/v2 | python -mjson.tool
```

You _should_ get an error on the last call to `oscurl` which is querying Manila's `/v2` endpoint. However, the second-to-last call is querying Manila's root URL which is returning major and minor versions for both v1 and v2.

If you were to put `apiversions` in v2 and if someone wanted to add `apiversions` to v1 (or v3 in the future), wouldn't the only difference in code be changing the number "2"?

jtopjian:
Actually, as soon as I posted, I saw that

```
oscurl http://localhost:8786/v2 | python -mjson.tool
```

was missing a `/` at the end of the URL and it looks like adding the `/` makes a difference:

```
oscurl http://localhost:8786/v2/ | python -mjson.tool
```

That will return all of the v2 results. However, these results are already returned as part of the larger result set when doing:

```
oscurl http://localhost:8786/ | python -mjson.tool
```

So I still think putting `apiversions` in a non-versioned directory is still applicable.

pospispa:
@jtopjian I agree that having `apiversions` in non-versioned directory for querying `http://localhost:8786/`
is the correct way.

However, I need endpoint specific, i.e. major version specific microversion.

I can think of 2 possible ways how to get the endpoint specific microversion.
1. Using `apiversions` package:
a. Query `http://localhost:8786/` to get all available versions `allVersions`(note: this query expects HTTP status code 300).
b. Parse endpoint URL to get it's major version `epMajorVer`, e.g. `v2`.
c. Find `epMajorVer` specific records `epRec` in `allVersions`, choose one record from `epRec` and get the endpoint specific microversion from it.

2. Using `v2/microversion` package:
a. Query `http://localhost:8786/v2/` to get endpoint specific versions `epVersions`(note: this query expects HTTP status code 200).
b. Choose one record from `epVersions` and get the endpoint specific microversion from it.

I prefer the second way (to query `http://localhost:8786/v2/`) because IMHO it's easier.

@jtopjian please, what's your opinion.

I understand that you have concerns that in case `v2/microversion` package is added that there may be code duplication at the time `v3/microversion` package is added or when `apiversions` package is added because extraction of the results is the same and requests are same except URL and expected HTTP status code.

pospispa:
The PR #409 implements the Get microversion request in a better way. That's why I'm closing this PR.

Issue #179, opened 2016-11-03
Hypervisors Support


spjmurray:
Gophercloud is missing the os-hypervisors API.  This would allow operators to collate global cluster statistics to monitor CPU and memory utilization (and over-commit) and also enable capacity planning based on trends.

spjmurray:
Basic list functionality started in https://github.com/gophercloud/gophercloud/pull/178

jtopjian:
Resolved by #178 and #344

PR #178, opened 2016-11-03
Nova Hypervisors API

spjmurray:
Adds in support to list all available hypervisors (detailed) for applications
such as monitoring and capacity planning.

API: https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/hypervisors.py#L46

Pay particular attention to line 71 where the horrible cpu_info hack is introduced

For #179

PR #344, opened 2017-05-18
compute/v2: Hypervisors List API

jtopjian:
For #179
Supersedes #178


jtopjian:
@jrperritt Quick question: SinglePage returns still follow the same `EachPage` and `AllPage` patterns as other page types, right? Other SinglePage returns in Gophercloud are doing this, just want to explicitly ask :)

jrperritt:
Yes. `AllPages` returns immediately for `SinglePage`s (with the first (and only) page).

jtopjian:
This looks good to me. I'm going to go ahead and merge it. @spjmurray Thank you for your work on this :)

jtopjian:
Ping @majewsky - This might be useful for Limes (which I'm looking at because we have a similar in-house quota management system 😄 )

majewsky:
Thanks for pinging. @auhlig Do you want to adjust the Nova capacity plugin to take advantage of this?

Re Limes, I'd be delighted to see someone outside SAP use it. If you have any questions or requirements, feel free to open an issue over in our repo.

auhlig:
Sure. Thanks for the hint @majewsky

Issue #315, opened 2017-03-29
Shared file systems: Share Actions


pospispa:
This issue is raised to track the development of the [shares](http://developer.openstack.org/api-ref/shared-file-systems/?expanded=#shares) - actions in the shared file systems api v2

- [x] Grant access
- [x] Revoke acess
- [x] List access rules
- [ ] Reset share state
- [ ] Force-delete share
- [ ] Extend share
- [ ] Shrink share
- [ ] Unmanage share
- [ ] Revert share to snapshot

kayrus:
Extend and Shrink were implemented in #1175

PR #346, opened 2017-05-22
Shared Filesystems: Grant Access

pospispa:
For #315 

code in this PR:

https://github.com/openstack/manila/blob/stable/newton/manila/api/v1/shares.py#L393

### DIFF
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -117,7 +117,7 @@ func GrantAccess(client *gophercloud.ServiceClient, id string, opts GrantAccessO
 		return
 	}
 	_, r.Err = client.Post(grantAccessURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
-		OkCodes: []int{200, 201},
+		OkCodes: []int{200},
 	})
 	return
 }
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -110,7 +110,7 @@ func (opts GrantAccessOpts) ToGrantAccessMap() (map[string]interface{}, error) {
 
 // GrantAccess will grant access to a Share based on the values in GrantAccessOpts. To extract
 // the GrantAccess object from the response, call the Extract method on the GrantAccessResult.
-func GrantAccess(client *gophercloud.ServiceClient, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
+func GrantAccessRequest(client *gophercloud.ServiceClient, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
 	b, err := opts.ToGrantAccessMap()
 	if err != nil {
 		r.Err = err
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -94,16 +94,12 @@ type GrantAccessOptsBuilder interface {
 // For more information about these parameters, please, refer to the shared file systems API v2,
 // Share Actions, Grant Access documentation
 type GrantAccessOpts struct {
-	// The UUID of the share to which you are granted or denied access.
-	//ShareID string `json:"share_id"`
 	// The access rule type that can be "ip", "cert" or "user".
 	AccessType string `json:"access_type"`
 	// The value that defines the access that can be a valid format of IP, cert or user.
 	AccessTo string `json:"access_to"`
 	// The access level to the share is either "rw" or "ro".
 	AccessLevel string `json:"access_level"`
-	// The UUID of the tenant to which this share belongs to
-	//TenantID string `json:"tenant_id"`
 }
 
 // ToGrantAccessMap assembles a request body based on the contents of a
@@ -114,7 +110,7 @@ func (opts GrantAccessOpts) ToGrantAccessMap() (map[string]interface{}, error) {
 
 // GrantAccess will grant access to a Share based on the values in GrantAccessOpts. To extract
 // the GrantAccess object from the response, call the Extract method on the GrantAccessResult.
-func GrantAccess(client *gophercloud.ServiceClient, opts GrantAccessOptsBuilder, id string) (r GrantAccessResult) {
+func GrantAccess(client *gophercloud.ServiceClient, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
 	b, err := opts.ToGrantAccessMap()
 	if err != nil {
 		r.Err = err
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -1,8 +1,6 @@
 package shares
 
-import (
-	"github.com/gophercloud/gophercloud"
-)
+import "github.com/gophercloud/gophercloud"
 
 // CreateOptsBuilder allows extensions to add additional parameters to the
 // Create request.
@@ -83,5 +81,47 @@ func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
 // GetExportLocations will get shareID's export locations
 func GetExportLocations(client *gophercloud.ServiceClient, id string) (r GetExportLocationsResult) {
 	_, r.Err = client.Get(getExportLocationsURL(client, id), &r.Body, nil)
+	return
+}
+
+// GrantAccessOptsBuilder allows extensions to add additional parameters to the
+// GrantAccess request.
+type GrantAccessOptsBuilder interface {
+	ToGrantAccessMap() (map[string]interface{}, error)
+}
+
+// GrantAccessOpts contains the options for creation of an GrantAccess request.
+// For more information about these parameters, please, refer to the shared file systems API v2,
+// Share Actions, Grant Access documentation
+type GrantAccessOpts struct {
+	// The UUID of the share to which you are granted or denied access.
+	//ShareID string `json:"share_id"`
+	// The access rule type that can be "ip", "cert" or "user".
+	AccessType string `json:"access_type"`
+	// The value that defines the access that can be a valid format of IP, cert or user.
+	AccessTo string `json:"access_to"`
+	// The access level to the share is either "rw" or "ro".
+	AccessLevel string `json:"access_level"`
+	// The UUID of the tenant to which this share belongs to
+	//TenantID string `json:"tenant_id"`
+}
+
+// ToGrantAccessMap assembles a request body based on the contents of a
+// GrantAccessOpts.
+func (opts GrantAccessOpts) ToGrantAccessMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "allow_access")
+}
+
+// GrantAccess will grant access to a Share based on the values in GrantAccessOpts. To extract
+// the GrantAccess object from the response, call the Extract method on the GrantAccessResult.
+func GrantAccess(client *gophercloud.ServiceClient, opts GrantAccessOptsBuilder, id string) (r GrantAccessResult) {
+	b, err := opts.ToGrantAccessMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Post(grantAccessURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 201},
+	})
 	return
 }
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -108,7 +108,7 @@ func (opts GrantAccessOpts) ToGrantAccessMap() (map[string]interface{}, error) {
 	return gophercloud.BuildRequestBody(opts, "allow_access")
 }
 
-// GrantAccess will grant access to a Share based on the values in GrantAccessOpts. To extract
+// GrantAccessRequest will grant access to a Share based on the values in GrantAccessOpts. To extract
 // the GrantAccess object from the response, call the Extract method on the GrantAccessResult.
 func GrantAccessRequest(client *gophercloud.ServiceClient, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
 	b, err := opts.ToGrantAccessMap()
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -108,9 +108,9 @@ func (opts GrantAccessOpts) ToGrantAccessMap() (map[string]interface{}, error) {
 	return gophercloud.BuildRequestBody(opts, "allow_access")
 }
 
-// GrantAccessRequest will grant access to a Share based on the values in GrantAccessOpts. To extract
+// GrantAccess will grant access to a Share based on the values in GrantAccessOpts. To extract
 // the GrantAccess object from the response, call the Extract method on the GrantAccessResult.
-func GrantAccessRequest(client *gophercloud.ServiceClient, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
+func GrantAccess(client *gophercloud.ServiceClient, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
 	b, err := opts.ToGrantAccessMap()
 	if err != nil {
 		r.Err = err
--- a/openstack/sharedfilesystems/v2/shares/results.go
+++ b/openstack/sharedfilesystems/v2/shares/results.go
@@ -168,7 +168,7 @@ type GrantAccessRes struct {
 }
 
 // ExtractGrantAccess will get the GrantAccess object from the commonResult
-func (r commonResult) ExtractGrantAccess() (*GrantAccessRes, error) {
+func (r GrantAccessResult) ExtractGrantAccess() (*GrantAccessRes, error) {
 	var s struct {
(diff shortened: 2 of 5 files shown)


pospispa:
@jtopjian IMHO, this PR is ready for review, PTAL.

pospispa:
@jtopjian In this PR I started using microversions merged in PR https://github.com/gophercloud/gophercloud/pull/409
I'll appreciate if you comment on my approach so that we can eventually discuss it.
Thank you in advance.

jrperritt:
@pospispa I appreciate the work you've put into this PR, but something like this can't be merged. You've got several third-party libraries including one of your own. Design changes to Gophercloud are OK sometimes, but they need to be discussed separately. As an example, we won't be using Glide, nor the package for parsing versions, nor your `openstackmicroversions` package. Indeed, without proper justification, we probably won't be adding any third-party libraries. If you can make this PR small like all the others, we can certainly discuss it, but I think 316 files and 80,000+ lines is unactionable.

jtopjian:
I agree with @jrperritt.

With regard to the `openstackmicroversions` package, it's small enough that the content of it can be moved to either the `internal` package or one of the `openstack/utils` package. I would recommend doing this as a separate PR and include some unit tests.

For discussion, I think we can simplify this PR down to two topics:

1. Adding the Grant functionality. This is just a standard feature addition and I don't think there's anything to discuss here.
2. Restricting Grant to only work when a specific microversion has been set. This deserves some discussion since it'll be a first in Gophercloud.

In order to do 2, you need to check if the client has a microversion set and then compare that microversion with the minimum version needed for the service. I think this check can be turned into a function so that `GetAccess` (and in theory, `Create`, `Delete`, etc) would just have something like this:

```go
func GrantAccess(client *gophercloud.ServiceClient, opts GrantAccessOptsBuilder, id string) (r GrantAccessResult) {
    if err := supportedMicroversion(client) {
        r.Err = err
        return
    }

    ...
}
```

Maybe there's a better way of doing that, though -- that was just the first idea that came to me.

pospispa:
@jrperritt I'm pretty new to `gophercloud` library and to OpenStack. That's why I decided to implement the features I need first and discuss my approach later even though this approach may result in substantial rework.
I'll split this PR into several small ones.
I'll "copy/paste" the `openstackmicroversions` code into `openstack/utils` as suggested by @jtopjian so there will be no need for vendoring.

pospispa:
@jtopjian the below:
>>> Edit: Alternatively, we can forgo any microversion check and just let the API service return an error if the client was not configured correctly.
is a good comment. I've not thought about this option.

I tried to set `X-OpenStack-Manila-API-Version` header to below values and tried how the Manila API responds:
- No `X-OpenStack-Manila-API-Version` header results in error: "Resource not found"
- `X-OpenStack-Manila-API-Version` header set to "2.1" value results in error: "Resource not found"
- `X-OpenStack-Manila-API-Version` header set to "2.8" value results in error: "Resource not found"
- `X-OpenStack-Manila-API-Version` header set to "2.9" value succeeds in granting access rights.

So firstly, in case the microversion is not checked in the gophercloud library we end up with:
- more Manila API calls.
- currently, a generic error message that does not give a clue what exactly caused the error.

Secondly, in case we check the microversion in the gophercloud library we end up with:
- less Manila API calls.
- specific error message that clearly says what caused the error.
- quite a lot of code that parses and compares the microversions.

I prefer the second option, i.e. to check the microversions in the gophercloud library.

pospispa:
@jrperritt @jtopjian I started splitting this PR into small ones. I've chipped away PRs #435 #436 and #437 that are ready for review/discussion. I plan to continue tomorrow.

jtopjian:
> I prefer the second option, i.e. to check the microversions in the gophercloud library.

I agree. The microversion specs define a specific error that should be returned if there's an invalid microversion, but it's safe to say that it'll either take a while for all services to conform to the spec or possibly won't ever. Because we can't rely on a good error message, handling it in the client is probably good for a user experience.

pospispa:
Microversions are now discussed as a whole in https://github.com/gophercloud/gophercloud/issues/441
As soon as the discussion in https://github.com/gophercloud/gophercloud/issues/441 come to a conclusion the microversion part of this PR will be reworked.

pospispa:
> For discussion, I think we can simplify this PR down to two topics:
> 1. Adding the Grant functionality. This is just a standard feature addition and I don't think there's anything to discuss here.
> 2. Restricting Grant to only work when a specific microversion has been set. This deserves some discussion since it'll be a first in Gophercloud.

> Edit: Alternatively, we can forgo any microversion check and just let the API service return an error if the client was not configured correctly.

@jtopjian it seems that the microversion design discussions in https://github.com/gophercloud/gophercloud/issues/441 will take longer than I expected.
So can we just add the Grant Access functionality without microversion check and let the server return an error in case the client did not use a proper microversion?

Should I create an issue that the microversion check shall be added later?

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -110,3 +110,39 @@ type DeleteResult struct {
 type GetResult struct {
 	commonResult
 }
+
+// GrantAccessRes contains all information associated with an OpenStack share Grant Access
+type GrantAccessRes struct {
+	// The UUID of the share to which you are granted or denied access.
+	ShareID string `json:"share_id"`
+	// Timestamp when the share was created
+	CreatedAt time.Time `json:"created_at, omitempty"`
```
jtopjian: nit: `json:"created_at,omitempty"`

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -110,3 +110,39 @@ type DeleteResult struct {
...
+	// Timestamp when the share was updated
+	UpdatedAt time.Time `json:"updated_at, omitempty"`
```
jtopjian: nit: `json:"updated_at, omitempty"`

jtopjian:
> Should I create an issue that the microversion check shall be added later?

That would be the best way to go if you want to get basic Grant functionality in place quickly.

jtopjian:
Also, for review reference, here's the structure of the result body:

https://github.com/openstack/manila/blob/stable/newton/manila/api/views/share_accesses.py

I'm assuming the `update_version_resourced_dict` is what is adding the `CreatedAt` and `UpdatedAt` fields.

pospispa:
>> Should I create an issue that the microversion check shall be added later?

> That would be the best way to go if you want to get basic Grant functionality in place quickly.

@jtopjian I created: https://github.com/gophercloud/gophercloud/issues/447

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -110,3 +110,39 @@ type DeleteResult struct {
...
+	// The access level to the share is either "rw" or "ro".
+	AccessLevel string `json:"access_level,omitempty"`
+	// The state of the access rule
+	State string `json:"state,omitempty"`
+	// The access rule ID.
+	ID string `json:"id"`
+}
+
+// ExtractGrantAccess will get the GrantAccess object from the commonResult
+func (r commonResult) ExtractGrantAccess() (*GrantAccessRes, error) {
```
jtopjian: Per discussion in #381, this should be changed from `commonResult` to `GrantAccessResult`.

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -110,3 +110,39 @@ type DeleteResult struct {
 type GetResult struct {
 	commonResult
 }
+
+// GrantAccessRes contains all information associated with an OpenStack share Grant Access
+type GrantAccessRes struct {
```
pospispa: The name cannot be `GrantAccess` because there is a `func GrantAccess`.

openstack/sharedfilesystems/v2/shares/results.go
pospispa: Well, in case I have:
```
CreatedAt time.Time json:"created_at, omitempty"
```
I get an error:
```
syntax error: unexpected json, expecting semicolon, newline, or }
```
so I don't understand what you mean.

openstack/sharedfilesystems/v2/shares/results.go
pospispa: Well, in case I have:
```
UpdatedAt time.Time json:"updated_at, omitempty"
```
I get an error:
```
syntax error: unexpected json, expecting semicolon, newline, or }
```
so I don't understand what you mean.

pospispa:
> Also, for review reference, here's the structure of the result body:
> https://github.com/openstack/manila/blob/stable/newton/manila/api/views/share_accesses.py
> I'm assuming the `update_version_resourced_dict` is what is adding the `CreatedAt` and `UpdatedAt` fields.

I assume the same.

pospispa:
> One other area I'm curious about is the timestamps. The existing `Share` result has a `CreatedAt` field that has a custom unmarshaller. Would this be required for the `CreatedAt` and `UpdatedAt` fields in the `GrantAccessResult`, too? I would imagine that the timestamp format being returned is the same.

@jtopjian in constrast with [the documentation](https://developer.openstack.org/api-ref/shared-file-systems/?expanded=grant-access-detail#grant-access) I used wireshark to capture packet with `Grant Access` response and there're no `created_at` and `updated_at` fields.

So I looked again at [manila server code](https://github.com/openstack/manila/tree/stable/newton) and I didn't found the `created_at` and `updated_at` fields here neither.

That's why I think these fields aren't being sent so I suggest I'll remove them.

Please, what's your opinion?

pospispa:
@jtopjian I had to rebase this PR.

I believe I've fixed all your comments except a couple of nits and the `CreatedAt` and `UpdatedAt` fields.

Please, let me know if you have further comments.

openstack/sharedfilesystems/v2/shares/results.go
jtopjian: There is nowhere else in Gophercloud where "Res" is used. `GrantAccessRes` is not a suitable name. I suggest browsing Gophercloud and finding precedence for a good name to use here.

openstack/sharedfilesystems/v2/shares/results.go
jtopjian: Please remove the space between `created_at,` and `omitempty`. The full line should read:

```go
CreatedAt time.Time `json:"created_at,omitempty"`
```

jtopjian:
> in constrast with the documentation I used wireshark to capture packet with Grant Access response and there're no created_at and updated_at fields.

You don't need to use wireshark. You can use:

```shell
$ manila --debug ...
```

> Please, what's your opinion?

I appreciate you finishing the due diligence required to make sure this PR has all of the correct fields. If can find no evidence of those fields being returned, then they cannot be included here.

openstack/sharedfilesystems/v2/shares/results.go
pospispa: I renamed the `func GrantAccess` to `func GrantAccessRequest` and the `type GrantAccessRes` to `type GrantAccess`.

pospispa:
> I appreciate you finishing the due diligence required to make sure this PR has all of the correct fields. If can find no evidence of those fields being returned, then they cannot be included here.

@jtopjian I didn't find any evidence of those fields being returned so I deleted them.

Please, check if everything looks good to you.

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -85,3 +83,41 @@ func GetExportLocations(client *gophercloud.ServiceClient, id string) (r GetExpo
...
+
+// ToGrantAccessMap assembles a request body based on the contents of a
+// GrantAccessOpts.
+func (opts GrantAccessOpts) ToGrantAccessMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "allow_access")
+}
+
+// GrantAccessRequest will grant access to a Share based on the values in GrantAccessOpts. To extract
+// the GrantAccess object from the response, call the Extract method on the GrantAccessResult.
+func GrantAccessRequest(client *gophercloud.ServiceClient, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
```
jrperritt: Adding "Request" (or "Req", or "Res") is a non-starter. We'd like the functions to read well and be succinct. Users will know that this is a request because it starts with a verb ("Grant", which admittedly is very confusing, since "grant" can also be used as a noun). Let's call this `GrantAccess`.

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -144,3 +144,35 @@ func (r GetExportLocationsResult) Extract() ([]ExportLocation, error) {
 	err := r.ExtractInto(&s)
 	return s.ExportLocations, err
 }
+
+// GrantAccess contains all information associated with an OpenStack share Grant Access
+type GrantAccess struct {
```
jrperritt: Following on from my comment above on `GrantAccessRequest`, response objects should be nouns. Here, something as simple as `Access` or `AccessRight` would work fine (again, admittedly confusing because "access" can be both a noun and a verb).

Perhaps @jtopjian has a better suggestion for the name

pospispa: @jrperritt thank you for the suggestions. I personally like `AccessRight` more.

jtopjian: `AccessRight` works for me as well. I did not consider how nuanced the words in this PR could be, so I apologize about that.

One thing to keep in mind is that Manila has an API call to list access rules, which is a list of all `GrantAccess` calls that have been performed. Upon first glance, the body of these entries are slightly different than what `AccessRight` would be. Of course, it's always possible the docs are wrong.

https://developer.openstack.org/api-ref/shared-file-systems/?expanded=list-access-rules-detail#list-access-rules

If that API call gets implemented and new result structs need created, it might be a good idea to carry the `AccessRight` naming pattern to be consistent. Perhaps `AccessRightEntries` which is made up of individual `AccessRightEntry` (assuming `AccessRightEntry` is different than `AccessRight`).

Just thinking ahead. No action needs made.

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -85,3 +83,41 @@ func GetExportLocations(client *gophercloud.ServiceClient, id string) (r GetExpo
...
+// GrantAccess will grant access to a Share based on the values in GrantAccessOpts. To extract
+// the GrantAccess object from the response, call the Extract method on the GrantAccessResult.
+func GrantAccess(client *gophercloud.ServiceClient, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
+	b, err := opts.ToGrantAccessMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Post(grantAccessURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 201},
```
jtopjian: Do you know when a 201 would be returned? Have you seen a 201 in any `manila --debug` output? I have only been able to find a default 200 response [here](https://github.com/openstack/manila/blob/stable/newton/manila/api/openstack/wsgi.py#L369).

These return codes are usually the hardest pieces to find. I'm OK leaving both 200 and 201 here unless you're confident it's safe to drop 201.

pospispa: 201 is a copy/paste error from `func Create`.

pospispa:
@jtopjian @jrperritt please, do it look good to you? Should I squash the commits?

openstack/sharedfilesystems/v2/shares/requests.go
jtopjian: Yeah, `OkCodes` are usually just taken from the API docs, which we know to be incorrect from time to time. However, IMO, return codes are the hardest thing to search for, so I'm not very strict with them when doing a review.

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -144,3 +144,36 @@ func (r GetExportLocationsResult) Extract() ([]ExportLocation, error) {
...
+	// The state of the access rule
+	State string `json:"state,omitempty"`
+	// The access rule ID.
+	ID string `json:"id"`
+}
+
+// Extract will get the GrantAccess object from the commonResult
+func (r GrantAccessResult) Extract() (*AccessRight, error) {
+	var s struct {
+		AccessRight *AccessRight `json:"access"`
```
jtopjian: :+1: 

You're correct with `*AccessRight` here. I mistyped my last comment.

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -144,3 +144,36 @@ func (r GetExportLocationsResult) Extract() ([]ExportLocation, error) {
...
 	return s.ExportLocations, err
 }
+
+// AccessRight contains all information associated with an OpenStack share
+// Grant Access Response
+type AccessRight struct {
+	// The UUID of the share to which you are granted or denied access.
+	ShareID string `json:"share_id"`
+	// The access rule type that can be "ip", "cert" or "user".
+	AccessType string `json:"access_type,omitempty"`
```
jtopjian: @jrperritt General question: Does `omitempty` provide any use here?

If I do:

```shell
$ grep -R omitempty ~/go/src/github.com/gophercloud/gophercloud | grep result
```

I see that there are a good number of results.

jrperritt: > Does omitempty provide any use here?

No; I think `omitempty` is only relevant for encoding. But it doesn't hurt either, so... meh.

pospispa: `omitempty` is probably useless here.

pospispa:
@jtopjian thank you for merging the PR.

Issue #347, opened 2017-05-24
Issue with Security Group update


jtopjian:
I must have been half asleep. The way update was implemented in #307 is missing the secgroup ID in the resource URL.

jrperritt:
closed by #348

PR #348, opened 2017-05-24
Networking v2: Security Group Update Fix

jtopjian:
Fixes incorrect implementation of security group update.

For #347

200 response code is defined [here](https://github.com/openstack/neutron/blob/19a069d99f7249854fab9ff2706a38fe8b66bb1e/neutron/wsgi.py#L689). Since there's a result returned, the code is 200.

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/security_test.go
+++ b/acceptance/openstack/networking/v2/extensions/security_test.go
@@ -33,7 +33,7 @@ func TestSecurityGroupsList(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


jtopjian:
@jrperritt oops :( 

Ready for review.

jrperritt:
> oops :(

I'm not at all bothered by things like this :) If it doesn't work, people can't use it, and it can be fixed easily without breaking anyone. It's the design decisions that are hard to fix later.

LGTM

jtopjian:
Sounds good. Thanks!

Issue #350, opened 2017-05-29
Create function to determine extra fields in responses


jtopjian:
Per discussion in #228, it would be useful to have a function that can determine if a response body contains extra fields that are not defined in the resource's result struct.

jrperritt:
closed by #351

PR #351, opened 2017-05-29
internal: Add RemainingKeys function

jtopjian:
This commit adds the RemainingKeys function which can be used to
detect fields returned in response bodies but are not defined in
the resource's result struct.

For #350

### DIFF
--- a/internal/util.go
+++ b/internal/util.go
@@ -1,0 +1,25 @@
+package internal
+
+import (
+	"reflect"
+)
+
+// RemainingKeys will inspect a struct and compare it to a map. Any key that
+// is defined in a JSON tag of the struct will be removed from the map. The
+// remaining map and keys are then returned.
+//
+// This is useful for determining the extra fields returned in response bodies
+// for resources that can contain an arbitrary or dynamic number of fields.
+func RemainingKeys(s interface{}, m map[string]interface{}) map[string]interface{} {
+	valueOf := reflect.ValueOf(s)
+	typeOf := reflect.TypeOf(s)
+	for i := 0; i < valueOf.NumField(); i++ {
+		field := typeOf.Field(i)
+		tagValue := field.Tag.Get("json")
+		if _, ok := m[tagValue]; ok {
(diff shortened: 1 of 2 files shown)


jtopjian:
@jrperritt First draft of this. Thoughts?

internal/util.go
```diff
@@ -0,0 +1,25 @@
...
+	"reflect"
+)
+
+// RemainingKeys will inspect a struct and compare it to a map. Any key that
+// is defined in a JSON tag of the struct will be removed from the map. The
+// remaining map and keys are then returned.
+//
+// This is useful for determining the extra fields returned in response bodies
+// for resources that can contain an arbitrary or dynamic number of fields.
+func RemainingKeys(s interface{}, m map[string]interface{}) map[string]interface{} {
```
jrperritt: At first glance, I'd prefer one of the following:

1. `RemoveKnownKeys(s interface{}, m map[string]interface{})` (this one modifies `m`)
2. `RemainingKeys(s interface{}, m map[string]interface{}) (extras map[string]interface{})` (this one doesn't modify `m`)

I'd prefer 2

jtopjian:
Like that?

jtopjian:
Also, I have no idea why the Travis test is failing on Go 1.7. I'm unable to reproduce the reported problem.

jrperritt:
> Like that?

Yes

> Also, I have no idea why the Travis test is failing on Go 1.7

I'm not sure either. This is the only thing that I noticed from the Travis output:
`go build github.com/gophercloud/gophercloud/internal/testing: no buildable Go source files in /home/travis/gopath/src/github.com/gophercloud/gophercloud/internal/testing`
Not a concern though: Go 1.7 should probably be removed from the builder script anyway.

jtopjian:
@jrperritt Anything I can do to move this one forward? It'd allow me to continue work with the Identity v3 users as well as Glance properties.

jrperritt:
Yes, you can subtly remind me that I'd forgotten about this ;)

LGTM

PR #352, opened 2017-05-30
[WIP] Fix Microversion Header

pospispa:
For #53 

code in this PR:
https://github.com/openstack/manila/blob/master/manila/common/config.py#L194
https://github.com/openstack/manila/blob/master/manila/common/config.py#L203


pospispa:
Solution in https://github.com/gophercloud/gophercloud/pull/355 was chosen that's why this PR is closed without merge.

Issue #349, opened 2017-05-26
Block Storage (Cinder) Snapshot Support


rootfs:
There is no Cinder Snapshot API support yet. Snapshot API will be needed in Kubernetes volume snapshot feature.

@jrperritt

j-griffith:
I'll have a look at V3 shortly, no changes required for the Snapshot (V3.0 is actually V2), but it would be nice to get support for the new Attachment API sooner rather than later I think.

jtopjian:
Implemented by #354

PR #353, opened 2017-06-01
Add Cinder V2 Snapshot support

j-griffith:
This adds the base Cinder V2 Snapshot object to Gophercloud.
While it'd be nice to jump straight to Cinder's V3, there's
likely no shortage of folks still running Mitake or older.

For #349 

code in this PR:

https://developer.openstack.org/api-ref/block-storage/v2/index.html#volume-snapshots-snapshots

### DIFF
--- a/openstack/blockstorage/v2/snapshots/doc.go
+++ b/openstack/blockstorage/v2/snapshots/doc.go
@@ -1,0 +1,5 @@
+// Package snapshots provides information and interaction with snapshots in the
+// OpenStack Block Storage service. A snapshot is a point in time copy of the
+// data contained in an external storage volume, and can be controlled
+// programmatically.
+package snapshots
--- a/openstack/blockstorage/v2/snapshots/requests.go
+++ b/openstack/blockstorage/v2/snapshots/requests.go
@@ -1,0 +1,158 @@
+package snapshots
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// CreateOptsBuilder allows extensions to add additional parameters to the
+// Create request.
+type CreateOptsBuilder interface {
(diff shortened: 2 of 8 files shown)


jtopjian:
@j-griffith Thanks for this.

Just doing a cursory scan, this looks like it was taken from the Rackspace repo. The overall design has changed quite a bit, but shouldn't be too hard to change. I recommend taking a look at, say, the [volumes](https://github.com/gophercloud/gophercloud/tree/master/openstack/blockstorage/v2/volumes) package and comparing. For example:

* New import path
* JSON tags in the Create / Update options
* Removal of mapstructure

Also, it helps _a lot_ to have code links/references to any schema, objects, routes, etc from the Cinder service to support the code here. See [here](https://github.com/gophercloud/gophercloud/pull/271) and [here](https://github.com/gophercloud/gophercloud/issues/157) for examples. It might be a little presumptuous, but you might be able to dig up those references faster than one of the maintainers here :)

jrperritt:
@j-griffith Firstly, thank you for the PR. Just by skimming it, there are a few things you'll need to address before this can be reviewed:

> Links to the line numbers/files in the OpenStack source code that support the
code in this PR

That part is literal. Namely, the docs aren't good enough. We need the lines in the github.com/openstack/cinder source code. Have a look at any of @jtopjian 's PRs.

I noticed that you originally posted the issue to the `rackspace/gophercloud` repo and maybe you just copied it over? If so, things are a bit more stringent stylistically in this repo, but we've laid it out in a [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md).

You can also browse through the source code for examples.

jrperritt:
Wow, @jtopjian. You beat me by 1 minute, but we seem to be very much in sync.

j-griffith:
Sounds good, I'll get it turned around, thanks!

PR #354, opened 2017-06-01
Add Cinder V2 Snapshot support

j-griffith:
Note that there is little difference between V1 and V2, and
subsequently even V3.

For #349

Links to source code regarding the API additions that include params:
  * Create
    https://github.com/openstack/cinder/blob/master/cinder/api/v2/snapshots.py#L135

  * ViewBuilder for snapshots (API Snapshot object from Cinder)
    https://github.com/openstack/cinder/blob/master/cinder/api/views/snapshots.py#L45

  * Snapshot Metadata controller
    https://github.com/openstack/cinder/blob/master/cinder/api/v2/snapshot_metadata.py

Example usage from the cinderclient to show options and url's used (note
the client does hacky things like import other version objects,
fortunately there's no delta between base V3.0 and v2):
  * Create
    https://github.com/openstack/python-cinderclient/blob/master/cinderclient/v3/volume_snapshots.py#L82

### DIFF
--- a/acceptance/openstack/blockstorage/v2/snapshots_test.go
+++ b/acceptance/openstack/blockstorage/v2/snapshots_test.go
@@ -1,0 +1,58 @@
+// +build acceptance blockstorage
+
+package v2
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v2/snapshots"
+)
+
+func TestSnapshotsList(t *testing.T) {
+	client, err := clients.NewBlockStorageV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create a blockstorage client: %v", err)
+	}
+
+	allPages, err := snapshots.List(client, snapshots.ListOpts{}).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to retrieve snapshots: %v", err)
+	}
+
+	allSnapshots, err := snapshots.ExtractSnapshots(allPages)
+	if err != nil {
+		t.Fatalf("Unable to extract snapshots: %v", err)
+	}
+
+	for _, snapshot := range allSnapshots {
+		tools.PrintResource(t, snapshot)
+	}
+}
+
+func TestSnapshotsCreateDelete(t *testing.T) {
+	client, err := clients.NewBlockStorageV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create a blockstorage client: %v", err)
+	}
+
+	volume, err := CreateVolume(t, client)
+	if err != nil {
+		t.Fatalf("Unable to create volume: %v", err)
+	}
+	defer DeleteVolume(t, client, volume)
+
+	snapshot, err := CreateSnapshot(t, client, volume)
+	if err != nil {
+		t.Fatalf("Unable to create snapshot: %v", err)
+	}
+	defer DeleteSnapshot(t, client, snapshot)
+
+	newSnapshot, err := snapshots.Get(client, snapshot.ID).Extract()
+	if err != nil {
+		t.Errorf("Unable to retrieve snapshot: %v", err)
+	}
+
+	tools.PrintResource(t, newSnapshot)
+}
--- a/openstack/blockstorage/v2/snapshots/requests.go
+++ b/openstack/blockstorage/v2/snapshots/requests.go
@@ -1,0 +1,158 @@
+package snapshots
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// CreateOptsBuilder allows extensions to add additional parameters to the
+// Create request.
+type CreateOptsBuilder interface {
+	ToSnapshotCreateMap() (map[string]interface{}, error)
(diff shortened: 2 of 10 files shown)


openstack/blockstorage/v2/snapshots/requests.go
```diff
@@ -0,0 +1,158 @@
...
+// extract the Snapshot object from the response, call the Extract method on the
+// CreateResult.
+func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResult) {
+	b, err := opts.ToSnapshotCreateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Post(createURL(client), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 201},
```
jtopjian: I did a quick copy of the [v1 snapshot acceptance tests](https://github.com/gophercloud/gophercloud/tree/master/acceptance/openstack/blockstorage/v1) to test v2 and I got the following error:

```
=== RUN   TestSnapshotsList
--- PASS: TestSnapshotsList (0.29s)
=== RUN   TestSnapshotsCreateDelete
--- FAIL: TestSnapshotsCreateDelete (5.46s)
        blockstorage.go:53: Attempting to create volume: ACPTTESTIY8ygDIM6NjjZdWr
        blockstorage.go:25: Attempting to create snapshot ACPTTESTEPU7UzMfgx0Z1lD1 based on volume 72f8d398-d682-4bd1-b3f3-b8ecc32a387a
        snapshots_test.go:48: Unable to create snapshot: Expected HTTP response code [200 201] when accessing [POST https://foo:8776/v2/9aa5f9f66b4b417d84d778a23acdf45b/snapshots], but got 202 instead
                {"snapshot": {"status": "creating", "description": null, "created_at": "2017-06-01T04:38:54.860173", "metadata": {}, "volume_id": "72f8d398-d682-4bd1-b3f3-b8ecc32a387a", "size": 1, "id": "29be3eb4-738c-4db1-bad6-8d18082dbb45", "name": "ACPTTESTEPU7UzMfgx0Z1lD1"}}
        blockstorage.go:138: Unable to delete volume 72f8d398-d682-4bd1-b3f3-b8ecc32a387a: Invalid request due to incorrect syntax or missing required parameters.
FAIL
```

I updated the `OkCodes` with just `202` (to match v2 volumes) and the tests passed. Is 202 the right code to use here (and below)?

openstack/blockstorage/v2/snapshots/results.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Snapshot contains all the information associated with a Cinder Snapshot.
+type Snapshot struct {
+	// Unique identifier.
+	ID string `json:"id"`
+
+	// Date created.
+	CreatedAt gophercloud.JSONRFC3339MilliNoZ `json:"created_at"`
```
jtopjian: @jrperritt [v1 snapshots](https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v1/snapshots/results.go) (and v1 volumes) aren't set to `time.Time` -- was that missed during the "time" update? I can't think of a reason that they would follow a different pattern.

@j-griffith If `time.Time` should be used, you'll want to set `CreatedAt` and `UpdatedAt` to `time.Time` here and then do the conversion to `gophercloud.JSONRFC3339MilliNoZ` in an `UnmarshalJSON` function like [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v2/volumes/results.go#L80-L97).

The unit tests will also need updated like [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v2/volumes/testing/requests_test.go#L69).

jrperritt: @jtopjian yes, the `v1/snapshots` package was missed.

openstack/blockstorage/v2/snapshots/requests.go
j-griffith: My bad, yes it's a 202, I'll update and link to the API doc as reference

j-griffith:
Verified/Updated the response codes, modified the handling of time-stamps and updated the acceptance test.  Let me know if there's still some missing detail, and thanks again!

acceptance/openstack/blockstorage/v2/snapshots_test.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		t.Fatalf("Unable to create volume: %v", err)
+	}
+	defer DeleteVolume(t, client, volume)
+
+	snapshot, err := CreateSnapshot(t, client, volume)
+	if err != nil {
+		t.Fatalf("Unable to create snapshot: %v", err)
+	}
+	defer DeleteSnapshotshot(t, client, snapshot)
```
jtopjian: `s/DeleteSnapshotshot/DeleteSnapshot/`

j-griffith: DOHHHH!!!!

jtopjian:
And now Travis is reporting some gofmt/goimports changes. So close :)

jtopjian:
This looks good to me. Thank you for submitting this and for your help with it :)

j-griffith:
Thanks Joe!

PR #355, opened 2017-06-01
fix microversion header

jrperritt:
For #53

code in this PR:

https://github.com/openstack/manila/blob/stable/newton/manila/common/config.py#L196
https://github.com/openstack/nova/blob/acfbec7db9a3df5a3183a2730510fa1f0fea92bf/nova/api/openstack/wsgi.py#L68

I think those are the only 2 services that Gophercloud currently supports that offer microversions. It does look like the `X-OpenStack-<SERVICE>-API-Version` headers are being replaced by `OpenStack-API-Version`, but are still around for backwards-compatibility, so I'm adding both. Hopefully that won't cause a problem.

This is an alternative implementation to the ones in #341 and #352 . I think it's cleaner and doesn't force users to identify the service type.

It also refactors some of the common code when making requests and creating service clients.

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -182,9 +182,10 @@ func v3auth(client *gophercloud.ProviderClient, endpoint string, opts tokens3.Au
 // NewIdentityV2 creates a ServiceClient that may be used to interact with the v2 identity service.
 func NewIdentityV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
 	endpoint := client.IdentityBase + "v2.0/"
+	clientType := "identity"
 	var err error
 	if !reflect.DeepEqual(eo, gophercloud.EndpointOpts{}) {
-		eo.ApplyDefaults("identity")
+		eo.ApplyDefaults(clientType)
 		endpoint, err = client.EndpointLocator(eo)
 		if err != nil {
 			return nil, err
@@ -194,15 +195,17 @@ func NewIdentityV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOp
 	return &gophercloud.ServiceClient{
 		ProviderClient: client,
 		Endpoint:       endpoint,
+		Type:           clientType,
 	}, nil
 }
 
 // NewIdentityV3 creates a ServiceClient that may be used to access the v3 identity service.
 func NewIdentityV3(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
 	endpoint := client.IdentityBase + "v3/"
+	clientType := "identity"
 	var err error
 	if !reflect.DeepEqual(eo, gophercloud.EndpointOpts{}) {
-		eo.ApplyDefaults("identity")
+		eo.ApplyDefaults(clientType)
 		endpoint, err = client.EndpointLocator(eo)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


jtopjian:
> I think those are the only 2 services that Gophercloud currently supports that offer microversions.

I did a quick inventory here: https://gist.github.com/jtopjian/183bf8c7cb81d0c29717a42b78b9b762

It looks like Cinder v3 supports microversions, but Gophercloud doesn't have v3 support.

Glance is questionable. No idea what it's doing.

>  Hopefully that won't cause a problem.

I think it's OK.

I modified the `TestKeypairsList` acc test with:

```go
client.Microversion = "2.42"
```

and also did a dump of the request. I saw both headers and the request was successful. I then did an invalid request of:

```go
client.Microversion = "2.52"
```

And successfully go:

```
        keypairs_test.go:26: Unable to retrieve keypairs: Expected HTTP response code [200 204] when accessing [GET http://127.0.0.1:8774/v2.1/a99e9b4e620e4db09a2dfb6e42a01e66/os-keypairs], but got 406 instead
                {"computeFault": {"message": "Version 2.52 is not supported by the API. Minimum is 2.1 and maximum is 2.42.", "code": 406}}
```

This implementation looks good to me. It's a very clean implementation and I was able to digest it quickly. Absolutely no offence is meant to the other PRs, though.

I think this is a fine implementation. I have some other thoughts / questions that I'll add to #53 since they're not code related, but this could easily be a foundation for some other things, if applicable.

jtopjian:
Also, I'll do another review and some more testing later and then sign off on it.

service_client.go
```diff
@@ -37,105 +43,77 @@ func (client *ServiceClient) ServiceURL(parts ...string) string {
...
+	switch client.Type {
+	case "compute":
+		opts.MoreHeaders["X-OpenStack-Nova-API-Version"] = client.Microversion
+	case "sharev2":
+		opts.MoreHeaders["X-OpenStack-Manila-API-Version"] = client.Microversion
 	}
-	opts.MoreHeaders["X-OpenStack-Nova-API-Version"] = client.Microversion
-
-	return client.Request("DELETE", url, opts)
+	opts.MoreHeaders["OpenStack-API-Version"] = client.Type + " " + client.Microversion
```
pospispa: I would rather check whether `client.Type != ""`

openstack/client.go
```diff
@@ -212,125 +215,81 @@ func NewIdentityV3(client *gophercloud.ProviderClient, eo gophercloud.EndpointOp
...
+	eo.ApplyDefaults(clientType)
 	url, err := client.EndpointLocator(eo)
 	if err != nil {
-		return nil, err
+		return sc, err
 	}
-	return &gophercloud.ServiceClient{ProviderClient: client, Endpoint: url}, nil
+	sc.ProviderClient = client
+	sc.Endpoint = url
+	sc.Type = clientType
```
pospispa: It is assumed that `client.EndpointLocator` will return an endpoint URL that's service type equals the requested `clientType`. I think this is fine and it's how `client.EndpointLocator` works now.
However, there are other service information that are available in [`func V2EndpointURL`](https://github.com/gophercloud/gophercloud/blob/master/openstack/endpoint_location.go#L15) and [`func V3EndpointURL`](https://github.com/gophercloud/gophercloud/blob/master/openstack/endpoint_location.go#L63) and are not returned by the functions and that may be useful in the `gophercloud.ServiceClient` struct in the future.
IMHO, we may open the doors for returning these information from the functions [by adding one more return value to the `EndpointLocator` type](https://github.com/gophercloud/gophercloud/pull/352/commits/e8c06e99db6f207b542e75bba521e2b8a978cc84).

service_client.go
```diff
@@ -37,105 +43,77 @@ func (client *ServiceClient) ServiceURL(parts ...string) string {
 	return client.ResourceBaseURL() + strings.Join(parts, "/")
 }
 
-// Get calls `Request` with the "GET" HTTP verb.
-func (client *ServiceClient) Get(url string, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
-	if opts == nil {
-		opts = &RequestOpts{}
+func (client *ServiceClient) initReqOpts(url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) {
```
pospispa: I like how the duplicate code is replaced by this function.

openstack/client.go
jrperritt: For me, what you've suggested in #352 is too complex for what's required here. I don't want to throw all information returned from an endpoint into a `struct`.

service_client.go
jrperritt: That's reasonable

service_client.go
```diff
@@ -37,105 +43,77 @@ func (client *ServiceClient) ServiceURL(parts ...string) string {
...
+	return client.Request("DELETE", url, opts)
+}
 
-	if opts.MoreHeaders == nil {
-		opts.MoreHeaders = make(map[string]string)
+func (client *ServiceClient) setMicroversionHeader(opts *RequestOpts) {
+	switch client.Type {
+	case "compute":
+		opts.MoreHeaders["X-OpenStack-Nova-API-Version"] = client.Microversion
+	case "sharev2":
```
jrperritt: True. That may be something to add later when there are several more client types supporting microversions.

pospispa: I agree.

jrperritt:
We're going to go forward with this one. It adds the least amount of complexity while also allowing users to simply supply a service's microversion. If more information is needed by the `ServiceClient` in the future, we can address that then. Thank y'all for the reviews @jtopjian @pospispa.

openstack/client.go
pospispa: Currently, there is no mechanism for returning information from endpoint that would be easily extensible. That's why I'm proposing to add a struct as a new endpoint return value. The struct can be easily extended in the future and doesn't have to contain all currently available information.

As you introduced the `func initClientOpts`, the `client.EndpointLocator` is called at less places than before. So it's now easier to add new return values to the `client.EndpointLocator` function.

I agree with you solution.

Issue #360, opened 2017-06-02
Identity v3: Add Extract{User,Role,Project} to token


kwapik:
There should be a way to extract following attributes from token:
- Roles - https://github.com/openstack/keystone/blob/32577e08b61b31cda045513f06535b474a54d2a2/keystone/models/token_model.py#L242
- Project - https://github.com/openstack/keystone/blob/32577e08b61b31cda045513f06535b474a54d2a2/keystone/models/token_model.py#L140
- User - https://github.com/openstack/keystone/blob/32577e08b61b31cda045513f06535b474a54d2a2/keystone/models/token_model.py#L89

jrperritt:
closed by #361

PR #361, opened 2017-06-02
[RFR] Extract values from v3 token

kwapik:
For #360

code in this PR:

- Roles - https://github.com/openstack/keystone/blob/32577e08b61b31cda045513f06535b474a54d2a2/keystone/models/token_model.py#L242
- Project - https://github.com/openstack/keystone/blob/32577e08b61b31cda045513f06535b474a54d2a2/keystone/models/token_model.py#L140
- User - https://github.com/openstack/keystone/blob/32577e08b61b31cda045513f06535b474a54d2a2/keystone/models/token_model.py#L89

### DIFF
--- a/openstack/identity/v3/tokens/results.go
+++ b/openstack/identity/v3/tokens/results.go
@@ -122,10 +122,10 @@ func (r commonResult) ExtractRoles() (*Roles, error) {
 // ExtractProject returns Project to which User is authorized.
 func (r commonResult) ExtractProject() (*Project, error) {
 	var s struct {
-		Project `json:"project"`
+		*Project `json:"project"`
 	}
 	err := r.ExtractInto(&s)
-	return &s.Project, err
+	return s.Project, err
 }
 
 // CreateResult defers the interpretation of a created token.
--- a/openstack/identity/v3/tokens/results.go
+++ b/openstack/identity/v3/tokens/results.go
@@ -64,6 +64,12 @@ type Roles struct {
 	Roles []*Role `json:"roles"`
 }
 
+// Project provides information about project to which User is authorized.
+type Project struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
 // commonResult is the deferred result of a Create or a Get call.
 type commonResult struct {
 	gophercloud.Result
@@ -110,6 +116,15 @@ func (r commonResult) ExtractRoles() (*Roles, error) {
 	var s Roles
 	err := r.ExtractInto(&s)
 	return &s, err
+}
+
+// ExtractProject returns Project to which User is authorized.
+func (r commonResult) ExtractProject() (*Project, error) {
+	var s struct {
+		Project `json:"project"`
+	}
+	err := r.ExtractInto(&s)
+	return &s.Project, err
 }
 
 // CreateResult defers the interpretation of a created token.
--- a/openstack/identity/v3/tokens/results.go
+++ b/openstack/identity/v3/tokens/results.go
@@ -106,7 +106,7 @@ func (r commonResult) ExtractServiceCatalog() (*ServiceCatalog, error) {
 // ExtractUser returns the User that is the owner of the Token.
 func (r commonResult) ExtractUser() (*User, error) {
 	var s struct {
-		*User `json:"user"`
+		User *User `json:"user"`
 	}
 	err := r.ExtractInto(&s)
 	return s.User, err
--- a/openstack/identity/v3/tokens/results.go
+++ b/openstack/identity/v3/tokens/results.go
@@ -106,10 +106,10 @@ func (r commonResult) ExtractServiceCatalog() (*ServiceCatalog, error) {
 // ExtractUser returns the User that is the owner of the Token.
 func (r commonResult) ExtractUser() (*User, error) {
 	var s struct {
-		User `json:"user"`
+		*User `json:"user"`
 	}
 	err := r.ExtractInto(&s)
-	return &s.User, err
+	return s.User, err
 }
 
 // ExtractRoles returns Roles to which User is authorized.
--- a/openstack/identity/v3/tokens/results.go
+++ b/openstack/identity/v3/tokens/results.go
@@ -122,7 +122,7 @@ func (r commonResult) ExtractRoles() (*Roles, error) {
 // ExtractProject returns Project to which User is authorized.
 func (r commonResult) ExtractProject() (*Project, error) {
 	var s struct {
-		*Project `json:"project"`
+		Project *Project `json:"project"`
 	}
 	err := r.ExtractInto(&s)
 	return s.Project, err
--- a/openstack/identity/v3/tokens/results.go
+++ b/openstack/identity/v3/tokens/results.go
@@ -61,7 +61,7 @@ type Role struct {
 }
 
 type Roles struct {
-	Roles []*Role `json:"roles"`
+	Roles []Role `json:"roles"`
 }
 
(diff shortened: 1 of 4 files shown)


kwapik:
Hi @jrperritt ,

could you take a look?
Do I need to implement some acceptance tests as well?

jtopjian:
@kwapik Acceptance tests aren't required but I usually end up making them at some point just to verify usability / compatibility. If you're able to write a few quick tests, go for it :)

jtopjian:
@kwapik In general this looks good. I have a few comments:

The code you linked to looks like the equivalent of what you're implementing here, just in Python. That's OK. I think what's also important is how the token response is being built. It looks like that is done throughout here:

https://github.com/openstack/keystone/blob/49b74dd90a5bccd03b8432978ab92a7ef4ef727c/keystone/token/providers/common.py

For example:

https://github.com/openstack/keystone/blob/49b74dd90a5bccd03b8432978ab92a7ef4ef727c/keystone/token/providers/common.py#L268

https://github.com/openstack/keystone/blob/49b74dd90a5bccd03b8432978ab92a7ef4ef727c/keystone/token/providers/common.py#L241

https://github.com/openstack/keystone/blob/49b74dd90a5bccd03b8432978ab92a7ef4ef727c/keystone/token/providers/common.py#L110

Secondly, I think the unit tests might be better if you used an entire token response as a fixture, rather than just testing small pieces of the response. Here's a fixture you can use:

https://gist.github.com/jtopjian/a7168e0418140d618e7e10ee1c80197b

Third, the example response above shows that `project` also contains `domain`. Should this be accounted for, too?

kwapik:
Hi @jtopjian,

Thanks for your quick review.
I wasn't sure to which part of keystone code urls should point. I'm not really familiar with keystone code itself.

> Secondly, I think the unit tests might be better if you used an entire token response as a fixture, rather than just testing small pieces of the response. Here's a fixture you can use:
>
> https://gist.github.com/jtopjian/a7168e0418140d618e7e10ee1c80197b

Updated test to use this fixture. I deleted some catalog entries, IMO there is no point to check so many of them.

> Third, the example response above shows that project also contains domain. Should this be accounted for, too?

Included this also.

> Acceptance tests aren't required but I usually end up making them at some point just to verify usability / compatibility. If you're able to write a few quick tests, go for it :)

Created simple one. I'm not sure if they are run by CI tool though.

jtopjian:
@kwapik Thanks!

Just one small change to the acceptance tests. Instead of:

```go
client, err := clients.NewIdentityV3UnauthenticatedClient()
```

do

```go
client, err := clients.NewIdentityV3Client()
```

I think I was originally doing the former to test authentication via a token. But that's generating an unscoped token and causing the tests to fail. The latter version makes all tests pass.

Other than that, this looks good to me and can be merged once the above is done. :)

openstack/identity/v3/tokens/results.go
```diff
@@ -67,12 +97,37 @@ func (r commonResult) ExtractToken() (*Token, error) {
...
+func (r commonResult) ExtractRoles() (*Roles, error) {
+	var s Roles
+	err := r.ExtractInto(&s)
+	return &s, err
+}
+
+// ExtractProject returns Project to which User is authorized.
+func (r commonResult) ExtractProject() (*Project, error) {
+	var s struct {
+		Project `json:"project"`
```
jrperritt: `Project *Project`

openstack/identity/v3/tokens/results.go
```diff
@@ -67,12 +97,37 @@ func (r commonResult) ExtractToken() (*Token, error) {
...
+func (r commonResult) ExtractServiceCatalog() (*ServiceCatalog, error) {
 	var s ServiceCatalog
 	err := r.ExtractInto(&s)
 	return &s, err
 }
 
+// ExtractUser returns the User that is the owner of the Token.
+func (r commonResult) ExtractUser() (*User, error) {
+	var s struct {
+		User `json:"user"`
```
jrperritt: `User *User`

openstack/identity/v3/tokens/results.go
```diff
@@ -41,6 +41,36 @@ type ServiceCatalog struct {
...
+}
+
+// Role provides information about roles to which User is authorized.
+type Role struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+type Roles struct {
+	Roles []*Role `json:"roles"`
```
jrperritt: `Roles []Role`

jtopjian: Why `[]Role` here?

jrperritt: 1. Consistency
2. Slices are already pointers

jtopjian: > Slices are already pointers

👍 

Thanks for catching the other ones, too. I'm going to see if there's a `vet` or lint gate test that can catch those.

kwapik: Done.

openstack/identity/v3/tokens/results.go
kwapik: Done.

openstack/identity/v3/tokens/results.go
kwapik: Done.

kwapik:
@jrperritt @jtopjian I fixed the code according to your comments.

Side note: It is not stated in Contribution Guide how changes after review should be commited. I created `Fixup` commits and will rebase them after approval, but I could also amend previous ones or create new like `Review changes`. Personally I prefer the first approach, because it's outcome is the cleaner during review and after merge. Which one do you prefer and recommend in this repo?

jtopjian:
> Personally I prefer the first approach, because it's outcome is the cleaner during review and after merge. Which one do you prefer and recommend in this repo?

That's the correct approach. You also don't need to squash or rebase prior to a merge. Github can handle that now. 

The STYLEGUIDE has the following note:

> Unless explicitly asked, do not squash commits in the middle of a review; only append. It makes it difficult for the reviewer to see what's changed from one review to the next.

Although maybe it should be moved / copied to the CONTRIBUTING guide.

openstack/identity/v3/tokens/results.go
jrperritt: Not quite: this is what it should look like:
```go
User *User  `json:"user"`
```

Same with `Project` below.

kwapik: Sorry, missed the second comment. Fixed it properly now. Project related change also.

kwapik:
@jrperritt 
code should be ok now.

openstack/identity/v3/tokens/results.go
```diff
@@ -67,12 +97,37 @@ func (r commonResult) ExtractToken() (*Token, error) {
...
+func (r commonResult) ExtractUser() (*User, error) {
+	var s struct {
+		User *User `json:"user"`
+	}
+	err := r.ExtractInto(&s)
+	return s.User, err
+}
+
+// ExtractRoles returns Roles to which User is authorized.
+func (r commonResult) ExtractRoles() (*Roles, error) {
```
jrperritt: Instead of `*Roles`, let's just return `[]Role` and remove the `Roles` type.

kwapik: @jrperritt Done.

BTW I probably messed something up while creating fixups and they won't squash automatically. This would need manual work.

kwapik:
@jrperritt could you take a look one more time?

jrperritt:
LGTM

Issue #367, opened 2017-06-06
Snapshot CreateOpts Metadata type should be consistent with other resources


j-griffith:
Initial implementation of Snapshots in BlockStorage used type map[string]interface{} for the Metadata entry.  This can work, but we should be consistent and match the type we use with Volumes (map[string]string) otherwise it's kinda goofy for the end-user.

jtopjian:
Fixed by #368

PR #368, opened 2017-06-06
Use map[string]string for snapshot createopt

j-griffith:
Snapshot create opts are using map[string]interface{} for the
metadata opt.  This is doable, but we should be consistent with
other objects/resources and use map[string]string.

I missed this on the initial commit, so this patch just modifies
the CreateOption struct to make it match the types we use in Volumes.

For #367

### DIFF
--- a/openstack/blockstorage/v2/snapshots/requests.go
+++ b/openstack/blockstorage/v2/snapshots/requests.go
@@ -15,11 +15,11 @@ type CreateOptsBuilder interface {
 // the snapshots.Create function. For more information about these parameters,
 // see the Snapshot object.
 type CreateOpts struct {
-	VolumeID    string                 `json:"volume_id" required:"true"`
-	Force       bool                   `json:"force,omitempty"`
-	Name        string                 `json:"name,omitempty"`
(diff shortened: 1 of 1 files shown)


jtopjian:
I'm having a hard time seeing where a string is being enforced, but this should be enough proof that the value is being sent as a string:

```shell
$ cinder --debug snapshot-create --metadata foo=2 --name foo-snap 1c671e57-9954-40c3-ada3-0142d72b3b77
...
DEBUG:keystoneauth:REQ: curl -g -i -X POST https://example.com:8776/v2/9aa5f9f66b4b417d84d778a23acdf45b/snapshots -H "User-Agent: python-cinderclient" -H "Content-Type: application/json" -H "Accept: application/json" -H "X-Auth-Token: {SHA1}c9100a031e406007fdea758f04ddb2abe54b18be" -d '{"snapshot": {"description": null, "metadata": {"foo": "2"}, "force": false, "name": "foo-snap", "volume_id": "1c671e57-9954-40c3-ada3-0142d72b3b77"}}'
...
```

Thanks for quickly catching this.

Issue #366, opened 2017-06-05
ImageService V2 images.List is broken on 2nd page retrieval


tatsuhiro-t:
I have a problem getting list of images using ImageService v2 images.List.  First page is OK, but the calculation of the next URL is broken, and it gives me "Resource not found" error.

In particular, in [nextPageURL](https://github.com/gophercloud/gophercloud/blob/d8ccf0f660340eb612d757d632214cc48f87e962/openstack/imageservice/v2/images/urls.go#L41) function, if currentURL is http://example.com/v2/images, and next is /v2/images?marker=..., then the next URL becomes http://example.com/v2/v2/images?marker=....  Notice that extra "/v2" in path component.

tatsuhiro-t:
The following changes work for me:

```go
diff --git a/openstack/imageservice/v2/images/results.go b/openstack/imageservice/v2/images/results.go
index 09996f4..462f28f 100644
--- a/openstack/imageservice/v2/images/results.go
+++ b/openstack/imageservice/v2/images/results.go
@@ -161,7 +161,7 @@ func (r ImagePage) NextPageURL() (string, error) {
                return "", nil
        }
 
-       return nextPageURL(r.URL.String(), s.Next), nil
+       return nextPageURL(r.URL.String(), s.Next)
 }
 
 // ExtractImages interprets the results of a single page from a List() call, producing a slice of Image entities.
diff --git a/openstack/imageservice/v2/images/urls.go b/openstack/imageservice/v2/images/urls.go
index 58cb8f7..bf7cea1 100644
--- a/openstack/imageservice/v2/images/urls.go
+++ b/openstack/imageservice/v2/images/urls.go
@@ -1,7 +1,7 @@
 package images
 
 import (
-       "strings"
+       "net/url"
 
        "github.com/gophercloud/gophercloud"
 )
@@ -38,7 +38,14 @@ func deleteURL(c *gophercloud.ServiceClient, imageID string) string {
 }
 
 // builds next page full url based on current url
-func nextPageURL(currentURL string, next string) string {
-       base := currentURL[:strings.Index(currentURL, "/images")]
-       return base + next
+func nextPageURL(currentURL string, next string) (string, error) {
+       base, err := url.Parse(currentURL)
+       if err != nil {
+               return "", err
+       }
+       rel, err := url.Parse(next)
+       if err != nil {
+               return "", err
+       }
+       return base.ResolveReference(rel).String(), nil
 }
```

[...]

PR #369, opened 2017-06-07
Fix broken next URI of ImageService v2 images.List

tatsuhiro-t:
For #366

### DIFF
--- a/openstack/imageservice/v2/images/results.go
+++ b/openstack/imageservice/v2/images/results.go
@@ -161,7 +161,7 @@ func (r ImagePage) NextPageURL() (string, error) {
 		return "", nil
 	}
 
-	return nextPageURL(r.URL.String(), s.Next), nil
+	return nextPageURL(r.URL.String(), s.Next)
 }
 
 // ExtractImages interprets the results of a single page from a List() call, producing a slice of Image entities.
--- a/openstack/imageservice/v2/images/urls.go
+++ b/openstack/imageservice/v2/images/urls.go
@@ -1,7 +1,7 @@
 package images
 
 import (
-	"strings"
+	"net/url"
 
 	"github.com/gophercloud/gophercloud"
 )
@@ -38,7 +38,14 @@ func deleteURL(c *gophercloud.ServiceClient, imageID string) string {
 }
 
 // builds next page full url based on current url
-func nextPageURL(currentURL string, next string) string {
-	base := currentURL[:strings.Index(currentURL, "/images")]
-	return base + next
+func nextPageURL(currentURL string, next string) (string, error) {
+	base, err := url.Parse(currentURL)
+	if err != nil {
+		return "", err
+	}
+	rel, err := url.Parse(next)
+	if err != nil {
+		return "", err
+	}
+	return base.ResolveReference(rel).String(), nil
 }


jtopjian:
@tatsuhiro-t Thanks - I was coincidentally looking at #366 :)

A more simple solution is to just do this:

```go
base := currentURL[:strings.Index(currentURL, "/v2/images")]
```

This works against a running OpenStack environment (see #370), but it does not work with the unit tests. I believe this is because the fake service client is omitting the `/v2` path. 

This solution satisfies both forms of tests. It might also help in other OpenStack environments that, for whatever reason, are also omitting the `/v2` path.

@jrperritt Thoughts?

jrperritt:
> Thoughts?

Seems fine to me; whatever works :) I think I still don't have an Images v2 service to test this against.

jtopjian:
OK, merging.

@tatsuhiro-t Thank you again!

jtopjian:
> I think I still don't have an Images v2 service to test this against.

Terraform is pushing for acceptance CI on all providers, so I've been looking at getting some testing infrastructure up and running. Whatever works for Terraform can just as well work for Gophercloud. So that might be something cool to investigate.

jrperritt:
> OK, merging.

OK, that's fine too. I originally meant that your idea seemed fine to implement. If you like you can send a PR for it.

> Terraform is pushing for acceptance CI on all providers, 

That's interesting (and warranted).

> so I've been looking at getting some testing infrastructure up and running

Is all the Terraform CI infrastructure going to be owned by Terraform, or are you lending them hardware?

jtopjian:
> OK, that's fine too. I originally meant that your idea seemed fine to implement. If you like you can send a PR for it.

duh. Sorry about that. I might have been unclear above. The simple solution that just changes the `strings.Index` function works with the acceptance tests but does not work with the unit tests. I spent around 30m trying to get it to work, but it might be inherent with the fake service client and the way Glance is providing links.

@tatsuhiro-t's solution works with both. 

> Is all the Terraform CI infrastructure going to be owned by Terraform, or are you lending them hardware?

It'll all be virtualized. I have an all-in-one image that I've been working on. The build process takes around 90m but the resulting image is a fully functional environment. I'm working out the automated deployment and test runs now. Once it's working, I'm going to port it to AWS for HashiCorp to integrate into their testing environment, but could easily continue to host it on my own resources and link it to this repository.

Once it's set up, maintenance _should_ just be regenerating the image every 6 months with a new release (the image uses RDO - devstack is just too unstable, even with a stable branch).

I've also thought about hosting a standalone all-in-one that certain people can either run tests against manually or log directly into via SSH.

jrperritt:
Ah, I read this the other night and forgot to comment on it.

> duh. Sorry about that. I might have been unclear above.

No worries, I wasn't clear about that.

> It'll all be virtualized.

Right :) I guess I was wondering whether the cloud environments like OpenStack were going to be running on hardware that's owned by Hashicorp, or if they were just going to get a tenant in your cloud. It doesn't really matter; I was just curious if they were "investing" in OpenStack, so to speak (considering there are several other more popular platforms).

> I've also thought about hosting a standalone all-in-one that certain people can either run tests against manually or log directly into via SSH.

Yeah, I've considered doing something like that before. One of the things that stopped me was not knowing enough about the security of something like that. Another one was getting the permissions right for permitted users (i.e. I don't want to let any person create 100 VMs). Basically, I guess I wasn't very motivated to get it done.

PR #370, opened 2017-06-07
ImageService v2: Acceptance tests for listing

jtopjian:
Partial for #366


jtopjian:
I'm going to merge these tests now, but they'll fail until #366 is resolved.

Issue #371, opened 2017-06-07
Run vet on Gophercloud


jtopjian:
It might be beneficial to run `go vet` on Gophercloud to clean up any minor mistakes. It might also be beneficial to run this in Travis.

jrperritt:
closed by #372 . We can open another issue for adding `go vet` to Travis.

PR #372, opened 2017-06-07
go vet fixes

jtopjian:
For #371


jtopjian:
@jrperritt This cleans up all but a few reported issues.

The ones I can't figure out are:

```
auth_options.go:30: struct field DomainID repeats json tag "id" also at auth_options.go:24
openstack/identity/v3/tokens/requests.go:9: struct field DomainID repeats json tag "scope.project.id" also at openstack/identity/v3/tokens/requests.go:7
openstack/identity/v3/tokens/requests.go:10: struct field DomainName repeats json tag "scope.project.id" also at openstack/identity/v3/tokens/requests.go:7
openstack/identity/v3/tokens/requests.go:39: struct field DomainID repeats json tag "id" also at openstack/identity/v3/tokens/requests.go:33
openstack/networking/v2/extensions/testing/delegate_test.go:51: github.com/gophercloud/gophercloud/openstack/networking/v2/extensions.Extension composite literal uses unkeyed fields
```

For the JSON tags, I think these can be removed? It looks like the maps are all manually built.

For the `extensions.Extension` one, I can't figure it out for the life of me.

Do you think this cleanup is worthwhile?

jrperritt:
> For the JSON tags, I think these can be removed? It looks like the maps are all manually built.

Yes. That was something I was experimenting with when I was porting the old repo over, but decided not to use it.

> For the JSON tags, I think these can be removed? It looks like the maps are all manually built.

It looks like it's embedding a `common.Extension`, so something like
```go
Extension: common.Extension{
...
}
```
might appease it.

> Do you think this cleanup is worthwhile?

Oh sure. I think using field names makes the code easier for users to understand.

jtopjian:
`go vet ./...` now passes.

With regard to the JSON tags, I only removed what was necessary to get vet to pass.

jrperritt:
This looks good. Thanks for fixing it up.

jtopjian:
No problem at all. I'll look at adding a vet test to travis.

What triggered this was trying to catch the use of anonymous fields in structs. But after some investigation, it's probably not possible to flag this. Anonymous fields are too much of a core Go function to be able to flag as a style error.

jrperritt:
> Anonymous fields are too much of a core Go function to be able to flag as a style error.

Right, and sometimes you want to embed a type in a struct, for example to inherit the type's methods.

Issue #373, opened 2017-06-08
Problem with maps and deepdiffequal


jtopjian:
The `testhelper.deepDiffEqual` function is feeling especially generous to maps in that it considers any comparison valid :)

jrperritt:
closed by #374

PR #374, opened 2017-06-08
Fix testhelper.deepDiffEqual

jtopjian:
For #373

### DIFF
--- a/internal/testing/util_test.go
+++ b/internal/testing/util_test.go
@@ -33,22 +33,10 @@ func TestRemainingKeys(t *testing.T) {
 		"custom_field": "foo",
 	}
 
-	notExpected := map[string]interface{}{
-		"created_at":   "2017-06-08T02:49:03.000000",
-		"is_admin":     "true",
-		"custom_field": "foo",
-		"bar":          "baz",
-	}
-
 	actual := internal.RemainingKeys(User{}, userResponse)
 
 	isEqual := reflect.DeepEqual(expected, actual)
 	if !isEqual {
 		t.Fatalf("expected %s but got %s", expected, actual)
-	}
-
-	isEqual = reflect.DeepEqual(notExpected, actual)
-	if isEqual {
-		t.Fatalf("expected %s but got %s", notExpected, actual)
 	}
 }
--- a/internal/testing/util_test.go
+++ b/internal/testing/util_test.go
@@ -1,36 +1,54 @@
 package testing
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/gophercloud/gophercloud/internal"
-	th "github.com/gophercloud/gophercloud/testhelper"
 )
 
 func TestRemainingKeys(t *testing.T) {
 	type User struct {
-		FirstName string `json:"first_name"`
-		LastName  string `json:"last_name"`
-		City      string
+		UserID    string `json:"user_id"`
+		Username  string `json:"username"`
+		Location  string `json:"-"`
+		CreatedAt string `json:"-"`
+		Status    string
+		IsAdmin   bool
 	}
 
-	userStruct := User{
-		FirstName: "John",
-		LastName:  "Doe",
-	}
-
-	userMap := map[string]interface{}{
-		"first_name": "John",
-		"last_name":  "Doe",
-		"city":       "Honolulu",
-		"state":      "Hawaii",
+	userResponse := map[string]interface{}{
+		"user_id":      "abcd1234",
+		"username":     "jdoe",
+		"location":     "Hawaii",
+		"created_at":   "2017-06-08T02:49:03.000000",
+		"status":       "active",
+		"is_admin":     "true",
+		"custom_field": "foo",
 	}
 
 	expected := map[string]interface{}{
-		"city":  "Honolulu",
-		"state": "Hawaii",
+		"created_at":   "2017-06-08T02:49:03.000000",
+		"is_admin":     "true",
+		"custom_field": "foo",
 	}
 
-	actual := internal.RemainingKeys(userStruct, userMap)
-	th.AssertDeepEquals(t, expected, actual)
+	notExpected := map[string]interface{}{
+		"created_at":   "2017-06-08T02:49:03.000000",
+		"is_admin":     "true",
+		"custom_field": "foo",
+		"bar":          "baz",
+	}
+
+	actual := internal.RemainingKeys(User{}, userResponse)
+
+	isEqual := reflect.DeepEqual(expected, actual)
+	if !isEqual {
+		t.Fatalf("expected %s but got %s", expected, actual)
+	}
+
+	isEqual = reflect.DeepEqual(notExpected, actual)
+	if isEqual {
+		t.Fatalf("expected %s but got %s", notExpected, actual)
+	}
 }
--- a/internal/util.go
+++ b/internal/util.go
@@ -12,13 +12,17 @@ import (
 // for resources that can contain an arbitrary or dynamic number of fields.
 func RemainingKeys(s interface{}, m map[string]interface{}) (extras map[string]interface{}) {
 	extras = make(map[string]interface{})
+	for k, v := range m {
+		extras[k] = v
+	}
+
 	valueOf := reflect.ValueOf(s)
 	typeOf := reflect.TypeOf(s)
 	for i := 0; i < valueOf.NumField(); i++ {
 		field := typeOf.Field(i)
 		tagValue := field.Tag.Get("json")
-		if _, ok := m[tagValue]; !ok {
-			extras[tagValue] = m[tagValue]
+		if _, ok := extras[tagValue]; ok {
+			delete(extras, tagValue)
 		}
 	}
 
--- a/internal/util.go
+++ b/internal/util.go
@@ -34,7 +34,6 @@ func RemainingKeys(s interface{}, m map[string]interface{}) (extras map[string]i
 
 			if _, ok := extras[tagValue]; ok {
 				delete(extras, tagValue)
-				continue
 			}
 		}
 	}
--- a/internal/util.go
+++ b/internal/util.go
@@ -23,18 +23,10 @@ func RemainingKeys(s interface{}, m map[string]interface{}) (extras map[string]i
 		field := typeOf.Field(i)
 
 		lowerField := strings.ToLower(field.Name)
-		if _, ok := extras[lowerField]; ok {
-			delete(extras, lowerField)
-		}
+		delete(extras, lowerField)
 
-		if tagValue := field.Tag.Get("json"); tagValue != "" {
-			if tagValue == "-" {
-				continue
-			}
-
-			if _, ok := extras[tagValue]; ok {
-				delete(extras, tagValue)
-			}
+		if tagValue := field.Tag.Get("json"); tagValue != "" && tagValue != "-" {
+			delete(extras, tagValue)
 		}
 	}
 
--- a/internal/util.go
+++ b/internal/util.go
@@ -2,11 +2,12 @@ package internal
 
 import (
 	"reflect"
+	"strings"
 )
 
-// RemainingKeys will inspect a struct and compare it to a map. Any key that
-// is not defined in a JSON tag of the struct will be added to the extras map
-// and returned.
+// RemainingKeys will inspect a struct and compare it to a map. Any struct
+// field that does not have a JSON tag that matches a key in the map or
+// a matching lower-case field in the map will be returned as an extra.
 //
 // This is useful for determining the extra fields returned in response bodies
 // for resources that can contain an arbitrary or dynamic number of fields.
@@ -20,9 +21,21 @@ func RemainingKeys(s interface{}, m map[string]interface{}) (extras map[string]i
 	typeOf := reflect.TypeOf(s)
 	for i := 0; i < valueOf.NumField(); i++ {
 		field := typeOf.Field(i)
-		tagValue := field.Tag.Get("json")
-		if _, ok := extras[tagValue]; ok {
-			delete(extras, tagValue)
+
+		lowerField := strings.ToLower(field.Name)
+		if _, ok := extras[lowerField]; ok {
+			delete(extras, lowerField)
+		}
+
+		if tagValue := field.Tag.Get("json"); tagValue != "" {
+			if tagValue == "-" {
+				continue
+			}
+
+			if _, ok := extras[tagValue]; ok {
+				delete(extras, tagValue)
+				continue
+			}
 		}
 	}
 
--- a/openstack/cdn/v1/base/testing/requests_test.go
+++ b/openstack/cdn/v1/base/testing/requests_test.go
@@ -17,16 +17,18 @@ func TestGetHomeDocument(t *testing.T) {
 	th.CheckNoErr(t, err)
 
 	expected := base.HomeDocument{
-		"rel/cdn": map[string]interface{}{
-			"href-template": "services{?marker,limit}",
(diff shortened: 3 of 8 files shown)


jtopjian:
@jrperritt Let me know your thoughts on this.

The other two commits are bundled in this PR so travis passes.

internal/util.go
```diff
@@ -12,13 +12,17 @@ import (
...
+
 	valueOf := reflect.ValueOf(s)
 	typeOf := reflect.TypeOf(s)
 	for i := 0; i < valueOf.NumField(); i++ {
 		field := typeOf.Field(i)
 		tagValue := field.Tag.Get("json")
-		if _, ok := m[tagValue]; !ok {
-			extras[tagValue] = m[tagValue]
+		if _, ok := extras[tagValue]; ok {
+			delete(extras, tagValue)
```
jrperritt: Why does the logic in this function change?

testhelper/convenience.go
```diff
@@ -167,15 +167,13 @@ func deepDiffEqual(expected, actual reflect.Value, visited map[visit]bool, path
 
 		for _, k := range keys {
 			expectedValue := expected.MapIndex(k)
-			actualValue := expected.MapIndex(k)
+			actualValue := actual.MapIndex(k)
```
jrperritt: Nice!

openstack/compute/v2/servers/testing/fixtures.go
```diff
@@ -870,17 +870,17 @@ var ListAddressesExpected = map[string][]servers.Address{
 	"public": []servers.Address{
 		{
 			Version: 4,
-			Address: "80.56.136.39",
+			Address: "50.56.176.35",
```
jrperritt: Are all of these small changes required to make the unit tests pass (i.e. these unit tests have always been broken)?

jtopjian: Yes

internal/util.go
jtopjian: The function was actually wrong to begin with. I just thought it was correct because the unit test said so. It was only after I started implementing the Keystone v3 User support that I noticed it was wrong.

The way the previous function worked was that it was looking a json tag. However, the function is supposed to collect fields that have no json tags. `tagValue` is then `""`.

So I switched the logic around to delete known json fields and return the remaining result.

jtopjian: You can verify this by reverting to the original function (but with the `deepDiffEqual` fix in place, then purposely making the `RemainingKeys` unit test error out (like use `Hawaiiii` for expected).

You'll notice that the test still passes. This is a side effect of the `continue` lines I added to the `deepDiffEqual`. The `continue` lines are skipping nil / invalid values... Do you know a better way of handling that?

testhelper/convenience.go
```diff
@@ -167,15 +167,13 @@ func deepDiffEqual(expected, actual reflect.Value, visited map[visit]bool, path
...
 
 			if !expectedValue.IsValid() {
-				logDifference(path, nil, actual.Interface())
-				return
```
jtopjian: Sorry, stream of consciousness. 

What about:

```go
if !expectedValue.IsValid() && !actualValue.IsValid() {
  continue
}

if !expectedValue.IsValid() {
  logDifference(path, nil, actual.Interface())
  return
}
if !actualValue.IsValid() {
  logDifference(path, expected.Interface(), nil)
  return
}
```

jtopjian: No, that won't work. A good test to look at to trigger this area of code is `gophercloud/testing/params_test.go`.

internal/util.go
jrperritt: > However, the function is supposed to collect fields that have no json tags

I think it should be different than that; namely, I think we want a function that will return keys that are in the response JSON `map` but are not fields in the defined `struct` response object. `UnmarshalJSON` can work just fine without struct tags.
For example, in your unit test, `City` is a known field, so I wouldn't expect it to get returned from this function.

jtopjian: Yes, you're right.

jtopjian: But how is it possible to detect what is returned in the map but not defined in the struct without a JSON tag to do the mapping?

jrperritt: Lower-case the struct field name. `UnmarshalJSON` uses that be default.

jtopjian: I'm sorry, but I'm not following. :(

To start from scratch: From my point of view, the objective of this function is to find extra fields that were sent in a JSON response. For example:

```json
{
  "user_id": "abcd1234",
  "username": "John",
  "custom_field": "foo"
}
```

This response needs to be unmarshaled into into a struct, but it's impossible to have a definitive struct schema for an arbitrary amount of unknown fields. So we have a struct like this:

```go
type User struct {
  UserID string
  Username string
  Extra map[string]interface{}
}
```

Because Go's style of naming fields can differ from the JSON response, and to ensure the correct JSON field maps to the struct field, JSON tags are added:

```go
type User struct {
  UserID string `json:"user_id"`
  Username string `json:"username"`
  Extra map[string]interface{}
}
```

All of Gophercloud's results contain the correct JSON tags, effectively creating a confident bridge between the struct field and the JSON response.

So that's why I chose to use a method of reduction. My approach was to make a `map[string]interface{}` out of the JSON response, which effectively creates a simple map where the keys are the field in the JSON response. The known keys can be confidently referenced to a result struct, which includes mappings like `UserID` to `user_id`.

Anything that can't be mapped can be considered unknown to Gophercloud and thus be considered an "extra" field.

I'm having trouble trying to apply your suggested way to this method. If I were to just use the struct fields, that would cause problems with names such as `UserID`.

jrperritt: Example:

You have this response
```json
{
  "user_id": "abcd1234",
  "username": "John",
  "status": "current",
  "custom_field": "foo",
  "is_admin": true,
}
```
and this object to unmarshal it into:
```go
type User struct {
  UserID string `json:"user_id"`
  Username string `json:"username"`
  Status string
  IsAdmin bool
}
```

Both `custom_field` and `is_admin` should be in `Extra`.

Logic would be something like

```
for each struct field {
  if field has a struct tag {
    if struct tag is "-" {
      skip field
    } else {
      check map keys for struct tag
    }
  } else {
    check map keys for lower-cased field name
  }
}
```

testhelper/convenience.go
jtopjian: I spent several hours trying to sort this one out. What I have in the PR is not correct. Similarly, my "What about..." idea is incorrect, too.

The problem is when a map contains a nested map. The nested map will be turned into an initiated map and return `false` for `IsNil()`.

For example:

```
--- FAIL: TestBuildRequestBody (0.00s)
        convenience.go:251: Failure in params_test.go, line 202:at [auth] expected <nil>, but got map[string]interface {}{"token":map[string]interface {}{}, "or_fields":map[string]interface {}{}, "passwordCredentials":map[string]interface {}{"username":"me", "password":"s
wordfish"}}
        convenience.go:35: Failure in params_test.go, line 202: The structures were different.
```

The above is being triggered [here](https://github.com/jtopjian/gophercloud/blob/ea65cfab418dcd5b5e245ef4c28eff74641ef104/testhelper/convenience.go#L168) when trying to test against [this](https://github.com/jtopjian/gophercloud/blob/ea65cfab418dcd5b5e245ef4c28eff74641ef104/testing/params_test.go#L167-L182), specifically when the `for` loop hits [`token`](https://github.com/jtopjian/gophercloud/blob/ea65cfab418dcd5b5e245ef4c28eff74641ef104/testing/params_test.go#L158).

`expected` is set to "Invalid" but "actual" has `"token":map[string]interface {}{}`.

I tried a different route by changing the [struct](https://github.com/jtopjian/gophercloud/blob/ea65cfab418dcd5b5e245ef4c28eff74641ef104/testing/params_test.go#L146-L161) in the test to this:

```go
type AuthOptions struct {
		*PasswordCredentials `json:"passwordCredentials,omitempty" xor:"TokenCredentials"`

		// The TenantID and TenantName fields are optional for the Identity V2 API.
		// Some providers allow you to specify a TenantName instead of the TenantId.
		// Some require both. Your provider's authentication policies will determine
		// how these fields influence authentication.
		TenantID   string `json:"tenantId,omitempty"`
		TenantName string `json:"tenantName,omitempty"`

		// TokenCredentials allows users to authenticate (possibly as another user) with an
		// authentication token ID.
		*TokenCredentials `json:"token,omitempty" xor:"PasswordCredentials"`

		OrFields *orFields `json:"or_fields,omitempty"`
	}
```

Which made the test pass, but then I run into an error here:

```
$ go test -v openstack/cdn/v1/base/testing/*
=== RUN   TestGetHomeDocument
--- FAIL: TestGetHomeDocument (0.00s)
        convenience.go:256: Failure in requests_test.go, line 34: at  expected base.HomeDocument{"rel/cdn":map[string]interface {}{"href-template":"services{?marker,limit}", "href-vars":map[string]interface {}{"marker":"param/marker", "limit":"param/limit"}, "hints":map[string]interface {}{"allow":[]string{"GET"}, "formats":map[string]interface {}{"application/json":interface {}(nil)}}}}, but got <nil>
```

Would it be best to close this PR and break the two points of discussion (`RemainingKeys` and this part of the code) into separate issues? Or just focus on one for the moment.

internal/util.go
jtopjian: OK! And just to confirm: `IsAdmin` will be ignored altogether and that's correct? 

If so, then I fully understand. I thought you wanted to ignore tags altogether. :)

jrperritt: `IsAdmin` wouldn't be ignored, but it would go in `Extra` because:
1. it has no struct tag, **and**
2. there is no `isAdmin` field in the JSON response.

jtopjian: Right, what I meant by "ignored" is `User.IsAdmin` would be `false` and `User.Extra["is_admin"]` would be true.

jrperritt: I think we're on the page. I just want to make one more comment:

> User.IsAdmin would be false and User.Extra["is_admin"] would be true.

The _values_ of the fields (both in the JSON response and the struct) are irrelevant (e.g. "true", "false", "nil", 2). It's the field names themselves that are important. In my example `IsAdmin` doesn't work for `is_admin` but would work for `isAdmin`.

internal/testing/util_test.go
```diff
@@ -1,36 +1,54 @@
...
+	if !isEqual {
+		t.Fatalf("expected %s but got %s", expected, actual)
 	}
 
-	actual := internal.RemainingKeys(userStruct, userMap)
-	th.AssertDeepEquals(t, expected, actual)
+	isEqual = reflect.DeepEqual(notExpected, actual)
+	if isEqual {
+		t.Fatalf("expected %s but got %s", notExpected, actual)
+	}
```
jtopjian: This could technically be removed. It was more of a sanity check from the `th.AssertDeepEquals` bug.

internal/util.go
```diff
@@ -20,9 +21,21 @@ func RemainingKeys(s interface{}, m map[string]interface{}) (extras map[string]i
 	typeOf := reflect.TypeOf(s)
 	for i := 0; i < valueOf.NumField(); i++ {
 		field := typeOf.Field(i)
-		tagValue := field.Tag.Get("json")
-		if _, ok := extras[tagValue]; ok {
-			delete(extras, tagValue)
+
+		lowerField := strings.ToLower(field.Name)
```
jtopjian: Should only the first letter be lowercase? You were using `isAdmin` in your examples.

internal/testing/util_test.go
```diff
@@ -1,36 +1,54 @@
...
+		"custom_field": "foo",
 	}
 
-	userMap := map[string]interface{}{
-		"first_name": "John",
-		"last_name":  "Doe",
-		"city":       "Honolulu",
-		"state":      "Hawaii",
+	expected := map[string]interface{}{
+		"created_at":   "2017-06-08T02:49:03.000000",
```
jrperritt: :+1:

internal/util.go
```diff
@@ -20,9 +21,21 @@ func RemainingKeys(s interface{}, m map[string]interface{}) (extras map[string]i
...
+		if _, ok := extras[lowerField]; ok {
```
jrperritt: This `if` condition (and the one below) aren't necessary: `delete` either succeeds or is a no-op.

internal/util.go
jtopjian: I think all lower should be OK. The docs say:

> preferring an exact match but also accepting a case-insensitive match.

internal/util.go
```diff
@@ -2,23 +2,39 @@ package internal
...
-		tagValue := field.Tag.Get("json")
-		if _, ok := m[tagValue]; !ok {
-			extras[tagValue] = m[tagValue]
+
+		lowerField := strings.ToLower(field.Name)
+		if _, ok := extras[lowerField]; ok {
+			delete(extras, lowerField)
+		}
+
+		if tagValue := field.Tag.Get("json"); tagValue != "" {
```
jrperritt:
```go
if tagValue := field.Tag.Get("json"); tagValue != ""  && tagValue != "-"{
  delete(extras, tagValue)
}
```

openstack/cdn/v1/base/testing/requests_test.go
```diff
@@ -17,16 +17,18 @@ func TestGetHomeDocument(t *testing.T) {
...
-			"href-template": "services{?marker,limit}",
-			"href-vars": map[string]interface{}{
-				"marker": "param/marker",
-				"limit":  "param/limit",
-			},
-			"hints": map[string]interface{}{
-				"allow": []string{"GET"},
-				"formats": map[string]interface{}{
-					"application/json": map[string]interface{}{},
+		"resources": map[string]interface{}{
```
jtopjian: It took me _way_ too long to notice that.

jtopjian:
OK, this entire PR is ready for a proper review.

To summarize:

The version of `RemainingKeys` that I wrote was invalid from the start, however, the unit test was passing.

The cause was a typo in `testhelper/convenience.go` which was comparing the same map against itself, thus considering all maps to be considered the same and valid:

```diff
-			actualValue := expected.MapIndex(k)
+			actualValue := actual.MapIndex(k)
```

Once this was fixed, this shed light on a handful of tests that were always broken. It just took me a while to confidently determine that the tests themselves were broke and it wasn't further bugs in `testhelper/convenience.go`.

At this point, I'm quite sure `testhelper` is fine, short of the single line. The solutions I've proposed for the unit tests (notably, `testing/params_test.go`) could use a review.

openstack/cdn/v1/base/testing/requests_test.go
jrperritt: You? 2 years in, and I still haven't noticed it.

jrperritt:
What specifically in `testing/params_test.go` did you want me to look at? It looks like you just changed some fields to pointers in that file, which is fine, and the compiler will tell you if something's wrong with that.

jtopjian:
> It looks like you just changed some fields to pointers in that file, which is fine

Yep, just looking for that confirmation :)

I wasn't sure if those weren't pointers to begin with for a reason.

internal/testing/util_test.go
```diff
@@ -1,36 +1,42 @@
...
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/gophercloud/gophercloud/internal"
-	th "github.com/gophercloud/gophercloud/testhelper"
 )
 
 func TestRemainingKeys(t *testing.T) {
```
jrperritt: Good test

jrperritt:
Great find on the bug. The fix LGTM

jtopjian:
:tada: 

Thank you for your help and patience on this one :)

PR #376, opened 2017-06-13
ImageService v2: Collect all properties of images

jtopjian:
For #260 

While investigating this, I discovered that `virtual_size` was a first-class field:

* https://github.com/openstack/glance/blob/48ee8ef4793ed40397613193f09872f474c11abe/glance/db/sqlalchemy/migrate_repo/versions/034_add_virtual_size.py
* https://github.com/openstack/glance/blob/61046c8f247d52d5cd506ad465dcd0e70b2cb4b2/glance/domain/__init__.py#L51

`self` is being deleted from the JSON response since it's included in every response and part of the pagination system of Glance.


jrperritt:
LGTM

Issue #380, opened 2017-06-14
Shared file systems: Share Export Locations


pospispa:
This issue is raised to track the development of the [shares](http://developer.openstack.org/api-ref/shared-file-systems/?expanded=#shares) - export locations in the shared file systems api v2

- [X] List export locations
- [ ] Show single export location

PR #381, opened 2017-06-14
Shared Filesystems: Share Export Locations

pospispa:
For #380

code in this PR:

https://github.com/openstack/manila/blob/master/manila/api/v2/share_export_locations.py#L25

### DIFF
--- a/openstack/sharedfilesystems/v2/shares/results.go
+++ b/openstack/sharedfilesystems/v2/shares/results.go
@@ -117,21 +117,26 @@ type ExportLocation struct {
 	Path string `json:"path"`
 	// The UUID of the share instance that this export location belongs to.
 	ShareInstanceID string `json:"share_instance_id"`
-	// Defines purpose of an export location. If set to true, then it is expected to be used for service needs and by administrators only. If it is set to false, then this export location can be used by end users.
+	// Defines purpose of an export location.
+	// If set to true, then it is expected to be used for service needs
+	// and by administrators only.
+	// If it is set to false, then this export location can be used by end users.
 	IsAdminOnly bool `json:"is_admin_only"`
 	// The share export location UUID.
 	ID string `json:"id"`
-	// Drivers may use this field to identify which export locations are most efficient and should be used preferentially by clients. By default it is set to false value. New in version 2.14
+	// Drivers may use this field to identify which export locations are
+	// most efficient and should be used preferentially by clients.
+	// By default it is set to false value. New in version 2.14
 	Preferred bool `json:"preferred"`
 }
 
 // ExtractExportLocations will get the Export Locations from the commonResult
 func (r commonResult) ExtractExportLocations() ([]ExportLocation, error) {
 	var s struct {
-		GetExportLocationsRes []ExportLocation `json:"export_locations"`
+		ExportLocations []ExportLocation `json:"export_locations"`
 	}
 	err := r.ExtractInto(&s)
-	return s.GetExportLocationsRes, err
+	return s.ExportLocations, err
 }
 
 // GetExportLocationsResult contains the result.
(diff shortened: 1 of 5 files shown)


openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -110,3 +110,31 @@ type DeleteResult struct {
...
 	commonResult
 }
+
+// ExportLocation contains all information associated with a share export location
+type ExportLocation struct {
+	// The export location path that should be used for mount operation.
+	Path string `json:"path"`
+	// The UUID of the share instance that this export location belongs to.
+	ShareInstanceID string `json:"share_instance_id"`
+	// Defines purpose of an export location. If set to true, then it is expected to be used for service needs and by administrators only. If it is set to false, then this export location can be used by end users.
```
jtopjian: nit: Can you wrap this comment to around 80 characters?

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -110,3 +110,31 @@ type DeleteResult struct {
...
+	IsAdminOnly bool `json:"is_admin_only"`
+	// The share export location UUID.
+	ID string `json:"id"`
+	// Drivers may use this field to identify which export locations are most efficient and should be used preferentially by clients. By default it is set to false value. New in version 2.14
```
jtopjian: nit: Can you wrap this comment to around 80 characters?

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -110,3 +110,31 @@ type DeleteResult struct {
...
+	Preferred bool `json:"preferred"`
+}
+
+// ExtractExportLocations will get the Export Locations from the commonResult
+func (r commonResult) ExtractExportLocations() ([]ExportLocation, error) {
+	var s struct {
+		GetExportLocationsRes []ExportLocation `json:"export_locations"`
```
jtopjian: I think this should be:

```go
ExportLocations  []ExportLocation `json:"export_locations"`
```

pospispa:
@jtopjian I believe I have addressed all your comments and I created https://github.com/gophercloud/gophercloud/issues/447 in order to add a microversion check later.

jtopjian:
This looks good to me. I was able to get a share created in a test environment and the API results matched with what's here and in the referenced Manila code.

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -110,3 +110,36 @@ type DeleteResult struct {
...
+	// The share export location UUID.
+	ID string `json:"id"`
+	// Drivers may use this field to identify which export locations are
+	// most efficient and should be used preferentially by clients.
+	// By default it is set to false value. New in version 2.14
+	Preferred bool `json:"preferred"`
+}
+
+// ExtractExportLocations will get the Export Locations from the commonResult
+func (r commonResult) ExtractExportLocations() ([]ExportLocation, error) {
```
jrperritt: Why is this a `commonResult` instead of a `GetExportLocationResult`? Can every response type extract this data?

jtopjian: I caught that as well last night and I saw something else in the sharedfilesystems packages that made me think it was ok, but I most likely got it mixed up with the same occurrence in #346.

I doubt this is function is applicable to all. I don't see how a `share` result can extract this information. I can follow-up with a PR to fix.

jtopjian: Ah! This is what it was: https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/v2/sharenetworks/results.go#L172-L182

But upon reading further, I see that those actions return a share network and not some other result. So this is my mistake for not reading further.

Issue #382, opened 2017-06-14
Identity v3: List user's groups


kars7e:
It would be great to have support for retrieving groups a user belongs to. I could provide a PR for that, I'm just wondering whether it would be better to add this as a part of "users" package (something like users.ListGroups(userId) ), or better create new package "groups", that could deal with managing groups in general, and have a method like groups.ListByUser(userId).

jtopjian:
I think the best way to determine would be to look at the URL used when making that API call. Is it hitting `/users` or `/groups`?

I did a quick look at the Keystone controller code, and confusingly [`list_groups_for_user`](https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/controllers.py#L335) is defined in the Group controller and [`list_users_in_group`](https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/controllers.py#L236) is under the User controller. I would have imagined that being reversed.

jrperritt:
For this issue, `users.ListGroups`.

`groups.ListUsers` would be the complementary function.

jtopjian:
Sounds right:

https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/routers.py#L77-L84

mfranc213:
I've written groups.ListUsers.  Will try to create a PR for this soon.

kars7e:
I've added users.ListGroups in #386, please take a look. Thanks!

jrperritt:
closed by #386

PR #386, opened 2017-06-14
Add ListGroups method to Retrieve groups user belongs to

kars7e:
Adds a method to v3 bindings to retrieve groups user belongs to,
based on his/her ID.

For #382 

code in this PR:

groups resource definition in keystone request router: https://github.com/openstack/keystone/blob/a3aee6ccb52d85eac1deedec31724a955d47fa96/keystone/identity/routers.py#L77-L84

list_group_for_users general implementation https://github.com/openstack/keystone/blob/16f6ed14df136eb9c283a5415b9f18fcb0834350/keystone/identity/core.py#L1281

list_group_for_users SQL driver implementation https://github.com/openstack/keystone/blob/7754f170aa0eb42bea356e7f912bc241832eb1f3/keystone/identity/backends/sql.py#L312

Based on: https://developer.openstack.org/api-ref/identity/v3/?expanded=list-groups-to-which-a-user-belongs-detail,list-users-detail#list-groups-to-which-a-user-belongs

### DIFF
--- a/acceptance/openstack/identity/v3/users_test.go
+++ b/acceptance/openstack/identity/v3/users_test.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/groups"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/users"
 )
 
@@ -119,5 +120,37 @@ func TestUserCRUD(t *testing.T) {
 
 	tools.PrintResource(t, newUser)
 	tools.PrintResource(t, newUser.Extra)
+}
 
+func TestUsersListGroups(t *testing.T) {
+	client, err := clients.NewIdentityV3Client()
+	if err != nil {
+		t.Fatalf("Unable to obtain an identity client: %v", err)
+	}
+	allUserPages, err := users.List(client, nil).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to list users: %v", err)
+	}
+
+	allUsers, err := users.ExtractUsers(allUserPages)
+	if err != nil {
+		t.Fatalf("Unable to extract users: %v", err)
+	}
+
+	user := allUsers[0]
+
+	allGroupPages, err := users.ListGroups(client, user.ID).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to list groups: %v", err)
+	}
+
+	allGroups, err := groups.ExtractGroups(allGroupPages)
+	if err != nil {
+		t.Fatalf("Unable to extract groups: %v", err)
+	}
+
+	for _, group := range allGroups {
+		tools.PrintResource(t, group)
+		tools.PrintResource(t, group.Extra)
+	}
 }
--- a/openstack/identity/v3/groups/results.go
+++ b/openstack/identity/v3/groups/results.go
@@ -1,0 +1,108 @@
+package groups
+
+import (
+	"encoding/json"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/internal"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Group helps manage related users.
+type Group struct {
+	// Description describes the group purpose.
+	Description string `json:"description"`
+
+	// DomainID is the domain ID the group belongs to.
+	DomainID string `json:"domain_id"`
+
+	// ID is the unique ID of the group.
+	ID string `json:"id"`
+
+	// Extra is a collection of miscellaneous key/values.
+	Extra map[string]interface{} `json:"-"`
+
+	// Links contains referencing links to the group.
+	Links map[string]interface{} `json:"links"`
+
+	// Name is the name of the group.
+	Name string `json:"name"`
+}
+
+func (r *Group) UnmarshalJSON(b []byte) error {
+	type tmp Group
+	var s struct {
+		tmp
+		Extra map[string]interface{} `json:"extra"`
+	}
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*r = Group(s.tmp)
+
+	// Collect other fields and bundle them into Extra
+	// but only if a field titled "extra" wasn't sent.
+	if s.Extra != nil {
+		r.Extra = s.Extra
+	} else {
+		var result interface{}
+		err := json.Unmarshal(b, &result)
+		if err != nil {
+			return err
+		}
+		if resultMap, ok := result.(map[string]interface{}); ok {
+			r.Extra = internal.RemainingKeys(Group{}, resultMap)
+		}
+	}
+
+	return err
+}
+
+type groupResult struct {
+	gophercloud.Result
+}
+
+// GroupPage is a single page of Group results.
+type GroupPage struct {
+	pagination.LinkedPageBase
+}
+
+// IsEmpty determines whether or not a page of Groups contains any results.
+func (r GroupPage) IsEmpty() (bool, error) {
+	groups, err := ExtractGroups(r)
+	return len(groups) == 0, err
+}
+
+// NextPageURL extracts the "next" link from the links section of the result.
+func (r GroupPage) NextPageURL() (string, error) {
+	var s struct {
+		Links struct {
+			Next     string `json:"next"`
+			Previous string `json:"previous"`
+		} `json:"links"`
+	}
+	err := r.ExtractInto(&s)
+	if err != nil {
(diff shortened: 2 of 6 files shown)


jrperritt:
@kars7e 
> Links to the line numbers/files in the OpenStack source code

You've provided a link to the docs; that's not the same and not sufficient. This won't be reviewed until the source code links are provided.

kars7e:
Thanks @jrperritt and sorry for missing this requirement, I have updated the PR description.

jtopjian:
@kars7e Be careful when working with the Identity API. Rabbits love to dig there 😄 

A few comments/heads-up:

* Check out all of the resources I've collected in https://github.com/gophercloud/gophercloud/issues/287
* Also checkout the recent merge of Users Get/List: https://github.com/gophercloud/gophercloud/pull/288

It looks like the `Group` struct you made would be the exact same as what would go into a `Group` package. Can you confirm that? If so, I wonder if it's possible to define the `Group` result struct in a `groups` package and have it extracted in the `users.ListGroups` function. This would cut down on duplicated code.

If the groups returned in the `users.ListGroups` result is the same structure as what would be in a call to `groups.Get`, then note the following below:

* Note the `Extra` field in the above. It looks like Groups support extra attributes, too (see why we can't trust the API documentation? 😄 ):

```sql
MariaDB [keystone]> describe `group`;
+-------------+-------------+------+-----+---------+-------+
| Field       | Type        | Null | Key | Default | Extra |
+-------------+-------------+------+-----+---------+-------+
| id          | varchar(64) | NO   | PRI | NULL    |       |
| domain_id   | varchar(64) | NO   | MUL | NULL    |       |
| name        | varchar(64) | NO   |     | NULL    |       |
| description | text        | YES  |     | NULL    |       |
| extra       | text        | YES  |     | NULL    |       |
+-------------+-------------+------+-----+---------+-------+
5 rows in set (0.00 sec)
```

Now, you can't add `extra` data through the command-line, but you can do:

```sql
update `group` set extra = '{"hello":"world"}' where id = '7cf1576e744844c6886af09e495e62ba';
```

And in a response you get:

```json
{"links": 
  {"self": "http://127.0.0.1:35357/v3/groups?name=foo", "previous": null, "next": null}, 
  "groups": [
    {"description": "", 
    "links": {"self": "http://127.0.0.1:35357/v3/groups/7cf1576e744844c6886af09e495e62ba"}, 
    "hello": "world", 
    "id": "7cf1576e744844c6886af09e495e62ba", 
    "domain_id": 
    "default", 
    "name": "foo"
  }]
}
```

* Also be mindful of an `Options` field, too. Groups might support that one, as well.

jtopjian:
From playing with the API, it does look like what is returned in `/users/<id>/groups` is the same as `/groups/<id>`:

```shell
$ curl -s -H "X-Auth-Token: $token" http://127.0.0.1:35357/v3/users/dc2a24ccdeb046b988c52f1a6d8cf608/groups | python -mjson.tool
```

```json
{
    "groups": [
        {
            "description": "",
            "domain_id": "default",
            "hello": "world",
            "id": "7cf1576e744844c6886af09e495e62ba",
            "links": {
                "self": "http://127.0.0.1:35357/v3/groups/7cf1576e744844c6886af09e495e62ba"
            },
            "name": "foo"
        }
    ],
    "links": {
        "next": null,
        "previous": null,
        "self": "http://127.0.0.1:35357/v3/users/dc2a24ccdeb046b988c52f1a6d8cf608/groups"
    }
}
```

```shell
$ curl -s -H "X-Auth-Token: $token" http://127.0.0.1:35357/v3/groups/7cf1576e744844c6886af09e495e62ba
```

```json
{
    "group": {
        "description": "",
        "domain_id": "default",
        "hello": "world",
        "id": "7cf1576e744844c6886af09e495e62ba",
        "links": {
            "self": "http://127.0.0.1:35357/v3/groups/7cf1576e744844c6886af09e495e62ba"
        },
        "name": "foo"
    }
}
```

I was thinking only a subset of fields would be returned with the `/users/<id>/groups` call, ignoring the extra fields, but it looks like they're all there.

jrperritt:
@jtopjian All good points. One thing to think about here would be the import that you suggested: if we have `users` importing `groups` and (presumably soon) `groups` importing `users`, we'll get an import cycle, and the code won't compile.

kars7e:
Thanks @jtopjian for a thorough review. I've updated PR with support for `extra` in groups, I also added an acceptance test. 
WRT `options`, based on the content of this: https://github.com/openstack/keystone/blob/master/keystone/identity/backends/resource_options.py I understand `options` currently exists only for `users`. 
And finally, the `Group` type. I like the idea of defining this in `groups`, but it doesn't exist yet. If I was to create it, then it would make sense to add some actual methods to this package as well. Eventually, we get to the point where we get cyclic imports, as @jrperritt noted. I'm thinking how to solve that. Should we define some separate package for shared types?

jtopjian:
> I understand options currently exists only for users.

:+1: 

Thank you for looking into that.

> Should we define some separate package for shared types?

That was my thought as well. The shared package (`v3/shared`?) would contain the `User` and `Group` structs with their `UnmarshalJSON` method.

The other option is to just have duplicates in both the `users` and `groups` packages.

Unless there's a third option, my vote is for the shared package.

> Eventually, we get to the point where we get cyclic imports

This might be sooner than we think, especially if @mfranc213 follows up with the `groups.ListUsers` functionality.

And I can confirm that `groups.ListUsers` does return a full-fledged user and not a subset of user fields.

We should probably just figure out the best long-term structure now. @jrperritt ?

jtopjian:
Just a quick thought: would we be able to avoid the shared package as well as cyclic dependencies if the functions were placed in the opposite packages?

* `groups.ListGroupsOfUser(client, userid)`
* `users.ListUsersInGroup(client, groupid)`

The downside is that it's unintuitive.

jrperritt:
Alternatively, both functions could be in the `users` package:
`users.ListGroups(client, userid)`
`users.ListByGroup(client, groupid)`

jtopjian:
That's a good point. Since you suggested it, I'm in favor of it just to start being decisive. :)

But hopefully not delving into bike-shedding, `users.ListByGroup` gave me the immediate impression of "list all users, by group", implying more than just a single group membership would be returned.

What about: `users.ListInGroup`

Or more explicit: `users.ListMembersOfGroup`, though `Members` possibly adds a confusing synonym here.

jrperritt:
> implying more than just a single group membership would be returned

Understandable. I think the `groupid` parameter remedies that, but I think I'd be fine with `ListInGroup` instead.

jtopjian:
@kars7e Just a quick ping on this. It looks like this will need a rebase from other merges that have happened. Also, with regard to the discussion, let's just add everything to the `users` package.

kars7e:
Thanks @jtopjian for reminding me, I rebased the patch. Should I add the `ListInGroup` method in the same PR, or create a new one for it?

jtopjian:
@kars7e Thanks! It'd be best to open a new PR for `ListInGroup`.

jtopjian:
> let's just add everything to the users package.

This is my mistake and I apologize. I meant the two functions/methods being discussed would go in the users package, but everything else group-related (the result struct) should go in a `groups` package. This is so when someone implements CRUD functionality on groups, that information wouldn't be duplicated.

acceptance/openstack/identity/v3/users_test.go
```diff
@@ -119,5 +119,37 @@ func TestUserCRUD(t *testing.T) {
...
+	client, err := clients.NewIdentityV3Client()
+	if err != nil {
+		t.Fatalf("Unable to obtain an identity client: %v", err)
+	}
+	allUserPages, err := users.List(client, nil).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to list users: %v", err)
+	}
 
+	allUsers, err := users.ExtractUsers(allPages)
```
jtopjian: `allUserPages`

kars7e:
@jtopjian moved groups related things to groups package, take a look when you have some time

acceptance/openstack/identity/v3/users_test.go
jtopjian: Sorry, what I meant was:

```go
allUsers, err := users.ExtractUsers(allUserPages)
```

kars7e: Wow, I really need to stop reviewing code at 1AM. Sorry, fixed!

jtopjian: haha - no problem at all.

jtopjian:
This looks good to me.

Disclaimer: I did not do an in-depth review of the Groups API since we're technically only concerned about a group structure. I validated the fields, links, and result bodies from the Keystone schema and from interaction against a Keystone service. Everything looks good to me.

@jrperritt Thoughts / Comments on this PR?

Issue #388, opened 2017-06-15
Bug: token validate does not work with AllowReauth = true


asmund-hvalman:
I see problems with this line of code
https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/tokens/requests.go#L185

HTTP 401 is included in "ok codes". That means, that Request method would never enter this branch https://github.com/gophercloud/gophercloud/blob/master/provider_client.go#L188 and would never trigger reauth https://github.com/gophercloud/gophercloud/blob/master/provider_client.go#L211

jtopjian:
Considering `401` is "Unauthorized" and especially given the context of a Token, I'm inclined to say you're correct here.

jrperritt:
@asmund-hvalman thanks for catching this.

closed by #389

PR #389, opened 2017-06-16
Identity v3: 401 is not OK for tokens

jtopjian:
For #388

### DIFF
--- a/openstack/identity/v3/tokens/requests.go
+++ b/openstack/identity/v3/tokens/requests.go
@@ -182,7 +182,7 @@ func Get(c *gophercloud.ServiceClient, token string) (r GetResult) {
 func Validate(c *gophercloud.ServiceClient, token string) (bool, error) {
 	resp, err := c.Request("HEAD", tokenURL(c), &gophercloud.RequestOpts{
 		MoreHeaders: subjectTokenHeaders(c, token),
-		OkCodes:     []int{200, 204, 400, 403, 404},
+		OkCodes:     []int{200, 204, 404},
 	})
 	if err != nil {
 		return false, err
--- a/openstack/identity/v3/tokens/requests.go
+++ b/openstack/identity/v3/tokens/requests.go
@@ -182,7 +182,7 @@ func Get(c *gophercloud.ServiceClient, token string) (r GetResult) {
 func Validate(c *gophercloud.ServiceClient, token string) (bool, error) {
 	resp, err := c.Request("HEAD", tokenURL(c), &gophercloud.RequestOpts{
 		MoreHeaders: subjectTokenHeaders(c, token),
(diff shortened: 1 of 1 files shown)


jtopjian:
Side thought: If an OpenStack environment was, possibly incorrectly, configured to return a 401 on something other than a token validation (for example, to deny access to a certain feature (such as preventing the creation of Load Balancers)), would that trigger a loop [here](https://github.com/gophercloud/gophercloud/blob/master/provider_client.go#L210)?

openstack/identity/v3/tokens/requests.go
```diff
@@ -182,7 +182,7 @@ func Get(c *gophercloud.ServiceClient, token string) (r GetResult) {
 func Validate(c *gophercloud.ServiceClient, token string) (bool, error) {
 	resp, err := c.Request("HEAD", tokenURL(c), &gophercloud.RequestOpts{
 		MoreHeaders: subjectTokenHeaders(c, token),
-		OkCodes:     []int{200, 204, 400, 401, 403, 404},
+		OkCodes:     []int{200, 204, 400, 403, 404},
```
jrperritt: Yeah, I'm not sure why all these are in here. I think 200, 204, and 404 should be the only ones.

jtopjian: The PR that implemented this is #203. This is was my misunderstanding of `OkCodes`. I was under the impression that any valid response (no matter success or failure), should be listed as a valid code.

jrperritt:
> would that trigger a loop here?

Yes, that's why users should limit re-auth attempts in a custom `http.Client`. See #230.

openstack/identity/v3/tokens/requests.go
jtopjian: While we're on the topic (but I can open a separate Issue/PR), the only other place I see 400 codes listed as "OK" are here:

https://github.com/gophercloud/gophercloud/blob/c5c590a7ee87267c978586fdd1655db0dd4b6e05/openstack/imageservice/v2/members/requests.go#L21

Should those be removed, too?

To confirm, an "OK Code" is one that would only signify a successful action? Not one that is necessarily used in the API service itself?

jrperritt: > Should those be removed, too?

Yes

> To confirm, an "OK Code" is one that would only signify a successful action? Not one that is necessarily used in the API service itself?

Correct. In fact, the only reason the `Validate` function accepts 404 is for a legacy reason. It returns a `bool`, and 404 there would signify `false` (i.e. unauthorized).

jtopjian: Perfect - thanks! And the 404 was the odd one out that I was _hoping_ had a special case :)

Issue #390, opened 2017-06-16
ImageService v2: Remove 400 codes from being OK


jtopjian:
Per discussion in #388 , the 400 codes found in the `imageservice/v2/members` package should be removed.

jrperritt:
closed by #391

PR #391, opened 2017-06-16
ImageService v2: Remove 400 codes from members

jtopjian:
For #390


jrperritt:
Thanks for catching these

PR #392, opened 2017-06-16
Identity v3: Users Create

jtopjian:
For #287

### DIFF
--- a/openstack/identity/v3/users/requests.go
+++ b/openstack/identity/v3/users/requests.go
@@ -5,6 +5,18 @@ import (
 	"github.com/gophercloud/gophercloud/pagination"
 )
 
+// Option is a specific option defined at the API to enable features
+// on a user account.
+type Option string
+
+const (
+	IgnoreChangePasswordUponFirstUse Option = "ignore_change_password_upon_first_use"
+	IgnorePasswordExpiry             Option = "ignore_password_expiry"
+	IgnoreLockoutFailureAttempts     Option = "ignore_lockout_failure_attempts"
+	MultiFactorAuthRules             Option = "multi_factor_auth_rules"
+	MultiFactorAuthEnabled           Option = "multi_factor_auth_enabled"
+)
+
 // ListOptsBuilder allows extensions to add additional parameters to
 // the List request
 type ListOptsBuilder interface {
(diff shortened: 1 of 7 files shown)


jtopjian:
With regard to how "extras" are handled here, it's hard to determine from the Keystone code. The `openstack` client supports setting "properties" on projects, so I used the debug output to determine that extra fields are mixed into the POST/PUT request.

This also means that I'll modify `v3/projects` in the future to support extra fields.

jtopjian:
Found a nit: changing `Options` to `Option`.

openstack/identity/v3/users/requests.go
```diff
@@ -61,3 +73,67 @@ func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
...
+	// DomainID is the ID of the domain the user belongs to.
+	DomainID string `json:"domain_id,omitempty"`
+
+	// Enabled sets the user status to enabled or disabled.
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// Extra are free-form extra key/value pairs to describe the user.
+	Extra map[string]interface{} `json:"-"`
+
+	// Options are coded 4-character options defined by the API.
```
jrperritt: What does "coded 4-character options" mean?

openstack/identity/v3/users/requests.go
```diff
@@ -61,3 +73,67 @@ func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
...
+	Options map[Options]interface{} `json:"options,omitempty"`
```
jrperritt: Is this map's value type an `interface` only because of `MultiFactorAuthRules`?

openstack/identity/v3/users/testing/fixtures.go
```diff
@@ -75,6 +75,55 @@ const GetOutput = `
...
+{
+    "user": {
+        "default_project_id": "263fd9",
+        "domain_id": "1789d1",
+        "enabled": true,
+        "name": "jsmith",
+        "password": "secretsecret",
+        "email": "jsmith@example.com",
+        "options": {
+            "ignore_password_expiry": true
```
jrperritt: Let's have at least the non-boolean option in here as an example.

openstack/identity/v3/users/requests.go
jtopjian: You're too quick :)

Yes, I'll fix this. In the Keystone code, these are actually stored as 4-character codes. But then I found out that the readable-name is used and forgot to change this.

openstack/identity/v3/users/requests.go
jtopjian: Yes, otherwise it would have been strict bool.

jrperritt:
LGTM

jtopjian:
Thanks!

Issue #395, opened 2017-06-18
Identity v2: Enhance tenants


mozhuli:
The tenants of identity v2 only has the List method, it is necessary to support Create, Update, Delete, Get methods for some users want to do use identity v2. I will try to push a PR for this soon.

jtopjian:
@mozhuli First, thank you for submitting this. 

Please check out the [STYLEGUIDE](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md) on how to submit changes like this. In particular, there needs to be one PR per "route". So you'll need to break #396 up into separate PRs: one for Create, one for Update, etc.

You'll also want to collect links from the actual Keystone code that validate correct fields.

Check out https://github.com/gophercloud/gophercloud/issues/287 as a template/example and let me know if you have any questions. :)

PR #396, opened 2017-06-18
Add tenants of identity V2 Create Update Delete Get support

mozhuli:
For #395

### DIFF
--- a/openstack/identity/v2/tenants/requests.go
+++ b/openstack/identity/v2/tenants/requests.go
@@ -27,3 +27,77 @@ func List(client *gophercloud.ServiceClient, opts *ListOpts) pagination.Pager {
 		return TenantPage{pagination.LinkedPageBase{PageResult: r}}
 	})
 }
+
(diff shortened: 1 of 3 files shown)


openstack/identity/v2/tenants/requests.go
```diff
@@ -27,3 +27,77 @@ func List(client *gophercloud.ServiceClient, opts *ListOpts) pagination.Pager {
 		return TenantPage{pagination.LinkedPageBase{PageResult: r}}
 	})
 }
+
+// CommonOpts are the parameters that are shared between CreateOpts and UpdateOpts
+type CommonOpts struct {
+	// Name is the name of the project.
+	Name string `json:"name,required"`
```
jtopjian: Just a heads up: since it's possible to rename both a user and a project, the `Name` field can't be required, and thus you need two separate sets of options for `Create` and `Update`. Otherwise, yes, they could have been shared :)

Issue #397, opened 2017-06-19
Identity v2: Tenants CRUD


mozhuli:
- [x] create 
- [x] delete 
- [x] update 
- [x] get 

General API reference: https://developer.openstack.org/api-ref/identity/v2-admin/index.html#tenants

API processing:

* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/core.py#L237
* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/backends/sql.py

Schema:

* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/models.py#L50

Create:
* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/core.py#L293
* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/backends/sql.py#L331

Delete:
* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/core.py#L314
* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/backends/sql.py#L354

Update:
* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/core.py#L305
* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/backends/sql.py#L340

Get:
* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/core.py#L283
* https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/backends/sql.py#L161

jtopjian:
@mozhuli Thank you for your work on this!

PR #398, opened 2017-06-19
Add tenants V2 create support

mozhuli:
For #397

code in this PR:

create_tenant general implementation 
https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/core.py#L293

create_tenant sql driver implementation 
https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/backends/sql.py#L331

### DIFF
--- a/openstack/identity/v2/tenants/requests.go
+++ b/openstack/identity/v2/tenants/requests.go
@@ -27,3 +27,36 @@ func List(client *gophercloud.ServiceClient, opts *ListOpts) pagination.Pager {
 		return TenantPage{pagination.LinkedPageBase{PageResult: r}}
 	})
 }
+
+// CreateOpts represents the options needed when creating new tenant.
+type CreateOpts struct {
+	// Name is the name of the tenant.
+	Name string `json:"name,required"`
+	// Description is the description of the tenant.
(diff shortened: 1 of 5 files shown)


jtopjian:
LGTM - thank you for submitting this as well as providing all of the requested information :)

side note: I made an acceptance test to try this out. I will submit it later.

openstack/identity/v2/tenants/requests.go
```diff
@@ -27,3 +27,36 @@ func List(client *gophercloud.ServiceClient, opts *ListOpts) pagination.Pager {
 		return TenantPage{pagination.LinkedPageBase{PageResult: r}}
 	})
 }
+
+// CreateOpts represents the options needed when creating new tenant.
+type CreateOpts struct {
+	// Name is the name of the tenant.
+	Name string `json:"name,required"`
```
jrperritt: This will not work: it needs to be `json:"name" required:"true"`.

jtopjian: :man_facepalming: 

I'll patch this now.

PR #399, opened 2017-06-19
Add tenants V2 delete support

mozhuli:
For #397 

code in this PR:

delete_tenant general implementation
https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/core.py#L314

delete_tenant sql driver implementation
https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/backends/sql.py#L354

### DIFF
--- a/openstack/identity/v2/tenants/results.go
+++ b/openstack/identity/v2/tenants/results.go
@@ -69,3 +69,8 @@ func (r tenantResult) Extract() (*Tenant, error) {
 type CreateResult struct {
 	tenantResult
 }
+
+// DeleteResult temporarily contains the response from the Delete call.
+type DeleteResult struct {
+	gophercloud.ErrResult
+}
--- a/openstack/identity/v2/tenants/testing/requests_test.go
+++ b/openstack/identity/v2/tenants/testing/requests_test.go
@@ -55,3 +55,13 @@ func TestCreateTenant(t *testing.T) {
 
 	th.AssertDeepEquals(t, expected, tenant)
 }
+
+func TestDeleteTenant(t *testing.T) {
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+
+	mockDeleteTenantResponse(t)
+
+	err := tenants.Delete(client.ServiceClient(), "2466f69cd4714d89a548a68ed97ffcd4").ExtractErr()
+	th.AssertNoErr(t, err)
+}
--- a/openstack/identity/v2/tenants/requests.go
+++ b/openstack/identity/v2/tenants/requests.go
@@ -60,3 +60,9 @@ func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r Create
 	})
 	return
 }
+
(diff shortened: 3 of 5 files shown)


jtopjian:
@mozhuli Please rebase when you get a chance and ping when ready for review.

mozhuli:
@jtopjian Rebased

openstack/identity/v2/tenants/results.go
```diff
@@ -69,3 +69,8 @@ func (r tenantResult) Extract() (*Tenant, error) {
 type CreateResult struct {
 	tenantResult
 }
+
+// DeleteResult temporarily contains the response from the Delete call.
+type DeleteResult struct {
+	tenantResult
+}
```
jtopjian: I noticed this is also implemented the same way in `v2/users`, but I think this might be incorrect. Delete calls that only return an error and not a resource usually wrap `gophercloud.ErrResult`:

```go
type DeleteResult struct {
  gophercloud.ErrResult
}
```

 This way, you can do the following:

```go
err := tenants.Delete(client, id).ExtractErr()
if err != nil {
  ...
}
```

@jrperritt Do you know of a reason a "raw" result would be returned for `v2/users` Delete?

openstack/identity/v2/tenants/testing/requests_test.go
```diff
@@ -55,3 +55,13 @@ func TestCreateTenant(t *testing.T) {
...
 	th.AssertDeepEquals(t, expected, tenant)
 }
+
+func TestDeleteTenant(t *testing.T) {
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+
+	mockDeleteTenantResponse(t)
+
+	res := tenants.Delete(client.ServiceClient(), "2466f69cd4714d89a548a68ed97ffcd4")
```
jtopjian: This should also be changed to reflect the previous change:

```go
err := tenants.Delete(client.ServiceClient(), "2466f69cd4714d89a548a68ed97ffcd4").ExtractErr()
```

mozhuli:
@jtopjian updated

jtopjian:
@mozhuli Thanks!

I just want to wait for confirmation from @jrperritt on whether `ErrResult` is correct here. I suspect it is unless it's a special case.

openstack/identity/v2/tenants/results.go
jrperritt: Probably just an oversight. I took a quick look at each of those commits (`users.Delete` and `ErrResult`), and it looks like they were made 2 days apart. So the implementation of `users.Delete may have predated the creation of `ErrResult`. In addition, the documentation may have shown a user object in the result of a delete operation. It looks like it should be an `ErrResult` (I haven't tested), so we'll accept a PR for it.

jtopjian: That all makes sense. Thank you!

jtopjian:
@mozhuli Thank you again!

Issue #400, opened 2017-06-20
Fix incorrect syntax of required fields


jtopjian:
Per #398, a JSON tag had the incorrect syntax for flagging a required field.

I also found two other occurrences in Gophercloud, which also need fixed.

jrperritt:
closed by #401

PR #401, opened 2017-06-20
Fixing occurrences of incorrect required field tags

jtopjian:
For #400

### DIFF
--- a/openstack/dns/v2/zones/requests.go
+++ b/openstack/dns/v2/zones/requests.go
@@ -73,7 +73,7 @@ type CreateOpts struct {
 	Description string `json:"description,omitempty"`
 
 	// Name of the zone.
-	Name string `json:"name,required"`
+	Name string `json:"name" required:"true"`
 
 	// Masters specifies zone masters if this is a secondary zone.
 	Masters []string `json:"masters,omitempty"`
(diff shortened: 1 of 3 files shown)


jtopjian:
@jrperritt Is there a reason that `omitempty` is shared with the JSON tag while the others (`required`, `xor`) have their own?

Although it would be a rather large change, would it make sense to have `omitempty:"true"` to better be able to spot incorrect tagging?

Doing a `grep -R ,required`, I see three similar occurrences of `,required`:

```
testing/params_test.go:         R  string    `q:"r,required"`
testing/params_test.go:         R  string    `q:"r,required"`
testing/params_test.go:         Num    int    `h:"Number,required"`
```

Should these be fixed as well?

jtopjian:
Oh, nevermind. `omitempty` is a Go feature. I did not know that :)

jrperritt:
> omitempty is a Go feature

Right. It _would_ make sense for Go to have something akin to a `required` tag, and I think there is an open issue about it. But yes, as of right now, anything other than `<FIELDNAME>,omitempty` in the `json` struct tag should be changed, since it's not being recognized.

jrperritt:
Thanks for catching these others as well.

Issue #405, opened 2017-06-23
Identity v2: Add acceptance tests for Tenants


jtopjian:
With the recent Identity v2 tenant additions, it would be useful to have some acceptance tests.

jrperritt:
closed by #406

PR #406, opened 2017-06-23
Identity v2: Tenant Acceptance Tests

jtopjian:
For #405


jrperritt:
Thanks for adding these

Issue #408, opened 2017-06-24
SharedFileSystem: apiversions


jtopjian:
To better handle microversion discovery, Shared File Systems should be able to parse the available API versions.

https://github.com/openstack/manila/blob/4d45755ab7a69f9d2c12e0025a562b5b83e6efaf/manila/api/openstack/api_version_request.py

https://github.com/openstack/manila/blob/76742b9ea657e4335d9a1ffb49ec5ecc02f91b87/manila/api/versions.py

jtopjian:
Fixed by #409

PR #409, opened 2017-06-24
SharedFileSystems: apiversions package

jtopjian:
For #408 

@jrperritt There's some discussion here:

https://github.com/jtopjian/gophercloud/commit/5bf58ab1beb18b3220bd0b63f74104a56e1f273e
https://github.com/jtopjian/gophercloud/commit/7f48e23cca311be0fb1796e61bbded54cfefb7d6

This implementation could definitely use some critique when you have time.

Should proper error types be created?

The most notable oddity is that when retrieving a major version, an array is returned. I can't see why more than one result would be returned, but is that a safe assumption?

If it's best to handle multiple results, then this gets a little weird. The way I see it, a Get should be treated like a List and there should then be some convenience functions to extract the best result (or specify the type of result). But since there's no example of what multiple results would look like, it's really hard to think ahead.

/cc @pospispa

### DIFF
--- a/openstack/sharedfilesystems/apiversions/errors.go
+++ b/openstack/sharedfilesystems/apiversions/errors.go
@@ -1,0 +1,23 @@
+package apiversions
+
+import (
+	"fmt"
+)
+
+// ErrVersionNotFound is the error when the requested API version
+// could not be found.
+type ErrVersionNotFound struct{}
+
+func (e ErrVersionNotFound) Error() string {
+	return fmt.Sprintf("Unable to find requested API version")
+}
+
+// ErrMultipleVersionsFound is the error when a request for an API
+// version returns multiple results.
+type ErrMultipleVersionsFound struct {
+	Count int
+}
+
+func (e ErrMultipleVersionsFound) Error() string {
+	return fmt.Sprintf("Found %d API versions", e.Count)
+}
--- a/openstack/sharedfilesystems/apiversions/doc.go
+++ b/openstack/sharedfilesystems/apiversions/doc.go
@@ -1,0 +1,3 @@
+// Package apiversions provides information and interaction with the different
+// API versions for the Shared File System service, code-named Manila.
+package apiversions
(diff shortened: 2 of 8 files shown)


pospispa:
LGTM.

pospispa:
@jrperritt I'll appreciate if you have a look at this PR as there are two other PRs #346 and #381 that depend on this PR.

pospispa:
@jtopjian is it possible to merge this PR or do you prefer to wait for review from @jrperritt

jtopjian:
Implemented Error types to better conform to how other errors are handled.

openstack/sharedfilesystems/apiversions/errors.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"fmt"
+)
+
+// ErrVersionNotFound is the error when the requested API version
+// could not be found.
+type ErrVersionNotFound struct{}
+
+func (e ErrVersionNotFound) Error() string {
```
pospispa: There is a kind of [framework](https://github.com/gophercloud/gophercloud/blob/master/errors.go) for errors in the gophercloud library.
This framework is used for defining errors, e.g. in https://github.com/gophercloud/gophercloud/blob/master/openstack/errors.go

However, [the framework](https://github.com/gophercloud/gophercloud/blob/master/errors.go) is not used here so it feels inconsistent.

jtopjian: > There is a kind of framework for errors in the gophercloud library.

I am aware of this.

> However, the framework is not used here so it feels inconsistent.

Can you provide a more concrete example of how it can be used? I did not see an existing error type that would best fit the errors defined here and I didn't want to use `BaseError` just to say I used `BaseError`.

pospispa: I agree that there is no existing error that would fit.

My point is that `BaseError` and `func choseErrString` are not used. However, I understand that you don't see much benefit using `BaseError` and `func choseErrString`.

Issue #410, opened 2017-07-10
Servers v2: List servers in arbitrary tenant


wuisaacwu:
OpenStack API supports listing servers in any arbitrary tenant but it is currently not supported in this project. The change is adding a single parameter in the servers.ListOpts.

PR: #411

Usage wanted:
```
serverOpts := servers.ListOpts{AllTenants: true, TenantID: <tenant-id>}
pager := servers.List(client, serverOpts)
```

jtopjian:
Fixed by #411

PR #411, opened 2017-07-10
Allows a user with privilege to list servers for any particular tenant

wuisaacwu:
A very simple change that allows listing of servers for any particular tenant.

Issue #410

Here is how the server listing would work:
```
serverOpts := servers.ListOpts{AllTenants: true, TenantID: <tenant-id>}
pager := servers.List(client, serverOpts)
```

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -50,6 +50,9 @@ type ListOpts struct {
 
 	// Bool to show all tenants
 	AllTenants bool `q:"all_tenants"`
+
+	// List servers for a particular tenant. Setting "AllTenants = true" is required.
+	TenantID string `q:"tenant_id"`
 }
 
 // ToServerListQuery formats a ListOpts into a query string.


jtopjian:
@wuisaacwu Thank you for submitting this.

Per the PR template and [STYLEGUIDE](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md), can you please link to the actual Nova Python source code where this is allowed?

wuisaacwu:
@jtopjian Thanks for taking a look. I have attached the requested Nova Python source code.

Please see: https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/servers.py#L244

L244-L268 states that when all_tenants=1 (and this has to be in Admin privilege), you can specify the additional tenant_id query field to get servers for a particular tenant.

jtopjian:
@wuisaacwu Thanks for digging that up. 

For posterity, the next call that is made is here:

https://github.com/openstack/nova/blob/stable/ocata/nova/compute/api.py#L2373

I also verified via `nova --debug` and wrote a quick test with Gophercloud to verify this works.

LGTM - thanks for submitting this :)

wuisaacwu:
@jtopjian Great! Thanks for the help! :)

Issue #412, opened 2017-07-15
There is no support for Cinder microversions


j-griffith:
Cinder has adopted microversions and moved to a base version of V3.  Gophercloud should be updated to support this so we can start using shiny new things in Cinder.

urzds:
Was this implemented in #415?

jtopjian:
Yes, I think this issue is closable.

Best practices about microversions are described here: https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md. Gophercloud's Cinder support has some microversion-enabled features already added.

Please let me know if you have any questions.

PR #413, opened 2017-07-15
Add v3 and microversion support to cinder

j-griffith:
This change pulls the existing V2 implementations into V3.  Note that
the base V3 version (3.0) is a direct map back to V2.  As a result to
use the base V3 endpoint for Cinder we can don't need to change anything
other than the URL.  Also clean up some simple lint errors in
requests.go

One thing that the base V3 version does buy us here though is
Microversions (for better or worse).  So, this change also adds in the
ability to specify MicroVersion in the additional-header of the
ServiceClient.  This will be important going forward and so we can now
add things like the new standalone Attachment API.

For #412

Volume V3 just points to V2 unless a microversioned call is issued:
    https://github.com/openstack/cinder/blob/master/cinder/api/v3/volumes.py#L25
    https://github.com/openstack/cinder/blob/master/cinder/api/v3/volumes.py#L50

An example of a micro versioned call (not implemented in gophercloud):
    https://github.com/openstack/cinder/blob/master/cinder/api/v3/volumes.py#L212

Even the base V3 view builder just utilizes V2 (only adding additional
info for specific versions):
    https://github.com/openstack/cinder/blob/master/cinder/api/v3/views/volumes.py#L19
    https://github.com/openstack/cinder/blob/master/cinder/api/v3/views/snapshots.py#L19


jtopjian:
@j-griffith Thanks!

These two commits look kind of weird. The second commit seems to only be removing files and adding some commits -- nothing about adding microversion support. No doubt the commits themselves are valid, but from a historical log perspective, it might not make sense.

Since v3 is pretty much a copy of v2, I think we can do a mass commit of it. Would you be able do add some github inline comments on the changes between the two? I can also diff the files manually, but having the changes explicitly pointed out will help.

Also, with regard to your comments about the public methods: yes, all of those methods need to be public. They're not used by the user, but they're used by code in other packages.

j-griffith:
Ahhh!  My bad:
Funky commits -
    Yeah, this is part of a larger set that added attachments, I pulled that out but should've removed the dependency so it doesn't look so screwy.

Public items -
    Noted, I'll remove those comments, thanks!

I think in the interest of being "clean" on this I'll just submit a new PR for this and "rewrite" my git history a bit that way.

jtopjian:
Sounds good to me!

j-griffith:
Abort, Abort!!!

PR #415, opened 2017-07-15
Add Cinder V3 and MicroVersion Support

j-griffith:
This change pulls the existing V2 implementations into V3. Note that
the base V3 version (3.0) is a direct map back to V2. As a result to
use the base V3 endpoint for Cinder we can don't need to change anything
other than the URL. Also clean up some simple lint errors in
requests.go

One thing that the base V3 version does buy us here though is
Microversions (for better or worse). So, this change also adds in the
ability to specify MicroVersion in the additional-header of the
ServiceClient. This will be important going forward and so we can now
add things like the new standalone Attachment API.

For #412

Volume V3 just points to V2 unless a microversioned call is issued:
https://github.com/openstack/cinder/blob/master/cinder/api/v3/volumes.py#L25
https://github.com/openstack/cinder/blob/master/cinder/api/v3/volumes.py#L50

An example of a micro versioned call (not implemented in gophercloud):
https://github.com/openstack/cinder/blob/master/cinder/api/v3/volumes.py#L212

Even the base V3 view builder just utilizes V2 (only adding additional
info for specific versions):
https://github.com/openstack/cinder/blob/master/cinder/api/v3/views/volumes.py#L19
https://github.com/openstack/cinder/blob/master/cinder/api/v3/views/snapshots.py#L19

### DIFF
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -147,6 +147,25 @@ func NewBlockStorageV2Client() (*gophercloud.ServiceClient, error) {
 	})
 }
 
+// NewBlockStorageV3Client returns a *ServiceClient for making calls
+// to the OpenStack Block Storage v3 API. An error will be returned
+// if authentication or client creation was not possible.
+func NewBlockStorageV3Client() (*gophercloud.ServiceClient, error) {
+	ao, err := openstack.AuthOptionsFromEnv()
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := openstack.AuthenticatedClient(ao)
+	if err != nil {
+		return nil, err
+	}
+
+	return openstack.NewBlockStorageV3(client, gophercloud.EndpointOpts{
+		Region: os.Getenv("OS_REGION_NAME"),
+	})
+}
+
 // NewBlockStorageV2NoAuthClient returns a noauth *ServiceClient for
 // making calls to the OpenStack Block Storage v2 API. An error will be
 // returned if client creation was not possible.
@@ -159,7 +178,24 @@ func NewBlockStorageV2NoAuthClient() (*gophercloud.ServiceClient, error) {
 		return nil, err
 	}
 
-	return noauth.NewBlockStorageV2(client, noauth.EndpointOpts{
+	return noauth.NewBlockStorageNoAuth(client, noauth.EndpointOpts{
+		CinderEndpoint: os.Getenv("CINDER_ENDPOINT"),
+	})
+}
+
+// NewBlockStorageV3NoAuthClient returns a noauth *ServiceClient for
+// making calls to the OpenStack Block Storage v2 API. An error will be
+// returned if client creation was not possible.
+func NewBlockStorageV3NoAuthClient() (*gophercloud.ServiceClient, error) {
+	client, err := noauth.NewClient(gophercloud.AuthOptions{
+		Username:   os.Getenv("OS_USERNAME"),
+		TenantName: os.Getenv("OS_TENANT_NAME"),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return noauth.NewBlockStorageNoAuth(client, noauth.EndpointOpts{
 		CinderEndpoint: os.Getenv("CINDER_ENDPOINT"),
 	})
 }
--- a/openstack/blockstorage/v3/snapshots/results.go
+++ b/openstack/blockstorage/v3/snapshots/results.go
@@ -1,0 +1,121 @@
+package snapshots
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Snapshot contains all the information associated with a Cinder Snapshot.
+type Snapshot struct {
+	// Unique identifier.
+	ID string `json:"id"`
+
+	// Date created.
+	CreatedAt time.Time `json:"-"`
+
+	// Date updated.
+	UpdatedAt time.Time `json:"-"`
+
+	// Display name.
+	Name string `json:"name"`
+
+	// Display description.
+	Description string `json:"description"`
+
+	// ID of the Volume from which this Snapshot was created.
+	VolumeID string `json:"volume_id"`
+
+	// Currect status of the Snapshot.
+	Status string `json:"status"`
+
+	// Size of the Snapshot, in GB.
+	Size int `json:"size"`
+
+	// User-defined key-value pairs.
+	Metadata map[string]string `json:"metadata"`
+}
+
+// CreateResult contains the response body and error from a Create request.
+type CreateResult struct {
+	commonResult
+}
+
+// GetResult contains the response body and error from a Get request.
+type GetResult struct {
+	commonResult
+}
+
+// DeleteResult contains the response body and error from a Delete request.
+type DeleteResult struct {
+	gophercloud.ErrResult
+}
+
+// SnapshotPage is a pagination.Pager that is returned from a call to the List function.
+type SnapshotPage struct {
+	pagination.SinglePageBase
+}
+
+// UnmarshalJSON converts our JSON API response into our snapshot struct
+func (r *Snapshot) UnmarshalJSON(b []byte) error {
+	type tmp Snapshot
+	var s struct {
+		tmp
+		CreatedAt gophercloud.JSONRFC3339MilliNoZ `json:"created_at"`
+		UpdatedAt gophercloud.JSONRFC3339MilliNoZ `json:"updated_at"`
+	}
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*r = Snapshot(s.tmp)
+
+	r.CreatedAt = time.Time(s.CreatedAt)
+	r.UpdatedAt = time.Time(s.UpdatedAt)
+
+	return err
+}
+
+// IsEmpty returns true if a SnapshotPage contains no Snapshots.
+func (r SnapshotPage) IsEmpty() (bool, error) {
+	volumes, err := ExtractSnapshots(r)
+	return len(volumes) == 0, err
+}
+
+// ExtractSnapshots extracts and returns Snapshots. It is used while iterating over a snapshots.List call.
+func ExtractSnapshots(r pagination.Page) ([]Snapshot, error) {
+	var s struct {
+		Snapshots []Snapshot `json:"snapshots"`
+	}
+	err := (r.(SnapshotPage)).ExtractInto(&s)
+	return s.Snapshots, err
+}
+
+// UpdateMetadataResult contains the response body and error from an UpdateMetadata request.
+type UpdateMetadataResult struct {
+	commonResult
+}
+
+// ExtractMetadata returns the metadata from a response from snapshots.UpdateMetadata.
+func (r UpdateMetadataResult) ExtractMetadata() (map[string]interface{}, error) {
+	if r.Err != nil {
+		return nil, r.Err
+	}
+	m := r.Body.(map[string]interface{})["metadata"]
+	return m.(map[string]interface{}), nil
+}
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// Extract will get the Snapshot object out of the commonResult object.
+func (r commonResult) Extract() (*Snapshot, error) {
+	var s struct {
+		Snapshot *Snapshot `json:"snapshot"`
+	}
+	err := r.ExtractInto(&s)
+	return s.Snapshot, err
+}
--- a/openstack/blockstorage/v3/volumes/requests.go
+++ b/openstack/blockstorage/v3/volumes/requests.go
@@ -1,0 +1,182 @@
+package volumes
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// CreateOptsBuilder allows extensions to add additional parameters to the
+// Create request.
+type CreateOptsBuilder interface {
+	ToVolumeCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts contains options for creating a Volume. This object is passed to
+// the volumes.Create function. For more information about these parameters,
+// see the Volume object.
+type CreateOpts struct {
+	// The size of the volume, in GB
+	Size int `json:"size" required:"true"`
+	// The availability zone
+	AvailabilityZone string `json:"availability_zone,omitempty"`
+	// ConsistencyGroupID is the ID of a consistency group
+	ConsistencyGroupID string `json:"consistencygroup_id,omitempty"`
+	// The volume description
+	Description string `json:"description,omitempty"`
+	// One or more metadata key and value pairs to associate with the volume
+	Metadata map[string]string `json:"metadata,omitempty"`
+	// The volume name
+	Name string `json:"name,omitempty"`
+	// the ID of the existing volume snapshot
+	SnapshotID string `json:"snapshot_id,omitempty"`
+	// SourceReplica is a UUID of an existing volume to replicate with
+	SourceReplica string `json:"source_replica,omitempty"`
+	// the ID of the existing volume
+	SourceVolID string `json:"source_volid,omitempty"`
+	// The ID of the image from which you want to create the volume.
+	// Required to create a bootable volume.
+	ImageID string `json:"imageRef,omitempty"`
+	// The associated volume type
+	VolumeType string `json:"volume_type,omitempty"`
+}
+
+// ToVolumeCreateMap assembles a request body based on the contents of a
+// CreateOpts.
+func (opts CreateOpts) ToVolumeCreateMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "volume")
+}
+
+// Create will create a new Volume based on the values in CreateOpts. To extract
+// the Volume object from the response, call the Extract method on the
+// CreateResult.
+func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResult) {
+	b, err := opts.ToVolumeCreateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Post(createURL(client), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+	})
+	return
+}
+
+// Delete will delete the existing Volume with the provided ID.
+func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
+	_, r.Err = client.Delete(deleteURL(client, id), nil)
+	return
+}
+
+// Get retrieves the Volume with the provided ID. To extract the Volume object
+// from the response, call the Extract method on the GetResult.
+func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
+	_, r.Err = client.Get(getURL(client, id), &r.Body, nil)
+	return
+}
+
+// ListOptsBuilder allows extensions to add additional parameters to the List
+// request.
+type ListOptsBuilder interface {
+	ToVolumeListQuery() (string, error)
+}
+
+// ListOpts holds options for listing Volumes. It is passed to the volumes.List
+// function.
+type ListOpts struct {
+	// admin-only option. Set it to true to see all tenant volumes.
+	AllTenants bool `q:"all_tenants"`
+	// List only volumes that contain Metadata.
+	Metadata map[string]string `q:"metadata"`
+	// List only volumes that have Name as the display name.
+	Name string `q:"name"`
+	// List only volumes that have a status of Status.
+	Status string `q:"status"`
+}
+
+// ToVolumeListQuery formats a ListOpts into a query string.
+func (opts ListOpts) ToVolumeListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
+// List returns Volumes optionally limited by the conditions provided in ListOpts.
+func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pager {
+	url := listURL(client)
+	if opts != nil {
+		query, err := opts.ToVolumeListQuery()
+		if err != nil {
+			return pagination.Pager{Err: err}
+		}
+		url += query
+	}
+
+	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
+		return VolumePage{pagination.SinglePageBase(r)}
+	})
+}
+
+// UpdateOptsBuilder allows extensions to add additional parameters to the
+// Update request.
+type UpdateOptsBuilder interface {
+	ToVolumeUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts contain options for updating an existing Volume. This object is passed
+// to the volumes.Update function. For more information about the parameters, see
+// the Volume object.
+type UpdateOpts struct {
+	Name        string            `json:"name,omitempty"`
+	Description string            `json:"description,omitempty"`
+	Metadata    map[string]string `json:"metadata,omitempty"`
+}
+
+// ToVolumeUpdateMap assembles a request body based on the contents of an
+// UpdateOpts.
+func (opts UpdateOpts) ToVolumeUpdateMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "volume")
+}
+
+// Update will update the Volume with provided information. To extract the updated
+// Volume from the response, call the Extract method on the UpdateResult.
+func Update(client *gophercloud.ServiceClient, id string, opts UpdateOptsBuilder) (r UpdateResult) {
+	b, err := opts.ToVolumeUpdateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Put(updateURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	return
+}
+
+// IDFromName is a convienience function that returns a server's ID given its name.
+func IDFromName(client *gophercloud.ServiceClient, name string) (string, error) {
+	count := 0
+	id := ""
+	pages, err := List(client, nil).AllPages()
+	if err != nil {
+		return "", err
+	}
+
+	all, err := ExtractVolumes(pages)
+	if err != nil {
+		return "", err
+	}
+
+	for _, s := range all {
+		if s.Name == name {
+			count++
+			id = s.ID
+		}
+	}
+
+	switch count {
+	case 0:
+		return "", gophercloud.ErrResourceNotFound{Name: name, ResourceType: "volume"}
+	case 1:
+		return id, nil
+	default:
+		return "", gophercloud.ErrMultipleResourcesFound{Name: name, Count: count, ResourceType: "volume"}
+	}
+}
--- a/openstack/blockstorage/v3/volumes/results.go
+++ b/openstack/blockstorage/v3/volumes/results.go
@@ -1,0 +1,159 @@
+package volumes
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Attachment represents a Volume Attachment record
+type Attachment struct {
+	AttachedAt   time.Time `json:"-"`
+	AttachmentID string    `json:"attachment_id"`
+	Device       string    `json:"device"`
+	HostName     string    `json:"host_name"`
+	ID           string    `json:"id"`
+	ServerID     string    `json:"server_id"`
+	VolumeID     string    `json:"volume_id"`
+}
+
+// UnmarshalJSON is our unmarshalling helper
+func (r *Attachment) UnmarshalJSON(b []byte) error {
+	type tmp Attachment
+	var s struct {
+		tmp
+		AttachedAt gophercloud.JSONRFC3339MilliNoZ `json:"attached_at"`
+	}
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*r = Attachment(s.tmp)
+
+	r.AttachedAt = time.Time(s.AttachedAt)
+
+	return err
+}
+
+// Volume contains all the information associated with an OpenStack Volume.
+type Volume struct {
+	// Unique identifier for the volume.
+	ID string `json:"id"`
+	// Current status of the volume.
+	Status string `json:"status"`
+	// Size of the volume in GB.
+	Size int `json:"size"`
+	// AvailabilityZone is which availability zone the volume is in.
+	AvailabilityZone string `json:"availability_zone"`
+	// The date when this volume was created.
+	CreatedAt time.Time `json:"-"`
+	// The date when this volume was last updated
+	UpdatedAt time.Time `json:"-"`
+	// Instances onto which the volume is attached.
+	Attachments []Attachment `json:"attachments"`
+	// Human-readable display name for the volume.
+	Name string `json:"name"`
+	// Human-readable description for the volume.
+	Description string `json:"description"`
+	// The type of volume to create, either SATA or SSD.
+	VolumeType string `json:"volume_type"`
+	// The ID of the snapshot from which the volume was created
+	SnapshotID string `json:"snapshot_id"`
+	// The ID of another block storage volume from which the current volume was created
+	SourceVolID string `json:"source_volid"`
+	// Arbitrary key-value pairs defined by the user.
+	Metadata map[string]string `json:"metadata"`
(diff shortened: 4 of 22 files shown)


j-griffith:
Hopefully that makes it much more clear.  I'll add some comments regarding any differences, I took the liberty of adding some missing comments on public structs and removed occurences of naked-returns just because they bug me, and that's about it.

openstack/blockstorage/v3/snapshots/results.go
```diff
@@ -0,0 +1,121 @@
...
+type DeleteResult struct {
+	gophercloud.ErrResult
+}
+
+// SnapshotPage is a pagination.Pager that is returned from a call to the List function.
+type SnapshotPage struct {
+	pagination.SinglePageBase
+}
+
+// UnmarshalJSON converts our JSON API response into our snapshot struct
```
j-griffith: Added missing comment on public struct here

openstack/blockstorage/v3/volumes/requests.go
```diff
@@ -0,0 +1,182 @@
...
+	})
+	return r
+}
+
+// Delete will delete the existing Volume with the provided ID.
+func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
+	_, r.Err = client.Delete(deleteURL(client, id), nil)
+	return r
+}
+
```
j-griffith: In Create and Delete above I made the returns explicit "return r" instead of the naked returns that were in the old version.

openstack/blockstorage/v3/volumes/requests.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		r.Err = err
+		return r
+	}
+	_, r.Err = client.Put(updateURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	return r
+}
+
```
j-griffith: Same modification to Update, remove the naked returns and explicitly return 'r'

openstack/blockstorage/v3/volumes/results.go
```diff
@@ -0,0 +1,159 @@
...
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Attachment represents a Volume Attachment record
```
j-griffith: Added the public struct comment

openstack/blockstorage/v3/volumes/results.go
```diff
@@ -0,0 +1,159 @@
...
+type Attachment struct {
```
j-griffith: Another public struct comment added here

openstack/blockstorage/v3/volumes/results.go
```diff
@@ -0,0 +1,159 @@
...
+	AttachedAt   time.Time `json:"-"`
+	AttachmentID string    `json:"attachment_id"`
+	Device       string    `json:"device"`
+	HostName     string    `json:"host_name"`
+	ID           string    `json:"id"`
+	ServerID     string    `json:"server_id"`
+	VolumeID     string    `json:"volume_id"`
+}
+
+// UnmarshalJSON is our unmarshalling helper
```
j-griffith: Public function so added the comment above

openstack/blockstorage/v3/volumes/results.go
```diff
@@ -0,0 +1,159 @@
...
+	pagination.SinglePageBase
+}
+
+// IsEmpty returns true if a ListResult contains no Volumes.
+func (r VolumePage) IsEmpty() (bool, error) {
+	volumes, err := ExtractVolumes(r)
+	return len(volumes) == 0, err
+}
+
+// ExtractVolumes extracts and returns Volumes. It is used while iterating over a volumes.List call.
```
j-griffith: Public function, needed above comment

openstack/blockstorage/v3/volumes/results.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// Extract will get the Volume object out of the commonResult object.
+func (r commonResult) Extract() (*Volume, error) {
+	var s Volume
+	err := r.ExtractInto(&s)
+	return &s, err
+}
+
+// ExtractInto converts our response data into a volume struct
```
j-griffith: Public function, needed above comment

openstack/blockstorage/v3/volumes/results.go
```diff
@@ -0,0 +1,159 @@
...
+func (r commonResult) ExtractInto(v interface{}) error {
+	return r.Result.ExtractIntoStructPtr(v, "volume")
+}
+
+// ExtractVolumesInto similar to ExtractInto but operates on a `list` of volumes
```
j-griffith: Public function, needed above comment

openstack/blockstorage/v3/volumes/requests.go
jtopjian: I'm not sure how familiar you are with Go (and I'm no expert, either), so I apologize if you already know this: the "naked" return is known as a "named return". It comes off as a little odd at first. I've only recently started to appreciate it and am going through a phase of "name all the returns".

If you wanted to make this an explicit return, I think it'd be best to also change the function signature, too. From:

```go
func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResult) {
```

to

```go
func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) CreateResult {
```

However, I think it'd be best to leave the returns as they are, just because it would break consistency from how the other files are set up:

```shell
$ cd ~/go/src/github.com/gophercloud/gophercloud/openstack
$ grep -R return$ * | wc
    441     888   21740
```

jtopjian:
@j-griffith I'm so sorry! I thought there would have been more changes to the actual API interaction between v2 and v3 than... zero :)

Is it safe to say that interaction between v2 and v3 will remain the same? Or will v3, at some point, deviate from v2? The reason I ask is because if v2 and v3 would remain the same, we could probably just rename the `v2` package to something more generic and just create a v3 client. That would have the benefit of not duplicating all of the v2 code as v3.

But if v3 will, at some point, take off from v2, then it'd be best to make this copy.

openstack/blockstorage/v3/volumes/requests.go
j-griffith: @jtopjian I'm pretty familiar with Golang, but I think it's always good to make suggestions and sync up terminology etc.  One of my greatest pet peeves has become "naked" return statements in functions, initially I thought they were "awesome", but after a while I find them to be annoying, vague and often times confusing.  It's not very "idomatic" as they say, but if you would prefer it in the interest of keeping things consistent that's fine, it is your project after all.  It would be great if you'd reconsider though, just because it's "consistent" isn't necessarily a great argument IMO.

Your point about removing the name altogether is a good one, and I agree.  Let me know if you're willing to reconsider at least for the BlockStorage modules going further, if not like I said that's completly fair.  I can change it all back.

j-griffith:
@jtopjian So V3 "will" diverge but only when you start tacking on microversions.  The base 3.0 version is in fact truly a 1:1 mapping to V2.  That was actually intentional as part of the microversion adoption in Cinder.  I'm actually not convinced of the best way to handle microversions from an SDK perspective, but that's beside the point for this I think.  

One thing I did think about when working on this was actually starting the V3 modules off as just interfaces that call directly into V2 if there is no microversion specified?  That might actually make more sense here.  

You're suggestion of "making V2" something more generic makes sense too, the only thing is we've now got this nice pattern of using the directory structure to build the URL; that's certainly easy to change.

Maybe we could sync up on IRC and chat?  I'd like to get your input (and anybody else that has opinions).  Maybe I'll put together a gist of some different options and post it to see what you and others think?

openstack/blockstorage/v3/volumes/requests.go
jtopjian: I think the consistency is the most important thing here. There are parts of Gophercloud that not everyone agrees with, but adhering to a strict set of rules has helped keep it simple and relatively easy to implement additions. So I'm going to have to be the bad guy and say "keep the named return".

jtopjian:
> Maybe I'll put together a gist of some different options and post it to see what you and others think?

That's probably the best option - I usually can't find the time to jump on IRC throughout the day. But yeah, I'm definitely interested in any ideas you have.

On one hand, like you say the directory structure is a nice pattern. And I think that should be the default option. 

But since v2 support isn't fully implemented, I'm curious if there's a way that a change applicable to both v2 and v3 can be done in one spot. However, if making that happen will make overall management more complicated (meaning it would take more time to maintain the combined packages than just copying and pasting a change applicable to both v2 and v3), let's just defer to separate packages.

jtopjian:
Also, will `volumeactions` be affected by microversion feature additions at some point? IIRC, `volumeactions` was moved to its own non-versioned package because v1 and v2 could both interact with those actions.

j-griffith:
Sounds good, I'll see if I can find some time in the next day or two to come up with some different ideas.  The fact that there's only 2 modules (other than the contrib [volumeactions]) does make it a good time to consider this).  I hated the copy/paste in V2 of Cinder so I totally agree it's annoying to repeat it here.

The contrib (extension) items like volumeactions for now are pretty static and don't fall under either versioning or microversioning.  That being said however one of the goals we're working on right now is deprecating and removing the current volume-actions attach related methods.  They'll certainly be around for at least a couple more releases but they really need to go away.

That's actually why I added this change was to have a foundation to pull in the new attachment API's.  They've landed in Cinder last release, and we're currently working on updating Nova here:
    https://review.openstack.org/#/c/330285/

The idea is that when that lands we can deprecate and move forward and get things like multi-attach support finally (both Nova and BareMetal).

jtopjian:
Cool - thanks!

Also, if discussing / designing a long-term structure for v2/v3/contrib starts preventing you from actually using Gophercloud and v3, I'm more than happy to merge a straight copy first and then work on the long-term design.

service_client.go
```diff
@@ -114,6 +114,8 @@ func (client *ServiceClient) setMicroversionHeader(opts *RequestOpts) {
 		opts.MoreHeaders["X-OpenStack-Nova-API-Version"] = client.Microversion
 	case "sharev2":
 		opts.MoreHeaders["X-OpenStack-Manila-API-Version"] = client.Microversion
+	case "volume":
+		opts.MoreHeaders["X-OpenStack-Volume-API-Version"] = client.Microversion
```
pospispa: The `client.Microversion` is not set anywhere currently so it's an empty string. Shouldn't CinderV3 API reject packets that contain an empty string in the `X-OpenStack-Volume-API-Version` field.

acceptance/clients/clients.go
```diff
@@ -136,6 +136,25 @@ func NewBlockStorageV2Client() (*gophercloud.ServiceClient, error) {
...
+		return nil, err
+	}
+
+	client, err := openstack.AuthenticatedClient(ao)
+	if err != nil {
+		return nil, err
+	}
+
+	return openstack.NewBlockStorageV3(client, gophercloud.EndpointOpts{
+		Region: os.Getenv("OS_REGION_NAME"),
```
jtopjian: Having an empty region name is completely valid.

service_client.go
jtopjian: If `client.Microversion` is an empty string, this piece of code will never be reached. See here: https://github.com/j-griffith/gophercloud/blob/2af0cfeba7c8f0e894f945fb89ab8b0b0f7096a6/service_client.go#L61-L63

pospispa: @jtopjian yes, you're right, this piece of code will never be reached if the string is empty. I've overlooked the condition. I'm sorry for that.

However, how does CinderV3 behave in case the `X-OpenStack-Volume-API-Version` is not present? Are such packets rejected?

jtopjian: No problem.

My understanding is that it will act like the other services (Nova, Manila) that implement microversions. A user won't be able to use a feature that requires a specific microversion. So the request won't be outright rejected, but certain features won't be usable.

pospispa: Yes, you're are right. Everything that is available in version 3.0 is available. Anything that requires 3.1+ version is not available.

pospispa:
LGTM

pospispa:
@rootfs I think you may be interested.

jrperritt:
If reasonable, I'd like to see something in-between what's here now: I like the idea of creating a `v3` directory since inevitably there will be divergence, but complete copy-paste seems unnecessary. Alternatively, how much of the code in `v3` could we delegate to `v2` for now and amend if the time comes (as in, for example, wrapping function calls and casting the result)?

rootfs:
@j-griffith will Attachment API be here too?

j-griffith:
@rootfs Yes, and that was in fact that motivation for this PR.  

I know the wholesale copy/paste is lame; but sadly that's what we did in Cinder, Nova etc as well to get to that initial compatible microversion.  I've considered trying something clever to get some re-use but honestly I'm not sure that's such a good idea.  One thing to keep in mind, V3 *should* be it; there should never be another wholesale copy again, only additions for new microversions within the V3 endpoint functions.

rootfs:
If in the future v3 and v2 do diverges, delegation is probably less desirable.

@jrperritt is there any release plan to pick up this pr? thanks.

jtopjian:
@j-griffith Understood. I'm totally fine doing a copy/paste for v3. 

IIRC, I think the only issue with this PR was the named returned stuff. After that, it should be good to go.

jrperritt:
@jtopjian see my comment above for someone who is not yet on board with copy/paste

j-griffith:
@jrperritt Yeah, I'm not saying your idea isn't better.  The problem is that I haven't been able to come up with a clean way of doing that without doing some sort of trickery which IMHO is never good.  That being said, if you have an idea of a clean way to point back until updates that is easy to follow and doesn't involve "magic" I'd be thrilled to learn and implement it.

I know this has sat a while because I was going to go off and try to come up with a happy medium, I didn't have much luck though without making a mess, the best I came up with was just interfaces in the V3 dir that called V2 and that seemed lame... but maybe that's ok??

The other option (and probably the right option) was to move all of the requests and not have the V2, V3 namespace be the controlling factor here.  That's certainly doable, and makes a lot of sense I think; but I didn't like it because I was concerned about introducing compatibility issues for folks already using Gophercloud after performing upgrades.  Maybe there's a way around that... I can look at it some more but if you have something in mind let me know.

jtopjian:
> see my comment above for someone who is not yet on board with copy/paste

I apologize. I didn't interpret what you said as being against copy/paste. I was under the impression that if a better solution wasn't put forward, a copy/paste solution would be fine.

jtopjian:
> The other option (and probably the right option) was to move all of the requests and not have the V2, V3 namespace be the controlling factor here. 

I would imagine that if there was some way of combining v2 and v3 (maybe even v1) requests, they would be organized under a versionless package. In that case, is there harm in doing a copy/paste of v2 to v3 with the agreement that any further solution would go into a new package as to not break compatibility?

I realize this idea is not as clean and methodical as Gophercloud decisions usually are. However, it would allow v3 development to proceed in the meantime.

jrperritt:
> I was under the impression that if a better solution wasn't put forward, a copy/paste solution would be fine.

More or less, that's correct. But until today I had not heard any experience reports about trying anything other than copy/paste.

I'd like to think about this some more. If I (or anyone else) can't think of anything better by Monday, we can go forward with copy/paste.

rootfs:
@jrperritt great, looking forward to more update!

jtopjian:
One idea I tried was to create the equivalent `v2/volumes` functions in a `v3/volumes` package. These functions simply returned the `v2` version:

```go
func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r v2.CreateResult) {
  return v2.Create(client, opts)
}
```

I did the same with the result types:

```go
type Attachment struct {
  v2.Attachment         
}                       
                        
type Volume struct {    
  v2.Volume             
}                       
```

The compiler was accepting of this, but it didn't work out in practice. The embedded structs were difficult to work with:

```go
vol1 := volumes.Volume{}
attachment1 := volumes.Attachment{}
vol1.Attachments = append(vol1.Attachments, attachment1)
```

yields:

```
cannot use attachment1 (type "github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes".Attachment) as type "github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes".Attachment in append
```

Even if this worked, I see two issues with this method:

1. This might cause problems with the recent "extensions" support which was recently added. It makes some assumptions about the existence of an embedded/anonymous struct only occurring in extensions.

2. The moment a v3 function or result struct deviates from v2, there will be a mixture of this method and normal functions/structs.

A second idea was to proxy requests somehow:

https://github.com/tmrts/go-patterns/blob/master/structural/proxy.md

This is beyond my skill to cleanly implement, though.

FengyunPan:
@j-griffith Cool, sounds good, except this PR work fine. LGTM

j-griffith:
@jtopjian Yeah, everything I came up with ended up being horrible user experience; the proxy idea is ultimately what I'd like and would be somewhat in line with how cinder works.  I'm happy to take a look at it again next week or of course if you're enjoying the challenge just help out in any way I can.

jtopjian:
@j-griffith It's all yours! I've got plenty of other items to work on, so this would end up getting perpetually postponed.

j-griffith:
@jtopjian Ok, so I spent a little time on this; and frankly didn't like anything that I came up with all that much except for one thing.  I played around with moving the resources into a version independent directory (ie blockstorage/volumes,snapshots) and used a single common requuests.go for both V2 and V3.  This works just fine, so now I do something like:

import (
    "github.com/gophercloud/gophercloud/openstack/blockstorage/volumes"
    "github.com/gophercloud/gophercloud/openstack"
)
v2Client, _ := openstackNewBlockStorageV2(provider, eopts)
v3Client, _ := openstackNewBlockStorageV3(provider, eopts)
v2VolPager := volumes.List(v2Client, nil)
v3VolPager := volumes.List(v3Client, nil)

Microversions are handled as an optional addition to the request struct (so you'd have to use the opts struct to get/use microversions other than current max).

While I personally liked this and thought it was clean and a bit more friendly going forward; there is an obvious problem with compatability for existing deployments, and the other more general problem around microversions themeselves is that all of this gets ridiculously painful to deal and work with when you have the joy of filtering through 44 micro-versions in a single Cinder release.

Anyway, I'd be curious what you think; I did mess with proxies and interfaces a little but honestly it created more code and complexity than I think it was worth and I didn't get it fully working (although I think it's possible if that's what we have to do).  It was close, but it was cumbersome rewriting interfaces for everything; given all the more that's implemented in Cinder here right now it doesn't seem worthwhile.

jtopjian:
@j-griffith That's really cool that you got a working solution. Would you be able to make your work available to view?

> While I personally liked this and thought it was clean and a bit more friendly going forward

That's good news.

>  there is an obvious problem with compatability for existing deployments,

This deserves discussion, but let's table it for now.

> and the other more general problem around microversions themeselves is that all of this gets ridiculously painful to deal and work with when you have the joy of filtering through 44 micro-versions in a single Cinder release.

This is a concern.

First, we've been having an in-depth discussion of Gophercloud and Microversions over at #441. We're not entirely sure how, or to what extent, Gophercloud will implement microversion support. One issue is that we simply don't have enough knowledge of what a client really needs in order to adequately work with microversions. So I'm definitely keen to see the work you've done.

But that's kind of tangential here. For the sake of discussion, let's categorize "microversions" as "the big difference between Cinder v2 and v3".

If it's possible to create a sane, version independent implementation of `blockstorage` in Gophercloud that won't bite anyone down the road, that's probably the best way to go. If it's a matter of doing a bunch of upfront work to create new functions, structs, methods, etc, I am happy to help with that.

But will a version independent `blockstorage` only provide a short-term benefit? In twelve months, will we be wishing we did a clean separation of v2 and v3?

Those are kind of rhetorical questions. I think, at this point, it's a matter of reviewing ideas that have been tried and making a call. Notably, @jrperritt will need to weigh in on the attempted methods plus any that he might have looked into himself.

j-griffith:
@jtopjian Here's what I'm thinking is a good solution assuming we're ok with the compatability issue around the pkg imports, ideally folks are using pkg management so this should be "ok":
    https://github.com/j-griffith/gophercloud/commit/6523e4bfaa611ef5b045fd313450dff4574b9f23
So this would be the base commit; from there we can decide when/where (or if eve) we want to add a micro-version specific option to a call and then implement that.  My suggestion would be that we don't go down the path of exposing micro-versions directly to the consumer of the SDK, but instead if there are features that people want/need we can wrap them up with specific requests and check if the endpoint supports the version or not.  Trying to chase all the micro-versions is a bit of a nightmare, this past release alone introduced 44 versions to Cinder; most of which aren't overly significant.

Anyway, would love to know what you and @jrperritt think and if you'd be good with moving forward in this direction instead.  I think it fits better with the architecture personally.

jtopjian:
@j-griffith The commit you listed doesn't have much to it - just a few file renames and a new client. Was there supposed to be anything more?

j-griffith:
@jtopjian No, actually other than adding the micro-version header capability that's about it.  I've updated that in the commit.  So part of the point is that since V2 and V3.0 are exactly the same this consolidates everything and "new" micro-versioned things can then just be added on from there.

The result of moving this around like this is that now you just setup the client version you want (V2 or V3) and just access the volumes or snapshots pkg functions and Version is no longer relevant.

https://github.com/j-griffith/gophercloud/commit/9dbdb7417244f14b07673482a4e5aafe1d460fd8

It seems to me that it's that, or stick with this PR and wholesale copy/paste which is annoying but works without changing any existing behavior.

jtopjian:
ahh - I understand! I apologize.

My immediate impression is that this makes a lot of sense.

Do you have any ideas in mind about how v3/microversion additions would be added to this new combined package? We're currently hashing out some ideas in #441, but like I mentioned before, we're all really new to microversions and are trying to make sense of them with limited experience.

j-griffith:
No worries, granted it looks like sort of a dumb commit on the surface.

So YES, I have strong opinions about micro-versions and their usage in this sort of context, I'll add my opinions in the referenced PR.  Thanks!

j-griffith:
@jtopjian @jrperritt @pospispa so what do you guys think?  Are we moving any closer to a sort of consensus or do you think we need to spend some more time on the microversion topic and hammer some more details and possibly prototypes out?  I'd like to flush out the approach here:  j-griffith/gophercloud@9dbdb74 a bit more but if we don't think that approach is a good way to go I'd like to focus the limited time I have on something more likely to work in the long run.

jtopjian:
@j-griffith Would you be able to do a quick proof-of-concept on how a v3-specific feature (agnostic of a microversion, if possible for simplicity) would look with your current approach?

j-griffith:
@jtopjian 
> @j-griffith Would you be able to do a quick proof-of-concept on how a v3-specific feature 
> (agnostic of a microversion, if possible for simplicity) would look with your current approach?
For sure!  If there's interest in this approach I'm more than happy to work out a POC for it.  I'll start putting some things together.

jtopjian:
@j-griffith Cool. I think your current approach is the best non-copy/paste approach we've seen. Depending on how cleanly v2 and v3 can live together, it's probably the way to go. Additionally, since v2 isn't feature complete, submitting a patch for v2 would also help complete v3 (and vice versa where applicable).

pospispa:
> I'd like to flush out the approach here: [j-griffith/gophercloud@9dbdb74](https://github.com/j-griffith/gophercloud/commit/9dbdb74)

@j-griffith I agree with the approach.

IMHO, it'd be beneficial if you show how an implementation of a Cinder V3 specific feature will be approached.

j-griffith:
@jrperritt @jtopjian @pospispa Here's an example of what I have in mind for making Cinder use non versioned resource objects and also implementing Microversion support.  The trivial example is the addition of the "force" option to volume.delete (version 3.23).

This approach just introduces a new function to the package and deals with setting and modifying the request as needed without the consumer necessarily having any insight into the details of what's happening.  The only indication for the end-user is the doc heading in the function itself.  We will want to add some error reporting etc for version availability problems but this should give the gist of things:
https://github.com/j-griffith/gophercloud/tree/non-versioned-cinder-objects

Example usage:
```
func main() {
   
    opts := gophercloud.AuthOptions{
        IdentityEndpoint: "http://10.117.36.41/identity",
        Username:         "admin",
        Password:         "nomoresecret",
        TenantID:         "d9a93fdf01fc46fb8f25dcc652e3a145",
        DomainName:       "default",
    }
    provider, err := openstack.AuthenticatedClient(opts)
    if err != nil {
        fmt.Printf("Bummer dude: %+v", err)
    }
    eopts := gophercloud.EndpointOpts{
        Region: "RegionOne",
    }

    client, _ := openstack.NewBlockStorageV3(provider, eopts)
    volumes.ForceDelete(client, "87138bdb-849e-4720-b64b-a430b30c4e4b")
}
```

jtopjian:
@j-griffith Thanks for putting that together.

I took your initial commit and modified the acceptance tests to handle the v2/v3 merge. The result can be seen here: https://github.com/gophercloud/gophercloud/compare/master...jtopjian:cinderv3?expand=1

The acceptance tests pass with both a v2 and v3 client, so that validates the approach.

With regard to the `ForceDelete` example:

>  We will want to add some error reporting etc

Yes, agreed. I think, at a minimum, any v3-only function should first check if the passed Client is a v2 client and, if so, return an error.

The only concern I have about this approach is if there is a v3 revision that would significantly alter a v2-original API call and turn it into a total mess in order to handle v2+. I really wanted to avoid mixing microversions into the initial v3 support, but that might be unavoidable. At the same time, making the yet-to-be-decided microversion support not mess up, say, `volumes.Create`, might be a good mandatory rule.

Protecting v2 from v3 microversion alterations is, IMO, the only benefit of doing a copy/paste. But if you figure v3.0 is susceptible to the same microversion complications, then protecting v2 is no different than protecting v3.0.

Another thing to keep in mind is if a future v4 would be so different that it would require its own versioned package (meaning: go back to a `blockstorage/v4` directory). I'm willing to ignore that for now as it might be an issue that never happens.

With all of that in mind, I am in favor of this approach, but I'm hesitant to give it an official +1/+2 until microversions have been sorted out just in case the solution would negate this combined v2/v3 approach. Unfortunately for me that means I need to jump on getting microversions sorted out.

j-griffith:
@jtopjian 
> Yes, agreed. I think, at a minimum, any v3-only function should first check if the passed Client is a > v2 client and, if so, return an error.

Yeah, that sounds great; seems reasonable to also do that if it's a micro version as well.  I think you're "non-versioned" version PR will be great for that.

> The only concern I have about this approach is if there is a v3 revision that would significantly 
> alter a v2-original API call and turn it into a total mess in order to handle v2+. I really wanted to 
> avoid mixing micro versions into the initial v3 support, 

Actually, I'd agree with you on this;  there's a mix of things showing up in micro versions for cinder thus far; if you exclude the things like filtering and metadata related mv's though the bulk of the mv's point to newly introduced calls so I think this might be "ok".  I agree with you on keeping things clean so to speak.  Honestly looking at most of the mv's that modify existing commands I don't really believe that they're even worth adding at this point.  I chose the force option as a trivial example and thought it might be useful but I wouldn't propose going crazy adding every option under the sun (if any).

> but that might be unavoidable. At the same time, making the yet-to-be-decided micro version 
> support not mess up, say, volumes.Create, might be a good mandatory rule.

Absolutely!  I think this needs to be a requirement.  I did play around a bit with some of the more complex calls and versioning; if we need to we can add options fairly easily to request struct using omit empty and it keeps things from breaking.

> Protecting v2 from v3 micro version alterations is, IMO, the only benefit of doing a copy/paste. But > if you figure v3.0 is susceptible to the same micro version complications, then protecting v2 is no > different than protecting v3.0.

This is exactly the decision I arrived at.  Initially, I was going to push harder on the first patch that just does copy/paste, but it doesn't eliminate the problem.  The only thing it does is keeps us from breaking existing users (ie import .../v2/volumes --> import .../volumes).  I could still go either way on this one FWIW.

As far as needing to sort things out, I totally get it.  One other option to consider is short and long term approaches.  In other words short term we copy/paste and roll with a new v3 pkg (original patch) and plan on consolidating later.  The problem is that it doesn't really do us any good if we don't have a good strategy for micro versions though because in that case, just use V2.

I would like to introduce things like the new attach API's, but those are going to depend on a micro version strategy.

Anyway, don't hesitate to ping me if I can help out with some of the things you're working through or  possibly brain storm some more things together.

pospispa:
> This approach just introduces a new function to the package and deals with setting and modifying the request as needed without the consumer necessarily having any insight into the details of what's happening. The only indication for the end-user is the doc heading in the function itself. We will want to add some error reporting etc for version availability problems but this should give the gist of things: https://github.com/j-griffith/gophercloud/tree/non-versioned-cinder-objects

@j-griffith I agree with the approach.

I assume that in
```
func ForceDelete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
    client.Microversion = "3.23"
```
the `client.Microversion = "3.23"` will be replaced with a check whether the `client.Microversion` is greater or equal to `3.23`.

j-griffith:
Been away for a bit, looks like this approach is dead no?  Was there ever an agreement on the MV topic?  FWI I still think the version independent resource structure is worth while.

jtopjian:
Hi all,

Apologies for the delay with this. The microversion discussion is still ongoing but we wanted to at least get Cinder v3 moving. 

@j-griffith Go ahead and do a straight copy of v2 to v3. IIRC the only request I had from the previous copy was to leave the named returns in place. That can all be housed under one PR. Either separate commits per resource or a single commit is fine with me.

Next, you can add whatever new v3 features you need on top of that (I've seen "multi-attach" mentioned a bunch). Our only request is to keep the additions minimal and stay within the same release/microversion as they will most likely be modified when a microversion solution has been reached.

Each feature should follow the normal house rules: Please create an issue (similar to [this](https://github.com/gophercloud/gophercloud/issues/583)) and open one PR per method. I'll gladly make it a priority to review them.

Our hope with this is to unblock Cinder v3 usage while the microversion discussion is still ongoing.

Please let me know if you have questions or concerns. Again, we do apologize for the delay.

j-griffith:
@jtopjian Thanks for the update, and no worries on the delay; there's a lot to be figured out here.  I'll pick back up on the copy/update approach and hit the items you referenced.  Thanks!!

openstack/blockstorage/v3/volumes/requests.go
```diff
@@ -0,0 +1,182 @@
...
+// Create will create a new Volume based on the values in CreateOpts. To extract
+// the Volume object from the response, call the Extract method on the
+// CreateResult.
+func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResult) {
+	b, err := opts.ToVolumeCreateMap()
+	if err != nil {
+		r.Err = err
+		return r
```
jtopjian: `s/return r/return/`

j-griffith: got it, thanks!

jrperritt:
@jtopjian if you're good with this, feel free to merge

jtopjian:
Apologies for the delay -- I was waiting for Travis to pass. The error it is reporting is:

```
# github.com/gophercloud/gophercloud/acceptance/clients
acceptance/clients/clients.go:198: undefined: noauth.NewBlockStorageV3
```

I've confirmed only `noauth.NewBlockStorageV2` exists. Since this is for the `noauth` client, and since an Endpoint must be specified, it might make sense to just change this to `noauth.NewBlockStorageClient`? That can probably addressed separately. For now, adding `noauth.NewBlockStorageV3` is probably the quickest way to move forward.

Once that's fixed up, the next error is:

```
# github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes/testing
openstack/blockstorage/v3/volumes/testing/requests_test.go:105:3: undefined: volumetenants.VolumeExt
```

`volumetenants.VolumeExt` was renamed to `volumetenants.VolumeTenantExt` a while back. 

Once those two changes are made, all tests should pass. Further, doing:

```shell
$ diff -r openstack/blockstorage/v2 openstack/blockstorage/v3
```

Looks clean.

j-griffith:
Thanks all, my bad on this; I haven't been looking for the last couple weeks.  I'll go ahead and rework things with a general NewBlockStorageClient in noauth and fix up the other issue as well.

Sorry I've been MIA for a bit.

jtopjian:
@j-griffith The non-versioned approach may very well be the best long-term solution. The problem is that it implies a microversion solution, for which there isn't one yet.

I wanted to proceed with a straight copy of v2-to-v3 because we know it works and it'll get v3 support into applications now. It buys us some time to get a solution for microversions. I realize that this means there might be a change in the way Cinder v3 (maybe even v2) is used within Gophercloud. However, that could be in 1 month or 6...

The caveat here is to not add too many features to the v3 copy that would introduce breakage within the v3 API (meaning: let's try to keep this as v3.0 compatible as possible).

j-griffith:
@jtopjian Absolutely!!!  Thank you, looking forward to working together more in the near future.

pospispa:
@j-griffith I was away on PTOs. I'm glad to see Cinder V3 merged.

Issue #416, opened 2017-07-16
Add method to get image_id from "Upload volume to image" action result


vitkhab:
It's good to have and image id after starting a process of creating an image from volume using volume actions.
It might be useful for monitoring image creation process

jtopjian:
Fixed by #417

PR #417, opened 2017-07-16
Added method to get info from Volume actions volume_upload_image method response

vitkhab:
I've added response body to UploadImageResults and added method Extract to extract the response into an object.

For #416 

https://github.com/openstack/cinder/blob/master/cinder/api/contrib/volume_actions.py#L304
https://github.com/openstack/cinder/blob/master/cinder/api/contrib/volume_actions.py#L290-L293
https://github.com/openstack/cinder/blob/master/cinder/volume/api.py#L1272-L1281

### DIFF
--- a/openstack/blockstorage/extensions/volumeactions/results.go
+++ b/openstack/blockstorage/extensions/volumeactions/results.go
@@ -1,6 +1,10 @@
 package volumeactions
 
-import "github.com/gophercloud/gophercloud"
+import (
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+)
 
 // AttachResult contains the response body and error from a Get request.
 type AttachResult struct {
@@ -50,10 +54,32 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
 	return s.ConnectionInfo, err
 }
 
+// VolumeImage contains information about volume upload to an image service.
+type VolumeImage struct {
+	// The ID of a volume an image is created from.
+	VolumeId string `json:"id"`
+	// Container format, may be bare, ofv, ova, etc.
+	ContainerFormat string `json:"container_format"`
+	// Disk format, may be raw, qcow2, vhd, vdi, vmdk, etc.
+	DiskFormat string `json:"disk_format"`
+	// Human-readable description for the volume.
+	Description string `json:"display_description"`
+	// The ID of an image being created.
+	ImageId string `json:"image_id"`
+	// Human-readable display name for the image.
+	ImageName string `json:"image_name"`
+	// Size of the volume in GB.
+	Size int `json:"size"`
+	// Current status of the volume.
+	Status string `json:"status"`
+	// The date when this volume was last updated.
+	UpdatedAt time.Time `json:"-"`
+}
+
 // Extract will get an object with info about image being uploaded out of the UploadImageResult object.
-func (r UploadImageResult) Extract() (map[string]interface{}, error) {
+func (r UploadImageResult) Extract() (VolumeImage, error) {
 	var s struct {
-		VolumeUploadImage map[string]interface{} `json:"os-volume_upload_image"`
+		VolumeUploadImage VolumeImage `json:"os-volume_upload_image"`
 	}
 	err := r.ExtractInto(&s)
 	return s.VolumeUploadImage, err
--- a/openstack/blockstorage/extensions/volumeactions/results.go
+++ b/openstack/blockstorage/extensions/volumeactions/results.go
@@ -53,7 +53,7 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
 // Extract will get an object with info about image being uploaded out of the UploadImageResult object.
 func (r UploadImageResult) Extract() (map[string]interface{}, error) {
 	var s struct {
-		VolumeUploadImage   map[string]interface{} `json:"os-volume_upload_image"`
+		VolumeUploadImage map[string]interface{} `json:"os-volume_upload_image"`
 	}
 	err := r.ExtractInto(&s)
 	return s.VolumeUploadImage, err
--- a/openstack/blockstorage/extensions/volumeactions/results.go
+++ b/openstack/blockstorage/extensions/volumeactions/results.go
@@ -69,7 +69,7 @@ type ImageVolumeType struct {
 	ExtraSpecs map[string]interface{} `json:"extra_specs"`
 	// ID of quality of service specs.
 	QosSpecsID string `json:"qos_specs_id"`
-	// Flag for deletion status of volume tpe.
+	// Flag for deletion status of volume type.
 	Deleted bool `json:"deleted"`
 	// The date when volume type was deleted.
 	DeletedAt time.Time `json:"-"`
--- a/openstack/blockstorage/extensions/volumeactions/results.go
+++ b/openstack/blockstorage/extensions/volumeactions/results.go
@@ -55,10 +55,10 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
 	return s.ConnectionInfo, err
 }
 
-// VolumeImage contains information about volume upload to an image service.
-type VolumeImageVolumeType struct {
+// ImageVolumeType contains volume type object obtained from UploadImage action.
+type ImageVolumeType struct {
 	// The ID of a volume type.
-	Id string `json:"id"`
+	ID string `json:"id"`
 	// Human-readable display name for the volume type.
 	Name string `json:"name"`
 	// Human-readable description for the volume type.
@@ -67,8 +67,8 @@ type VolumeImageVolumeType struct {
 	IsPublic bool `json:"is_public"`
 	// Extra specifications for volume type.
 	ExtraSpecs map[string]interface{} `json:"extra_specs"`
-	// Id of quality of service specs.
-	QosSpecsId string `json:"qos_specs_id"`
+	// ID of quality of service specs.
+	QosSpecsID string `json:"qos_specs_id"`
 	// Flag for deletion status of volume tpe.
 	Deleted bool `json:"deleted"`
 	// The date when volume type was deleted.
@@ -79,8 +79,8 @@ type VolumeImageVolumeType struct {
 	UpdatedAt time.Time `json:"-"`
 }
 
-func (r *VolumeImageVolumeType) UnmarshalJSON(b []byte) error {
-	type tmp VolumeImageVolumeType
+func (r *ImageVolumeType) UnmarshalJSON(b []byte) error {
+	type tmp ImageVolumeType
 	var s struct {
 		tmp
 		CreatedAt gophercloud.JSONRFC3339MilliNoZ `json:"created_at"`
@@ -91,7 +91,7 @@ func (r *VolumeImageVolumeType) UnmarshalJSON(b []byte) error {
 	if err != nil {
 		return err
 	}
-	*r = VolumeImageVolumeType(s.tmp)
+	*r = ImageVolumeType(s.tmp)
 
 	r.CreatedAt = time.Time(s.CreatedAt)
 	r.UpdatedAt = time.Time(s.UpdatedAt)
@@ -103,7 +103,7 @@ func (r *VolumeImageVolumeType) UnmarshalJSON(b []byte) error {
 // VolumeImage contains information about volume upload to an image service.
 type VolumeImage struct {
 	// The ID of a volume an image is created from.
-	VolumeId string `json:"id"`
+	VolumeID string `json:"id"`
 	// Container format, may be bare, ofv, ova, etc.
 	ContainerFormat string `json:"container_format"`
 	// Disk format, may be raw, qcow2, vhd, vdi, vmdk, etc.
@@ -111,7 +111,7 @@ type VolumeImage struct {
 	// Human-readable description for the volume.
 	Description string `json:"display_description"`
 	// The ID of an image being created.
-	ImageId string `json:"image_id"`
+	ImageID string `json:"image_id"`
 	// Human-readable display name for the image.
 	ImageName string `json:"image_name"`
 	// Size of the volume in GB.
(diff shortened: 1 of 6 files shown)


jtopjian:
@vitkhab Thanks for submitting this and for including all of the proper links!

Since this returns a standard set of data, I think this should also include a defined type:

```go
type VolumeImage struct {
  ID String `json:"id"`
  CreatedAt time.Time `json:"-"`
  Status String `json:"status"`
  ...
}
```

The reason why `ConnectionInfo` just returns a generic `map[string]interface{}` is because the results are too varied from driver to driver to be able to make a standard structure out of the results. But since a defined structure is being returned here, we should leverage that.

vitkhab:
@jtopjian Thank you for informative response.
I added VolumeImage structure into volumeactions. Also added example response for UploadImageResponse into tests (I hope I did it right).

I have a question about volume_type. I have considerations:
- As I understand (and I might be wrong) it may have different structure depending on the api version.
- Also VolumeType struct is defined in Volume package for API v1. In API v2 VolumeType used as a string. My cloud provider has complex structure for it (you can see it in the reponse in MockUploadImageResponse method)
- It should be possible to get info about volume type another way if you know the volume id.

Will it be better to implement VolumeType inside VolumeImage struct? Do you have any advice how to do it?

jtopjian:
@vitkhab From what I can see, `Volume.VolumeType` as a `String` under the `v2/volumes` package is correct. Here's the result I see when doing a general `cinder --debug show foo`:

```json
{
    "volume": {
        "attachments": [],
        "availability_zone": "nova",
        "bootable": "false",
        "consistencygroup_id": null,
        "created_at": "2017-07-18T17:13:45.000000",
        "description": null,
        "encrypted": false,
        "id": "de795ef9-105a-4646-abf4-d5c66153a3ab",
        "links": [
            {
                "href": "http://127.0.0.1:8776/v2/a99e9b4e620e4db09a2dfb6e42a01e66/volumes/de795ef9-105a-4646-abf4-d5c66153a3ab",
                "rel": "self"
            },
            {
                "href": "http://127.0.0.1:8776/a99e9b4e620e4db09a2dfb6e42a01e66/volumes/de795ef9-105a-4646-abf4-d5c66153a3ab",
                "rel": "bookmark"
            }
        ],
        "metadata": {},
        "migration_status": null,
        "multiattach": false,
        "name": "foo",
        "os-vol-host-attr:host": "packstack-standard-ocata.novalocal@nfs#nfs",
        "os-vol-mig-status-attr:migstat": null,
        "os-vol-mig-status-attr:name_id": null,
        "os-vol-tenant-attr:tenant_id": "a99e9b4e620e4db09a2dfb6e42a01e66",
        "replication_status": null,
        "size": 1,
        "snapshot_id": null,
        "source_volid": null,
        "status": "available",
        "updated_at": "2017-07-18T17:13:47.000000",
        "user_id": "0fe36e73809d46aeae6705c39077b1b3",
        "volume_type": "nfs"
    }
}
```

You can see that `volume_type` above is just a simple string.

However, `volume_type` being returned as an upload response is a full-fledged object:

```json
{
    "os-volume_upload_image": {
        "container_format": "bare",
        "disk_format": "raw",
        "display_description": null,
        "id": "de795ef9-105a-4646-abf4-d5c66153a3ab",
        "image_id": "f82eb2c4-de94-4270-9608-197f7f3107d6",
        "image_name": "foo",
        "is_public": false,
        "protected": false,
        "size": 1,
        "status": "uploading",
        "updated_at": "2017-07-18T17:13:47.000000",
        "volume_type": {
            "created_at": "2017-05-17T23:01:06.000000",
            "deleted": false,
            "deleted_at": null,
            "description": null,
            "extra_specs": {
                "volume_backend_name": "nfs"
            },
            "id": "f9fa3496-a0f5-4bf6-94f0-b9ec17df36f8",
            "is_public": true,
            "name": "nfs",
            "qos_specs_id": null,
            "updated_at": null
        }
    }
}
```

Now, you're right when mentioning that there are other ways of getting a "type". There's a whole API endpoint for types:

```shell
$ cinder type-show x
```

```json
{
    "volume_type": {
        "description": null,
        "extra_specs": {
            "volume_backend_name": "nfs"
        },
        "id": "f9fa3496-a0f5-4bf6-94f0-b9ec17df36f8",
        "is_public": true,
        "name": "nfs",
        "os-volume-type-access:is_public": true,
        "qos_specs_id": null
    }
}
```

_But_, that result is different than what was returned with the Upload action (_whyyy?_) so, IMO, it would be best to define a `VolumeImageVolumeType` struct... At least the good news is you don't have to create a whole new `types` package as a prerequisite to this. 

I hope that helps!

Also, before you implement `VolumeImageVolumeType`, you'll have to dig into the Cinder code and find where that's being defined so there's reference to what fields are being used.

openstack/blockstorage/extensions/volumeactions/results.go
```diff
@@ -50,6 +54,37 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
...
+	// The ID of an image being created.
+	ImageId string `json:"image_id"`
+	// Human-readable display name for the image.
+	ImageName string `json:"image_name"`
+	// Size of the volume in GB.
+	Size int `json:"size"`
+	// Current status of the volume.
+	Status string `json:"status"`
+	// The date when this volume was last updated.
+	UpdatedAt time.Time `json:"-"`
```
jtopjian: @vitkhab Also, in your structs, you will need to implement a custom unmarshaller for parsing the time using `UnmarshalJSON`. Please see how it's implemented in the `volumes` package as a reference.

vitkhab:
VolumeType class fields are defined here:
https://github.com/openstack/cinder/blob/master/cinder/objects/volume_type.py#L39-L49

I assume as VolumeType is extension to CinderPersistentObject class it also contains these fields:
https://github.com/openstack/cinder/blob/master/cinder/objects/base.py#L249-L255

I didn't find the exact method how fields variable is extended but it seems response contains fields both from CinderPersistentObject and VolumeType

Cinder and Openstack clients access API through another methods and output is being trimmed
https://github.com/openstack/cinder/blob/master/cinder/api/v2/types.py#L35-L57
https://github.com/openstack/cinder/blob/master/cinder/api/v2/views/types.py#L25-L37

vitkhab:
I haven't implement qos_specs object (it seems excessive inside volumeactions package) and projects list (simply, didn't find how to test it) in `VolumeImageVolumeType` struct.

@jtopjian Is there anything else be done in scope of this pull request?

jtopjian:
> I haven't implement qos_specs object (it seems excessive inside volumeactions package) and projects list (simply, didn't find how to test it) in VolumeImageVolumeType struct.

I agree. These do not look like they are being returned in the responses, so no need to acknowledge them.

Thank you for putting all of this together - no doubt this turned into more work than you expected! But it looks like everything required is now in place.

I'll do a first pass at a code review.

openstack/blockstorage/extensions/volumeactions/results.go
```diff
@@ -50,6 +55,101 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
 	return s.ConnectionInfo, err
 }
 
+// VolumeImage contains information about volume upload to an image service.
+type VolumeImageVolumeType struct {
+	// The ID of a volume type.
+	Id string `json:"id"`
```
jtopjian: This should be `ID` instead of `Id`.

openstack/blockstorage/extensions/volumeactions/results.go
```diff
@@ -50,6 +55,101 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
...
+	// Human-readable display name for the volume type.
+	Name string `json:"name"`
+	// Human-readable description for the volume type.
+	Description string `json:"display_description"`
+	// Flag for public access.
+	IsPublic bool `json:"is_public"`
+	// Extra specifications for volume type.
+	ExtraSpecs map[string]interface{} `json:"extra_specs"`
+	// Id of quality of service specs.
+	QosSpecsId string `json:"qos_specs_id"`
```
jtopjian: `QosSpecsID`

openstack/blockstorage/extensions/volumeactions/results.go
```diff
@@ -50,6 +55,101 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
...
+	r.UpdatedAt = time.Time(s.UpdatedAt)
+	r.DeletedAt = time.Time(s.DeletedAt)
+
+	return err
+}
+
+// VolumeImage contains information about volume upload to an image service.
+type VolumeImage struct {
+	// The ID of a volume an image is created from.
+	VolumeId string `json:"id"`
```
jtopjian: `VolumeID`

openstack/blockstorage/extensions/volumeactions/results.go
```diff
@@ -50,6 +55,101 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
...
+	// Container format, may be bare, ofv, ova, etc.
+	ContainerFormat string `json:"container_format"`
+	// Disk format, may be raw, qcow2, vhd, vdi, vmdk, etc.
+	DiskFormat string `json:"disk_format"`
+	// Human-readable description for the volume.
+	Description string `json:"display_description"`
+	// The ID of an image being created.
+	ImageId string `json:"image_id"`
```
jtopjian: `ImageID`

openstack/blockstorage/extensions/volumeactions/results.go
```diff
@@ -50,6 +55,101 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
 	return s.ConnectionInfo, err
 }
 
+// VolumeImage contains information about volume upload to an image service.
```
jtopjian: `VolumeImageVolumeType`

(side note: I'm not a fan of this name, but I can't think of a better one. Suggestions are welcomed, or else we'll just use what's here)

openstack/blockstorage/extensions/volumeactions/results.go
```diff
@@ -50,6 +55,101 @@ func (r commonResult) Extract() (map[string]interface{}, error) {
...
+
+	r.UpdatedAt = time.Time(s.UpdatedAt)
+
+	return err
+}
+
+// Extract will get an object with info about image being uploaded out of the UploadImageResult object.
+func (r UploadImageResult) Extract() (VolumeImage, error) {
+	var s struct {
+		VolumeUploadImage VolumeImage `json:"os-volume_upload_image"`
```
jtopjian:
```go
VolumeImage VolumeImage `json:"os-volume_upload_image"
```

openstack/blockstorage/extensions/volumeactions/testing/requests_test.go
```diff
@@ -55,7 +55,7 @@ func TestUploadImage(t *testing.T) {
 		Force:           true,
 	}
 
-	err := volumeactions.UploadImage(client.ServiceClient(), "cd281d77-8217-4830-be95-9528227c105c", options).ExtractErr()
+	_, err := volumeactions.UploadImage(client.ServiceClient(), "cd281d77-8217-4830-be95-9528227c105c", options).Extract()
```
jtopjian: Can you add a more detailed unit test with a mock `VolumeImage` struct?

openstack/blockstorage/extensions/volumeactions/results.go
jrperritt: This is what I wanted to address. What does this type represent (descriptively if possible)?

jtopjian: It represents the information stored in the `volume_type` object returned by the `volume_upload_image` result (An alternative name that is `VolumeUploadImage` rather than `VolumeImage`)

So it's `VolumeType` by nature, _but_ it's different than what would be stored under a `volumes/vx/types` package. So my suggestion of name was to make sure it's differentiated from that.

 I also suggested it because I didn't want someone to become confused by seeing `VolumeType` and wondering why it was only accessible through a `VolumeImage` result.

Third, I haven't scanned the rest of the volume actions to see if there isn't another action which would return yet another `VolumeType`, possibly using the same fields as `volumes/vx/types`. If I was confident that was not a possibility, I'd be fine with calling this `volumeactions.VolumeType` and handling questions about why two `VolumeType`s exist.

So, in the end, it's just trying to make sanity out of inconsistencies. :/

Thoughts / suggestions?

jrperritt: Because it's in the `volumeactions` package, it seems like at least one of the `Volume`s in the type are redundant. Are `VolumeImageType` and `ImageVolumeType` both misleading/inaccurate?

jtopjian: Yeah, that's a good point.

So the result is information about a volume which has just turned into an image. I like `ImageVolumeType`. It keeps `VolumeType`, which I think is important, and the `Image` part easily conveys that it's restricted to just the uploaded image.

Thank you :)

vitkhab:
@jtopjian I've made all requested changes.

vitkhab:
@jtopjian gophercloud.Result doesn't have method ExtractErr() which gophercloud.ErrResult has. My biggest concern here is backward compatibility. Is it okay to change?

I propose using commonResult instead of gophercloud.Result. As it's done in other packages, example
https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v2/volumes/results.go#L136-L154

jtopjian:
Yes, it's OK to change - this is an acceptable breaking change.

The problem with `commonResult` is that it's incorrectly tied to `ConnectionInfo` at the moment:

https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/extensions/volumeactions/results.go#L44-L51

I was going to fix this up once this PR was merged.

jtopjian:
@vitkhab Looks good to me. Thank you so much for your patience on this.

@jrperritt Any thoughts?

jtopjian:
bah - Travis isn't happy:

```acceptance/openstack/blockstorage/extensions/extensions.go:30: volumeactions.UploadImage(client, volume.ID, uploadImageOpts).ExtractErr undefined (type volumeactions.UploadImageResult has no field or method ExtractErr)```

jtopjian:
@vitkhab I actually have a branch with the required acceptance changes done that I was going to merge after this. Let me submit a PR to you. Should save you some time.

vitkhab:
@jtopjian thank you for PR.

jtopjian:
No problem! _Now_ everything looks good.

@jrperritt Any other thoughts/comments?

jtopjian:
okie dokie - merging.

@vitkhab Again, thank you for your work on this :)

Issue #418, opened 2017-07-18
Floating IPs cannot be filtered by status


BenoitKnecht:
`floatingips.ListOpts` does not contain a `Status` field, which means floating IPs cannot be filtered by status when being listed.

jtopjian:
Fixed by #419

PR #419, opened 2017-07-18
floatingips: Add Status field to ListOpts

BenoitKnecht:
It lets `floatingips.List()` filter by `Status` (which can be `ACTIVE`,
`DOWN` or `ERROR`).

For #418

code in this PR:

https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/l3.py#L134-L135


jtopjian:
This looks good to me and I've verified it works.

It would be nice to pinpoint where the filtering is being done, but I can't find it in the neutron code. My guess is because either the request is being turned into a port filter request or a general subnet pool list (because technically a floating ip is just a regular IP on a subnet marked as public). 

But in any case, this just follows the same pattern / feature of other neutron resources with the ability to list/filter by status.

@BenoitKnecht Thanks for submitting this

Issue #420, opened 2017-07-18
Support Pause/Unpause for OpenStack compute instance


zionwu:
The OpenStack server package(openstack/compute/v2/servers) does not support pause/unpause operation, which would be nice to have

jtopjian:
Fixed by #422

PR #422, opened 2017-07-18
support pause/unpause for compute instance

zionwu:

For #420

### DIFF
--- a/openstack/compute/v2/extensions/pauseunpause/doc.go
+++ b/openstack/compute/v2/extensions/pauseunpause/doc.go
@@ -1,0 +1,5 @@
+/*
+Package pauseunpause provides functionality to pause and unpause servers that have
+been provisioned by the OpenStack Compute service.
+*/
+package pauseunpause
--- a/openstack/compute/v2/extensions/pauseunpause/requests.go
+++ b/openstack/compute/v2/extensions/pauseunpause/requests.go
@@ -1,0 +1,19 @@
+package pauseunpause
+
+import "github.com/gophercloud/gophercloud"
+
+func actionURL(client *gophercloud.ServiceClient, id string) string {
+	return client.ServiceURL("servers", id, "action")
+}
+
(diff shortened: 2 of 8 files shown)


jtopjian:
@zionwu Thanks for submitting this.

Per the PR template and [STYLEGUIDE](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md), you need to also include links to the OpenStack Python source code where this functionality is implemented.

zionwu:
@jtopjian 

The following is the OpenStack source code:
pause: https://github.com/openstack/nova/blob/stable/newton/nova/compute/api.py#L3105
unpause: https://github.com/openstack/nova/blob/stable/newton/nova/compute/api.py#L3115

jtopjian:
@zionwu Thanks! I was double checking and found this:

https://github.com/openstack/nova/blob/stable/ocata/nova/api/openstack/compute/pause_server.py

which defines `os-pause-server`, which is an extension.

This can also be seen by doing:

```shell
$ nova list-extensions
```

So I think this functionality is better suited in the `extensions` package. The good news is that you can use the `startstop` package as a template:

https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions/startstop

zionwu:
@jtopjian Thank you for your help and suggestion. I have made the change accordingly.

jtopjian:
This looks good to me. Thank you for contributing this!

jrperritt:
Er, why is there an entire package just for pause/unpause?

jtopjian:
The same reason there's a package for stop/stop :)

It's a dedicated extension.

jrperritt:
Hmm, yeah, looks like the precedent's been set

jtopjian:
> Hmm, yeah, looks like the precedent's been set

Yeah. That was the deciding factor. I actually don't mind it, even though certain extensions are quite small. Being able to decisively put whatever the output of `nova list-extensions` is under `extensions` has some benefit.

Issue #106, opened 2016-09-28
`ChooseVersion` doesn't match recent Keystone builds


Quentin-M:
Hi,

[ChooseVersion](https://github.com/gophercloud/gophercloud/blob/0bc5578dc193f747667cdeee035460f4a168e015/openstack/utils/choose_version.go#L23), as it is used in [Authenticate](https://github.com/gophercloud/gophercloud/blob/0bc5578dc193f747667cdeee035460f4a168e015/openstack/client.go#L65) tries to explicitly match the `ID` field of the Keystone response to `v2.0` and `v3.0`.

However, my Keystone API returns the following:

```
{
    "versions": {
        "values": [
            {
                "status": "stable",
                "updated": "2016-10-06T00:00:00Z",
                "media-types": [
                    {
                        "base": "application/json",
                        "type": "application/vnd.openstack.identity-v3+json"
                    }
                ],
                "id": "v3.7",
                "links": [
                    {
                        "href": "http://identity.openstack.cluster:30080/v3/",
                        "rel": "self"
                    }
                ]
            },
            {
                "status": "deprecated",
                "updated": "2016-08-04T00:00:00Z",
                "media-types": [
                    {
                        "base": "application/json",
                        "type": "application/vnd.openstack.identity-v2.0+json"
                    }
                ],
                "id": "v2.0",
                "links": [
                    {
                        "href": "http://identity.openstack.cluster:30080/v2.0/",
                        "rel": "self"
                    },
                    {
                        "href": "http://docs.openstack.org/",
                        "type": "text/html",
                        "rel": "describedby"
                    }
                ]
            }
        ]
    }
}
```

The v3 API can't be matched due to the ID being `v3.7` and the v2 API can't either as it is deprecated.

jrperritt:
`Authenticate` is more of a convenience function (though it probably doesn't seem convenient to you at the moment). It may now make sense to parse the version `id`s and look for something reasonable. I suspect I'll probably not get around to doing that (but I'd be willing to review a PR for it). However, if you know you want to use v3, the suggested way is to create a new client and directly authenticate to v3:
```go
pc, err := gophercloud.NewClient("authEndpoint")
err := gophercloud.AuthenticateV3(pc, ao, eo)
```

jtopjian:
This was resolved in #508. However, if you're still running into problems, let me know.

PR #423, opened 2017-07-18
Fix use with deployment having a path in the Keystone URL

MatMaul:
For #424

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/url"
 	"reflect"
+	"strings"
 
 	"github.com/gophercloud/gophercloud"
 	tokens2 "github.com/gophercloud/gophercloud/openstack/identity/v2/tokens"
@@ -25,19 +26,22 @@ func NewClient(endpoint string) (*gophercloud.ProviderClient, error) {
 	if err != nil {
 		return nil, err
 	}
-	hadPath := u.Path != ""
-	u.Path, u.RawQuery, u.Fragment = "", "", ""
-	base := u.String()
+	u.RawQuery, u.Fragment = "", ""
 
-	endpoint = gophercloud.NormalizeURL(endpoint)
-	base = gophercloud.NormalizeURL(base)
-
-	if hadPath {
-		return &gophercloud.ProviderClient{
-			IdentityBase:     base,
-			IdentityEndpoint: endpoint,
-		}, nil
+	segments := strings.Split(u.Path, "/")
+	// if the url ends by / we have an empty string as last segment
+	if segments[len(segments)-1] == "" {
+		segments = segments[:len(segments)-1]
 	}
+
+	lastSegmentIndex := len(segments) - 1
+	lastSegment := segments[lastSegmentIndex]
+	if lastSegment == "v2.0" || lastSegment == "v3" {
+		segments = segments[:lastSegmentIndex]
+		u.Path = strings.Join(segments, "/")
+	}
+
+	base := gophercloud.NormalizeURL(u.String())
 
 	return &gophercloud.ProviderClient{
 		IdentityBase:     base,


jtopjian:
@MatMaul Thanks for submitting this.

Per the PR template and [STYLEGUIDE](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md), can you also open an Issue to describe the problem? If this change involves communicating with an OpenStack API service, you'll also need to include links to the OpenStack Python source code to validate the change.

MatMaul:
I'll open a bug. I can't pinpoint code server side, it is a bug in the client logic that just don't deal correctly with a Keystone URL that include a path (other than the version postfix).

jtopjian:
@MatMaul No problem, that makes sense.

jtopjian:
@MatMaul Just wanted to touch base about this. It looks like Travis is throwing an error. Might be as simple as running `gofmt` against the changes.

MatMaul:
Indeed thanks I didn't see it. gofmt should be happy now, let's see if I broke a test.

jtopjian:
oh boy... talk about uncovering a rabbit hole.

Thank you for identifying this and raising it as an issue :)

I tested this out with a devstack build and I can confirm that Keystone is now listening on `http://<ip>/identity` - an endpoint with no version. Similarly, `OS_AUTH_URL` is now set to that URL, too, which means that will eventually be the standard. I can also confirm that you are correct about Gophercloud stripping away the `/identity/` portion of the URL.

However, I'm unable to get acceptance tests to work with this patch. Even if I set `OS_AUTH_URL` to `http://<ip>/identity/v3`, tests still aren't working. You can try this yourself by doing:

```shell
source /path/to/openrc
export OS_TENANT_NAME=$OS_PROJECT_NAME
export OS_DOMAIN_NAME=default
go test -v -tags "fixtures acceptance" -run "TestKeypairsList" github.com/gophercloud/gophercloud/acceptance/openstack/compute/v2
```

When I run the above, I get:

```
=== RUN   TestKeypairsList
--- FAIL: TestKeypairsList (0.02s)
        keypairs_test.go:19: Unable to create a compute client: No supported version available from endpoint http://10.1.11.33/identity/
FAIL
exit status 1
FAIL    github.com/gophercloud/gophercloud/acceptance/openstack/compute/v2      0.032s
```

If I change `OS_AUTH_URL` to `http://<ip>/identity/v3/`, I get the same error.

The reason this error is happening is because you're not setting `IdentityEndpoint` in your patch. If this was set to match `OS_AUTH_URL`, then [this](https://github.com/gophercloud/gophercloud/blob/2804b72cf099b41d2e25c8afcca786f9f962ddee/openstack/utils/choose_version.go#L54-L59) piece of code would be triggered and all would probably work correctly.

However, I think we have a more long-term issue. I started digging into this and I found that Keystone is now replying with version `v3.8` instead of `v3.0`:

```
$ curl http://10.1.11.33/identity | python -mjson.tool                          
{                                                                               
    "versions": {                                                               
        "values": [                                                             
            {                                                                   
                "id": "v3.8",                                                   
                "links": [                                                      
                    {                                                           
                        "href": "http://10.1.11.33/identity/v3/",               
                        "rel": "self"                                           
                    }                                                           
                ],                                                              
                "media-types": [                                                
                    {                                                           
                        "base": "application/json",                             
                        "type": "application/vnd.openstack.identity-v3+json"    
                    }                                                           
                ],                                                              
                "status": "stable",                                             
                "updated": "2017-02-22T00:00:00Z"                               
            },                                                                  
            {                                                                   
                "id": "v2.0",                                                   
                "links": [                                                      
                    {                                                           
                        "href": "http://10.1.11.33/identity/v2.0/",             
                        "rel": "self"                                           
                    },                                                          
                    {                                                           
                        "href": "https://docs.openstack.org/",                  
                        "rel": "describedby",                                   
                        "type": "text/html"                                     
                    }                                                           
                ],                                                              
                "media-types": [                                                
                    {                                                           
                        "base": "application/json",                             
                        "type": "application/vnd.openstack.identity-v2.0+json"  
                    }                                                           
                ],                                                              
                "status": "deprecated",                                         
                "updated": "2016-08-04T00:00:00Z"                               
            }                                                                   
        ]                                                                       
    }                                                                           
}                                                                               
```

This means that [`utils.ChooseVersion`](https://github.com/gophercloud/gophercloud/blob/2804b72cf099b41d2e25c8afcca786f9f962ddee/openstack/utils/choose_version.go) will become less and less relevant as time goes on, to the point where `OS_AUTH_URL` will always need explicitly set to a version for authentication to work in Gophercloud. Gophercloud can't work with a versionless Keystone endpoint.

I started working on a more long-term fix and came up with [this](https://github.com/jtopjian/gophercloud/commit/d73f672536cf3ceff82d3f91e986941b6a4f20df). It works for all services except, ironically enough, Identity services, since the `ServiceURL` needs fixed. I thought I'd stop for now before I got too deep.

@jrperritt Thoughts on what's going on?

MatMaul:
Let's go down the hole :)
So I am not at ease with the semantic difference between IdentityBase and IdentityEndpoint.
Base is without the version and endpoint the full path ?
I'll do some tests with the acceptance suite.

For the version I don't think we really care actually. If I understand correctly it is v3.8 since now it supports micro-versioning but the base API is still v3 compatible. We just need to change a bit the code in ChooseVersion to handle correctly this use case and we can safely ignore the .8 part, at least while we don't depend on a micro version change/addition. In this use case we probably have to add some headers.
So I think that the approach in your patch is fine.

jtopjian:
> So I am not at ease with the semantic difference between IdentityBase and IdentityEndpoint.
> Base is without the version and endpoint the full path ?

That's correct

> For the version I don't think we really care actually.

Yes and no. If the `OS_AUTH_URL` setting in the `openrc` file bundled with devstack retained the `/v3`, I could see ignoring parsing of the Keystone version. But since `OS_AUTH_URL` is becoming versionless, too, this will become a larger usability issue within the next year as more and more deployments move to this format.

I consider sourcing an `openrc` file to be standard practice when working with any OpenStack-compatible application, so whatever the values are considered de-facto standard need to be considered here.

I believe that patching `NewClient` is acceptable in the short-term. I just sort of amended this issue with the long-term issue :)

jrperritt:
> Thoughts on what's going on?

Yes, see https://github.com/gophercloud/gophercloud/issues/106 and my response there. Unless there is more to this?

jtopjian:
Ah, I remember that Issue.

> Unless there is more to this?

I don't think so. This is similar to #106. The difference is that Keystone is now publishing a versionless endpoint which increases the likelihood that `NewClient` and `ChooseVersion` will fail.

With regard to a solution, is the progress I made [here](https://github.com/jtopjian/gophercloud/commit/d73f672536cf3ceff82d3f91e986941b6a4f20df) going in the right direction?

MatMaul:
@jtopjian can you push a PR with your WIP ? Seems fine to me.

jrperritt:
> is the progress I made here going in the right direction?

Yes, that looks like the right direction, with the caveat that, unless there's a compelling reason to use regexes, since we know what format to expect, I'd prefer to just parse the version string.

jtopjian:
> Yes, that looks like the right direction, with the caveat that, unless there's a compelling reason to use regexes, since we know what format to expect, I'd prefer to just parse the version string.

regexes are being used in two places. The first is in `NewClient` which is to detect a wide range of version endpoints (anything from `v2.0/` to `v2.0` to `v3` to `v3.8/`). I think a regex makes sense here given the range. Thoughts?

The second place they're used is in the `ChooseVersion` function. I can see replacing them with simple `strings.Contains`, looking for generic `v2` and `v3` strings.

jtopjian:
Each time I thought I've had this sorted out, I stumble upon another issue :)

I've updated my branch [here](https://github.com/jtopjian/gophercloud/commits/identity-versionless-endpoint) with a working solution. This branch has two commits:

The first commit fixes the issue you've reported about Gophercloud not working with versionless identity URLs or URLs that contain a path. It enables the client to successfully authenticate to a wider range of identity URLs. The client can then proceed to work with various OpenStack services.

Except for Identity Services. The reason is because of how Gophercloud parses the service catalog. Currently, if a service is publishing a versionless endpoint, Gophercloud attaches the well-known version suffix to the end of the published URL. For example, [here](https://github.com/gophercloud/gophercloud/blob/3172ddf2336933bc38b355df89cb015e0954df67/openstack/client.go#L248).

It's not doing this for Identity because Identity has, until recently, published a versioned URL.

The work done in the first commit should also be done during creation of the Identity ServiceClient (technically _all_ clients) [here](https://github.com/gophercloud/gophercloud/blob/3172ddf2336933bc38b355df89cb015e0954df67/openstack/endpoint_location.go), but that's a much larger issue that ties into another discussion happening about microversion discovery. 

These two commits _should_ be a good solution. The work in the second commit will most likely change down the road.

@MatMaul Would you be able to test this and see if it resolves the problem you ran into which caused you to report this issue?

If @jrperritt is OK with the use of the regex in the first commit, I can either submit a PR to you or just open a new PR here.

MatMaul:
I don't have something to test right now but the code LGTM.
@jrperritt can we move on ?

MatMaul:
@jtopjian you should probably send a new PR and reference this one. Thanks for the work.

jtopjian:
@MatMaul I apologize - this slipped my inbox. I just opened #508.

jrperritt:
superseded by #508

PR #430, opened 2017-07-22
Add Nova interface-list support

FengyunPan:
Add support to list interface for a server
through a GET on: /v2.1/{tenant_id}/servers/{server_id}/os-interface

The same operation with OpenStack CLI is done with:
  nova interface-list <server_id>

Fix gophercloud issue:
https://github.com/gophercloud/gophercloud/issues/431

Fix kubernetes issue: 

   https://github.com/kubernetes/kubernetes/issues/43909

### DIFF
--- a/acceptance/openstack/compute/v2/servers_test.go
+++ b/acceptance/openstack/compute/v2/servers_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/attachinterfaces"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/availabilityzones"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/pauseunpause"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
@@ -72,6 +73,20 @@ func TestServersCreateDestroy(t *testing.T) {
 
 	for network, address := range allAddresses {
 		t.Logf("Addresses on %s: %+v", network, address)
+	}
+
+	allInterfacePages, err := attachinterfaces.List(client, server.ID).AllPages()
+	if err != nil {
+		t.Errorf("Unable to list server Interfaces: %v", err)
+	}
+
+	allInterfaces, err := attachinterfaces.ExtractInterfaces(allInterfacePages)
+	if err != nil {
+		t.Errorf("Unable to extract server Interfaces: %v", err)
+	}
+
+	for _, Interface := range allInterfaces {
+		t.Logf("Interfaces: %+v", Interface)
 	}
 
 	allNetworkAddressPages, err := servers.ListAddressesByNetwork(client, server.ID, choices.NetworkName).AllPages()
--- a/openstack/compute/v2/extensions/attachinterfaces/doc.go
+++ b/openstack/compute/v2/extensions/attachinterfaces/doc.go
@@ -1,0 +1,3 @@
+// Package attachinterfaces provides the ability to manage network interfaces through
+// nova-network
+package attachinterfaces
--- a/openstack/compute/v2/extensions/attachinterfaces/requests.go
+++ b/openstack/compute/v2/extensions/attachinterfaces/requests.go
@@ -1,0 +1,13 @@
+package attachinterfaces
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// List makes a request against the nova API to list the servers interfaces.
+func List(client *gophercloud.ServiceClient, serverID string) pagination.Pager {
+	return pagination.NewPager(client, listInterfaceURL(client, serverID), func(r pagination.PageResult) pagination.Page {
+		return InterfacePage{pagination.SinglePageBase(r)}
+	})
+}
--- a/openstack/compute/v2/extensions/attachinterfaces/results.go
+++ b/openstack/compute/v2/extensions/attachinterfaces/results.go
@@ -1,0 +1,43 @@
+package attachinterfaces
+
+import (
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// FixedIP represents a Fixed IP Address.
+type FixedIP struct {
+	SubnetID  string `json:"subnet_id"`
+	IPAddress string `json:"ip_address"`
+}
+
+// Interface represents a network interface on an instance.
+type Interface struct {
+	PortState string    `json:"port_state"`
+	FixedIPs  []FixedIP `json:"fixed_ips"`
+	PortID    string    `json:"port_id"`
+	NetID     string    `json:"net_id"`
+	MACAddr   string    `json:"mac_addr"`
+}
+
+// InterfacePage abstracts the raw results of making a List() request against the API.
+// As OpenStack extensions may freely alter the response bodies of structures returned
+// to the client, you may only safely access the data provided through the ExtractInterfaces call.
+type InterfacePage struct {
+	pagination.SinglePageBase
+}
+
+// IsEmpty returns true if an InterfacePage contains no interfaces.
+func (r InterfacePage) IsEmpty() (bool, error) {
+	interfaces, err := ExtractInterfaces(r)
+	return len(interfaces) == 0, err
+}
+
+// ExtractInterfaces interprets the results of a single page from a List() call,
+// producing a map of interfaces.
+func ExtractInterfaces(r pagination.Page) ([]Interface, error) {
+	var s struct {
+		Interfaces []Interface `json:"interfaceAttachments"`
+	}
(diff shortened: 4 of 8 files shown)


FengyunPan:
@jtopjian Hi jtopjian, I'd like you to do a code review, thank you.
Currently the openstack cloud provider of kubernetes can't get port by ip, so we need support node interface-list for it, then openstack cloud provider can get port by interface and ip.

jtopjian:
@FengyunPan Thank you for submitting this. Per the PR template and the [STYLEGUIDE](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md) you'll need to provide links to the python source code which implements this feature in the OpenStack service. Also, a corresponding Gophercloud issue needs to be opened. Please let me know if you have any questions.

FengyunPan:
@jtopjian I have added it, need other information?

jtopjian:
@FengyunPan I had some time to look into this in more detail.

It looks like the ability to "get" interfaces is part of the `os-interface-attachment`. This can be seen in the Nova routes [here](https://github.com/openstack/nova/blob/bf0a5bc48519d97f104d3e0ea199f6308d390a1a/nova/api/openstack/compute/routes.py#L833-L835) and [here](https://github.com/openstack/nova/blob/bf0a5bc48519d97f104d3e0ea199f6308d390a1a/nova/api/openstack/compute/routes.py#L339-L340). That ends up linking to the `attach_interfaces.py` file you linked to earlier:

https://github.com/openstack/nova/blob/4f91ed3a547965ed96a22520edcfb783e7936e95/nova/api/openstack/compute/attach_interfaces.py

This is actually utterly confusing because I had no idea this ability existed hidden in the `os-attach-interface` extension -- I could have used this a long time ago, so thank you for submitting it.

Since it's an extension, it needs to be packaged under the `compute/v2/extensions` directory, titled `attachinterfaces`. It's relatively easy to bundle this code as an extension. See the newly added [`pauseunpause`](https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions/pauseunpause) package as a template.

As for the code itself, you'll need to modify the field names to use "camel case" rather than the current format which is using underscores. For example, `Fixed_id` needs renamed to `FixedID`, `Port_state` needs changed to `PortState`, etc.

Once all of that is done, I can do another review.

FengyunPan:
@jtopjian Thank for your comment, I will update it soon.

openstack/compute/v2/extensions/attachinterfaces/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package attachinterfaces provides the ability to manage attach interfaces through
```
jtopjian: How about:

"the ability to manage interfaces"

openstack/compute/v2/extensions/attachinterfaces/requests.go
```diff
@@ -0,0 +1,13 @@
+package attachinterfaces
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// ListInterfaces makes a request against the nova API to list the servers interfaces.
+func ListInterfaces(client *gophercloud.ServiceClient, id string) pagination.Pager {
```
jtopjian: The attach-interfaces extension only lists one type of resource (interfaces), so this can be renamed to just `List`.

In addition, the `id` is the id of the server, so let's rename `id` to `serverID` so it's not mistaken as an "interface id".

openstack/compute/v2/extensions/attachinterfaces/results.go
```diff
@@ -0,0 +1,43 @@
+package attachinterfaces
+
+import (
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Interface represents an interface.
+type FixedID struct {
+	SubnetID  string `json:"subnet_id"`
+	IpAddress string `json:"ip_address"`
```
jtopjian: `IPAddress`

openstack/compute/v2/extensions/attachinterfaces/results.go
```diff
@@ -0,0 +1,43 @@
+package attachinterfaces
+
+import (
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Interface represents an interface.
```
jtopjian: // FixedIP represents a Fixed IP Address.

openstack/compute/v2/extensions/attachinterfaces/results.go
```diff
@@ -0,0 +1,43 @@
...
+type FixedID struct {
```
jtopjian: `FixedIP`

openstack/compute/v2/extensions/attachinterfaces/results.go
```diff
@@ -0,0 +1,43 @@
...
+	SubnetID  string `json:"subnet_id"`
+	IpAddress string `json:"ip_address"`
+}
+
+// Interface represents an interface.
```
jtopjian: How about:

// Interface represents a network interface on an instance.

Just so it's not confused as a golang interface.

openstack/compute/v2/extensions/attachinterfaces/results.go
```diff
@@ -0,0 +1,43 @@
...
+type Interface struct {
+	PortState string    `json:"port_state"`
+	FixedIPs  []FixedID `json:"fixed_ips"`
```
jtopjian:
```golang
FixedIPs []FixedIP `json:"fixed_ips"`
```

openstack/compute/v2/extensions/attachinterfaces/results.go
```diff
@@ -0,0 +1,43 @@
...
+	PortID    string    `json:"port_id"`
+	NetID     string    `json:"net_id"`
+	MacAddr   string    `json:"mac_addr"`
+}
+
+// InterfacePage abstracts the raw results of making a ListInterfaces() request against the API.
```
jtopjian: `List()`

acceptance/openstack/compute/v2/servers_test.go
```diff
@@ -74,6 +75,20 @@ func TestServersCreateDestroy(t *testing.T) {
 		t.Logf("Addresses on %s: %+v", network, address)
 	}
 
+	allInterfacePages, err := attachinterfaces.ListInterfaces(client, server.ID).AllPages()
```
jtopjian: `List`

openstack/compute/v2/extensions/attachinterfaces/doc.go
FengyunPan: Yeah, thank you. Done.

openstack/compute/v2/extensions/attachinterfaces/requests.go
FengyunPan: Done.

openstack/compute/v2/extensions/attachinterfaces/results.go
FengyunPan: Done.

openstack/compute/v2/extensions/attachinterfaces/results.go
FengyunPan: Done.

openstack/compute/v2/extensions/attachinterfaces/results.go
FengyunPan: Done.

openstack/compute/v2/extensions/attachinterfaces/results.go
FengyunPan: Done.

openstack/compute/v2/extensions/attachinterfaces/results.go
FengyunPan: Done.

openstack/compute/v2/extensions/attachinterfaces/results.go
FengyunPan: Done.

acceptance/openstack/compute/v2/servers_test.go
FengyunPan: Done.

openstack/compute/v2/extensions/attachinterfaces/results.go
```diff
@@ -0,0 +1,43 @@
...
+	IPAddress string `json:"ip_address"`
+}
+
+// Interface represents a network interface on an instance.
+type Interface struct {
+	PortState string    `json:"port_state"`
+	FixedIPs  []FixedIP `json:"fixed_ips"`
+	PortID    string    `json:"port_id"`
+	NetID     string    `json:"net_id"`
+	MacAddr   string    `json:"mac_addr"`
```
jtopjian: @FengyunPan I'm sorry, I just noticed this.

This should be `MACAddr`. There are several other areas in Gophercloud where MAC is used, so this will be consistent. It's better to change this now than later.

FengyunPan: Yeah, my mistake, I update it soon, thank you very much.

jtopjian:
Looks good to me! Thank you again for your patience and for contributing this!

PR #435, opened 2017-07-26
Add Microversion support to sharev2 client

pospispa:
For #340

code in this PR:

https://github.com/openstack/manila/blob/stable/newton/manila/api/versions.py#L88

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -8,6 +8,7 @@ import (
 	"github.com/gophercloud/gophercloud"
 	tokens2 "github.com/gophercloud/gophercloud/openstack/identity/v2/tokens"
 	tokens3 "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
+	sfsapi "github.com/gophercloud/gophercloud/openstack/sharedfilesystems/apiversions"
 	"github.com/gophercloud/gophercloud/openstack/utils"
 )
 
@@ -261,7 +262,19 @@ func NewBlockStorageV2(client *gophercloud.ProviderClient, eo gophercloud.Endpoi
 
 // NewSharedFileSystemV2 creates a ServiceClient that may be used to access the v2 shared file system service.
 func NewSharedFileSystemV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
-	return initClientOpts(client, eo, "sharev2")
+	sc, err := initClientOpts(client, eo, "sharev2")
+	if err != nil {
+		return sc, err
+	}
+
+	apiVersion, err := sfsapi.Get(sc, "v2").Extract()
+	if err != nil {
+		return sc, err
+	}
+
+	sc.Microversion = apiVersion.Version
+
+	return sc, nil
 }
 
 // NewCDNV1 creates a ServiceClient that may be used to access the OpenStack v1


pospispa:
This commit somehow disappeared from PR https://github.com/gophercloud/gophercloud/pull/409 after a rebase.
This commit initializes the `ServiceClient.Microversion` value so that the `X-OpenStack-Manila-API-Version` header value is filled correctly.

jtopjian:
This implementation assumes that using Gophercloud's `NewSharedFileSystemV2` function will automatically configure the client with a microversion. I can see advantages and disadvantages to this, though I'm leaning more on this being an advantage.

@jrperritt From a design perspective, any advice on this?

pospispa:
@jtopjian I can think only about 2 options:
1. `NewSharedFileSystemV2` function automatically configures the client with microversion. IMO, the tricky part is that in case `initClientOpts` function succeeds, but GET request for API version fails the `NewSharedFileSystemV2` function returns an error even though the client can be used for all 2.0 features. However, failing to get API version is IMO strange situation that should not happen often.
Is this change backward compatible? I'm afraid not.

2. Have a dedicated function like `func InitMicroversion(client) (Microversion, error)`. Gophercloud library user will have to call the `func InitMicroversion` immediately after `func NewSharedFileSystemV2` in case the user wants to use 2.1+ features.
This change is backward compatible as the `func NewSharedFileSystemV2` is not changed. However, the user has to write more code.

@jtopjian please, how strong is the backward compatibility argument for you?

jtopjian:
@pospispa breaking backwards compatibility is taken into account on a case-by-case basis. If adding microversion support causes backwards compatibility issues, IMO, that is acceptable. Microversions are core to the Gophercloud client and backwards incompatibility might be unavoidable. However, we need to make sure the implementation is done correctly as to not make follow-up changes that would cause further issues.

This is why these PR reviews have been very strict. Please don't take it personal 😄 

>  However, failing to get API version is IMO strange situation that should not happen often.

I agree

> Is this change backward compatible? I'm afraid not.

Technically Option 1 you describe _is_ backwards compatible. The developer using Gophercloud would not have to change their code. If there was a new option added to `NewSharedFileSystemV2`, for example:

```go
func NewSharedFileSystemV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts, getMicroVersion bool)
```

then that would be a backwards incompatible change.

As you mentioned, Option 1 increases the chance of an error being returned since it's doing new things internally. This is something the developer should be aware of, but it would not break code.

There's also Option 3 which is to not merge this PR at all. This would put the responsibility of setting a microversion entirely on the developer and the application they maintain. As discussed in other PRs, Gophercloud is simply a library to assist a developer with communicating with the OpenStack APIs. Letting the developer / application opt-in to microversions is a valid option. However, microversions are becoming pretty core/fundamental to OpenStack, so this is a debatable area.

I don't have enough experience to make a good call on this, so I'll wait until @jrperritt has some time to give input.

pospispa:
> This is why these PR reviews have been very strict. Please don't take it personal :-)

AFAIK, Linus Torvalds has a rule that the person who created the PR have to persuade him that Linus will be better-off merging the PR than not merging it. IMO, this is a fair approach.

> Letting the developer / application opt-in to microversions is a valid option. However, microversions are becoming pretty core/fundamental to OpenStack, so this is a debatable area.

Citing from [the Microversion description](https://developer.openstack.org/api-guide/compute/microversions.html): "If neither X-OpenStack-Nova-API-Version nor OpenStack-API-Version (specifying compute) is provided, act as if the minimum supported microversion was specified."

So in case the microversions are switched-off by default, i.e. are not initialized in `func NewSharedFileSystemV2` the `gophercloud` library won't send `X-OpenStack-Manila-API-Version` header so the Manila API server will behave as if the minimum supported microversion was specified. And this is 100% backward compatible behaviour.

After our above discussion I'm more inclined to having microversions switched-off by default and providing `gophercloud` users with a `shares.InitializeMicroversion()` function to make switching the microversions on easier for users.

@jtopjian do you prefer to close this PR or change the content of this PR completely in such a way that it implements a `shares.InitializeMicroversion()` function?

jrperritt:
> From a design perspective, any advice on this?

The part I'm missing is what we're trying to accomplish by implementing microversion logic in Gophercloud. For example, we can get the microversion from the `apiversions` endpoint, but we don't know which microversion the user wants to use. Additionally, how will we know if the user's request is supported before it gets sent? Maybe it would help me understand better if y'all could provide some code or pseudocode demonstrating how y'all see a user interacting with Gophercloud using microversions?

pospispa:
> Additionally, how will we know if the user's request is supported before it gets sent?

An OpenStack service specification specifies for which microversion range is a particular request supported.

> The part I'm missing is what we're trying to accomplish by implementing microversion logic in Gophercloud. 

My initial intentions changed during the discussions in this PR and in https://github.com/gophercloud/gophercloud/pull/346 and https://github.com/gophercloud/gophercloud/pull/436

Let me summarize what we already have merged into `gophercloud` library (I'll focus only on Manila as similar things apply to other services that use microversions):
- [`X-OpenStack-Manila-API-Version` header is filled in with `client.Microversion`](https://github.com/gophercloud/gophercloud/blob/master/service_client.go#L116). I interpret this code as the `client.Microversion` is the particular microversion a `gophercloud` library user wants to use.
- It's possible to [get minimum and maximum microversions](https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/apiversions/results.go#L69) supported by the particular Manila API the `gophercloud` library is talking to.

So `gophercloud` users can now successfully use microversions in the `gophercloud` library as is. However, I believe there're 3 scenarios that can deliver a better user experience.

First scenario: `gophercloud` library should return a better error message than just ["Resource not found"](https://github.com/gophercloud/gophercloud/pull/346#issuecomment-318015733).
User requires a "low" microversion, e.g.:
```
client.Microversion = "2.5"
```
and calls a `gophercloud` library method, e.g. `grantAccessReqResponse, err := shares.GrantAccess(client, grantAccessReq, createdShare.ID)` that sends a request that is supported from ["2.7+" microversion](https://developer.openstack.org/api-ref/shared-file-systems/?expanded=grant-access-detail)
In case the `func shares.GrantAccess` does not check that the user requested microversion is `>= "2.7"` it return a generic error "Resource not found". IMO, this is not enough. The error message should be better. That's why a check `gophercloud.CompatibleMicroversion` is discussed [here](https://github.com/gophercloud/gophercloud/pull/436#issuecomment-318567306). IMO, it's worth to have such checks at the beginning of functions like `shares.GrantAccess`.

Second scenario: as mentioned [here](https://github.com/gophercloud/gophercloud/pull/436#issuecomment-318641978) microversions may introduce small incompatibilities. So some `gophercloud` library code will be specific to certain microversion range. So check if the user requested microversion is within a certain range of microversions should be needed in future.

Third scenario: a user wants to check if the microversion requested by him/her is supported by the Manila API the `gophercloud` library is talking to. The user may do it himself/herself:
```
min, max, err := GetServerMicroversions()
if err != nil {
    return err
}
client.Microversion = "2.5" 
if client.Microversion is NOT within min max server microversion interval {
    return fmt.Errorf("...")
}
```
However, the `gophercloud` library should provide a function that will do the above for the user.

openstack/client.go
```diff
@@ -261,7 +262,19 @@ func NewBlockStorageV2(client *gophercloud.ProviderClient, eo gophercloud.Endpoi
...
+	if err != nil {
+		return sc, err
+	}
+
+	apiVersion, err := sfsapi.Get(sc, "v2").Extract()
+	if err != nil {
+		return sc, err
+	}
+
+	sc.Microversion = apiVersion.Version
```
jrperritt: The `Microversion` in `ServiceClient` is the version the user wants to use. Why set it here to the maximum version allowed by the OpenStack deployment?

jtopjian:
> So gophercloud users can now successfully use microversions in the gophercloud library as is. However, I believe there're 3 scenarios that can deliver a better user experience.

As we've been discussing over the past days, these are all valid scenarios. I'd like us to move this discussion to #441 since I think we need to look at microversions in Gophercloud as a whole rather than in separate issues/PR, each discussing one piece of how they will be used.

pospispa:

PR #436, opened 2017-07-26
[WIP] Methods for Parsing and Comparing Microversions

pospispa:
For #315, #380 

code in this PR:
N/A

### DIFF
--- a/openstack/utils/microversions.go
+++ b/openstack/utils/microversions.go
@@ -1,0 +1,65 @@
+package utils
+
+import (
+	"errors"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+// ErrIncompatible is an error in case Incompatible microversion occured.
+var ErrIncompatible = errors.New("Incompatible microversion")
+
+// CompatibleMicroversion checks input microversions to find out whether they are compatible or not. The input parameters are:
+// - minimumMV: minimum microversion supported by the called function.
+// - maximumMV: maximum microversion supported by the called function.
+// - requestedMV: microversion requested by the user of the gophercloud library.
+// - serverMaximumMV: maximum microversion supported by the particular server the gophercloud library is talking to.
+func CompatibleMicroversion(minimumMV, maximumMV, requestedMV, serverMaximumMV string) error {
+	if requestedMV == "latest" {
+		if maximumMV == "" {
+			return nil
+		}
+		maximumMajor, maximumMinor := splitMicroversion(maximumMV)
+		serverMaximumMajor, serverMaximumMinor := splitMicroversion(serverMaximumMV)
+		if (maximumMajor > serverMaximumMajor) || (maximumMajor == serverMaximumMajor && maximumMinor >= serverMaximumMinor) {
+			return nil
+		}
+		return ErrIncompatible
+	}
+	if err := validMicroversion(requestedMV); err != nil {
+		return err
+	}
+	minimumMajor, minimumMinor := splitMicroversion(minimumMV)
+	requestedMajor, requestedMinor := splitMicroversion(requestedMV)
+	if (requestedMajor > minimumMajor) || (requestedMajor == minimumMajor && requestedMinor >= minimumMinor) {
+		return nil
+	}
+	return ErrIncompatible
+}
+
+func splitMicroversion(mv string) (major, minor int) {
+	if err := validMicroversion(mv); err != nil {
+		return
+	}
+
+	mvParts := strings.Split(mv, ".")
+	major, _ = strconv.Atoi(mvParts[0])
+	minor, _ = strconv.Atoi(mvParts[1])
+
+	return
+}
+
+func validMicroversion(mv string) (err error) {
+	if mv == "latest" {
+		return
+	}
+
+	mvRe := regexp.MustCompile("^\\d+\\.\\d+$")
+	if v := mvRe.MatchString(mv); v {
+		return
+	}
+
+	err = ErrIncompatible
+	return
+}
--- a/openstack/utils/testing/microversions_test.go
+++ b/openstack/utils/testing/microversions_test.go
@@ -1,0 +1,36 @@
+package testing
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/gophercloud/gophercloud/openstack/utils"
+)
+
+func TestCompatibleMicroversion(t *testing.T) {
+	functionUnderTest := "utils.CompatibleMicroversion"
+	tests := []struct {
+		minMicroversion       string
+		maxMicroversion       string
+		requestedMicroversion string
+		serverMaxMicroversion string
+		want                  error
+	}{
+		{"2.7", "", "latest", "2.22", nil},
+		{"2.7", "", "2.5", "2.22", utils.ErrIncompatible},
+		{"2.7", "2.14", "latest", "2.22", utils.ErrIncompatible},
+		{"2.7", "2.14", "latest", "2.11", nil},
+		{"2.7", "", "2.11.5", "2.22", utils.ErrIncompatible},
+		{"2.7", "", "2.11", "2.22", nil},
+	}
+	for i, tt := range tests {
+		fmt.Printf("%v", i)
+		if got := utils.CompatibleMicroversion(tt.minMicroversion, tt.maxMicroversion, tt.requestedMicroversion, tt.serverMaxMicroversion); got != tt.want {
+			if got == nil {
+				t.Errorf("%v(%q, %q, %q, %q) = (%v) WANT (%v)", functionUnderTest, tt.minMicroversion, tt.maxMicroversion, tt.requestedMicroversion, tt.serverMaxMicroversion, got, tt.want)
+			} else {
+				t.Errorf("%v(%q, %q, %q, %q) = (%v) WANT (%v)", functionUnderTest, tt.minMicroversion, tt.maxMicroversion, tt.requestedMicroversion, tt.serverMaxMicroversion, got, tt.want)
+			}
+		}
+	}
+}


pospispa:
Instead of keeping `ServiceClient.Microversion` as `string` I propose to use a specialized type `Microversion` for parsing and comparing microversions.

jtopjian:
@pospispa Thank you for breaking this out into a smaller PR.

I think a lot of the functionality here can be simplified by basing it strictly on OpenStack's definition of a microversion:

* https://developer.openstack.org/api-guide/compute/microversions.html
* http://specs.openstack.org/openstack/api-wg/guidelines/microversion_specification.html

A collection of details from both links:

* The VERSION_STRING must match “^([1-9]d*).([1-9]d*|0)$”.
* If X-OpenStack-Nova-API-Version or OpenStack-API-Version has a value of latest (special keyword), act as if maximum was specified. 
*  Note that these versions numbers do not follow the rules of Semantic Versioning.

Given the above, I think doing a simple regex check is all that is needed:

```go
mvRe := regexp.MustCompile("^([1-9]d*).([1-9]d*|0)$")
if v := mvRe.FindStringSubmatch(microversion); v != nil {
  return true, nil
}

if microversion == "latest" {
  return true, nil
}

return nil, fmt.Errorf("Invalid microversion specified")
```

The above is just quick pseudo-code. The regex will not work under Go.

> Instead of keeping ServiceClient.Microversion as string I propose to use a specialized type Microversion for parsing and comparing microversions.

Is there a reason for this? Is there a feature that will take advantage of a struct that will have both the Major and Minor number defined separately? Or is just checking if the microversion is valid all required?

Thoughts?

pospispa:
@jtopjian thank you for posting the links. After reading the links again after several weeks 2 questions occur to me. I think we should agree on answers to these 2  questions before improving the implementation.

1. Will gophercloud library support in the `X-OpenStack-Manila-API-Version` header the `"latest"` string value? I personaly don't need it. In addition, there is a warning:
> The latest value is mostly meant for integration testing and would be dangerous to rely on in client code since microversions are not following semver and therefore backward compatibility is not guaranteed. Clients should always require a specific microversion but limit what is acceptable to the microversion range that it understands at the time.

That's why I'm against supporting the `"latest"` string in gophercloud library. Please, what's your opinion?

2. The specs allows client to use a different microversion in the `X-OpenStack-Manila-API-Version` header in every request. Do we need this now? My approach is to get the server's API microversion value and use this value in all follow-up requests. Only check if the server's API microversion is high enough, e.g. in PR https://github.com/gophercloud/gophercloud/pull/346 I check in `func GrantAccess` whether the server's API microversion is >=2.7.
So are we going to design the implementation in such a way that it's easy to change the value in the `X-OpenStack-Manila-API-Version` header in every request?

Regarding your question:
> Is there a reason for this? Is there a feature that will take advantage of a struct that will have both the Major and Minor number defined separately? Or is just checking if the microversion is valid all that is required?

The reason for a specialized type instead of string is to simplify implementation, however, currently I think we should first agree on answers to the above questions.

jtopjian:
> That's why I'm against supporting the "latest" string in gophercloud library. Please, what's your opinion?

Throwing an error if the version is set to "latest" will make it very difficult to use Gophercloud for development purposes. 

To be frank, if you are submitting features like microversion support and parsing of microversions, you need to think about _all_ use-cases of Gophercloud - not just those that fulfil your own cases. Gophercloud's primary purpose is to interact with the OpenStack APIs, not dictate how one should interact with the APIs. That's the job of the application which uses Gophercloud.

> So are we going to design the implementation in such a way that it's easy to change the value in the X-OpenStack-Manila-API-Version header in every request?

This should already be possible, right?

A service client has a Microversion field which can be set at any time.

> The reason for a specialized type instead of string is to simplify implementation, however, currently I think we should first agree on answers to the above questions.

But do you have a specific way this would be used? Can you give a clear way the use of the Microversion type could be used when a string could not?

pospispa:
OK, the version "latest" must be supported on client's side. Of course, the maximum version in server's response must not be "latest", but must be a specific X.Y.

>> So are we going to design the implementation in such a way that it's easy to change the value in the X-OpenStack-Manila-API-Version header in every request?

> This should already be possible, right?
> A service client has a Microversion field which can be set at any time.

Yes, it's possible now.
However, we should store the minimum and maximum versions received from server somewhere.
I personally would create a specialized data type:
```
type Microversion struct {
  SrvMajorOfMinVersion uint
  SrvMinorOfMinVersion uint
  SrvMajorOfMaxVersion uint
  SrvMinorOfMaxVersion uint
  ClientMajor uint
  ClientMinor uint
  ClientLatest bool
}
```
to hold all this information and methods that will make sure that the data comply to the below rules:
- Server maximum version is bigger than server minimum version.
- Client version is either "latest" or client version is equal or greater that server minimum version and client version is equal or lower that server maximum version.

The service client `Microversion` field would be of the `Microversion` type.

>> The reason for a specialized type instead of string is to simplify implementation, however, currently I think we should first agree on answers to the above questions.

> But do you have a specific way this would be used? Can you give a clear way the use of the Microversion type could be used when a string could not?

I believe that by making the data more complicated I would be able to make the code less complicated. I believe that:
- Methods `SetSrvVersions` and `SetClientVersion` will make sure that the data integrity is kept.
- It's easier to compare "2.9" with "2.10" in the form of two pairs of `uint`s instead of in the form of two `string`s.
- The pair of two `uint`s cannot hold an invalid microversion value while a `string` can.

jtopjian:
> OK, the version "latest" must be supported on client's side. Of course, the maximum version in server's response must not be "latest", but must be a specific X.Y.

Yes, this is correct.

When a Gophercloud package implements a feature that requires a certain microversion, that feature will be "guarded" by a minimum microversion, which will be a `const` in the package. We can safely assume that the `const` will be a valid microversion. No one should commit an incorrectly formatted microversion to Gophercloud.

A developer will invoke that feature like this:

```go
client.Microversion = "2.5" 
// Or 2.5 could be set automatically when NewClient is called, if 2.5 is advertised in apiversions
result, err := foo.Bar(client, createOpts)
```

Then inside Gophercloud:

```go
package foo

const minBarMicroversion = "2.4"

func Bar(client *gophercloud.ServiceClient, createOpts CreateOpts) (result, error) {
  if err := gophercloud.CompatibleMicroversion(minBarMicroversion, client.Microversion); err != nil {
    return nil, err
  }

  // continue as normal
}
```

This is how I envisioning this working, but maybe I'm missing something? However, so far I haven't seen a need to have the struct you are suggesting. Do let me know if I am missing something, though.

As for the actual logic to check for a valid microversion and to compare it, I put this together which seems to fulfill the above example: https://gist.github.com/jtopjian/bf12ebcd5b10ed02871e2a1913ec6ca0

pospispa:
Let me firstly discuss what the `gophercloud.CompatibleMicroversion` function should check.

Firstly, `gophercloud.CompatibleMicroversion` function checks that the `minBarMicroversion` is lower or equal to the `client.Microversion`, i.e. microversion the user wants to use. I agree with this.

Secondly, citing from [Microversion Specification](http://specs.openstack.org/openstack/api-wg/guidelines/microversion_specification.html): "Each version includes all the changes since minimum version was introduced. It is not possible to request the feature introduced at microversion X.Y without accepting all the changes before X.Y in the same request. For example, you cannot request the feature which was introduced at microversion 2.100 without backwards incompatible changes which were introduced in microversion 2.99 and earlier."
So `func Bar` may not require only a minimum microversion (`minBarMicroversion`), but it may be implemented up to certain microversion so `maxBarMicroversion` is also needed. However, `maxBarMicroversion` will be infinity in most cases.
Actually, I came across [List export locations](https://developer.openstack.org/api-ref/shared-file-systems/?expanded=list-export-locations-detail) that is available from microversion "2.9", but the field `preferred` is available from microversion "2.14".

Lastly, in https://github.com/gophercloud/gophercloud/pull/346#issuecomment-318258288 we agreed that we will check microversions in order to get a better error message than the one provided by Manila API. I tried to use a client's microversion that is bigger that the server's maximum microversion or that is lower than the server's minimum microversion and received the below error messages:
- "Expected HTTP response code [200 201] when accessing [POST http://controller:8786/v2/ecbc0da9369f41e3a8a17e49a425ff2d/shares], but got 406 instead
{"computeFault": {"message": "Version 2.50 is not supported by the API. Minimum is 2.0 and maximum is 2.22.", "code": 406}}"
- "Expected HTTP response code [200 201] when accessing [POST http://controller:8786/v2/ecbc0da9369f41e3a8a17e49a425ff2d/shares], but got 406 instead
{"computeFault": {"message": "Version 1.7 is not supported by the API. Minimum is 2.0 and maximum is 2.22.", "code": 406}}"

So it seems the `gophercloud.CompatibleMicroversion` function does not have to check whether the client's microversion is within the min-max server microversion boundaries as the error message from the server is good enough. Please, do you agree?

How will a `gophercloud` library user switch-on the microversions feature?
You propose:
```
client.Microversion = "2.5" 
```
My idea is bigger (in pseudo-code):
```
min, max, err := GetServerMicroversions()
if err != nil {
    return err
}
client.Microversion = "2.5" 
if client.Microversion is NOT within min max server microversion interval {
    return fmt.Errorf("...")
}
```
Does my approach make sense? Does it make sense to put my idea into a `func` that will be a part of the `gophercloud` library? Will it make sense to store `min, max` server microversions in a `gophercloud` data structure, for example in `gophercloud.ServiceClient`?

> This is how I envisioning this working, but maybe I'm missing something? However, so far I haven't seen a need to have the struct you are suggesting. Do let me know if I am missing something, though.

Every time the `func Bar` is called the microversion `string`s are validated (`func validMicroversion`) and converted to integers before their comparison. I'd like to store the microversions as `(u)int`s to validate and convert them into integers only once and compare them as `(u)int`s many times because I expect functions like `func Bar` will be called many times. So my argument is about saving CPU cycles.

jrperritt:
I may not fully understand the Microversion concept, but here's what I'd like to see unless there are compelling reasons otherwise:
1. Microversion validation without a regular expression: split at the ".", check for length == 2, and try to cast each of the two `string` elements as an `int`.
2. No special types: keep the microversions as `string`s and compare with them "==". When comparing strings, "1.7" < "2.2" < "2.2.1" < "latest"

jtopjian:
> When comparing strings, "1.7" < "2.2" < "2.2.1" < "latest"

https://play.golang.org/p/WNOHxF07ap

Is there a way to compare strings in a way that would return `true` here?

pospispa:
> https://play.golang.org/p/WNOHxF07ap
>
> Is there a way to compare strings in a way that would return true here?

I can't think of any.

jrperritt:
> Is there a way to compare strings in a way that would return true here?

Good point. It sounds like the method that's already in this PR is the way to go, except I don't think we don't need to check for negative versions.

pospispa:
@jtopjian I started working on an implementation that is based on your [example](https://gist.github.com/jtopjian/bf12ebcd5b10ed02871e2a1913ec6ca0).

pospispa:

PR #437, opened 2017-07-26
[Pending #435] [Pending #436] Switch service client.microversion from string to microversion type

pospispa:
For #315, #380 

code in this PR:

N/A


pospispa:
Started using `internal.Microversion` type for `ServiceClient.Microversion` instead of string in order to simplify parsing and comparing in functions that create a request that is available from a particular microversion.

pospispa:

PR #433, opened 2017-07-24
support suspend/resume for compute instance

zionwu:
Prior to a PR being reviewed, there needs to be a Github issue that the PR
addresses. Replace the brackets and text below with that issue number.

For #432 

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:

https://github.com/openstack/nova/blob/stable/ocata/nova/api/openstack/compute/suspend_server.py
https://github.com/openstack/nova/blob/stable/ocata/nova/compute/api.py#L3381

Issue #432, opened 2017-07-24
Compute V2:  Support Suspend/Resume 


zionwu:
It would be nice to have suspend/resume action supported in gophercloud.

jtopjian:
Implemented by #438 

Thank you again 😄

PR #438, opened 2017-07-27
support suspend/resume for compute instance

zionwu:

For #432

code in this PR:

https://github.com/openstack/nova/blob/stable/ocata/nova/api/openstack/compute/suspend_server.py
https://github.com/openstack/nova/blob/stable/ocata/nova/compute/api.py#L3381

The forked gophercloud repository is missing in #433 , recreate a new PR.


jtopjian:
This looks good to me. Thank you for submitting it!

Issue #439, opened 2017-07-28
Compute v2: Add Suspend Acceptance Test


jtopjian:
With the merge of #438, it would be nice to have an acceptance test to go along with it.

jtopjian:
Implemented by #440

PR #440, opened 2017-07-28
Compute v2: Add Suspend Acceptance Test

jtopjian:
For #439


jtopjian:
This is going to cause a merge conflict with #430, so I will wait until #430 is merged and then fix this PR.

PR #442, opened 2017-08-02
Domain scoping via name

jobrs:

For #252

code in this PR:

https://github.com/openstack/python-keystoneclient/blob/67b1cd3705c0e45f0304f8a9d008b03f554ec7d2/keystoneclient/auth/identity/v3/base.py#L165


jtopjian:
@jobrs Thank you for submitting this. I've finally had a chance to look this over.

The Travis error is being caused by some formatting issues. Can you run `go fmt` on the files which were changed?

In #252, you said:

> There is still not adequate AuthOptionsFromEnv function in openstack/identity/v3/tokens. I did not fill this gap since I do not need it.

What is still missing from `AuthOptionsFromEnv`? 

Finally, for historical reference, here is how `keystoneauth` is building scope:

https://github.com/openstack/keystoneauth/blob/1abfa7c843381cced3879aa260d2e0d2f137f3f3/keystoneauth1/identity/v3/base.py#L112-L179

jobrs:
@jtopjian sorry, I messed up this branch and had to move.

AuthOptionsFromEnv is missing support for V3 specific environment variables, e.g. OS_PROJECT_DOMAIN_NAME

Let's continue in https://github.com/gophercloud/gophercloud/pull/489

PR #443, opened 2017-08-05
Extensions Refactor

jtopjian:
For #300

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/extensions.go
+++ b/acceptance/openstack/networking/v2/extensions/extensions.go
@@ -21,12 +21,16 @@ func CreateExternalNetwork(t *testing.T, client *gophercloud.ServiceClient) (*ne
 
 	adminStateUp := true
 	isExternal := true
-	createOpts := external.CreateOpts{
-		External: &isExternal,
+
+	networkCreateOpts := networks.CreateOpts{
+		Name:         networkName,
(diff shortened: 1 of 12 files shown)


jtopjian:
I'm not sure if this PR could be broken up into separate ones. I believe a change in any of these will cause tests to fail in other areas.

jrperritt:
Thanks for digging into this! After skimming it, the first thing I'd like to see are several pass/fail scenarios run via unit tests for `extractIntoPtr` (located in a yet-to-be-created `/testing/results_test.go`). Having the `networking` unit tests that run through the function are a bonus, but really we want to know if the function itself stands up to all the things we want it to.

jtopjian:
> After skimming it, the first thing I'd like to see are several pass/fail scenarios run via unit tests for extractIntoPtr

Ah, yeah, that's a good point. I'll look at this sometime this weekend.

I think the Travis error is with Travis itself. All tests pass locally.

jrperritt:
From Travis:

`acceptance/openstack/networking/v2/extensions/extensions.go:24: undefined: external.CreateOpts`

jtopjian:
You have a better eye than me!

jtopjian:
@jrperritt Let me know if there are any other scenarios you had in mind.

jtopjian:
Yay! Thanks!

Issue #82, opened 2016-09-01
Migration


stgleb:
Is there currently ability to migrate server using gophercloud?

jrperritt:
Not that I know of

jtopjian:
@stgleb Sorry for the delay with this.

It looks like this can be implemented just like how [`ForceDelete`](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/requests.go#L283) is implemented. The way both are implemented in the API server is functionally the same, too:

* https://github.com/openstack/nova/blob/77b11a9e31e265c338857326418bfe8cd135c1f8/nova/api/openstack/compute/migrate_server.py
* https://github.com/openstack/nova/blob/77b11a9e31e265c338857326418bfe8cd135c1f8/nova/api/openstack/compute/deferred_delete.py

Unless you know of a reason why the options should be extended? Do you know of a deployed OpenStack that has extended the migrate action?

Let me know if you'd like to continue working on this one. If not, I'm happy to take it over.

jrperritt:
closed by #444

PR #444, opened 2017-08-10
Compute v2: Adding Migrate Action

jtopjian:
For #82 

Supersedes #84 

https://github.com/openstack/nova/blob/77b11a9e31e265c338857326418bfe8cd135c1f8/nova/api/openstack/compute/migrate_server.py

Note that this only implements what would be handled in the `_migrate` function, not the `_migrate_live` function.

### DIFF
--- a/openstack/compute/v2/extensions/migrate/requests.go
+++ b/openstack/compute/v2/extensions/migrate/requests.go
@@ -1,0 +1,11 @@
+package migrate
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// Migrate will initiate a migration of the instance to another host.
+func Migrate(client *gophercloud.ServiceClient, id string) (r gophercloud.ErrResult) {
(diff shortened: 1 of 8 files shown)


jtopjian:
@jrperritt ping for review. No rush, though.

openstack/compute/v2/extensions/migrate/requests.go
```diff
@@ -0,0 +1,11 @@
+package migrate
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// Migrate will initiate a migration of the instance to another host.
+func Migrate(client *gophercloud.ServiceClient, id string) (r gophercloud.ErrResult) {
```
jrperritt: Can we return a `MigrateResult` instead?

jtopjian: We certainly can. I'll get this cleaned up soon and ping.

jtopjian:
@jrperritt ping

Issue #448, opened 2017-08-11
SharedFileSystem v2: Fix GetExportLocations Result Type


jtopjian:
As a follow-up to #381, `ExtractExportLocations` should be restricted to the `GetExportLocations` result type only.

jrperritt:
closed by #449

PR #449, opened 2017-08-11
SharedFileSystems v2: Restrict ExtractExportLocations to only GetExportLocationsResults

jtopjian:
For #448

### DIFF
--- a/openstack/sharedfilesystems/v2/shares/results.go
+++ b/openstack/sharedfilesystems/v2/shares/results.go
@@ -96,18 +96,24 @@ func (r commonResult) Extract() (*Share, error) {
 	return s.Share, err
 }
 
-// CreateResult contains the result..
+// CreateResult contains the response body and error from a Create request.
 type CreateResult struct {
 	commonResult
 }
 
-// DeleteResult contains the delete results
+// DeleteResult contains the response body and error from a Delete request.
 type DeleteResult struct {
 	gophercloud.ErrResult
 }
 
-// GetResult contains the get result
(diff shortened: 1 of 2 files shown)


openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -96,21 +96,27 @@ func (r commonResult) Extract() (*Share, error) {
...
-// GetResult contains the get result
+// GetResult contains the response body and error from a Get request.
 type GetResult struct {
 	commonResult
 }
 
+// GetExportLocationsResult contains the result body and error from an
+// GetExportLocations request.
+type GetExportLocationsResult struct {
+	commonResult
```
jrperritt: This isn't a `commonResult`, is it? Does it return a `Share` in addition to an `ExportLocation`? If not, why should the method below stay `ExtractExportLocations` instead of `Extract`?

jtopjian:
All good questions. Thank you.

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -96,21 +96,27 @@ func (r commonResult) Extract() (*Share, error) {
...
-// GetResult contains the get result
+// GetResult contains the response body and error from a Get request.
 type GetResult struct {
 	commonResult
 }
 
+// GetExportLocationsResult contains the result body and error from an
+// GetExportLocations request.
+type GetExportLocationsResult struct {
+	gophercloud.Result
```
jrperritt: :+1:

Issue #446, opened 2017-08-10
Provide Better GoDoc Documentation


jtopjian:
This is a shell issue to house various PRs that will update the godoc documentation throughout Gophercloud.

jtopjian:
@jrperritt Would you mind if I merged the PRs that only contain doc changes or do you prefer to do a review of them? There's going to be a good number of these over the next week.

jrperritt:
> Would you mind

No, in fact I was going to suggest just that.

jtopjian:
Alright. That'll do for now. I'll circle back to BlockStorage once the v3 stuff has been sorted out.

jrperritt:
@jtopjian With #471 merging, it looks like all the referenced issues have merged. I think we can close this, but feel free to re-open if there's more to do. Thank you for all these docs

PR #454, opened 2017-08-11
BlockStorage v2: Updating docs for volumeactions

jtopjian:
For #446 

This also contains the fixes in #452, so this should be merged (and possibly rebased) after that.


jtopjian:
@jrperritt This one is ready to go.

j-griffith:
LGTM

Issue #455, opened 2017-08-11
BlockStorage v2: Cleanup volumetenants


jtopjian:
While reviewing the `volumetenants` package, I noticed it could use a cleanup to better conform to the extensions pattern.

jrperritt:
closed by #456

PR #456, opened 2017-08-11
BlockStorage v2: Clean-up volumetenants

jtopjian:
For #455 
For #446 

`VolumeExt` is renamed to `VolumeTenantExt` since `VolumeExt` is rather generic and having two extensions named `VolumeExt` cannot be used together.

The `UnmarshalJSON` function is not needed, so it has been removed.

Example documentation has also been added.


j-griffith:
LGTM

Issue #461, opened 2017-08-13
Add support for OS_PROJECT_* variables


jtopjian:
`openstack.AuthOptionsFromEnv` doesn't support the `OS_PROJECT_ID` and `OS_PROJECT_NAME` environment variables. A standard `openrc` file that a user downloads from Horizon (when using Keystone v3) will have `PROJECT` set instead of `TENANT`.

Adding support for `PROJECT` will make it possible for users to not have to modify their `openrc` file when using `openstack.AuthOptionsFromEnv`.

bugosiphionah:
I am applying for outreachy round 15 and starting to work on this issue as advised by Davanum Srinivas the mentor. I hope no one is working on it for now.

jtopjian:
@bugosiphionah Thank you for your interest in this! However, there is a PR already in place: #462 (referenced above).

bugosiphionah:
@jtopjian Thanks for the feedback I will let him know its been worked on so he can advise on another. If you have ideas for any first time bugs I can work on. You can also advise. Thanks

jtopjian:
@bugosiphionah Sure, I can try to find something. What's your current skill level with each:

* OpenStack APIs
* Go

brand new, some experience, familiar, very familiar?

bugosiphionah:
@jtopjian I am familiar with Go but brand new to OpenStack APIs but am flexible and willing to learn.

jrperritt:
closed by #462

PR #462, opened 2017-08-13
Support project env vars

jtopjian:
For #461 

This change should not affect scoping due to the underlying variable names not being changed, so the logic to determine scope stays the same.

### DIFF
--- a/openstack/auth_env.go
+++ b/openstack/auth_env.go
@@ -16,7 +16,12 @@ The following variables provide sources of truth: OS_AUTH_URL, OS_USERNAME,
 OS_PASSWORD, OS_TENANT_ID, and OS_TENANT_NAME.
 
 Of these, OS_USERNAME, OS_PASSWORD, and OS_AUTH_URL must have settings,
-or an error will result.  OS_TENANT_ID and OS_TENANT_NAME are optional.
+or an error will result.  OS_TENANT_ID, OS_TENANT_NAME, OS_PROJECT_ID, and
+OS_PROJECT_NAME are optional.
+
+OS_TENANT_ID and OS_TENANT_NAME are mutually exclusive to OS_PROJECT_ID and
+OS_PROJECT_NAME. If OS_PROJECT_ID and OS_PROJECT_NAME are set, they will
+still be referred as "tenant" in Gophercloud.
 
 To use this function, first set the OS_* environment variables (for example,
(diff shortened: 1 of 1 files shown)


openstack/auth_env.go
```diff
@@ -22,6 +39,16 @@ func AuthOptionsFromEnv() (gophercloud.AuthOptions, error) {
 	domainID := os.Getenv("OS_DOMAIN_ID")
 	domainName := os.Getenv("OS_DOMAIN_NAME")
 
+	// If OS_PROJECT_ID is set, overwrite tenantID with the value.
+	if v := os.Getenv("OS_PROJECT_ID"); v != "" {
```
pospispa: The source code comment says: "OS_TENANT_ID and OS_TENANT_NAME are mutually exclusive to OS_PROJECT_ID and OS_PROJECT_NAME."

Shouldn't be the mutual exclusivity checked and an error return respectively?

jtopjian: If Gophercloud had full support for `ProjectName` and `ProjectID` in `AuthOptions`, I could see throwing an error if both Tenant and Project were used. This kind of error check could be added in the future.

Terraform has been doing this kind of permissive setting for months and there hasn't been an issue raised about it, so I think it's pretty safe:

https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/provider.go#L44-L61

pospispa: > If Gophercloud had full support for `ProjectName` and `ProjectID` in `AuthOptions`, I could see throwing an error if both Tenant and Project were used. 

@jtopjian so gophercloud doesn't have full support for `ProjectName` and `ProjectID`. Please, would you elaborate more on what's missing?

jtopjian: Gophercloud uses `TenantName` and `TenantID` instead of `ProjectName` and `ProjectID`. Therefore, there's no way to have a conflict between `TenantName` and `ProjectName`.

pospispa: Yeap, you're right that `func AuthOptionsFromEnv` is the only place where both `OS_TENANT_*` and `OS_PROJECT_*` exist next to each other.

So I agree to keep the code as it is.

jtopjian:
@jrperritt ping for review when you have a moment.

jrperritt:
Missed this one too. This looks good

Issue #465, opened 2017-08-13
Compute v2: Clean up availabilityzones


jtopjian:
While reviewing the `availabilityzones` package, I noticed it could use a cleanup to better conform to the extensions pattern.

jrperritt:
closed by #466

PR #466, opened 2017-08-13
Compute v2: Clean up availabilityzones

jtopjian:
For #465 
For #446

### DIFF
--- a/openstack/compute/v2/extensions/availabilityzones/results.go
+++ b/openstack/compute/v2/extensions/availabilityzones/results.go
@@ -1,12 +1,8 @@
 package availabilityzones
 
-// ServerExt is an extension to the base Server object
-type ServerExt struct {
+// ServerAvailabilityZoneExt is an extension to the base Server result which
+// includes the Availability Zone information.
+type ServerAvailabilityZoneExt struct {
 	// AvailabilityZone is the availabilty zone the server is in.
 	AvailabilityZone string `json:"OS-EXT-AZ:availability_zone"`
-}
-
-// UnmarshalJSON to override default
-func (r *ServerExt) UnmarshalJSON(b []byte) error {
-	return nil
 }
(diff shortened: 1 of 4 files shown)


openstack/compute/v2/extensions/availabilityzones/results.go
```diff
@@ -1,12 +1,8 @@
 package availabilityzones
 
-// ServerExt is an extension to the base Server object
-type ServerExt struct {
+// ServerAvailabilityZoneExt is an extension to the base Server result which
+// includes the Availability Zone information.
+type ServerAvailabilityZoneExt struct {
```
jrperritt: I'm OK with renaming these, but this is a mouthful, especially when it's used in user code. Here, it seems (to me) like `availabilityzone.Ext` is clear. I'd like to hear suggestions, with an eye toward keeping it succinct yet clear when used in user code.

jtopjian: I agree that this one is a mouthful. Maybe `ServerAZExt`?

The naming pattern that I have in mind for these extensions is `ParentType-Extension-Ext`. The reason for this is to make sure the extension is easily recognizable with its parent type, since both will be under separate packages.

If this was named just `AvailabilityZoneExt` (or even `AvailabilityZone`), it wouldn't be immediately recognizable that it is intended to extend the `Server` result.

I suppose it's possible that an extension could be applied to more than one parent. I don't know if it's a good idea to enable/allow, that, though. That kind of flexibility might create confusion and odd ways of applying extensions (I don't have a specific extension in mind that is reusable, just thinking of a possible case).

With regard to `availabilityzone.Ext`, IIRC, there is an issue when embedded/anonymous structs have the same name. So something like this wouldn't work:

```go
var foo struct {
  servers.Server
  availabilityzone.Ext
  tenantid.Ext
}
```

jrperritt: > there is an issue when embedded/anonymous structs have the same name

That's right, I remember now. I'm now in favor of this.

Issue #470, opened 2017-08-14
Compute v2: Clean up diskconfig


jtopjian:
While reviewing the `diskconfig` package, I noticed it could use a cleanup to better conform to the extensions pattern.

jrperritt:
closed by #471

PR #471, opened 2017-08-14
Compute v2: Clean up diskconfig package

jtopjian:
For #470 
For #446 

A unit test for the extension can be added after #466 is merged since it modifies the same area of code. I tested it locally and it works as expected.


jtopjian:
@jrperritt ping for review. Unit test has been added now that #466 was merged.

PR #489, opened 2017-08-17
Domain name scoping

jobrs:

For #252

code in this PR:

https://github.com/openstack/python-keystoneclient/blob/67b1cd3705c0e45f0304f8a9d008b03f554ec7d2/keystoneclient/auth/identity/v3/base.py#L165

### DIFF
--- a/auth_options.go
+++ b/auth_options.go
@@ -338,7 +338,12 @@ func (opts *AuthOptions) ToTokenV3ScopeMap() (map[string]interface{}, error) {
 			},
 		}, nil
 	} else if scope.DomainName != "" {
-		return nil, ErrScopeDomainName{}
+		// DomainName
+		return map[string]interface{}{
+			"domain": map[string]interface{}{
+				"name": &scope.DomainName,
+			},
+		}, nil
 	}
 
 	return nil, nil
--- a/errors.go
+++ b/errors.go
@@ -393,13 +393,6 @@ func (e ErrScopeProjectIDAlone) Error() string {
 	return "ProjectID must be supplied alone in a Scope"
 }
 
-// ErrScopeDomainName indicates that a DomainName was provided alone in a Scope.
-type ErrScopeDomainName struct{ BaseError }
-
-func (e ErrScopeDomainName) Error() string {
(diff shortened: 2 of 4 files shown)


jobrs:
this is a replacement for PR https://github.com/gophercloud/gophercloud/pull/489

jtopjian:
@jobrs Thanks for submitting this.

It looks like the Travis errors are due to `gofmt` needing to be run on some of the files. Would you also be able to trim those `Merge` commits off? It's not a _huge_ deal since we can always squash them via Github.

> AuthOptionsFromEnv is missing support for V3 specific environment variables, e.g. OS_PROJECT_DOMAIN_NAME

Right. My understanding is that `OS_PROJECT_DOMAIN_NAME` and `OS_PROJECT_DOMAIN_ID` are used for scoping purposes. Gophercloud currently tries to determine scope based on the combination of the other auth options currently set.

While that works, it might be worthwhile to look at adding the full suite of auth options and env variables. That would provide common ground between Gophercloud and other OpenStack SDKs (notably the Python ones). 

However, that doesn't need to be done right now. Enabling `DomainName` in the scope should allow all auth/scope combinations and I think that's good progress.

jobrs:
@jtopjian gofmt was on my list. Now everything should be fine. I also managed to meld everything into a single commit.

I agree that providing a v3-version of AuthOptionsFromEnv should be treated separately. But since it is part of the original issue #252, that issue remain open or better be split into two. @majewsky any thoughts?

jobrs:
I believe we can ignore the -0,0006% decrease in coverage. It is happening because the removed code sections had a coverage slightly above average. All the added code is covered.

jtopjian:
> I believe we can ignore the -0,0006% decrease in coverage.

Yeah, that's just coveralls being coveralls :)

This looks good to me. For historical reference in this issue, here's the [server-side code](https://github.com/openstack/keystone/blob/285d6e408a7e84c6ee09a60f76fb74669409e289/keystone/common/tokenless_auth.py#L76-L79) that allows `domain_name` in the scope.

Issue #500, opened 2017-08-24
Support delete of flavors


fatmcgav:
Add support for deleting flavors.

jtopjian:
Fixed by #501

PR #501, opened 2017-08-24
Flavors support delete update

fatmcgav:
Add support for deleting flavors.

For #500


jtopjian:
The API implementation of this can be found here: https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavor_manage.py#L48

This looks good to me. I wrote a quick acceptance test and can confirm its functionality. Plus the debug output matches what's implemented here + in the `flavor_manage.py` file.

Issue #502, opened 2017-08-24
BlockStorage v2: Support Cinder "noauth" auth_strategy


tchughesiv:
Here's my current approach:
https://github.com/gophercloud/gophercloud/compare/master...tchughesiv:cinder-noauth

Example of Creating a Service Client
```golang
	ao, err := openstack.AuthOptionsFromEnv()
	provider, err := openstack.UnAuthenticatedClient(ao)
	client, err := openstack.NewBlockStorageV2(provider, gophercloud.EndpointOpts{
		Region:         os.Getenv("OS_REGION_NAME"),
		CinderEndpoint: os.Getenv("CINDER_ENDPOINT"),
	})
```

This is already supported in the python client:
https://github.com/openstack/python-cinderclient/commit/60f92db7049b4f66e5198b86bfecc1029b6cdccd

jtopjian:
@tchughesiv Would it be possible to implement this similar to swauth?

https://github.com/gophercloud/gophercloud/tree/master/openstack/objectstorage/v1/swauth

tchughesiv:
@jtopjian don't see why not... let me give it a go in a separate branch.

tchughesiv:
@jtopjian here's a first go add new route ... still no tests yet, but wanted initial feedback:
https://github.com/gophercloud/gophercloud/compare/master...tchughesiv:cinder-new-noauth

used like:
```golang
	ao, err := openstack.AuthOptionsFromEnv()
	client, err := noauth.NewBlockStorageV2(ao, noauth.EndpointOpts{
		CinderEndpoint: os.Getenv("CINDER_ENDPOINT"),
	})
```

jtopjian:
@tchughesiv I think it's a good start and it's worth opening a PR. 

Two comments:

1. Can you provide a link to the cinder-api / server-side implementation of this?
2. I think `ProviderClient` should be taken out of the `NewBlockStorageV2` function. I understand why you would want to bundle them together, but for consistency, it should be broken out. You can always write an app-specific `CinderNoAuth` function that bundles the two.

tchughesiv:
@jtopjian cinder api code:
https://github.com/openstack/cinder/blob/92980e4972356f16389539b4cc926d7b4ec1dc3c/cinder/api/middleware/auth.py#L117

PR coming soon... tks for feedback

jrperritt:
closed by #503

PR #503, opened 2017-08-25
BlockStorage v2: Support Cinder "noauth" auth_strategy

tchughesiv:
For https://github.com/gophercloud/gophercloud/issues/502

code in this PR:
https://github.com/openstack/cinder/blob/92980e4972356f16389539b4cc926d7b4ec1dc3c/cinder/api/middleware/auth.py#L117


openstack/blockstorage/v2/noauth/urls.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"fmt"
+
+	"github.com/gophercloud/gophercloud"
+)
+
+func getURL(client *gophercloud.ProviderClient, eo EndpointOpts, clientType string) (string, error) {
+	url := ""
+	if len(eo.CinderEndpoint) > 0 && clientType == "volumev2" {
```
jtopjian: How about moving this to `initClientOpts`? That way, when `getURL` is called, we can be confident that data is valid and just return a string.

jtopjian:
Also going to ping @j-griffith as he might be interested in this.

j-griffith:
@tchughesiv Awesome to see this, thanks!  Other than moving the checking out of getURL this LGTM.  

As far as the versioning thing in theory as @jtopjian mentioned it should work independently of versions but we have had issues over the years with this so I don't know for sure if it works or not.

Regardless I'm a fan of the the non-versioned name-spacing of the resources anyway so I fully support just having `blockstorage/noauth`.

tchughesiv:
@jtopjian & @j-griffith done... tks for the feedback

openstack/blockstorage/noauth/requests.go
```diff
@@ -0,0 +1,61 @@
...
+	return client, nil
+}
+
+func initClientOpts(client *gophercloud.ProviderClient, eo EndpointOpts, clientType string) (*gophercloud.ServiceClient, error) {
+	sc := new(gophercloud.ServiceClient)
+	if len(eo.CinderEndpoint) > 0 {
+		sc.Endpoint = getURL(client, eo.CinderEndpoint)
+	} else {
+		return nil, fmt.Errorf("Pass proper EndPointOpt")
+	}
```
jtopjian: Just a few nits about this one (I'm really sorry).

1. Since `CinderEndpoint` is a string, it's better to check with:

```go
if eo.CinderEndpoint != ""
```

2. The error message should be worded differently. It could be something as succinct as 

"CinderEndpoint is required"

tchughesiv:
@jtopjian went ahead and pushed some fixes... can rebase to same commit if preferred, once @jrperritt reviews.

tchughesiv:
@jtopjian @jrperritt friendly ping

jtopjian:
@tchughesiv Sorry for the delay with this feature.

I took a fresh look at it and saw a few areas that could be simplified and submitted a PR to your branch. If the changes look good to you, go ahead and merge them into your branch and then I'll merge everything into Gophercloud.

jrperritt:
> then I'll merge everything into Gophercloud.

Please don't. I'd like to see the changes first.

tchughesiv:
tks for the improvements @jtopjian

tchughesiv:
@jrperritt is there anything else you think this PR needs?

openstack/blockstorage/noauth/requests.go
```diff
@@ -0,0 +1,67 @@
...
+func NewClient(options gophercloud.AuthOptions) (*gophercloud.ProviderClient, error) {
+	client := &gophercloud.ProviderClient{
+		TokenID: fmt.Sprintf("%s:%s", options.Username, options.TenantName),
+	}
+
+	return client, nil
+}
+
+// UnAuthenticatedClient allows for standalone "noauth" Cinder usage.
+func UnAuthenticatedClient(options gophercloud.AuthOptions) (*gophercloud.ProviderClient, error) {
```
jrperritt: Can we remove this? It should be easy enough to pass in admin values to `NewClient`

tchughesiv: remove the entire "UnAuthenticatedClient" function?  are you saying just default to "admin" every time? 
As it is written today, someone can specify a user/tenant if they want to.

jrperritt: Why do we need this function?

jtopjian: Not to step on @jrperritt's feet, just dunno how busy he is to follow up :)

Since `UnAuthenticatedClient` and `NewClient` both take the same inputs and return the same values, the logic in `UnAuthenticatedClient` can easily be moved to `NewClient` like this:

```go
func NewClient(options gophercloud.AuthOptions) (*gophercloud.ProviderClient, error) {
	if options.Username == "" {
		options.Username = "admin"
	}
	if options.TenantName == "" {
		options.TenantName = "admin"
	}

	client := &gophercloud.ProviderClient{
		TokenID: fmt.Sprintf("%s:%s", options.Username, options.TenantName),
	}

	return client, nil
}
```

The returned `error` will always be `nil` with this combined function (technically it's always `nil` with `UnAuthenticatedClient` and `NewClient` anyway), but I think it's safe to leave -- maybe an actual error will be returned one day and the function signature won't have to be updated.

jtopjian: well that's bloody ironic.

tchughesiv: works for me... was just trying to follow existing framework... agree this is cleaner.

tchughesiv:
@jtopjian @jrperritt requested changes completed

openstack/blockstorage/noauth/requests.go
```diff
@@ -0,0 +1,57 @@
...
+
+// EndpointOpts specifies a "noauth" Cinder Endpoint.
+type EndpointOpts struct {
+	// CinderEndpoint [required] is currently only used with "noauth" Cinder.
+	// A cinder endpoint with "auth_strategy=noauth" is necessary, for example:
+	// http://example.com:8776/v2.
+	CinderEndpoint string
+}
+
+/*
```
jrperritt: This looks weird [to me] for such a small comment. Let's change to "//"-style like the others

tchughesiv:
@jrperritt added acceptance tests... let me know if i need to change anything.  thanks

acceptance/clients/clients.go
```diff
@@ -136,6 +137,23 @@ func NewBlockStorageV2Client() (*gophercloud.ServiceClient, error) {
...
+	client, err := noauth.NewClient(gophercloud.AuthOptions{
+		Username:   os.Getenv("OS_USERNAME"),
+		TenantName: os.Getenv("OS_TENANT_NAME"),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return noauth.NewBlockStorageV2(client, noauth.EndpointOpts{
+		CinderEndpoint: os.Getenv("CINDER_ENDPOINT"),
```
jrperritt: @tchughesiv I think OpenStack calls this CINDERCLIENT_BYPASS_URL and, if so, we'll want to use that. Can you confirm or disprove?

jtopjian: It's a little convoluted, but I _think_ `CINDER_ENDPOINT` will work:

https://github.com/openstack/python-cinderclient/blob/fd0a11e7d7bec10a18b8bc0b209e5620fca61590/cinderclient/shell.py#L209-L229

Both `CINDERCLIENT_BYPASS_URL` and `CINDER_ENDPOINT` ultimately get stored in `os_endpoint`.

tchughesiv: It appears CINDERCLIENT_BYPASS_URL has been depreciated in favor of CINDER_ENDPOINT.
https://docs.openstack.org/python-cinderclient/pike/cli/details.html

jtopjian: Actually, `CINDER_ENDPOINT` is the best choice. `CINDERCLIENT_BYPASS_URL` is marked as deprecated. `--os-endpoint` is the non-deprecated option, which `CINDER_ENDPOINT` is used as the environment variable.

jtopjian: @tchughesiv yep, agreed. arg, I guess Github doesn't automatically show updates for inline threads - just comments.

jrperritt: Fine by me

Issue #506, opened 2017-08-28
Compute v2: Flavor Access


jtopjian:
- [x] Flavor Access List (#507)
- [x] Flavor Access Add (#687)
- [x] Flavor Access Remove (#688)

API reference: 

* https://developer.openstack.org/api-ref/compute/#flavors-access-flavors-os-flavor-access

API code:

* General: https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavor_access.py
* List: https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavor_access.py#L32-L56
* Add: https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavor_access.py#L90-L113
* Remove: https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavor_access.py#L115-L137

PR #507, opened 2017-08-28
Compute v2: flavor access list

jtopjian:
For #506

### DIFF
--- a/openstack/compute/v2/flavors/requests.go
+++ b/openstack/compute/v2/flavors/requests.go
@@ -154,11 +154,13 @@ func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
 	return
 }
 
-// ListAccesses retrieves details about tenant access to a flavor. Use
-// ExtractAccesses to convert its result into a slice of FlavorAccess.
-func ListAccesses(client *gophercloud.ServiceClient, id string) (r AccessesResult) {
-	_, r.Err = client.Get(accessURL(client, id), &r.Body, nil)
-	return
+// ListAccesses retrieves the tenants which have access to a flavor.
+func ListAccesses(client *gophercloud.ServiceClient, id string) pagination.Pager {
+	url := accessURL(client, id)
+
+	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
+		return AccessPage{pagination.SinglePageBase(r)}
+	})
 }
 
 // IDFromName is a convienience function that returns a flavor's ID given its
--- a/openstack/compute/v2/flavors/requests.go
+++ b/openstack/compute/v2/flavors/requests.go
@@ -154,9 +154,9 @@ func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
 	return
 }
 
-// ListAccess retrieves details about tenant access to a flavor. Use
-// ExtractAccess to convert its result into a FlavorAccess.
-func ListAccess(client *gophercloud.ServiceClient, id string) (r AccessResult) {
+// ListAccesses retrieves details about tenant access to a flavor. Use
+// ExtractAccesses to convert its result into a slice of FlavorAccess.
+func ListAccesses(client *gophercloud.ServiceClient, id string) (r AccessesResult) {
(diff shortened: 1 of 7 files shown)


jtopjian:
@jrperritt I have a question about the design of this kind of implementation:

This implements a "List" as it returns a list of results. To conform with other lists, I used the whole pagination thing, but with a single page, given there's never going to be more than one page of results.

SinglePage works well for this type of API call, and AllPages provides a nice shortcut, but it still seems a bit too much code for what it is:

```
allPages, err := flavors.ListAccess(client, foo).AllPages()
allAccess, err := flavors.ExtractAccess(allPages)
```

I _could_ implement it this way:

```
allAccess, err := flavors.ListAccess(client, foo).Extract()
```

But using `List` in this context breaks a pattern.

Am I thinking too much into this? Is it better to deal with the extra few lines in favor of conformity?

openstack/compute/v2/flavors/results.go
```diff
@@ -28,6 +30,12 @@ type DeleteResult struct {
 	gophercloud.ErrResult
 }
 
+// AccessResult is the result from an Access operation. Call its Extract method
+// to interpret it as a FlavorAccess.
+type AccessResult struct {
+	gophercloud.Result
+}
```
jtopjian: This is leftover from me doing a test with `Extract` on `AccessResult`. I'll remove this at the same time any other changes are requested.

jtopjian:
After mulling this over for a few days, I decided to not do the whole pagination thing. Happy to revert, though.

This is ready for review.

openstack/compute/v2/flavors/requests.go
```diff
@@ -154,6 +154,13 @@ func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
 	return
 }
 
+// ListAccess retrieves details about tenant access to a flavor. Use
+// ExtractAccess to convert its result into a FlavorAccess.
```
jrperritt: `Extract`

jrperritt:
If there is no `Get` operation for this API, should we just make this operation the `Get`?

jtopjian:
Yeah, that's exactly what started the whole debate for me.

The way I see it, List is used for actions that contain more than one of something. Get is usually done to retrieve one resource. But I could just be making this more difficult than it needs to be.

jtopjian:
Actually, nevermind. GetAccessInfo works. I'll redo.

jrperritt:
Actually, I misread what you'd written originally. I thought you were saying the response will only ever be a single object (a slice of length at most 1), but maybe you were saying that a single-page response seems more like a `Get` operation? Looking at the OpenStack source, it looks like private flavors will return an entry for every project that can access that flavor.

If there can in fact be, say, hundreds of items coming back in the response, it does seem to me that `ListAccesses` would be most consistent. But, to what I think is your point, a single-page operation could just as well be a seen as a `Get`.

jtopjian:
> but maybe you were saying that a single-page response seems more like a Get operation? 

Yes, exactly. 

> Looking at the OpenStack source, it looks like private flavors will return an entry for every project that can access that flavor.

Yes, exactly again. To use a real-world context, `$work` has a series of `g1.x` flavors for GPU access. These flavors are given out upon request. So there is one access entry per tenant which will all be returned in this API call. It is not possible to return a single access entry.

> If there can in fact be, say, hundreds of items coming back in the response, it does seem to me that ListAccesses would be most consistent (with an accompanying ExtractAccesses). But, to what I think is your point, a single-page operation could just as well be a seen as a Get.

Right, exactly. So which direction? :)

jrperritt:
Because there is precedent in this package, I'd say `ListAccesses`/`ExtractAccesses`, and we could discuss using `Get`-style logic for single-page operations in future packages.

jtopjian:
@jrperritt ping when you have time.

jrperritt:
Sorry if I wasn't clear earlier: I think we should use pagination for this operation

jtopjian:
ahh - OK, no problem.

jtopjian:
@jrperritt How's this?

jtopjian:
Thanks for your help on this!

Issue #424, opened 2017-07-18
Not working if Keystone URL contains a path


MatMaul:
If a Keystone URL contains a path before the version qualifier it removes it and then obviously fails.
http://keystoneserver/v3/ => ok
http://keystoneserver/identity/v3/ => not ok. Generated URLs don't have "identity"

It is really easy to test since now devstack provides Keystone on /identity

jrperritt:
closed by #508

PR #508, opened 2017-08-28
Identity Versionless Endpoint

jtopjian:
For #424 

Supersedes #423, but there's useful discussion in #423 to reference.


j-griffith:
Glad I looked, I just started messing with this and luckily checked to see if anybody already fixed it up.  Tested it out very briefly and looks good.

👍

jtopjian:
Thanks!

Issue #509, opened 2017-08-30
Network v2: Security groups are wiped out on port update if not supplied


ccatlett2000:
If ports.Update is called without specifying a change to the security groups null is sent to neutron, which wipes out an instance's groups.

Example code to trigger bug:
```go
package main

import (
        "fmt"
        "github.com/gophercloud/gophercloud"
        "github.com/gophercloud/gophercloud/openstack"
        "github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
)

func main() {
        // Setup neutron client
        opts := gophercloud.AuthOptions{
                // Your OpenStack details here
                IdentityEndpoint: "https://openstack.example.com:35357/v3",
                Username: "example",
                Password: "pass",
                TenantName: "example",
                DomainName: "default",
        }
        provider, _ := openstack.AuthenticatedClient(opts)
        client, _ := openstack.NewNetworkV2(provider, gophercloud.EndpointOpts{
                Name:   "neutron",
                Region: "RegionOne", // Your region here
        })

        // Example address pair to update
        newPairs := []ports.AddressPair {ports.AddressPair{
                IPAddress: "10.244.0.0/24",
        }}

        // Run ports.Update
        // Although SecurityGroups is not specified, null will be passed to neutron
        // Change the ID to an existing port that has a security group attached
        ports.Update(client, "f5902ec0-9c5a-4438-ab65-bb64861668da", ports.UpdateOpts{
                AllowedAddressPairs: newPairs,
        }).Extract()
}
```

When I execute the above code (with my OS creds plugged in) NGINX logs this request body:
```
{"port":{"allowed_address_pairs":[{"ip_address":"10.244.0.0/24"}],"security_groups":null}}
```

jtopjian:
@ccatlett2000 Thanks for reporting this.

Can you provide the output when you use either the `openstack` or `neutron` command line tool with the `--debug` flag?

```shell
$ neutron --debug port-update ...
```

I suspect `security_groups` is omitted entirely from the request body.

ccatlett2000:
@jtopjian Yeah, I can confirm that it just doesn't submit `security_groups` at all, as I saw with horizon's requests. However, just in case you need the debug output still:

```
openstack port set --debug --allowed-address ip-address=10.244.0.0/24,mac-address=fa:16:3e:a2:14:a8 f5902ec0-9c5a-4438-ab65-bb64861668da    
[...]

PR #445, opened 2017-08-10
Network v2: portsbinding refactor

jtopjian:
For #298

PR #510, opened 2017-08-31
Networking v2: Port Fixes

jtopjian:
Port `SecurityGroups` and `AllowedAddressPairs` have a tertiary type of behavior.

During creation, if `security_groups` is omitted from the request body, then Neutron will apply the default security group. If an empty array is passed, no security groups will be applied to the port. If an array of security groups is passed, those security groups will be applied to the port.

For update, if `security_groups` is omitted from the request body, existing security groups on the port are unchanged. If an empty array is passed, all security groups are removed. If an array of security groups is passed, then the port will have those groups applied.

`allowed_address_pairs` has the same behavior, but only for updates.

For #509 

Amends #236

#236 has the appropriate links to the API code, but the client code is easier to understand for this one:

https://github.com/openstack/python-neutronclient/blob/bc56657633cb7b3938e8a3fedb70dd050e68805c/neutronclient/neutron/v2_0/port.py#L142-L147

https://github.com/openstack/python-neutronclient/blob/bc56657633cb7b3938e8a3fedb70dd050e68805c/neutronclient/neutron/v2_0/port.py#L215-L219

@jrperritt I can split this into 2 PRs, but 99% of this PR is tests. The actual change is quite small (though significant).

### DIFF
--- a/openstack/networking/v2/ports/requests.go
+++ b/openstack/networking/v2/ports/requests.go
@@ -110,13 +110,13 @@ type UpdateOptsBuilder interface {
 
 // UpdateOpts represents the attributes used when updating an existing port.
 type UpdateOpts struct {
-	Name                string        `json:"name,omitempty"`
-	AdminStateUp        *bool         `json:"admin_state_up,omitempty"`
-	FixedIPs            interface{}   `json:"fixed_ips,omitempty"`
-	DeviceID            string        `json:"device_id,omitempty"`
-	DeviceOwner         string        `json:"device_owner,omitempty"`
-	SecurityGroups      *[]string     `json:"security_groups,omitempty"`
-	AllowedAddressPairs []AddressPair `json:"allowed_address_pairs"`
+	Name                string         `json:"name,omitempty"`
+	AdminStateUp        *bool          `json:"admin_state_up,omitempty"`
+	FixedIPs            interface{}    `json:"fixed_ips,omitempty"`
+	DeviceID            string         `json:"device_id,omitempty"`
+	DeviceOwner         string         `json:"device_owner,omitempty"`
+	SecurityGroups      *[]string      `json:"security_groups,omitempty"`
+	AllowedAddressPairs *[]AddressPair `json:"allowed_address_pairs,omitempty"`
 }
 
 // ToPortUpdateMap builds a request body from UpdateOpts.
--- a/openstack/networking/v2/ports/requests.go
+++ b/openstack/networking/v2/ports/requests.go
@@ -81,7 +81,7 @@ type CreateOpts struct {
 	DeviceID            string        `json:"device_id,omitempty"`
 	DeviceOwner         string        `json:"device_owner,omitempty"`
 	TenantID            string        `json:"tenant_id,omitempty"`
-	SecurityGroups      []string      `json:"security_groups,omitempty"`
+	SecurityGroups      *[]string     `json:"security_groups,omitempty"`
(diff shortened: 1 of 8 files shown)


jtopjian:
@jrperritt Ping for review. The unit test clutter makes this one a little awkward to review. The cleanup I did in #445 would have helped, but I'd like to make this one a priority since this is more of a bug.

openstack/networking/v2/ports/requests.go
```diff
@@ -110,13 +110,13 @@ type UpdateOptsBuilder interface {
...
-	DeviceID            string        `json:"device_id,omitempty"`
-	DeviceOwner         string        `json:"device_owner,omitempty"`
-	SecurityGroups      []string      `json:"security_groups"`
-	AllowedAddressPairs []AddressPair `json:"allowed_address_pairs"`
+	Name                string         `json:"name,omitempty"`
+	AdminStateUp        *bool          `json:"admin_state_up,omitempty"`
+	FixedIPs            interface{}    `json:"fixed_ips,omitempty"`
+	DeviceID            string         `json:"device_id,omitempty"`
+	DeviceOwner         string         `json:"device_owner,omitempty"`
+	SecurityGroups      *[]string      `json:"security_groups,omitempty"`
```
jrperritt: :+1:  exactly the way to fix this

openstack/networking/v2/ports/requests.go
```diff
@@ -81,7 +81,7 @@ type CreateOpts struct {
 	DeviceID            string        `json:"device_id,omitempty"`
 	DeviceOwner         string        `json:"device_owner,omitempty"`
 	TenantID            string        `json:"tenant_id,omitempty"`
-	SecurityGroups      []string      `json:"security_groups,omitempty"`
+	SecurityGroups      *[]string     `json:"security_groups,omitempty"`
```
jrperritt: Why this one too?

jtopjian: The previous PR only modified the updating of security groups, but while researching this I found that security groups are handled the same way during create. I included it in here hoping to get all port SG/AAP quirks done and over with. If this is too much of a change for a single PR, I can submit this one in a separate one.

The PR description and commit message lay out the details. As for the code to prove this, the same functions are called both on the client and server side to trigger this behavior. You can see that port creation supports this behavior on the client-side [here](https://github.com/openstack/python-neutronclient/blob/bc56657633cb7b3938e8a3fedb70dd050e68805c/neutronclient/neutron/v2_0/port.py#L265).

I've included unit tests to mimic the create behavior. The acceptance tests were updated as well and my test environment behaves accordingly, though I only added a "warning" instead of a failure because I have heard of some clouds that modify this behavior (to account for the odd case that someone runs the acceptance test in one of these clouds).

jtopjian: One more note: what might make this ultra-confusing is that the client `CreatePort` inherits from the [`UpdatePortSecGroupMixin`](https://github.com/openstack/python-neutronclient/blob/bc56657633cb7b3938e8a3fedb70dd050e68805c/neutronclient/neutron/v2_0/port.py#L222)

jrperritt:
@jtopjian As a side note, if the commits for a PR are organized (like all yours seem to be), I have recently been (and plan to continue) just merging (instead of squashing and creating a new commit), with the assumption that should help with the hierarchy of "pending" PRs

jtopjian:
I like to tell a story with my commits :P

Thanks for the quick review and merge!

Issue #515, opened 2017-09-14
Networking v2: Routers GatewayInfo is missing external_fixed_ips


BenoitKnecht:
Neutron returns an `external_fixed_ips` list in the `external_gateway_info` of a router, but it's not being parsed by gophercloud.

jtopjian:
Implemented by #516

PR #516, opened 2017-09-14
Networking v2: Routers: Parse external_fixed_ips

BenoitKnecht:
For #515

code in this PR:

https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/l3.py#L93-L99

### DIFF
--- a/openstack/networking/v2/extensions/layer3/routers/results.go
+++ b/openstack/networking/v2/extensions/layer3/routers/results.go
@@ -8,7 +8,15 @@ import (
 // GatewayInfo represents the information of an external gateway for any
 // particular network router.
 type GatewayInfo struct {
-	NetworkID string `json:"network_id"`
+	NetworkID        string            `json:"network_id"`
+	ExternalFixedIPs []ExternalFixedIP `json:"external_fixed_ips,omitempty"`
+}
+
+// ExternalFixedIP is the IP address and subnet ID of the external gateway of a
+// router.
+type ExternalFixedIP struct {
+	IP       string `json:"ip_address"`
+	SubnetID string `json:"subnet_id"`
 }
 
 // Route is a possible route in a router.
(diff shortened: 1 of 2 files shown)


jtopjian:
@BenoitKnecht Thank you for submitting this. Is there somewhere else in the Neutron code that further defines `external_fixed_ips`? Specifically something that confirms `external_fixed_ips` contains only a `subnet_id` and `ip_address`.

BenoitKnecht:
@jtopjian The only thing I could find is [this description](https://github.com/openstack/neutron-lib/blob/master/api-ref/source/v2/parameters.yaml#L4844-L4848).

jtopjian:
We need to find the piece of code that sets those two attributes. Unfortunately this can sometimes be more difficult than making the PR itself. We're not _trying_ to be difficult, and it's not that I don't believe this PR is correct, but for code review and to have a historical reference to why PRs were done a certain way, we need to find the neutron code.

 From looking at the link you gave, I see this:

```
'type:fixed_ips': None,
```

Is it possible that this is a reference to another schema (maybe in another Neutron repo) for `fixed_ips`?

openstack/networking/v2/extensions/layer3/routers/results.go
```diff
@@ -8,7 +8,15 @@ import (
...
 type GatewayInfo struct {
-	NetworkID string `json:"network_id"`
+	NetworkID        string            `json:"network_id"`
+	ExternalFixedIPs []ExternalFixedIP `json:"external_fixed_ips,omitempty"`
+}
+
+// ExternalFixedIP is the IP address and subnet ID of the external gateway of a
+// router.
+type ExternalFixedIP struct {
+	IP       string `json:"ip_address"`
```
jtopjian: Let's change this to `IPAddress` for [consistency](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/ports/results.go#L45-L49)

BenoitKnecht: @jtopjian Thanks a lot for digging through Neutron and finding that reference!

I've renamed `IP` to `IPAddress` as requested.

jtopjian:
@BenoitKnecht Looks good! Thanks for submitting this.

Issue #345, opened 2017-05-22
SharedFilesystems v2: Change Specs to ExtraSpecs


jtopjian:
@ehdou I just came across this while doing a review of the last sfs PR:

https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/v2/sharetypes/requests.go#L116

I think `Specs` should be changed to `ExtraSpecs` here for consistency. Let me know if you'd like to clean that up or I can do it. :)

ehdou:
Hmm, that makes sense.
I have changed company some time ago and I don't really have a devstack setup anymore so it would be easier for me if you would take care of it :)
Let me know if I can assist you somehow.

jtopjian:
@bugosiphionah How about this one? It should be an easy one to get started with.

bugosiphionah:
@jtopjian  Sounds good. Working on this. I will set up and share a PR .

jtopjian:
Fixed by #517

PR #517, opened 2017-09-14
SharedFilesystems v2: Change Specs to ExtraSpecs

bugosiphionah:
For #345 

cc @jtopjian

### DIFF
--- a/openstack/sharedfilesystems/v2/sharetypes/requests.go
+++ b/openstack/sharedfilesystems/v2/sharetypes/requests.go
@@ -113,7 +113,7 @@ type SetExtraSpecsOptsBuilder interface {
 
 type SetExtraSpecsOpts struct {
 	// A list of all extra specifications to be added to a ShareType
-	Specs map[string]interface{} `json:"extra_specs"`
+	ExtraSpecs map[string]interface{} `json:"extra_specs" required:"true"`
 }
 
 // ToShareTypeSetExtraSpecsMap assembles a request body based on the contents of a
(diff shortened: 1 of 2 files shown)


openstack/sharedfilesystems/v2/sharetypes/requests.go
```diff
@@ -113,7 +113,7 @@ type SetExtraSpecsOptsBuilder interface {
 
 type SetExtraSpecsOpts struct {
 	// A list of all extra specifications to be added to a ShareType
-	Specs map[string]interface{} `json:"extra_specs"`
+	ExtraSpecs ExtraSpecsOpts `json:"extra_specs" required:"true"`
```
jtopjian: This should stay `map[string]interface{}`. According to the [Manila documentation](https://docs.openstack.org/manila/latest/contributor/capabilities_and_extra_specs.html), `driver_handles_share_servers` and `snapshot_support` are required for all drivers, which is most likely why they are part of `CreateOpts`.

But Manila can support other types of extra specs, depending on the driver in use. Since we'll never know the full list of supported options, `map[string]interface{}` can be used so users can add any extra spec they need.

OpenStack can be confusing like this. I would have done the same thing myself. I have a test environment that runs Manila and I noticed I could add other key/values, so that's what confirmed `map[string]interface{}`.

`required:"true"`: this is a good catch. It turns out the API will trigger an error if `ExtraSpecs` is left empty :)

bugosiphionah: I will make the necessary changes. Thanks

jtopjian:
Looks good! Thank you for working on this.

I'm going to try to open some Issues with more small/easy tasks. Keep an eye out in the next few days :)

Issue #518, opened 2017-09-15
Compute v2: Add Lock/Unlock Action


jtopjian:
Nova supports the ability to lock and unlock an instance:

* https://developer.openstack.org/api-ref/compute/#lock-server-lock-action
* https://developer.openstack.org/api-ref/compute/#unlock-server-unlock-action

This action would look very similar to the [pause/unpause](https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions/pauseunpause) action.

Additional work required: Locate the Nova API Server code that implements lock/unlock.

Bonus: Implement a lock/unlock [acceptance test](https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/compute/v2/servers_test.go#L447-L481) and verify it works.

bugosiphionah:
@jtopjian I Would love to work on this over the weekend. We can discuss my solution on IRC on Monday.

jtopjian:
Sure, go for it. :)

Github is the best way to discuss Gophercloud. I'm not normally on IRC.

jtopjian:
Implemented in #522 

@bugosiphionah Nice work on this!

PR #522, opened 2017-09-17
Compute v2: Add Lock/Unlock Action

bugosiphionah:
For #518 

https://github.com/openstack/nova/blob/stable/ocata/nova/api/openstack/compute/lock_server.py

cc @jtopjian


jtopjian:
@bugosiphionah Just a heads up that before this can be reviewed, this also needs done:

> Additional work required: Locate the Nova API Server code that implements lock/unlock.

See https://github.com/gophercloud/gophercloud/pull/422 for a discussion that relates how to look for this.

bugosiphionah:
@jtopjian Noted!

bugosiphionah:
@jtopjian  I am trying to figure out why the acceptance tests for these actions still fail as implemented in previous commits. I dont see why. I look to be implementing things the right way. Could I be missing some thing critical?
```
func TestServersActionLock(t *testing.T) {
	t.Parallel()

	client, err := clients.NewComputeV2Client()
	if err != nil {
		t.Fatalf("Unable to create a compute client: %v", err)
	}

	server, err := CreateServer(t, client)
	if err != nil {
		t.Fatal(err)
	}
	defer DeleteServer(t, client, server)

	t.Logf("Attempting to Lock server %s", server.ID)
	err = lockunlock.Lock(client, server.ID).ExtractErr()
	if err != nil {
		t.Fatal(err)
	}

	err = lockunlock.Unlock(client, server.ID).ExtractErr()
	if err != nil {
		t.Fatal(err)
	}

	err = WaitForComputeStatus(client, server, "ACTIVE")
	if err != nil {
		t.Fatal(err)
	}
}
```

jtopjian:
@bugosiphionah The acceptance test is working for me. What errors are you seeing when you run it?

bugosiphionah:
@jtopjian  fails on Travis CI for some reason but I look like I am implementing everything right.

jtopjian:
@bugosiphionah Travis doesn't actually run the acceptance tests. However, Travis does check the formatting of the code and will report an error if `gofmt` needs to be run. That looks to be the case with the last failure:

https://travis-ci.org/gophercloud/gophercloud/jobs/276870294

Check toward the bottom and you'll see the formatting errors.

Let me know if you have any questions. :)

bugosiphionah:
@jtopjian  I think you were right I ran` gofmt` on the test file and it now works!

jtopjian:
@bugosiphionah Is this ready for review?

bugosiphionah:
@jtopjian yes you can review.

openstack/compute/v2/extensions/lockunlock/requests.go
```diff
@@ -0,0 +1,19 @@
+package lockunlock
+
+import "github.com/gophercloud/gophercloud"
+
+func actionURL(client *gophercloud.ServiceClient, id string) string {
+	return client.ServiceURL("servers", id, "action")
+}
+
+// Lock is the operation responsible for locking a Compute server.
+func Lock(client *gophercloud.ServiceClient, id string) (r gophercloud.ErrResult) {
```
jtopjian: Instead of `gophercloud.ErrResult`, let's change this to `LockResult`. I believe the reason to do this is to localize the result strictly to the Lock functionality. This is _not_ done in the `pauseunpause` extension that I referenced in #518, which means `pauseunpause` should be updated to reflect this as well. See [here](https://github.com/gophercloud/gophercloud/pull/444#pullrequestreview-57964997) for more discussion. As well, you can use the `migrate` action as a template for how to implement the response.

openstack/compute/v2/extensions/lockunlock/requests.go
```diff
@@ -0,0 +1,19 @@
...
+	_, r.Err = client.Post(actionURL(client, id), map[string]interface{}{"lock": nil}, nil, nil)
+	return
+}
+
+// Unlock is the operation responsible for unlocking a Compute server.
+func Unlock(client *gophercloud.ServiceClient, id string) (r gophercloud.ErrResult) {
```
jtopjian: Similar here, but `UnlockResult`.

openstack/compute/v2/extensions/lockunlock/requests.go
jtopjian: Instead of gophercloud.ErrResult, let's change this to LockResult. I believe the reason to do this is to localize the result strictly to the Lock functionality. This is not done in the pauseunpause extension that I referenced in #518, which means pauseunpause should be updated to reflect this as well. See here for more discussion. As well, you can use the migrate action as a template for how to implement the response.

openstack/compute/v2/extensions/lockunlock/requests.go
jtopjian: Similar here, but `UnlockResult`.

openstack/compute/v2/extensions/lockunlock/requests.go
jtopjian: The above has been edited with the actual link to discussion. Copy/paste fail on my part.

jtopjian:
@bugosiphionah Also, with regard to the acceptance test, were you able to run it on an actual OpenStack environment to verify it worked? If not, it might be beneficial to try doing that. Acceptance tests aren't required, but it helps to have run the code on a real system -- sometimes you can spot bugs that didn't come up in the unit tests.

There's documentation on how to run the tests here:

https://github.com/gophercloud/gophercloud/tree/master/acceptance

If you don't have access to an OpenStack environment, setting up devstack is usually sufficient. Let me know if you run into any problems.

One addition to the acceptance test would be to do this:

```go
func TestServersActionLock(t *testing.T) {
	t.Parallel()

	client, err := clients.NewComputeV2Client()
	if err != nil {
		t.Fatalf("Unable to create a compute client: %v", err)
	}

	server, err := CreateServer(t, client)
	if err != nil {
		t.Fatal(err)
	}
	defer DeleteServer(t, client, server)

	t.Logf("Attempting to Lock server %s", server.ID)
	err = lockunlock.Lock(client, server.ID).ExtractErr()
	if err != nil {
		t.Fatal(err)
	}

        // Ensure the server is locked
        err := servers.Delete(client, server.ID).ExtractErr()
        if err == nil {
          t.Fatalf("Should not have been able to delete the server")
        }

	err = lockunlock.Unlock(client, server.ID).ExtractErr()
	if err != nil {
		t.Fatal(err)
	}

	err = WaitForComputeStatus(client, server, "ACTIVE")
	if err != nil {
		t.Fatal(err)
	}
}
```

By ensuring the server cannot be deleted, the acceptance test proves that `Lock` actually worked.

bugosiphionah:
Thanks for the review @jtopjian . I will commit changes as requested.

bugosiphionah:
@jtopjian ready for review.

jtopjian:
@bugosiphionah This looks good. Thank you for your work on this!

Issue #524, opened 2017-09-23
Compute v2: Small Extension Fixes


jtopjian:
Packages under `compute/v2/extensions` require some small fixes:

* lock/unlock doc fix
* Change `ErrResult` in suspend/resume
* migrate doc fix
* Change `ErrResult` in defsecrules
* Change `ErrResult` in pause/unpause
* Change `ErrResult` in start/stop
* Change `ErrResult` in secgroups

These are all small, minor changes, so I'm going to submit a PR with the collection.

jrperritt:
closed by #525

PR #525, opened 2017-09-23
Compute v2: Small Extension Fixes

jtopjian:
For #524


jtopjian:
@jrperritt ping for review

Issue #528, opened 2017-09-26
Identity v3: Add Project List for Users


davidaah:
OpenStack Identity API v3 allows you to get a list of projects that a user has access to via `/users/{user_id}/projects`. I am looking to add this functionality to gophercloud which has a similiar implementation for listing groups that a user has access to (`/users/{user_id}/groups`)

jtopjian:
Implemented in #529

PR #529, opened 2017-09-26
Identity v3 - Add Get Projects by User

davidaah:
For #528

Details: Add ListUsers function for Gophercloud's OpenStack Identity v3 API
- Allows a caller to get the list of projects that a specified user has access to
- Effectively wraps OpenStack Identity v3 API function /v3/users/{user_id}/projects
- Based on implementation of ListGroups

code in this PR:

https://developer.openstack.org/api-ref/identity/v3/#list-projects-for-user
https://github.com/openstack/keystone/blob/stable/pike/keystone/common/policies/user.py#L39
https://github.com/openstack/keystone/blob/stable/pike/keystone/auth/controllers.py#L363-L381

**Update** Source link identified by maintainer: https://github.com/openstack/keystone/blob/6b6e1a7322666fb4da871281bb8c58db199c7f49/keystone/assignment/core.py#L227-L240

### DIFF
--- a/openstack/identity/v3/users/doc.go
+++ b/openstack/identity/v3/users/doc.go
@@ -89,7 +89,7 @@ Example to List Projects a User Belongs To
 		panic(err)
 	}
 
-	allProjects, err := projects.ExtractGroups(allPages)
+	allProjects, err := projects.ExtractProjects(allPages)
 	if err != nil {
 		panic(err)
 	}
--- a/openstack/identity/v3/users/doc.go
+++ b/openstack/identity/v3/users/doc.go
@@ -71,7 +71,7 @@ Example to List Groups a User Belongs To
 		panic(err)
 	}
 
-	allGroups, err := users.ExtractGroups(allPages)
+	allGroups, err := groups.ExtractGroups(allPages)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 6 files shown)


davidaah:
@jtopjian thanks for the feedback, absolutely agree with what you are saying. The issue of documentation license attribution came up in my internal review of the code change where I mentioned that I followed the pattern of the `ListUserGroups` which leveraged the OpenStack documentation for test code. So to proceed cautiously on our side with that, it was requested that I add the required attribution. 

Here is my thinking, let me know if this works for you:
1) Open a Separate issue specifically for how/where the project should provide attribution to OS license
2) Use completely fake response/response you have provided and remove attribution (since it will no longer be related) for this PR so as not to gate the change

If that's acceptable to you, I will push an updated commit shortly for review and open the other issue for broader discussion.

Thanks!

jtopjian:
@davidaah Sounds good to me.

davidaah:
updated this PR as requested and opened #537

openstack/identity/v3/users/doc.go
```diff
@@ -71,13 +71,32 @@ Example to List Groups a User Belongs To
...
+Example to List Projects a User Belongs To
+
+	userID := "0fe36e73809d46aeae6705c39077b1b3"
+
+	allPages, err := users.ListProjects(identityClient, userID).AllPages()
+	if err != nil {
+		panic(err)
+	}
+
+	allProjects, err := projects.ExtractGroups(allPages)
```
jtopjian: `s/ExtractGroups/ExtractProjects`

jtopjian:
@davidaah Quick note: I think this is the service-side function that is being called:

https://github.com/openstack/keystone/blob/6b6e1a7322666fb4da871281bb8c58db199c7f49/keystone/assignment/core.py#L227-L240

openstack/identity/v3/users/doc.go
davidaah: fixed!

davidaah:
thanks for the src find, i'll update the initial comment above for completeness

jtopjian:
Looks good! Thank you for submitting this!

Issue #519, opened 2017-09-15
Compute v2: Add Evacuate Action


jtopjian:
Nova supports the ability to evacuate a compute node:

* https://developer.openstack.org/api-ref/compute/#evacuate-server-evacuate-action

This action would look similar to the [migrate](https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions/migrate) action.

Additional work required: Locate where evacuate is implemented in the Nova API service.

Bonus: Implement an evacuate [acceptance test](https://github.com/gophercloud/gophercloud/tree/master/acceptance/openstack/compute/v2). However, this might be difficult since it might require a multi-node OpenStack testing environment.

bugosiphionah:
@jtopjian Working on this.

jtopjian:
Implemented by #532

PR #532, opened 2017-09-27
Compute v2: Add Evacuate Action

bugosiphionah:
For #519 

https://github.com/openstack/nova/blob/stable/ocata/nova/api/openstack/compute/evacuate.py

cc @jtopjian

### DIFF
--- a/openstack/compute/v2/extensions/evacuate/doc.go
+++ b/openstack/compute/v2/extensions/evacuate/doc.go
@@ -1,0 +1,13 @@
+/*
+Package evacuate provides functionality to evacuates servers that have been
+provisioned by the OpenStack Compute service from a failed host to a new host.
+
+Example to Evacuate a Server
+
+	serverID := "b16ba811-199d-4ffd-8839-ba96c1185a67"
+	err := evacuate.Evacuate(computeClient, serverID).ExtractErr()
+	if err != nil {
+		panic(err)
+	}
+*/
+package evacuate
--- a/openstack/compute/v2/extensions/evacuate/doc.go
+++ b/openstack/compute/v2/extensions/evacuate/doc.go
@@ -2,10 +2,10 @@
 Package evacuate provides functionality to evacuates servers that have been
 provisioned by the OpenStack Compute service from a failed host to a new host.
 
-Example to Evacuate a Server
+Example to Evacuate a Server from a Host
 
 	serverID := "b16ba811-199d-4ffd-8839-ba96c1185a67"
-	err := evacuate.Evacuate(computeClient, serverID).ExtractErr()
+	err := evacuate.Evacuate(computeClient, serverID, nil).ExtractErr()
 	if err != nil {
 		panic(err)
 	}
--- a/openstack/compute/v2/extensions/evacuate/doc.go
+++ b/openstack/compute/v2/extensions/evacuate/doc.go
@@ -5,7 +5,7 @@ provisioned by the OpenStack Compute service from a failed host to a new host.
 Example to Evacuate a Server from a Host
 
 	serverID := "b16ba811-199d-4ffd-8839-ba96c1185a67"
-	err := evacuate.Evacuate(computeClient, serverID, nil).ExtractErr()
+	err := evacuate.Evacuate(computeClient, serverID, evacuate.EvacuateOpts{}).ExtractErr()
 	if err != nil {
 		panic(err)
 	}
--- a/openstack/compute/v2/extensions/evacuate/requests.go
+++ b/openstack/compute/v2/extensions/evacuate/requests.go
@@ -13,10 +13,10 @@ type EvacuateOptsBuilder interface {
 // EvacuateOpts specifies Evacuate action parameters.
 type EvacuateOpts struct {
 	// The name of the host to which the server is evacuated
-	Host string `json:"host"`
+	Host string `json:"host,omitempty"`
 
 	// Indicates whether server is on shared storage
-	onSharedStorage *bool `json:"onSharedStorage"`
+	OnSharedStorage bool `json:"OnSharedStorage"`
 
 	// An administrative password to access the evacuated server
 	AdminPass string `json:"adminPass,omitempty"`
--- a/openstack/compute/v2/extensions/evacuate/requests.go
+++ b/openstack/compute/v2/extensions/evacuate/requests.go
@@ -13,7 +13,7 @@ type EvacuateOptsBuilder interface {
 // EvacuateOpts specifies Evacuate action parameters.
 type EvacuateOpts struct {
 	// The name of the host to which the server is evacuated
-	Host string 	`json:"host"`
+	Host string `json:"host"`
 
 	// Indicates whether server is on shared storage
 	onSharedStorage *bool `json:"onSharedStorage"`
--- a/openstack/compute/v2/extensions/evacuate/requests.go
+++ b/openstack/compute/v2/extensions/evacuate/requests.go
@@ -4,8 +4,38 @@ import (
 	"github.com/gophercloud/gophercloud"
 )
 
+// EvacuateOptsBuilder allows extensions to add additional parameters to the
+// the Evacuate request.
+type EvacuateOptsBuilder interface {
+	ToEvacuateMap() (map[string]interface{}, error)
+}
+
+// EvacuateOpts specifies Evacuate action parameters.
+type EvacuateOpts struct {
+	// The name of the host to which the server is evacuated
+	Host string 	`json:"host"`
+
+	// Indicates whether server is on shared storage
+	onSharedStorage *bool `json:"onSharedStorage"`
+
+	// An administrative password to access the evacuated server
+	AdminPass string `json:"adminPass,omitempty"`
+}
+
+// ToServerGroupCreateMap constructs a request body from CreateOpts.
+func (opts EvacuateOpts) ToEvacuateMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "evacuate")
+}
+
 // Evacuate will Evacuate a failed instance to another host.
-func Evacuate(client *gophercloud.ServiceClient, id string) (r EvacuateResult) {
-	_, r.Err = client.Post(actionURL(client, id), map[string]interface{}{"evacuate": nil}, nil, nil)
+func Evacuate(client *gophercloud.ServiceClient, id string, opts EvacuateOptsBuilder) (r EvacuateResult) {
+	b, err := opts.ToEvacuateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Post(actionURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+	})
 	return
 }
--- a/openstack/compute/v2/extensions/evacuate/requests.go
+++ b/openstack/compute/v2/extensions/evacuate/requests.go
@@ -39,17 +39,3 @@ func Evacuate(client *gophercloud.ServiceClient, id string, opts EvacuateOptsBui
 	})
 	return
 }
-
-func (r EvacuateResult) ExtractAdminPass() (string, error) {
-	var s struct {
-		AdminPass string `json:"adminPass"`
-	}
-	err := r.ExtractInto(&s)
-	if err != nil && err.Error() == "EOF" {
-		return "", nil
-	}
-	if err != nil && err.Error() != "EOF" {
-		return s.AdminPass, err
-	}
-	return s.AdminPass, err
-}
--- a/openstack/compute/v2/extensions/evacuate/requests.go
+++ b/openstack/compute/v2/extensions/evacuate/requests.go
@@ -45,8 +45,11 @@ func (r EvacuateResult) ExtractAdminPass() (string, error) {
 		AdminPass string `json:"adminPass"`
 	}
 	err := r.ExtractInto(&s)
+	if err != nil && err.Error() == "EOF" {
+	  return "", nil
+	}
 	if err != nil && err.Error() != "EOF" {
 		return s.AdminPass, err
 	}
-	return s.AdminPass, nil
+	return s.AdminPass, err
 }
--- a/openstack/compute/v2/extensions/evacuate/requests.go
+++ b/openstack/compute/v2/extensions/evacuate/requests.go
@@ -1,0 +1,11 @@
+package evacuate
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// Evacuate will Evacuate a failed instance to another host.
+func Evacuate(client *gophercloud.ServiceClient, id string) (r EvacuateResult) {
+	_, r.Err = client.Post(actionURL(client, id), map[string]interface{}{"evacuate": nil}, nil, nil)
+	return
+}
--- a/openstack/compute/v2/extensions/evacuate/requests.go
+++ b/openstack/compute/v2/extensions/evacuate/requests.go
@@ -16,7 +16,7 @@ type EvacuateOpts struct {
 	Host string `json:"host,omitempty"`
 
 	// Indicates whether server is on shared storage
-	OnSharedStorage bool `json:"OnSharedStorage"`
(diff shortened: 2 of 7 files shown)


bugosiphionah:
@jtopjian Please  guide me on how to create multi-node OpenStack testing environment so that I can work on an acceptance test for this action.

jtopjian:
@bugosiphionah That's a good question.

I know that both [devstack](https://docs.openstack.org/devstack/latest/guides/multinode-lab.html) and [PackStack](https://www.rdoproject.org/install/adding-a-compute-node/) support multi-node configurations, but I haven't tried either. If you'd like to try either of those out yourself, go for it and let me know if you have any problems. I will try to do a test of a multi-node PackStack environment when I get a chance (it might take me a few days to get to it).

jtopjian:
@bugosiphionah I've got some good news and bad news:

The good news is that I was able to create a multi-node deployment with PackStack. The bad news is that it's not as easy as the docs say. I've spent the past year+ investigating how to create OpenStack environments for testing and I should really write a document about best practices. In the meantime, if you let me know how you currently test OpenStack code (whether you use devstack, packstack, virtualbox, an existing OpenStack cloud (openstack on openstack), AWS, etc), I can give you some pointers.

More good news is that I was able to get the environment to support "evacuation", but more bad news is that the `nova-compute` service on one node must be "down" for the evacuation to work. In order to do this through the API, Gophercloud needs to support the [`os-services`](https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/services.py) API calls, which it does not yet. Therefore, it's not possible to accurately create an acceptance test for evacuation.

So for now, we can proceed with this, but without an acceptance test. I can write a quick test to verify it works in a specially tuned test environment.

jtopjian:
@bugosiphionah With all of the above said, here are some notes on the code you have so far:

You found the correct source code file for this - great work there.

The evacuate action is able to take a set of options. If you look at the [API reference doc](https://developer.openstack.org/api-ref/compute/#evacuate-server-evacuate-action) and click on "detail", you'll see a table with the list of options. In order to ensure these options are valid, you need to correlate them to the source code file you specified. For example: 

* "host (Optional)" matches [here](https://github.com/openstack/nova/blob/stable/ocata/nova/api/openstack/compute/evacuate.py#L91)
* "onSharedStorage" matches [here](https://github.com/openstack/nova/blob/stable/ocata/nova/api/openstack/compute/evacuate.py#L94)

etc

To add support for these options, you'll want to bundle them in a struct. Let's call it `EvacuateOpts`. You can use something like [this](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/servergroups/requests.go#L16-L34) as a template, but instead of the name `Create`, use `Evacuate`.

When someone submits a PR for an API call that has options, it's sometimes acceptable to omit some options in the initial PR (maybe they were undocumented, confusing, too difficult to implement at the time, etc), but the important thing is to ensure the functionality is in place for others to later add missing pieces (meaning: ensure the struct exists and the function accepts the struct as a parameter).

bugosiphionah:
@jtopjian I use devstack but have had hickups setting up the multi-node environment. As we figure this out , let me correct this pull request and do as advised above.

Thanks for the review.

bugosiphionah:
@jtopjian  I have made edits . This is ready for review.

openstack/compute/v2/extensions/evacuate/doc.go
```diff
@@ -0,0 +1,13 @@
+/*
+Package evacuate provides functionality to evacuates servers that have been
+provisioned by the OpenStack Compute service from a failed host to a new host.
+
+Example to Evacuate a Server
```
jtopjian: Let's change this to

Example to Evacuate a Server from a Host

openstack/compute/v2/extensions/evacuate/doc.go
```diff
@@ -0,0 +1,13 @@
...
+
+	serverID := "b16ba811-199d-4ffd-8839-ba96c1185a67"
+	err := evacuate.Evacuate(computeClient, serverID).ExtractErr()
```
jtopjian: This will need updated for `EvacuateOpts`. At a minimum, you can just specify `nil`:

```go
err := evacuate.Evacuate(computeClient, serverID, nil).ExtractErr()
```

openstack/compute/v2/extensions/evacuate/requests.go
```diff
@@ -0,0 +1,41 @@
...
+	ToEvacuateMap() (map[string]interface{}, error)
+}
+
+// EvacuateOpts specifies Evacuate action parameters.
+type EvacuateOpts struct {
+	// The name of the host to which the server is evacuated
+	Host string `json:"host"`
+
+	// Indicates whether server is on shared storage
+	onSharedStorage *bool `json:"onSharedStorage"`
```
jtopjian: This needs to be `OnSharedStorage`

openstack/compute/v2/extensions/evacuate/requests.go
```diff
@@ -0,0 +1,41 @@
...
+// EvacuateOptsBuilder allows extensions to add additional parameters to the
+// the Evacuate request.
+type EvacuateOptsBuilder interface {
+	ToEvacuateMap() (map[string]interface{}, error)
+}
+
+// EvacuateOpts specifies Evacuate action parameters.
+type EvacuateOpts struct {
+	// The name of the host to which the server is evacuated
+	Host string `json:"host"`
```
jtopjian: According to both the API reference and Nova code, `host` is optional, so I _think_ this should be `json:"host,omitempty"`.

bugosiphionah:
@jtopjian  About adding a test for omniempty fields, I do not understand specifically what you mean. Kindly elaborate.

When I read the [documentation](https://developer.openstack.org/api-ref/compute/#evacuate-server-evacuate-action). I felt like testing this scenario:

`In the request body, if onSharedStorage is set, then do not set adminPass.`

But in the test, when I set both onSharedStorage and adminPass the call actually succeeds . I thought this should return an error from Openstack. This the only other test I could think of. 

Do you have any thoughts?

jtopjian:
@bugosiphionah About adding a test for omniempty fields, I do not understand specifically what you mean. Kindly elaborate.

Sure. What I mean is to add some more unit tests with different combinations of fields. For example, create a test with no options and create a test with only specifying `host`. These types of tests will validate that the JSON body is being build correctly.

> But in the test, when I set both onSharedStorage and adminPass the call actually succeeds . I thought this should return an error from Openstack. This the only other test I could think of.

Did you run an actual acceptance test or was this only a unit test? If it was just a unit test, then the reason you didn't see an error is because the unit tests are only verify the request and response bodies are correct. The unit tests don't actually connect to a running OpenStack environment and so they can't validate logic such as that.

But if you created an actual acceptance for this and it still worked, then it's possible that the API reference is incorrect.

I hope that helps.

bugosiphionah:
Ohh okay fair enough.. got it . I have been running unit tests. I havent written any acceptance test for this action because of the multinode environment.
I will quickly commit those test scenarios you advised.

openstack/compute/v2/extensions/evacuate/requests.go
```diff
@@ -0,0 +1,41 @@
...
+	ToEvacuateMap() (map[string]interface{}, error)
+}
+
+// EvacuateOpts specifies Evacuate action parameters.
+type EvacuateOpts struct {
+	// The name of the host to which the server is evacuated
+	Host string `json:"host,omitempty"`
+
+	// Indicates whether server is on shared storage
+	OnSharedStorage bool `json:"OnSharedStorage"`
```
jtopjian: The full line of this needs to be:

```go
OnSharedStorage bool `json:"onSharedStorage"`
```

Note how the Go field has a capital O but the JSON field is all lower-case.

jtopjian:
@bugosiphionah 

> I have been running unit tests. I havent written any acceptance test for this action because of the multinode environment.

Right. Unit tests are essential because they ensure the requests and responses are being built correctly. But since unit tests cannot emulate the full interaction with a running OpenStack environment, we need to use acceptance/integration tests.

Acceptance tests can be difficult to write because they require access to certain OpenStack environments (like a multinode test for this PR). Because of that, I'm okay with people not including acceptance tests, but the PR won't be merged until I'm able to verify it.

jtopjian:
@bugosiphionah I was able to get a multinode environment working and ran some tests.  The evacuate action had a few surprises, so there are some changes that need to be made.

The first change is simple: "evacuate" responds with a `200` response code, not `202`.

Next, _sometimes_ evacuate responds with a body and sometimes it doesn't. It depends if `nova.conf` has `enable_instance_password` set to true or false. This will vary from cloud to cloud, so we must account for both. 

If a body is sent, the body contains what is noted in the API reference:

```json
{
    "adminPass": "MySecretPass"
}
```

In my opinion, it would have made more sense for Nova to always respond with a body and just set the value to `null` if `enable_instance_password` was disabled... but oh well.

Because the evacuate action can sometimes respond with a body, we can no longer use `gophercloud.ErrResult` as the result type. Handling the evacuate method is a little tricky and I don't think Gophercloud has an existing example to use. To get started, change:

```go
type EvacuateResult struct {
       gophercloud.ErrResult
}                           
```

to

```go
type EvacuateResult struct {
       gophercloud.Result   
}                           
```

You'll then want to implement an `Extract()` method with this signature:

```go
func (r EvacuateResult) Extract() (string, error)
```

That method is the tricky part because you'll want to account for an `EOF` error. I suspect that's why you might have added `EOF` to your unit tests? If so, you can remove that from the unit tests and implement it in the `Extract()` method.

Let me know if you have any questions or if you need any help :)

jtopjian:
@bugosiphionah Actually, a better name for the method is `ExtractAdminPass` rather than `Extract`.

bugosiphionah:
@jtopjian When I run the unit tests while expecting a response of 200 it fails..apparently my requests keep returning 202 for response.

`Expected HTTP response code [200] when accessing [POST http://127.0.0.1:46163/servers/b16ba811-199d-4ffd-8839-ba96c1185a67/action], but got 202 instead`

Could this then mean I am making the request wrongly?

jtopjian:
@bugosiphionah Your unit tests have been configured to respond with a 202 [here](https://github.com/bugosiphionah/gophercloud/blob/d292d65c3207e5d68db063734801e4a6af532be6/openstack/compute/v2/extensions/evacuate/testing/fixtures.go#L25). You need to change this to either `200` or `http.StatusOK`.

bugosiphionah:

openstack/compute/v2/extensions/evacuate/requests.go
```diff
@@ -0,0 +1,51 @@
...
+
+func (r EvacuateResult) ExtractAdminPass() (string, error) {
+	if r.Err != nil && r.Err.Error() != "EOF" {
+		return "", r.Err
+	}
+	if l, ok := r.Header["adminPass"]; ok && len(l) > 0 {
+		return l[0], nil
+	}
+	return "", nil
+}
```
jtopjian: Nice work - this is really close. What you've done is correct, but we want to change it to utilize some core Gophercloud methods that already exist.

See [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/results.go#L138-L144) for a similar method to use an example. You will still want to check for the EOF error in the same way, but first extract `adminPass` like how it's done in the example.

bugosiphionah:
@jtopjian have a look. Thanks for the review:)  I am learning a lot about gophercloud. Future PRs will take shorter reviews as  am getting conversant.

openstack/compute/v2/extensions/evacuate/doc.go
```diff
@@ -0,0 +1,13 @@
+/*
+Package evacuate provides functionality to evacuates servers that have been
+provisioned by the OpenStack Compute service from a failed host to a new host.
+
+Example to Evacuate a Server from a Host
+
+	serverID := "b16ba811-199d-4ffd-8839-ba96c1185a67"
+	err := evacuate.Evacuate(computeClient, serverID, nil).ExtractErr()
```
jtopjian: It looks like `nil` won't work here. Instead, an empty EvacuateOpts needs to be passed in.

openstack/compute/v2/extensions/evacuate/requests.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (r EvacuateResult) ExtractAdminPass() (string, error) {
+	var s struct {
+		AdminPass string `json:"adminPass"`
+	}
+	err := r.ExtractInto(&s)
+	if err != nil && err.Error() != "EOF" {
+		return s.AdminPass, err
+	}
```
jtopjian: I think this should be

```go
if err != nil && err.Error() == "EOF" {
  return "", nil
}
```

This way, an empty admin password is returned if no body is returned.

openstack/compute/v2/extensions/evacuate/requests.go
```diff
@@ -0,0 +1,52 @@
...
+	return s.AdminPass, nil
```
jtopjian: and then this will be

```go
return s.AdminPass, err
```

This way, any error that is not EOF will be returned to the user. At the same time, if AdminPass was successfully parsed, it will be returned and `err` will be nil.

openstack/compute/v2/extensions/evacuate/testing/requests_test.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func TestEvacuateAdminpassResponse(t *testing.T) {
+	const serverID = "b16ba811-199d-4ffd-8839-ba96c1185a67"
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+
+	mockEvacuateAdminpassResponse(t, serverID)
+
+	_, err := evacuate.Evacuate(client.ServiceClient(), serverID, evacuate.EvacuateOpts{}).ExtractAdminPass()
```
jtopjian: For this test, instead of 

```go
_, err := evacuate.Evacuate(client.ServiceClient(), serverID, evacuate.EvacuateOpts{}).ExtractAdminPass()
```

How about

```go
actual, err := evacuate.Evacuate(client.ServiceClient(), serverID, evacuate.EvacuateOpts{}).ExtractAdminPass()
```

and then compare:

```
th.CheckEquals(t, "MySecretPass", actual)
```

This way, we are sure that the admin password is being parsed correctly.

jtopjian:
@bugosiphionah Also, just to clarify about a previous comment I made about giving hints and such: 

This Evacuate feature has been surprisingly difficult, which was not immediately apparent. The combination of not being able to easily test it and having to account for two different response types is uncommon compared other OpenStack features. So the difficulties in this PR are rather unique. 

Great job sticking with this until completion. I figure you'd want to try to solve most of these problems yourself, so I've only been trying to give helpful hints. But if you ever get stuck or frustrated, feel free to ask for help :)

bugosiphionah:

openstack/compute/v2/extensions/evacuate/requests.go
```diff
@@ -0,0 +1,55 @@
...
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Post(actionURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	return
+}
+
+func (r EvacuateResult) ExtractAdminPass() (string, error) {
```
jtopjian: I just noticed that this is in `requests.go`. It needs to be moved to `results.go`.

openstack/compute/v2/extensions/evacuate/results.go
```diff
@@ -0,0 +1,11 @@
+package evacuate
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// EvacuateResult is the response from an Evacuate operation. Call its ExtractErr
+// method to determine if the request suceeded or failed.
```
jtopjian: This needs updated to say "Call its ExtractAdminPass method to retrieve the admin password of the instance. The admin password will be an empty string if the cloud is not configured to inject admin passwords"

And the lines should be wrapped at 80-characters.

bugosiphionah:
@jtopjian please have a look.

jtopjian:
Looks good! Thank you for your work on this :)

Issue #385, opened 2017-06-14
Identity v3: List the users in a group


mfranc213:
This feature would list the users in a group.  I will try to push a PR for this soon.  

This functionality will complement the "list the groups for a user" functionality described in https://github.com/gophercloud/gophercloud/issues/382

jtopjian:
@mfranc213 Are you still interested in implementing this? No problem at all if you don't have the time.

dklyle:
I'm working on this, and I should have a pull request up in the next day or two.

dklyle:
@jtopjian I believe this can be closed.

jtopjian:
Yep, it sure can. Thanks, @dklyle.

PR #534, opened 2017-09-27
Adding Group CRUD support for groups

dklyle:
Adds support for basic operations on groups.

Includes tests for CRUD support.

Also adds the ability to list users in a group.

Did a little refactoring to enabling sharing result definitions from users and groups between the two.

For #385 

code in this PR:
https://developer.openstack.org/api-ref/identity/v3/#groups


jtopjian:
@dklyle Thank you for your work on this. A few notes:

* This PR is out of scope for #385. Not that I'm complaining about having full CRUD support for groups - that would be awesome - but it's a separate feature than what #385 is requesting, so it needs a separate Issue created.
* The different API calls for C, R, U, and D need to be broken out into individual PRs. See https://github.com/gophercloud/gophercloud/issues/287 as a template.
* We also need links to the actual Keystone service's implementation (reference to actual server-side python code). The above template can also be used as an example.

No doubt that most of the code you have here is correct. The above are [Gophercloud-specific procedures](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md) for adding large feature sets. 

The one other topic for discussion is the `userresults` and `groupresults` packages. I understand why you approached it this way, but I think we can avoid it. There was some past discussion about this in https://github.com/gophercloud/gophercloud/pull/386 where we would try to add both the ability to list groups a user belongs to _and_ list users of a group in the `users` package.

If it turns out that approach is still limiting, the approach of adding the functions in the [opposite packages](https://github.com/gophercloud/gophercloud/pull/386#issuecomment-308599066) might be the way to go. In https://github.com/gophercloud/gophercloud/issues/382, I mentioned that I thought it looked weird that Keystone does this, but I now understand why.

Also keeping in mind future support to add and remove users from groups, since those API calls only require a `user_id` string, it should be safe to have this in the `groups` package without the `userresults` and `groupresults` packages.

But I might be incorrect about this - do let me know if I am.

If it turns out `groupresults` and `userresults` packages are required, these will also need to be separate PRs and should probably be done as a prerequisite of adding full group CRUD support.

Again, thank you for the work you've put in to this. Let me know if you have any questions.

dklyle:
jtopjian, thanks for the feedback. I wasn't a fan of the new packages either, and fretted about doing it that way. I'll try stuffing the users per group in the users package.

One note, I did test all the functionality I added, but the API, much like for users doesn't expressly document the full capabilities that keystone supports and that are exposed via gophercloud. I'll try to dig up source to back up things, unless you prefer a stricter following of the API.

Are isolated PRs desired, or a dependent chain for the individual crud components?

Thanks

jtopjian:
> One note, I did test all the functionality I added, but the API, much like for users doesn't expressly document the full capabilities that keystone supports and that are exposed via gophercloud. I'll try to dig up source to back up things, unless you prefer a stricter following of the API.

If the reference API services (ie: vanilla keystone, nova et al) are accepting and responding with certain fields and actions, I consider it to be part of the API, just undocumented. Providing raw API requests and responses are usually good supplements to complex code in the OpenStack services.

If you run into any difficulties with code spelunking, please mention it before going insane :)

> Are isolated PRs desired, or a dependent chain for the individual crud components?

It's really your choice of style.

I'm a fan of the dependency chain. The benefit is that you can have all your work pushed to Github, even if it sits in queue for some time. This shows others that the work is done, just pending, so no one duplicates the effort. The disadvantage is that you have to rebase dependent branches when code review changes are made as well as when the parent branch is merged (I swear there was a time Github handled this automatically... but maybe I was just dreaming). But it's manageable. And usually by the 3rd or 4th PR, there's a rhythm and things go quickly.

jrperritt:
I think if you have organized commits such that we don't have to squash them (namely, if we can just merge it/them), Github doesn't create a new commit and the dependent commits shouldn't have to be rebased. When doing CRUD PRs, I think one commit is usually fine, maybe 2 (1 for operation, 1 for tests).

PR #538, opened 2017-09-28
Adding ability to list users for a group.

dklyle:
Adding ability to list users belonging to a group given the group id.

Although this is logically a group operation, mixing this operation into groups creates a circular dependency between groups and users OR requires adding packages for common components. This approach was deemed the least offensive.

For #385

code in this PR:

groups resource definition in keystone request router: 
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/routers.py#L50-L57

list_group_for_users general implementation 
https://github.com/openstack/keystone/blob/16f6ed14df136eb9c283a5415b9f18fcb0834350/keystone/identity/core.py#L1314

list_group_for_users SQL driver implementation 
https://github.com/openstack/keystone/blob/7754f170aa0eb42bea356e7f912bc241832eb1f3/keystone/identity/backends/sql.py#L320-L330

Based on: https://developer.openstack.org/api-ref/identity/v3/#list-users-in-group

### DIFF
--- a/openstack/identity/v3/users/doc.go
+++ b/openstack/identity/v3/users/doc.go
@@ -98,5 +98,26 @@ Example to List Projects a User Belongs To
 		fmt.Printf("%+v\n", project)
 	}
 
+Example to List Users in a Group
+
+	groupID := "bede500ee1124ae9b0006ff859758b3a"
+	listOpts := users.ListOpts{
+		DomainID: "default",
+	}
+
+	allPages, err := users.ListInGroup(identityClient, groupID, listOpts).AllPages()
+	if err != nil {
+		panic(err)
+	}
+
+	allUsers, err := users.ExtractUsers(allPages)
+	if err != nil {
+		panic(err)
+	}
+
+	for _, user := range allUsers {
+		fmt.Printf("%+v\n", user)
+	}
+
 */
 package users
--- a/openstack/identity/v3/users/requests.go
+++ b/openstack/identity/v3/users/requests.go
@@ -225,3 +225,18 @@ func ListProjects(client *gophercloud.ServiceClient, userID string) pagination.P
 		return projects.ProjectPage{pagination.LinkedPageBase{PageResult: r}}
 	})
 }
+
+// ListInGroup enumerates users that belong to a group.
+func ListInGroup(client *gophercloud.ServiceClient, groupID string, opts ListOptsBuilder) pagination.Pager {
+	url := listInGroupURL(client, groupID)
+	if opts != nil {
+		query, err := opts.ToUserListQuery()
+		if err != nil {
+			return pagination.Pager{Err: err}
+		}
+		url += query
+	}
+	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
+		return UserPage{pagination.LinkedPageBase{PageResult: r}}
+	})
+}
--- a/openstack/identity/v3/users/testing/fixtures.go
+++ b/openstack/identity/v3/users/testing/fixtures.go
@@ -461,3 +461,17 @@ func HandleListUserProjectsSuccessfully(t *testing.T) {
 		fmt.Fprintf(w, ListProjectsOutput)
 	})
 }
+
+// HandleListInGroupSuccessfully creates an HTTP handler at /groups/{groupID}/users
+// on the test handler mux that response with a list of two users
+func HandleListInGroupSuccessfully(t *testing.T) {
+	th.Mux.HandleFunc("/groups/ea167b/users", func(w http.ResponseWriter, r *http.Request) {
+		th.TestMethod(t, r, "GET")
+		th.TestHeader(t, r, "Accept", "application/json")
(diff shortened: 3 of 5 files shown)


jtopjian:
@dklyle The implementation looks really good.

I apologize if I'm bike-shedding on the name, but it's usually best to do this now than later...

In #386 we discussed a few variations on the name. The last one that was mentioned is `users.ListInGroup(client, group_id)`, which seems to be succinct and conveys what the function is doing. However, I admit I might have looked at these names for too long, so I'm open to counter-arguments or other suggestions (which is why I didn't do a formal review for this).

dklyle:
I did not add an acceptance test as it will depend on the list API functionality that has not been added yet. But will add this when list is added.

func TestUsersListInGroup(t *testing.T) {                                       
›   client, err := clients.NewIdentityV3Client()                                
›   if err != nil {                                                             
›   ›   t.Fatalf("Unable to obtain an identity client: %v", err)                
›   }                                                                           
›   allGroupPages, err := groups.List(client, nil).AllPages()                   
›   if err != nil {                                                             
›   ›   t.Fatalf("Unable to list groups: %v", err)                              
›   }                                                                           
                                                                                
›   allGroups, err := groups.ExtractGroups(allGroupPages)                       
›   if err != nil {                                                             
›   ›   t.Fatalf("Unable to extract groups: %v", err)                           
›   }                                                                           
                                                                                
›   group := allGroups[0]                                                       
                                                                                
›   allUserPages, err := users.ListInGroup(client, group.ID).AllPages()         
›   if err != nil {                                                             
›   ›   t.Fatalf("Unable to list users: %v", err)                               
›   }                                                                           
                                                                                
›   allUsers, err := users.ExtractUsers(allUserPages)                           
›   if err != nil {                                                             
›   ›   t.Fatalf("Unable to extract users: %v", err)                            
›   }                                                                           
                                                                                
›   for _, user := range allUsers {                                             
›   ›   tools.PrintResource(t, user)                                            
›   ›   tools.PrintResource(t, user.Extra)                                      
›   }                                                                           
}

jtopjian:
@dklyle Just a heads up: I merged a PR to support listing projects last night, so this will need rebased.

dklyle:
rebased and updated to ListInGroup()

jtopjian:
@dklyle The Identity API never ceases to surprise me :)

Looking at the API docs and the relevant code, it looks like this API call accepts a `password_expires_at` query string, which I'm making an assumption is where the `hints` parameter is coming from:

* https://github.com/openstack/keystone/blob/16f6ed14df136eb9c283a5415b9f18fcb0834350/keystone/identity/core.py#L1323
* https://github.com/openstack/keystone/blob/16f6ed14df136eb9c283a5415b9f18fcb0834350/keystone/identity/core.py#L975

If so, then that's going to modify the function signature with an additional field:

```go
allPages, err := users.ListInGroup(client, groupID, listOpts).AllPages()
// or
allPages, err := users.ListInGroup(client, groupID, nil).AllPages()
```

Because the function signature is different, we should make sure this is correct from the start.

Now, I had a funny feeling that not only is `password_expires_at` a valid query string, but everything in `users.ListOpts` would be valid, too. Sure enough, I think that's true:

```shell
#!/bin/bash

oscurl() {
  token=$(openstack token issue -c id -f value)
  curl -s -H "X-Auth-Token:$token" $1
}

openstack group create foobar

openstack user create --disable bill
openstack group add user foobar bill

openstack user create --enable ted
openstack group add user foobar ted

oscurl http://localhost:5000/v3/groups/<gid>/users | python -mjson.tool

oscurl http://localhost:5000/v3/groups/<gid>/users?domain_id=foo | python -mjson.tool
# no results

oscurl http://localhost:5000/v3/groups/<gid>/users?domain_id=default | python -mjson.tool

oscurl http://localhost:5000/v3/groups/<gid>/users?enabled=true
# ted

oscurl http://localhost:5000/v3/groups/<gid>/users?enabled=false
# bill
```

So I'm pretty sure we can safely re-use `users.ListOpts` as the additional parameter for the `users.ListInGroup` function. In a way, this makes sense and it's pretty clever of the Identity API.

Thoughts?

dklyle:
@jtopjian you are correct, I overlooked the hints. The ListOpts param has been added and works when testing.

jtopjian:
This looks good to me. I just want to do a quick check with @jrperritt to see if there's any issue (architecturally) with re-using `users.ListOpts` here.

@jrperritt tl;dr: Listing users within a group can take the same query parameters as listing users in general. Since both actions return the same type of result (users) and both support the same query parameters, IMO, it makes sense to just use the same list options. Thoughts?

jrperritt:
> to see if there's any issue (architecturally) with re-using users.ListOpts here

The only thing I can think of where that _could_ become a problem is if the options diverge. That's probably unlikely though. I think this should be fine as-is.

jtopjian:
@jrperritt Sounds good - thanks!

@dklyle Thank you for putting this together and for being patient with the changes. All looks good to me!

Issue #521, opened 2017-09-15
Compute v2: Add Reset State Action


jtopjian:
Nova supports the ability to reset an instance's state to either `active` or `error`:

* https://developer.openstack.org/api-ref/compute/#reset-server-state-os-resetstate-action

This action should look similar to the [migrate](https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions/migrate) action (but I might be wrong).

Additional work required: Locate where the reset state action is implemented in the Nova API service.

Bonus: Implement a reset state [acceptance test](https://github.com/gophercloud/gophercloud/tree/master/acceptance/openstack/compute/v2) and verify it works.

ss22ever:
@jtopjian I would like to work on this

jrperritt:
@ss22ever Great, it's yours

ss22ever:
@jtopjian @jrperritt I tried creating the reset state action and when I test it shows me 
`go test .
--- FAIL: TestResetState (0.01s)
        convenience.go:300: Expected JSON:
                {
                  "os-resetState": null
                }
        convenience.go:308: Actual JSON:
                {
                  "os-resetState": {
                    "state": "active"
                  }
                }
        convenience.go:39: Failure in http_responses.go, line 90: The generated JSON structure differed.
FAIL
FAIL    github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/resetstate/testing   0.191s
`
Even when it follows the example Reset Server State (os-resetState Action)
`{
    "os-resetState": {
        "state": "active"
    }
}`

jtopjian:
Implemented by #545

PR #545, opened 2017-10-06
Compute v2: Add Reset State Action

ss22ever:
For #521

Links to the line numbers/files in the OpenStack source code (Nova API service) that support the
code in this PR:
https://github.com/openstack/nova/blob/c2aa30b102808882c85d3d3f53d531c4510218cd/nova/api/openstack/compute/admin_actions.py#L74
https://github.com/openstack/nova/blob/c2aa30b102808882c85d3d3f53d531c4510218cd/nova/api/openstack/compute/admin_actions.py#L29-L32

### DIFF
--- a/openstack/compute/v2/extensions/resetstate/doc.go
+++ b/openstack/compute/v2/extensions/resetstate/doc.go
@@ -1,0 +1,13 @@
+/*
+Package resetstate provides functionality to reset the state of a server that has
+been provisioned by the OpenStack Compute service.
+
+Example to Reset a Server
+
+	serverID := "47b6b7b7-568d-40e4-868c-d5c41735532e"
+	err := resetstate.ResetState(client, id, resetstate.StateActive).ExtractErr()
+	if err != nil {
+		panic(err)
+	}
+*/
+package resetstate
--- a/openstack/compute/v2/extensions/resetstate/requests.go
+++ b/openstack/compute/v2/extensions/resetstate/requests.go
@@ -1,0 +1,23 @@
+package resetstate
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// ServerState refers to the states usable in ResetState Action
+type ServerState string
+
+const (
+	// StateActive returns the state of the server as active
+	StateActive ServerState = "active"
+
+	// StateError returns the state of the server as error
+	StateError ServerState = "error"
+)
+
+// ResetState will reset the state of a server
+func ResetState(client *gophercloud.ServiceClient, id string, state ServerState) (r ResetResult) {
+	stateMap := map[string]interface{}{"state": state}
+	_, r.Err = client.Post(actionURL(client, id), map[string]interface{}{"os-resetState": stateMap}, nil, nil)
+	return
+}
--- a/openstack/compute/v2/extensions/resetstate/results.go
+++ b/openstack/compute/v2/extensions/resetstate/results.go
@@ -1,0 +1,11 @@
+package resetstate
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// ResetResult is the response of a ResetState operation. Call its ExtractErr
+// method to determine if the request suceeded or failed.
+type ResetResult struct {
+	gophercloud.ErrResult
+}
--- a/openstack/compute/v2/extensions/resetstate/testing/doc.go
+++ b/openstack/compute/v2/extensions/resetstate/testing/doc.go
@@ -1,0 +1,1 @@
+package testing
--- a/openstack/compute/v2/extensions/resetstate/testing/fixtures.go
+++ b/openstack/compute/v2/extensions/resetstate/testing/fixtures.go
@@ -1,0 +1,19 @@
+package testing
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	th "github.com/gophercloud/gophercloud/testhelper"
+	"github.com/gophercloud/gophercloud/testhelper/client"
+)
+
+func mockResetStateResponse(t *testing.T, id string, state string) {
+	th.Mux.HandleFunc("/servers/"+id+"/action", func(w http.ResponseWriter, r *http.Request) {
+		th.TestMethod(t, r, "POST")
+		th.TestHeader(t, r, "X-Auth-Token", client.TokenID)
+		th.TestJSONRequest(t, r, fmt.Sprintf(`{"os-resetState": {"state": "%s"}}`, state))
+		w.WriteHeader(http.StatusAccepted)
+	})
+}
--- a/openstack/compute/v2/extensions/resetstate/testing/requests_test.go
+++ b/openstack/compute/v2/extensions/resetstate/testing/requests_test.go
@@ -1,0 +1,21 @@
+package testing
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/resetstate"
+	th "github.com/gophercloud/gophercloud/testhelper"
+	"github.com/gophercloud/gophercloud/testhelper/client"
+)
+
+const serverID = "b16ba811-199d-4ffd-8839-ba96c1185a67"
+
+func TestResetState(t *testing.T) {
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+
+	mockResetStateResponse(t, serverID, "active")
+
+	err := resetstate.ResetState(client.ServiceClient(), serverID, "active").ExtractErr()
+	th.AssertNoErr(t, err)
+}
--- a/openstack/compute/v2/extensions/resetstate/urls.go
+++ b/openstack/compute/v2/extensions/resetstate/urls.go
@@ -1,0 +1,9 @@
+package resetstate
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+func actionURL(client *gophercloud.ServiceClient, id string) string {
+	return client.ServiceURL("servers", id, "action")
+}


jtopjian:
@ss22ever This is a really good start - nice work. :)

I'll go through a few things in detail, which might help in general when working with Gophercloud.

The error you're getting when testing is due to Gophercloud building a request body in one format and the test expecting another. According to the API docs, the request body should look like:

```json
{
    "os-resetState": {
        "state": "active"
    }
}
```

But let's double-check that just to make sure. There are times when the API reference can be incorrect. If you have access to a devstack environment, you can do something like:

```shell
$ nova boot --flavor <id> --image <image name> test
$ nova show test # wait until it's ACTIVE
$ nova --debug reset-state --active <uuid> # must be done as the admin user
```

In that output, you'll see:

```
DEBUG (session:347) REQ: curl -g -i -X POST http://127.0.0.1:8774/v2.1/b01f4fa8661d4ffb8f7f4466cf3c3b5c/servers/2ed3d31e-e5ce-4003-86b7-62e2342bbb8a/action -H "Accept: application/json" -H "User-Agent: python-novaclient" -H "OpenStack-API-Version: compute 2.41" -H "X-OpenStack-Nova-API-Version: 2.41" -H "X-Auth-Token: {SHA1}b6df3d55bd31deaa649383c9d2510130694fab0f" -H "Content-Type: application/json" -d '{"os-resetState": {"state": "active"}}'
```

And if we clean it up:

```json
{"os-resetState": {"state": "active"}}
```

That confirms the API reference is correct.

Looking at your code, you're building the body correctly [here](https://github.com/ss22ever/gophercloud/blob/ecebc7e705b9f5d5dae36e5022a1bcca7fbcbfb4/openstack/compute/v2/extensions/resetstate/requests.go#L16) (And the test is reporting as such), but your [test](https://github.com/ss22ever/gophercloud/blob/ecebc7e705b9f5d5dae36e5022a1bcca7fbcbfb4/openstack/compute/v2/extensions/resetstate/testing/fixtures.go#L15) is looking for this:

```json
{"os-resetState": null}
```

So you'll want to fix that. Once it's fixed, your test will pass.

But I'll jump to the next part of the review.

Your [ResetState function](https://github.com/ss22ever/gophercloud/blob/ecebc7e705b9f5d5dae36e5022a1bcca7fbcbfb4/openstack/compute/v2/extensions/resetstate/requests.go#L10-L18) is taking a string as an input parameter and it's also checking if that string matches either `active` or `error`. This shows that this function can only accept a finite amount of values.

In a situation where there is only a finite amount of valid input, it's better to create a custom type and constants. There are several examples of this throughout Gophercloud. [Here's](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/flavors/requests.go#L22-L35) the first one I could think of. If you'd like, I'll let you try to implement this pattern in the `resetstate` package, but let me know if you need help or a hint.

One more note:

You gave the correct link to the Nova code which implements this. Technically we also need to prove that `active` and `error` are really the only two acceptable choices, which is defined above the line number you gave:

https://github.com/openstack/nova/blob/c2aa30b102808882c85d3d3f53d531c4510218cd/nova/api/openstack/compute/admin_actions.py#L29-L32

This sounds trivial, but it's important in this case. What if the API reference was wrong and _any_ state could be passed? Then we'd have to account for and allow a user to pass any state, and we'd want to define all of those states up front. But in this case, `active` and `error` are the only two we need to be concerned about.

When adding features to Gophercloud, most of the work is being a detective and proving the implementation is correct. We're very strict on that, but it pays off in the long run :)

I hope that helps and let me know if you have any questions.

ss22ever:
@jrperritt Thanks :), fixed the ResetState function, added the line number link that you mentioned, now trying to implement custom type and constants.

ss22ever:
@jtopjian have a look!

openstack/compute/v2/extensions/resetstate/doc.go
```diff
@@ -0,0 +1 @@
+package resetstate
```
jtopjian: Would you be able to add some extra documentation here? See [this](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/startstop/doc.go) as an example.

openstack/compute/v2/extensions/resetstate/requests.go
```diff
@@ -0,0 +1,30 @@
...
+	"github.com/gophercloud/gophercloud"
+)
+
+//ResetStatus refers to the states usable in ResetState Action
+
+type ResetStatus string
+
+const (
+	//ResetStateActive returns the state of the server as active
+	ResetStateActive ResetStatus = "active"
```
jtopjian: How about `StateActive` and `StateError` for the const names? This flows a little better when someone is using it:

```go
err := resetstate.ResetState(client, id, resetstate.StateActive).ExtractErr()
```

openstack/compute/v2/extensions/resetstate/requests.go
```diff
@@ -0,0 +1,30 @@
...
+
+	//ResetStateError returns the state o the server as error
+	ResetStateError ResetStatus = "error"
+)
+
+// ResetState will initiate to reset the state of a server
+func ResetState(client *gophercloud.ServiceClient, id string, state ResetStatus) (r ResetResult) {
+	if state != ResetStateActive && state != ResetStateError {
```
jtopjian: We can remove this `if` block. There's no way someone can pass any other values besides `StateActive` and `StateError` now, so we don't need to check for an error :)

openstack/compute/v2/extensions/resetstate/requests.go
```diff
@@ -0,0 +1,30 @@
+package resetstate
+
+import (
+	"errors"
+
+	"github.com/gophercloud/gophercloud"
+)
+
+//ResetStatus refers to the states usable in ResetState Action
```
jtopjian: In order for `godoc` to work properly, there can't be a blank line between this and `type ResetStatus string`. You can verify this yourself if you do:

```shell
$ cd openstack/compute/v2/extensions/resetstate
$ godoc ./ | less
```

Also, please use a space between `//` and `ResetStatus refers to...`

openstack/compute/v2/extensions/resetstate/doc.go
ss22ever: @jtopjian Added some extra documentation.

openstack/compute/v2/extensions/resetstate/doc.go
```diff
@@ -0,0 +1,11 @@
+/*
+Package resetstate provides functionality to reset the state of a server that have
+been provisioned by the OpenStack Compute service.
+Example to reset a Server
```
jtopjian: "reset" should be capitalized as Reset. 

There should also be a newline above and below this line.

openstack/compute/v2/extensions/resetstate/doc.go
```diff
@@ -0,0 +1,11 @@
+/*
+Package resetstate provides functionality to reset the state of a server that have
```
jtopjian: Should be "has" instead of "have".

openstack/compute/v2/extensions/resetstate/requests.go
```diff
@@ -0,0 +1,23 @@
+package resetstate
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// ResetStatus refers to the states usable in ResetState Action
+type ResetStatus string
```
jtopjian: I apologize - I didn't catch this before. I think having `Status` and `State` might be confusing. We can't use `ResetState` because that's the name of the function.

Using `State` by itself is ambiguous (the state of _what_?) So in that case, how about instead of `ResetStatus`, use `ServerState`. This removes the word "status" and gives more detail to what the status is referring to.

`StateActive` and `StateError` are both correct - just their type will now be `ServerState`.

ss22ever:
@jtopjian fixed small changes :).Currently, would love to see this get merged will work on the acceptance test for other issues that I have on my plate.

openstack/compute/v2/extensions/resetstate/requests.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+// ServerState refers to the states usable in ResetState Action
+type ServerState string
+
+const (
+	// StateActive returns the state of the server as active
+	StateActive ServerState = "active"
+
+	// StateError returns the state o the server as error
```
jamiehannaford: small nit: `of` not `o`

openstack/compute/v2/extensions/resetstate/requests.go
```diff
@@ -0,0 +1,23 @@
...
+	StateError ServerState = "error"
+)
+
+// ResetState will initiate to reset the state of a server
```
jamiehannaford: `ResetState will reset the state of a server`

openstack/compute/v2/extensions/resetstate/results.go
```diff
@@ -0,0 +1,11 @@
+package resetstate
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// ResetResult is the response from a ResetState operation. Call its ExtractErr
```
jamiehannaford: is the response of a

ss22ever: @jamiehannaford Thanks for review fixed super minor nits.

jtopjian:
@jamiehannaford Thanks for catching those.

@ss22ever This looks good! Thank you for your work and patience on this.

jamiehannaford:
Awesome! Thanks for the great reviewing @jtopjian 🚀

Issue #539, opened 2017-09-28
Identity v3: Groups CRUD


dklyle:
- [x] Add List / Get (#546) 
- [x] Add Create (#548)
- [x] Add Delete (#552)
- [x] Add Update (#553)

General API reference: https://developer.openstack.org/api-ref/identity/v3/index.html#groups

API processing:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L302
https://github.com/openstack/keystone/blob/master/keystone/identity/backends/sql.py

Schema:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/schema.py#L100-L104
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/backends/sql_model.py#L376

List:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L327

Get:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L340

Create:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L316
https://github.com/openstack/keystone/blob/master/keystone/identity/backends/sql.py#L346
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/schema.py#L106

Delete:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L354
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/backends/sql.py#L394

Update:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L345
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/backends/sql.py#L381
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/schema.py#L113

dklyle:
I have to refactor my previous pull request https://github.com/gophercloud/gophercloud/pull/534 into separate PRs per operation, but most of this is already complete. Will post shortly.

dklyle:
Sorry for the github butchery. Dependent branches were a mistake. I will wait for #546 to merge before further pull requests on this.

jtopjian:
:tada:

PR #546, opened 2017-10-06
Identity v3 Groups Get / List

dklyle:

For #539

code in this PR:

List:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L327

Get:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L340

### DIFF
--- a/acceptance/openstack/identity/v3/groups_test.go
+++ b/acceptance/openstack/identity/v3/groups_test.go
@@ -1,0 +1,62 @@
+// +build acceptance
+
+package v3
+
+import (
+	"testing"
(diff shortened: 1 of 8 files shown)


jtopjian:
@dklyle This is spot-on. Nice work.

One nit is the `TestGroupsGet` acceptance test: it will fail if there isn't at least one group to choose from. I'm OK with that for now, but when you add support for `Create`, can you either amend that test or remove the test in favor of an in-progress CRUD test?

PR #548, opened 2017-10-09
Identity v3: Add Group create

dklyle:
Changing group acceptance tests to a CRUD based test now that
create exists. This also reduces the risks of errant test failures
as group existence is not required in an openstack v3 deployment. The
Update and Delete will be added as the functionality is merged.

For #539 

code in this PR:

Create:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L316
https://github.com/openstack/keystone/blob/master/keystone/identity/backends/sql.py#L346
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/schema.py#L106

### DIFF
--- a/acceptance/openstack/identity/v3/groups_test.go
+++ b/acceptance/openstack/identity/v3/groups_test.go
@@ -10,16 +10,33 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/groups"
 )
 
-func TestGroupsList(t *testing.T) {
+func TestGroupCRUD(t *testing.T) {
 	client, err := clients.NewIdentityV3Client()
 	if err != nil {
 		t.Fatalf("Unable to obtain an identity client: %v", err)
 	}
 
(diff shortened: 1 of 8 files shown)


jtopjian:
@dklyle This looks good!

Just one comment on the acceptance test: it'd be better to make a generic `CreateGroup` (and later `DeleteGroup`) function here:

https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/identity/v3/identity.go

This allows other acceptance testing packages to import this package and easily create/delete groups. You can use the `CreateProject` or `CreateUser` functions as a template (random name and such).

Each of the acceptance test categories have similar files. For `compute/v2`, there's a `compute.go` file, for `networking/v2`, there's a `networking.go` file, etc.

dklyle:
Updated to extract a common CreateGroup for acceptance tests.

jtopjian:
@dklyle Looks good! Thank you :)

PR #552, opened 2017-10-10
Identity v3: group delete

dklyle:
For #539

code in this PR:

Delete:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L354
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/backends/sql.py#L394

### DIFF
--- a/openstack/identity/v3/groups/results.go
+++ b/openstack/identity/v3/groups/results.go
@@ -75,6 +75,12 @@ type CreateResult struct {
 	groupResult
 }
 
+// DeleteResult is the response from a Delete operation. Call its ExtractErr to
+// determine if the request succeeded or failed.
+type DeleteResult struct {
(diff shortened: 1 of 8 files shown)


openstack/identity/v3/groups/results.go
```diff
@@ -75,6 +75,12 @@ type CreateResult struct {
 	groupResult
 }
 
+// DeleteResult is the response from a Delete operation. Call its ExtractErr
+// method to interpret it as a Group.
```
jtopjian: Should be "Call its ExtractErr method to determine if the result succeeded or failed" or something along those lines.

dklyle:
updated the comment

jtopjian:
All good. Thank you!

PR #553, opened 2017-10-11
Identity v3: group update

dklyle:
For #539 

code in this PR:

Update:

https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/controllers.py#L345
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/backends/sql.py#L381
https://github.com/openstack/keystone/blob/stable/pike/keystone/identity/schema.py#L113


jtopjian:
Looks good! Nice work - thank you for putting this together.

Issue #543, opened 2017-10-03
acceptancetest_environments/*.sh setup scripts no longer work


khdegraaf:
scripts like keystonev3-lbaasv2.sh and keystonev2-lbaasv1.sh no longer work as they refer to mitaka branch in 

git clone https://git.openstack.org/openstack-dev/devstack -b stable/mitaka

but that branch is EOL and no longer there.  Is there an official environment setup script that works?  Just trying to sort out the fine details of getting the acceptance tests running 100%.  I tried it with stable/pike, but the script doesn't work there due some configuration changes, so I am trying older branches...

jtopjian:
@khdegraaf I was the original creator of those scripts and it's something I should really clean up. You're right - those scripts are _way_ out of date. I'll get that sorted out in the next few days.

Coincidentally, Gophercloud is currently tested on the exact same setup as Terraform. 

> I tried it with stable/pike, but the script doesn't work there due some configuration changes, so I am trying older branches

I ran into this the other day and pretended I didn't see it. I need to set aside a good full afternoon and/or evening to go through Pike and see what works and what has broke.

Everything should work in Ocata.

If you want to use packstack/RDO, to get Ocata installed, do the following:

```shell
yum install -y https://repos.fedorapeople.org/repos/openstack/openstack-ocata/rdo-release-ocata-3.noarch.rpm
yum install -y centos-release-openstack-ocata
yum-config-manager --enable openstack-ocata
yum update -y
yum install -y openstack-packstack
packstack --allinone
```

jtopjian:
@khdegraaf Also, if you're interested, I have begun documenting how the Gophercloud and Terraform test environments are created:

http://terrarum.net/blog/building-openstack-environments.html

That details deploying to AWS, but as I mentioned in the Terraform OpenStack repo, any isolated virtual machine should work.

The blog post covers how to easily set all of the required environment variables and such.

khdegraaf:
Thanks, that is really helpful.  I was going to ask you what if anything you do with the network (following https://www.rdoproject.org/install/packstack/), but I see you answer that in your blog post.  I'll try it out.

jtopjian:
@khdegraaf Thanks for reporting this. #554 removed the obsolete scripts and added some documentation on building environments. Let me know if you have any questions.

PR #554, opened 2017-10-12
Acceptance Test Doc Updates

jtopjian:
For #543


jtopjian:
@jrperritt Ping for a quick review

Issue #555, opened 2017-10-12
Identity v3 domains CRUD


dklyle:
- [x] List / Get (#556)
- [x] Create
- [x] Delete
- [x] Update

List:
 https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L140

Get:

- General project model/schema: 
  https://github.com/openstack/keystone/blob/master/keystone/resource/schema.py#L51

- Controller / router: 
 https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L148

Create:

- https://github.com/openstack/keystone/blob/master/keystone/resource/schema.py#L57
- https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L132

Update:

- https://github.com/openstack/keystone/blob/master/keystone/resource/core.py#L733
-
 https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L153

Delete:

- https://github.com/openstack/keystone/blob/master/keystone/resource/core.py#L751
- https://github.com/openstack/keystone/blob/master/keystone/resource/core.py#L780

- Controller / router:
https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L162

jtopjian:
@dklyle Thank you for taking this on :)

I've never worked outside of the Default domain, but the API seems simple and straightforward. My understanding is that Projects can now also act as a domain if their `is_domain` attribute is set. Out of curiosity, are Projects now the recommended way of creating domains? If so, that doesn't lessen the importance of adding this feature - just asking more for my curiosity.

dklyle:
That certainly is the direction Keystone has taken. And there is a partial implementation of hierarchical projects beyond that which is more reliant on the projects API. In my opinion, interacting with the domains API rather than the projects API has been more straightforward as the is_domain flag on projects was added later, but I have been using the domains API since before the great project/domain unification, so I may just be mentally stuck.

jtopjian:
:tada:

PR #556, opened 2017-10-13
Identity v3 domain list / get

dklyle:
For #555 

code in this PR:

List:

https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L123

Get:

General project model/schema:
https://github.com/openstack/keystone/blob/master/keystone/resource/schema.py#L51

Controller / router:
https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L189

### DIFF
--- a/acceptance/openstack/identity/v3/domains_test.go
+++ b/acceptance/openstack/identity/v3/domains_test.go
@@ -1,0 +1,62 @@
+// +build acceptance
+
+package v3
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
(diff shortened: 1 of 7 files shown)


jtopjian:
@dklyle Looks good!

Just a few notes for posterity:

The ListOpts are defined here: https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L140

And the Get function/controller/route is actually here: https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L148

PR #559, opened 2017-10-13
Identity v3: Domain update

dklyle:
For #555

code in this PR:

Update:

https://github.com/openstack/keystone/blob/master/keystone/resource/core.py#L733
https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/resource/controllers.py#L153


dklyle:
requested updates made.

Sadly I ran into the same issue in manual testing but didn't think back to the acceptance test setup.

The if problem was just me being sloppy.

jtopjian:
Looks good - nice work!

PR #560, opened 2017-10-14
fix user delete result comment

dklyle:
trivial fix to update inaccurate comment


jtopjian:
Good catch - thank you.

Issue #550, opened 2017-10-10
Add ListRoles function to Keystone v3


jamiehannaford:
Currently we don't support the [list roles](https://developer.openstack.org/api-ref/identity/v3/#list-roles) function. Ideally a PR would implement the full CRUD set for Roles, but list would be a good first step.

dklyle:
I can take this.

dklyle:
Looking at the role content that exists around role assignments. I will likely want to refactor the exiting RoleAssignment struct as it includes structs for Role, User, and Group which are defined in role/results.go As there are structs defined elsewhere that match these already and I will need to define a Role struct for the context of listing roles that has more information. Just want to make sure that such a internal plumbing change is OK before moving forward. Thoughts?

jtopjian:
Quite coincidentally, I'm working with the RoleAssignment API in Python at this moment.

From what I can see, the information contained in the result of a RoleAssignment entry is only a subset of a full domain, user, group:

```
<RoleAssignment 
  role={u'id': u'bdfee2f5869f43508a4f881164d04c16', u'name': u'Member'}, 
  scope={u'project': {u'domain': {u'id': u'default', u'name': u'Default'}, u'id': u'2f77533e2d6148189da24a3cbe707859', u'name': u'jdoe@gmail.com'}}, 
  user={u'domain': {u'id': u'default', u'name': u'Default'},  u'id': u'24cc044d675f45098afab4a0e3bb5b77', u'name': foobar'}>
```

This also seems to match the API docs: https://developer.openstack.org/api-ref/identity/v3/#id271

I haven't done a raw curl command to verify, though.

Since the `role_assignment` API is only returning a subset of information, I'm leaning on leaving the structs as they are. I get that they are _kind of_ duplicate, but also kind of not. I see them as two different types of result. 

If we were to swap these structs out with the more robust ones in other packages, IMO I think it would be more confusing to a end-user/dev to have a `projects.Project` result with half of the fields empty.

Unless I totally misunderstood what you were proposing :/

dklyle:
My biggest concern is that there is a role struct that is defined as just an ID. A true Role struct with name, domain_id, ID, etc is more correct. I would prefer to at a minimum rename the existing Role struct to something else or I'll have to invent a less natural name to match what I would consider the full Role struct. The others (user, group) would just be potentially confusing.

jtopjian:
[...]

PR #561, opened 2017-10-16
Identity v3 role list / get

dklyle:
For #550 

code in this PR:

List:

https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L335

Get:
https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L349

Renamed Role struct used in RoleAssignment to AssignedRole as the scope of the Role field in RoleAssignment is a subset of the full v3 role API result. 

Ideally, I would move the previous role_assignment code to a role_assignment package, but as that code has existed for over 2 years, that could break existing users/implementations.

I can also see refactoring the role_assignment tests to match the rest of the identity v3 test structure, but that should be an independent PR.

### DIFF
--- a/acceptance/openstack/identity/v3/roles_test.go
+++ b/acceptance/openstack/identity/v3/roles_test.go
@@ -1,0 +1,61 @@
+// +build acceptance
+
+package v3
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/roles"
+)
+
+func TestRolesList(t *testing.T) {
+	client, err := clients.NewIdentityV3Client()
+	if err != nil {
+		t.Fatalf("Unable to obtain an identity client: %v", err)
+	}
+
+	listOpts := roles.ListOpts{
+		DomainID: "default",
+	}
+
+	allPages, err := roles.List(client, listOpts).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to list roles: %v", err)
+	}
+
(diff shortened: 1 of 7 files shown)


jtopjian:
@dklyle Thank you for your continued work fleshing out the Identity v3 features :)

A few notes:

`ListOpts` are defined here: https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L334

The [role schema](https://github.com/openstack/keystone/blob/stable/pike/keystone/assignment/schema.py#L32-L34) is a little strange in that it's only defining a `name`. The [SQL schema](https://github.com/openstack/keystone/blob/stable/pike/keystone/assignment/role_backends/sql.py#L196) has the complete list. My guess is that role v3 clones/inherits from role v2 at some point.

It looks like roles can support the [`Extra`](https://github.com/openstack/keystone/blob/stable/pike/keystone/assignment/role_backends/sql.py#L201) field. Adding this later is completely fine (even by another person down the road). Given the `openstack` cli tool doesn't seem to be able to add arbitrary extra info, it's tough to test this without adding `Create` support.

All in all, this looks good to me. Just going to ping @jrperritt on the rename of `Role` to `AssignedRole` and @jamiehannaford to see if he has any thoughts since he opened the issue.

jtopjian:
@jrperritt Quick ping to see if you have any objections to the rename being done here.

jrperritt:
> Quick ping to see if you have any objections to the rename being done here.

No, I think that's fine

jtopjian:
@jrperritt Thanks!

@dklyle And thank you again!

Issue #549, opened 2017-10-09
Compute v2: Add detailed quota reporting


nate-johnston:
Nova supports the ability to get not just the current quota for a tenant, but also the current allocation against the quota [1] through a "detail" endpoint.  Please enable this collection.

[1] https://developer.openstack.org/api-ref/compute/#show-the-detail-of-quota

This action would look similar to the existing quota fetch action, with a more complex response set.

nate-johnston:
I am waiting for approval from a core contributor before creating a PR.

jtopjian:
@nate-johnston You're more than welcome to open a PR for this.

jtopjian:
Implemented in #564

PR #564, opened 2017-10-19
Compute v2 - Add Get Detailed Quota for Tenant (#549)

nate-johnston:
* Add GetQuotaDetailed function for the os-quota-sets extension in Gophercloud's OpenStack Compute v2 API

- Allows a caller to get the list of quota types with in_use, reserved, and limit values for each
- Effectively wraps OpenStack Compute v2 API function /v2/os-quota-sets/{tenant_id}/detail
- Based on implementation of GetQuota

Note: In line with "GetQuota" uses sample response body from https://developer.openstack.org/api-ref/compute/#show-the-detail-of-quota for test

Code citation: https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/quota_sets.py#L119

For #549

### DIFF
--- a/openstack/compute/v2/extensions/quotasets/requests.go
+++ b/openstack/compute/v2/extensions/quotasets/requests.go
@@ -9,14 +9,14 @@ import (
 // Get returns public data about a previously created QuotaSet.
 func Get(client *gophercloud.ServiceClient, tenantID string) GetResult {
 	var res GetResult
-	_, res.Err = client.Get(getDetailURL(client, tenantID), &res.Body, nil)
+	_, res.Err = client.Get(getURL(client, tenantID), &res.Body, nil)
 	return res
 }
 
 // GetDetail returns detailed public data about a previously created QuotaSet.
 func GetDetail(client *gophercloud.ServiceClient, tenantID string) GetDetailResult {
 	var res GetDetailResult
-	_, res.Err = client.Get(getDetailURL(client, fmt.Sprintf("%s/detail", tenantID)), &res.Body, nil)
+	_, res.Err = client.Get(getDetailURL(client, tenantID), &res.Body, nil)
 	return res
 }
 
--- a/openstack/compute/v2/extensions/quotasets/requests.go
+++ b/openstack/compute/v2/extensions/quotasets/requests.go
@@ -1,6 +1,8 @@
 package quotasets
 
 import (
+	"fmt"
+
 	"github.com/gophercloud/gophercloud"
 )
 
@@ -8,6 +10,13 @@ import (
 func Get(client *gophercloud.ServiceClient, tenantID string) GetResult {
 	var res GetResult
 	_, res.Err = client.Get(getURL(client, tenantID), &res.Body, nil)
+	return res
+}
+
+// GetDetail returns detailed public data about a previously created QuotaSet.
+func GetDetail(client *gophercloud.ServiceClient, tenantID string) GetResultDetail {
+	var res GetResultDetail
+	_, res.Err = client.Get(getURL(client, fmt.Sprintf("%s/detail", tenantID)), &res.Body, nil)
 	return res
 }
 
--- a/openstack/compute/v2/extensions/quotasets/requests.go
+++ b/openstack/compute/v2/extensions/quotasets/requests.go
@@ -9,14 +9,14 @@ import (
 // Get returns public data about a previously created QuotaSet.
 func Get(client *gophercloud.ServiceClient, tenantID string) GetResult {
 	var res GetResult
-	_, res.Err = client.Get(getURL(client, tenantID), &res.Body, nil)
+	_, res.Err = client.Get(getDetailURL(client, tenantID), &res.Body, nil)
 	return res
 }
 
 // GetDetail returns detailed public data about a previously created QuotaSet.
-func GetDetail(client *gophercloud.ServiceClient, tenantID string) GetResultDetail {
-	var res GetResultDetail
-	_, res.Err = client.Get(getURL(client, fmt.Sprintf("%s/detail", tenantID)), &res.Body, nil)
+func GetDetail(client *gophercloud.ServiceClient, tenantID string) GetDetailResult {
+	var res GetDetailResult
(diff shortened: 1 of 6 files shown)


openstack/compute/v2/extensions/quotasets/requests.go
```diff
@@ -11,6 +13,13 @@ func Get(client *gophercloud.ServiceClient, tenantID string) GetResult {
 	return res
 }
 
+// GetDetail returns detailed public data about a previously created QuotaSet.
+func GetDetail(client *gophercloud.ServiceClient, tenantID string) GetResultDetail {
+	var res GetResultDetail
+	_, res.Err = client.Get(getURL(client, fmt.Sprintf("%s/detail", tenantID)), &res.Body, nil)
```
jtopjian: I think this would look cleaner if there was a url (`getDetailURL`) added to `urls.go`

openstack/compute/v2/extensions/quotasets/results.go
```diff
@@ -55,6 +55,68 @@ type QuotaSet struct {
...
+	Instances QuotaDetail `json:"instances"`
+
+	// ServerGroups is the number of ServerGroups allowed for the project.
+	ServerGroups QuotaDetail `json:"server_groups"`
+
+	// ServerGroupMembers is the number of members for each ServerGroup.
+	ServerGroupMembers QuotaDetail `json:"server_group_members"`
+}
+
+// QuotaDetail is a set of details about a single operational limit that allows for control of compute usage.
```
jtopjian: nit: can you wrap comments to 80 characters?

openstack/compute/v2/extensions/quotasets/results.go
```diff
@@ -106,3 +168,23 @@ type UpdateResult struct {
 type DeleteResult struct {
 	quotaResult
 }
+
+type quotaResultDetail struct {
```
jtopjian: `quotaDetailResult` might be better here.

openstack/compute/v2/extensions/quotasets/results.go
```diff
@@ -106,3 +168,23 @@ type UpdateResult struct {
...
+	gophercloud.Result
+}
+
+// GetResult is the response from a Get operation. Call its Extract method to interpret it
+// as a QuotaSet.
+type GetResultDetail struct {
```
jtopjian: `GetDetailResult` might be better. Also the godoc comment two lines above needs updated as it still has `GetResult`.

openstack/compute/v2/extensions/quotasets/testing/fixtures.go
```diff
@@ -2,13 +2,12 @@ package testing
...
-	"net/http"
-	"testing"
-
+	"github.com/Comcast/gophercloud/openstack/compute/v2/extensions/quotasets"
 	"github.com/gophercloud/gophercloud"
-	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/quotasets"
 	th "github.com/gophercloud/gophercloud/testhelper"
 	"github.com/gophercloud/gophercloud/testhelper/client"
+	"net/http"
+	"testing"
```
jtopjian: Looks like this needs cleaned up: `Comcast` needs to be replaced with `gophercloud` (which is the cause of Travis failing). As well, the imports need to be re-arranged to the original ordering: std packages first, blank line, then github packages.

openstack/compute/v2/extensions/quotasets/testing/requests_test.go
```diff
@@ -4,8 +4,8 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/Comcast/gophercloud/openstack/compute/v2/extensions/quotasets"
```
jtopjian: `s/Comcast/gophercloud/`

openstack/compute/v2/extensions/quotasets/results.go
```diff
@@ -55,6 +55,68 @@ type QuotaSet struct {
 	ServerGroupMembers int `json:"server_group_members"`
 }
 
+// QuotaDetailSet is detail on both operational limits that allow for control
+// of compute usage and also currently measured allocation against those limits.
```
jtopjian: perhaps: "QuotaDetailSet represents details of both operational limits of compute resources and the current usage of those resources"

openstack/compute/v2/extensions/quotasets/results.go
```diff
@@ -55,6 +55,68 @@ type QuotaSet struct {
...
+type QuotaDetailSet struct {
+	// ID is tenant associated with this QuotaDetailSet.
```
jtopjian: ID is the tenant ID associated with this QuotaDetailSet

hydroflame:
@jtopjian all comments have been fixed

openstack/compute/v2/extensions/quotasets/urls.go
```diff
@@ -8,14 +8,14 @@ func resourceURL(c *gophercloud.ServiceClient) string {
 	return c.ServiceURL(resourcePath)
 }
 
-func getURL(c *gophercloud.ServiceClient, tenantID string) string {
+func getDetailURL(c *gophercloud.ServiceClient, tenantID string) string {
```
jtopjian: I apologize - I didn't explain this well enough. What I meant was to create a new URL in `urls.go` and keep the existing `getURL`.

So `getURL` would remain unchanged and `getDetailURL` would look like:

```go
func getDetailURL(c *gophercloud.ServiceClient, tenantID string) string {
        return c.ServiceURL(resourcePath, tenantID, "detail")
}
```

By doing this, the `Get` function in `requests.go` remains unchanged and `GetDetail` will now look like:

```go
// GetDetail returns detailed public data about a previously created QuotaSet.
func GetDetail(client *gophercloud.ServiceClient, tenantID string) GetDetailResult {
	_, res.Err = client.Get(getDetailURL(client, tenantID), &res.Body, nil)
 	return res
 }
```

jtopjian:
@hydroflame Perfect - thank you very much. @nate-johnston Thank you as well.

Issue #551, opened 2017-10-10
Compute v2: Add Simple Tenant Usage extension


nate-johnston:
Nova's os-simple-tenant-usage extension [1] provides the ability to show metered usage data for OpenStack clouds.  This is very useful information to create a billing model around OpenStack clouds.  This would be a new Compute v2 extension.

[1] https://developer.openstack.org/api-ref/compute/#usage-reports-os-simple-tenant-usage

nate-johnston:
I am waiting for approval from a core contributor before creating a PR.

jtopjian:
@nate-johnston Apologies - you're more than welcome to open a PR for this.

jtopjian:
Implemented in #565 and #1125

PR #565, opened 2017-10-19
Compute v2 - Implement Simple Tenant Usage Extension (#551)  …

nate-johnston:
* Add GetUsage, GetTenantUsage, and GetDetailedUsage functions to
implement the os-simple-tenant-usage extension in
Gophercloud's OpenStack Compute v2 API

- GetUsage allows a caller to get aggregate usage metrics for all
  tenants.
- GetTenantUsage allows a caller to specify a tenant, and get both
  aggregate usage metrics as well as a set of instance-specific
  usage metrics.
- GetDetailedUsage allows a caller to get both aggregate usage metrics
  as well as a set of instance-specific usage metrics per tenant for all
  tenants.
- GetUsage and GetDetailedUsage effectively wrap OpenStack Compute v2
  API function /v2/os-simple-tenant-usage/
- GetTenantUsage effectively wraps OpenStack Compute v2 API function
  /v2/os-simple-tenant-usage/{tenant_id}

Note: Tests use sample response bodies from
https://developer.openstack.org/api-ref/compute/#list-tenant-usage-statistics-for-all-tenants
and
https://developer.openstack.org/api-ref/compute/#show-usage-statistics-for-tenant

Code citation:
https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/simple_tenant_usage.py

For #551


jtopjian:
@nate-johnston Thanks for submitting this.

Looks like there are some package import issues. All of the comcast packages should be changed to `github.com/gophercloud/gophercloud`.

I haven't done a detailed review yet, but at first glance, the paging might need updated. For example, instead of defining a struct for Links, `gophercloud.Link` already exists for that purpose. See how paging is implemented in `compute/v2/flavors/results.go` or `compute/v2/servers/results.go` for examples.

jrperritt:
> Note: Tests use sample response bodies from 
https://developer.openstack.org/api-ref/compute/#list-tenant-usage-statistics-for-all-tenants

Wherever possible, we try not to use information from there, both due to historic inaccuracy and unknown copyright issues that may arise. Better would be to have your environment set up for this feature and query it with the `openstack` CLI to get the response.

hydroflame:
@jtopjian all comments addressed :)

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// TenantUsage is a set of usage information about a tenant over the sampling window
+type TenantUsage struct {
+	// ServerUsages is an array of ServerUsage maps
+	ServerUsages []ServerUsage `json:"server_usages"`
+
+	// Start is the beginning time to calculate usage statistics on compute and storage resources
+	Start string `json:"start"`
```
jtopjian: This field should be parsed as `time.Time`. Based on the result I'm seeing, it'll need a custom unmarshal function. There are many of these throughout gophercloud to use as an example. It looks like the time format is a "MilliNoZ`, so look for an example that uses that format.

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// Stop is the ending time to calculate usage statistics on compute and storage resources
+	Stop string `json:"stop"`
```
jtopjian: Same here: `time.Time` with a custom unmarshal.

openstack/compute/v2/extensions/simpletenantusage/requests.go
```diff
@@ -0,0 +1,72 @@
...
+	// The ending time to calculate usage statistics on compute and storage resources.
+	End time.Time `json:"end,omitempty"`
+
+	// The beginning time to calculate usage statistics on compute and storage resources.
+	Start time.Time `json:"start,omitempty"`
+
+	// Requests a page size of items. Calculate usage for the limited number of instances.
+	// Use the limit parameter to make an initial limited request and use the last-seen instance
+	// UUID from the response as the marker parameter value in a subsequent limited request.
+	Limit *int `json:"limit,omitempty"`
```
jtopjian: I recommend deleting this. This was introduced in Microversion 2.40. A decision about how microversions will be implemented in Gophercloud is still in progress.

In addition, `Limit` and `Marker` would cause the result to no longer be a single page.

openstack/compute/v2/extensions/simpletenantusage/requests.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// The last-seen item based on an earlier Limit query.
+	Marker string `json:"marker,omitempty"`
```
jtopjian: Same as `Limit`.

openstack/compute/v2/extensions/simpletenantusage/requests.go
```diff
@@ -0,0 +1,72 @@
...
+		return SimpleSingleTenantUsagePage{pagination.SinglePageBase(r)}
+	})
+}
+
+// Options for fetching the tenant usage for a Tenant.
+// All int-values are pointers so they can be nil if they are not needed
+// you can use gopercloud.IntToPointer() for convenience
+type GetOpts struct {
+	// Set to 1 to get detailed (i.e. "server usages") tenant usage; 0 otherwise.
+	Detailed *bool `json:"detailed,omitempty"`
```
jtopjian: The URL parameter expects `detailed=1` or `detailed=0`. `*bool` only supports `true/false/nil`. So either there needs to be an internal custom conversion of `true` to `1` or this needs to be set to a `int` with documentation explaining that only a value of 1 will yield a result.

Also, since this is a query string, you'll need to use `q` instead of `json`. This goes for all other fields in this struct.

openstack/compute/v2/extensions/simpletenantusage/requests.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// The ending time to calculate usage statistics on compute and storage resources.
+	End time.Time `json:"end,omitempty"`
+
+	// The beginning time to calculate usage statistics on compute and storage resources.
+	Start time.Time `json:"start,omitempty"`
```
jtopjian: There's a big issue with the `Start` and `End` fields: they never end up in the URL query. You can see this by inspecting the URL being build for the request or by looking at the Nova API service's log files -- start and end are never part of the URL. The reason is because the [`BuildQueryString`](https://github.com/gophercloud/gophercloud/blob/master/params.go#L303) function does not support time fields. 

There are two things we can do here:

1. Use `string`, though I don't think that's the best long-term solution
2. Change these to `*time.Time` and append `q` in `ToSimpleTenantUsageGetMap` with values if set:

```go
const TimeParseFormat = "2006-01-02T15:04:05.000000"
func (opts GetOpts) ToSimpleTenantUsageGetMap() (string, error) {
  params, err := url.ParseQuery(q.RawQuery)
  if err != nil {
    return "", err
  }

  if opts.Start != nil {
    params.Add("start", opts.Start.Format(TimeParseFormat))
  }

  q = &url.URL{RawQuery: params.Encode()}
  return q.String(), err
}
``` 

@jrperritt thoughts on this one?

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,130 @@
...
+	TotalMemoryMBUsage float64 `json:"total_memory_mb_usage"`
+
+	// TotalVCPUsUsage multiplies the number of virtual CPUs of the server by hours the server exists, and then adding that all together for each server
+	TotalVCPUsUsage float64 `json:"total_vcpus_usage"`
+}
+
+// ServerUsage is a detailed set of information about a specific instance inside a tenant
+type ServerUsage struct {
+	// EndedAt is the date and time when the server was deleted
+	EndedAt string `json:"ended_at"`
```
jtopjian: This will also need to be `time.Time` with a custom unmarshaller.

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,130 @@
...
+	LocalGB int `json:"local_gb"`
+
+	// MemoryMB is the memory size of the server (in MB)
+	MemoryMB int `json:"memory_mb"`
+
+	// Name is the name assigned to the server when it was created
+	Name string `json:"name"`
+
+	// StartedAt is the date and time when the server was started
+	StartedAt string `json:"started_at"`
```
jtopjian: This will also need to be `time.Time` with a custom unmarshaller.

jtopjian:
@nate-johnston I've generated some test fixtures here:

https://gist.github.com/jtopjian/aa773fb2d18185381d0a7128b7b306a8

jtopjian:
err... wrong PR. Going to close this one in favor of #624

Issue #567, opened 2017-10-23
Gogland IDE internal directory is not ignored by git


zimnx:

jtopjian:
Implemented in #568

PR #568, opened 2017-10-23
Ignore Gogland IDE internal directory

zimnx:
For #567


jtopjian:
I don't see any harm with this - thanks for submitting it :)

Issue #378, opened 2017-06-13
BootFromVolume extension should support `disk_bus` and `device_type` fields


fatmcgav:
As per the Openstack docs for [block_device_mapping](https://docs.openstack.org/developer/nova/block_device_mapping.html), it's possible to specify a `disk_bus` and `device_type` fields.

> disk_bus and device_type - low level details that some hypervisors (currently only libvirt) may support. Some example disk_bus values can be: _ide_, _usb_, _virtio_, _scsi_, while device_type may be _disk_, _cdrom_, _floppy_, _lun_. This is not an exhaustive list as it depends on the virtualization driver, and may change as more support is added. Leaving these empty is the most common thing to do.

Required for https://github.com/terraform-providers/terraform-provider-openstack/issues/18

jtopjian:
Additional work required: Locate where `disk_bus` and `device_type` are defined in `block_device_mapping` v2.

Bonus: add these attributes to an existing (or new) [acceptance test](https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/compute/v2/bootfromvolume_test.go) and verify it works.

yongfengdu:
I'll look into this issue

yongfengdu:
@jtopjian , my-HenryS and I are in one team, and I asked my-HenryS to provide this patch.
Thanks for your review.

jtopjian:
@yongfengdu That's perfect - thanks :smile:

jtopjian:
Implemented in #1129

PR #569, opened 2017-10-23
Issue 378

my-HenryS:
For #378 

code in this PR:

https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/schemas/block_device_mapping.py

### DIFF
--- a/openstack/compute/v2/extensions/bootfromvolume/requests.go
+++ b/openstack/compute/v2/extensions/bootfromvolume/requests.go
@@ -69,10 +69,10 @@ type BlockDevice struct {
 	VolumeSize int `json:"volume_size,omitempty"`
 
 	// DeviceType [optional] low level details that some hypervisors (currently only libvirt) may support. device_type value may be disk, cdrom, floppy, lun.
-	DeviceType string `json:"device_type"`
+	DeviceType string `json:"device_type,omitempty"`
 
 	// DiskBus values can be: ide, usb, virtio, scsi
-	DiskBus string `json:"disk_bus"`
+	DiskBus string `json:"disk_bus,omitempty"`
 }
 
 // CreateOptsExt is a structure that extends the server `CreateOpts` structure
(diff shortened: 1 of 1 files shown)


jtopjian:
@my-HenryS Thank you for submitting this. It looks like this is your first contribution to Gophercloud, so welcome! And for a first PR, this is quite good. I have a few comments, though:

1. @yongfengdu mentioned that they were working on this issue in #378. Granted it has been two weeks with no activity, it would still be polite to check and see if they plan to continue. Let's give them a day or two to respond (GitHub should email them with a notification from this comment).

2. "device_type value may be disk, cdrom, floppy, lun" and "DiskBus values can be: ide, usb, virtio, scsi" 

This deserves some extra research. Focusing on `DiskBus`, from searching the nova code, it looks like [hyperv](https://github.com/openstack/nova/blob/768d7cc0a632e1a880f00c5840c1ec8051e161be/nova/virt/hyperv/block_device_manager.py) is the only driver that is doing a strict check about what the `disk_bus` value should be. But I might be wrong.

This is where the importance of acceptance tests come in. If you are able to, I recommend creating an instance/server with your patch, but supplying a false value like `foobar` for the `disk_bus` and see if the Nova API service prevents the request. 

If it does, we know that the API is doing a check _somewhere_ and we need to find where. Then the best course of action is to create a dedicated string type for the known valid values (similar to [this](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/resetstate/requests.go#L7-L16)).

If the API doesn't give you an error (ignoring hyperv), then keeping this as `string` is acceptable, even if the value makes no sense (like `foobar`).

The same strategy applies to `DeviceType`

If you aren't able to run an acceptance test, I can do this when I have some time.

If you have already verified this, let me know.

3. The comments you made for the fields are inconsistent. Making sure they're consistent is important since they are used as [official documentation](http://godoc.org/github.com/gophercloud/gophercloud). 

For `DeviceType`, you marked it as `[optional]`. Since the other optional fields don't have this, I recommend removing it. You also mention `DeviceType`, but then refer to it as `device_type`. Both should be `DeviceType`. Finally, comments should (though not required) be wrapped at 80 characters (this isn't a big deal. If this was the only change needed, I wouldn't have mentioned it).

Again, thank you for working on this. Please let me know if you have any questions.

my-HenryS:
It is really nice of you for your detailed instructions. I will continue working on this.

jtopjian:
I've finished up this work as part of #1129. Thank you for the initial work on it.

Issue #566, opened 2017-10-20
Identity v3: Add assigning role to user on project


dongwoo1005:
I need a support for assigning role to user on project: https://developer.openstack.org/api-ref/identity/v3/#assign-role-to-user-on-project
If nobody is working on it at the moment, I will make a PR for this.

I can start working on this first. Then I might as well add 
- assign/unassign role to group on domain
- assign/unassign role to user on domain
- assign/unassign role to group on project
- assign/unassign role to user on project

May I start working on this PR?

dongwoo1005:
I can start with this first. Then I might as well add 
- assign/unassign role to group on domain
- assign/unassign role to user on domain
- assign/unassign role to group on project
- assign/unassign role to user on project
May I start working on this PR?

jtopjian:
@dongwoo1005 Thanks for offering. You're more than welcome to work on these items. Please make sure to open a separate PR for each one, though.

I'd also wait until #561 is merged before opening any PRs since you'll most likely need to rebase based on those changes.

dongwoo1005:
Yep. I forked from dklyle:role-list and my implementation is complete. 
As you said though, I need to wait for #561 to be merged for my acceptance test.
I'll make a PR once #561 is checked in.

jtopjian:
@dongwoo1005 You are clear to go.

Please use some of the other Role Issues/PRs as templates for including all of the required information.

jtopjian:
Implemented in #570

PR #570, opened 2017-10-23
Identity v3: Add assigning a role to a user on a project

dongwoo1005:
This is an implementation for assigning/unassining a role to a user on a project for #566.

code in this PR:
* [api doc for assigning](https://developer.openstack.org/api-ref/identity/v3/#assign-role-to-user-on-project)
* [api doc for unassining](https://developer.openstack.org/api-ref/identity/v3/#unassign-role-from-user-on-project)
* https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L638
* https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L677

This is a contribution from [Salesforce](https://www.salesforce.com/)

### DIFF
--- a/acceptance/openstack/identity/v3/identity.go
+++ b/acceptance/openstack/identity/v3/identity.go
@@ -147,21 +147,6 @@ func CreateRole(t *testing.T, client *gophercloud.ServiceClient, c *roles.Create
 	return role, nil
 }
 
-// AssignRoleToUserOnProject will grant a role to a user on a project. An error will be
-// returned if the grant was unsuccessful.
-func AssignRoleToUserOnProject(t *testing.T, client *gophercloud.ServiceClient, role *roles.Role, user *users.User, project *projects.Project) error {
-	t.Logf("Attempting to grant user %s role %s on project %s", user.Name, role.Name, project.Name)
-
-	err := roles.AssignToUserOnProject(client, role.ID, user.ID, project.ID).ExtractErr()
-	if err != nil {
-		return err
-	}
-
-	t.Logf("Granted user %s role %s on project %s", user.Name, role.Name, project.Name)
-
-	return nil
-}
-
 // DeleteProject will delete a project by ID. A fatal error will occur if
 // the project ID failed to be deleted. This works best when using it as
 // a deferred function.
@@ -210,17 +195,15 @@ func DeleteDomain(t *testing.T, client *gophercloud.ServiceClient, domainID stri
 	t.Logf("Deleted domain: %s", domainID)
 }
 
-// UnassignRoleFromUserOnProject will revoke a role of a user on a project. A fatal error will
-// occur if the revoke was unsuccessful. This works best when used as a deferred function
-func UnassignRoleFromUserOnProject(t *testing.T, client *gophercloud.ServiceClient, role *roles.Role, user *users.User, project *projects.Project) {
-	t.Logf("Attempting to remove role %s from user %s on project %s", role.Name, user.Name, project.Name)
-
-	err := roles.UnassignFromUserOnProject(client, role.ID, user.ID, project.ID).ExtractErr()
+// UnassignRole will delete a role assigned to a user/group on a project/domain
+// A fatal error will occur if it fails to delete the assignment.
+// This works best when using it as a deferred function.
+func UnassignRole(t *testing.T, client *gophercloud.ServiceClient, roleID string, opts *roles.AssignOpts) {
+	err := roles.Unassign(client, roleID, *opts).ExtractErr()
 	if err != nil {
-		t.Fatalf("Unable to remove role")
+		t.Fatalf("Unable to unassign a role %v on context %v: %v", roleID, *opts, err)
 	}
-
-	t.Logf("Removed role %s from user %s on project %s", role.Name, user.Name, project.Name)
+	t.Logf("Unassigned the role %v on context %v", roleID, *opts)
 }
 
 // FindRole finds all roles that the current authenticated client has access
@@ -248,32 +231,4 @@ func FindRole(t *testing.T, client *gophercloud.ServiceClient) (*roles.Role, err
 	t.Logf("Successfully found a role %s with ID %s", role.Name, role.ID)
 
 	return role, nil
-}
-
-// FindProject finds all projects that the current authenticated client has access
-// to and returns the first one found. An error will be returned if the lookup
-// was unsuccessful.
-func FindProject(t *testing.T, client *gophercloud.ServiceClient) (*projects.Project, error) {
-	t.Log("Attempting to find a project")
-
-	var project *projects.Project
-
-	allPages, err := projects.List(client, nil).AllPages()
-	if err != nil {
-		return nil, err
-	}
-
-	allProjects, err := projects.ExtractProjects(allPages)
-	if err != nil {
-		return nil, err
-	}
-
-	for _, p := range allProjects {
-		project = &p
-		break
-	}
-
-	t.Logf("Successfully found a project %s with ID %s", project.Name, project.ID)
-
-	return project, nil
 }
--- a/acceptance/openstack/identity/v3/identity.go
+++ b/acceptance/openstack/identity/v3/identity.go
@@ -201,9 +201,9 @@ func DeleteDomain(t *testing.T, client *gophercloud.ServiceClient, domainID stri
 func UnassignRole(t *testing.T, client *gophercloud.ServiceClient, roleID string, opts *roles.AssignOpts) {
 	err := roles.Unassign(client, roleID, *opts).ExtractErr()
 	if err != nil {
-		t.Fatalf("Unable to unassign a role %v on context %v: %v", roleID, *opts, err)
+		t.Fatalf("Unable to unassign a role %v on context %+v: %v", roleID, *opts, err)
 	}
-	t.Logf("Unassigned the role %v on context %v", roleID, *opts)
+	t.Logf("Unassigned the role %v on context %+v", roleID, *opts)
 }
 
 // FindRole finds all roles that the current authenticated client has access
--- a/acceptance/openstack/identity/v3/identity.go
+++ b/acceptance/openstack/identity/v3/identity.go
@@ -147,6 +147,21 @@ func CreateRole(t *testing.T, client *gophercloud.ServiceClient, c *roles.Create
 	return role, nil
 }
 
+// AssignRoleToUserOnProject will grant a role to a user on a project. An error will be
+// returned if the grant was unsuccessful.
+func AssignRoleToUserOnProject(t *testing.T, client *gophercloud.ServiceClient, role *roles.Role, user *users.User, project *projects.Project) error {
+	t.Logf("Attempting to grant user %s role %s on project %s", user.Name, role.Name, project.Name)
+
+	err := roles.AssignToUserOnProject(client, role.ID, user.ID, project.ID).ExtractErr()
+	if err != nil {
+		return err
+	}
+
+	t.Logf("Granted user %s role %s on project %s", user.Name, role.Name, project.Name)
+
+	return nil
+}
+
 // DeleteProject will delete a project by ID. A fatal error will occur if
 // the project ID failed to be deleted. This works best when using it as
 // a deferred function.
@@ -165,10 +180,10 @@ func DeleteProject(t *testing.T, client *gophercloud.ServiceClient, projectID st
 func DeleteUser(t *testing.T, client *gophercloud.ServiceClient, userID string) {
 	err := users.Delete(client, userID).ExtractErr()
 	if err != nil {
-		t.Fatalf("Unable to delete user %s: %v", userID, err)
+		t.Fatalf("Unable to delete user with ID %s: %v", userID, err)
 	}
 
-	t.Logf("Deleted user: %s", userID)
+	t.Logf("Deleted user with ID: %s", userID)
 }
 
 // DeleteGroup will delete a group by ID. A fatal error will occur if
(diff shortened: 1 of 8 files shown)


acceptance/openstack/identity/v3/identity.go
```diff
@@ -166,3 +182,71 @@ func DeleteDomain(t *testing.T, client *gophercloud.ServiceClient, domainID stri
...
 }
+
+// UnassignRoleFromUserOnProject will revoke a role of a user on a project. A fatal error will
+// occur if the revoke was unsuccessful. This works best when used as a deferred function
+func UnassignRoleFromUserOnProject(t *testing.T, client *gophercloud.ServiceClient, role *roles.Role, user *users.User, project *projects.Project) {
+	t.Logf("Attempting to remove role %s from user %s on project %s", role.Name, user.Name, project.Name)
+
+	err := roles.UnassignFromUserOnProject(client, role.ID, user.ID, project.ID).ExtractErr()
+	if err != nil {
+		t.Fatalf("Unable to remove role")
```
nurland: nit: maybe return error instead of Fatalf. Also please add role.ID/user.ID to the log message.

acceptance/openstack/identity/v3/identity.go
```diff
@@ -166,3 +182,71 @@ func DeleteDomain(t *testing.T, client *gophercloud.ServiceClient, domainID stri
...
+// FindRole finds all roles that the current authenticated client has access
+// to and returns the first one found. An error will be returned if the lookup
+// was unsuccessful.
+func FindRole(t *testing.T, client *gophercloud.ServiceClient) (*roles.Role, error) {
+	t.Log("Attempting to find a role")
+	var role *roles.Role
+
+	allPages, err := roles.List(client, nil).AllPages()
+	if err != nil {
+		return role, err
```
nurland: nit: `return nil, err`

acceptance/openstack/identity/v3/identity.go
```diff
@@ -166,3 +182,71 @@ func DeleteDomain(t *testing.T, client *gophercloud.ServiceClient, domainID stri
...
+	}
+
+	allRoles, err := roles.ExtractRoles(allPages)
+	if err != nil {
+		return role, err
```
nurland: same as above

acceptance/openstack/identity/v3/identity.go
```diff
@@ -166,3 +182,71 @@ func DeleteDomain(t *testing.T, client *gophercloud.ServiceClient, domainID stri
...
+// to and returns the first one found. An error will be returned if the lookup
+// was unsuccessful.
+func FindProject(t *testing.T, client *gophercloud.ServiceClient) (*projects.Project, error) {
+	t.Log("Attempting to find a project")
+
+	var project *projects.Project
+
+	allPages, err := projects.List(client, nil).AllPages()
+	if err != nil {
+		return project, err
```
nurland: nit: `return nil, err`

openstack/identity/v3/roles/urls.go
```diff
@@ -2,14 +2,24 @@ package roles
...
+	return client.ServiceURL(rolePath)
 }
 
 func getURL(client *gophercloud.ServiceClient, roleID string) string {
-	return client.ServiceURL("roles", roleID)
+	return client.ServiceURL(rolePath, roleID)
 }
 
 func listAssignmentsURL(client *gophercloud.ServiceClient) string {
 	return client.ServiceURL("role_assignments")
```
nurland: nit: consider moving `role_assignments` to const as well.

acceptance/openstack/identity/v3/identity.go
dongwoo1005: UnassignFromUserOnProject is most likely to be used as a deferred function, so I created a fatal here.

openstack/identity/v3/roles/urls.go
jtopjian: I'm not sure any of the `urls.go` files have constants. I might be wrong, but it's worth double-checking. If there isn't precedence, I'd suggest not using constants here.

dongwoo1005: In other `urls.go`, I found that any paths used more than once is defined as const and it's not otherwise.

jtopjian: > In some other urls.go, I found that any paths used more than once is defined as const and it's not otherwise.

```shell
$ find . -name urls.go -exec grep -l const {} \;
```

Well I guess there's a good amount of `const`'s being used. In various ways, too: some have all parts of the URL as a constant, some have only repeated parts mixed with strings.

I don't want to spend too much time nit-picking this one so it's totally up to you. 😄

dongwoo1005: Cool. I'll leave it as is atm 😄

jtopjian: Sounds good. Let me know when this PR is good to go for a full review. I saw a few messages going back and forth and may have prematurely jumped in :)

dongwoo1005: I think that it's good to go for a full review :)

acceptance/openstack/identity/v3/roles_test.go
```diff
@@ -41,21 +41,63 @@ func TestRolesGet(t *testing.T) {
...
 	tools.PrintResource(t, p)
 }
+
+func TestAssignToUserOnProject(t *testing.T) {
+	client, err := clients.NewIdentityV3Client()
+	if err != nil {
+		t.Fatalf("Unable to obtain an indentity client: %v", err)
+	}
+
+	project, err := FindProject(t, client)
```
jtopjian: It'd be better to create a temporary project using the `CreateProject(t, client, nil)` function. This localizes all testing to a disposable project.

jtopjian:
@dongwoo1005 Thanks for putting this together. Short of the one comment I left, this is spot on. The reason I didn't do an official review is because after seeing the Keystone code, I think there might be a more succinct way of doing all of these assignment operations. 

The API reference has 8 URLs for handling these assignment and unassignment actions. However, as you linked, they are all aggregated into master assign/unassign functions.

Implementing 8 distinct functions in Gophercloud is one way to do this and I'm fine with that if the following doesn't work:

Gophercloud supports a go tag of [`xor`](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/lbaas_v2/pools/requests.go#L92-L98). It's not used often, but I think it would work well here, and it would match what is being done in [Keystone](https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L641-L642).

Using `xor`, we might be able to simplify all of this to something like:

```go
type AssignOpts struct {
  UserID string `xor:"GroupID"`
  GroupID string `xor:"UserID"`
  Project string `xor:"ProjectID"`
  Domain string `xor:"DomainID"`
}

func Assign(client, roleID, assignOpts) AssignResult {
  // do something here to build the correct URL based on given options
}
```

I realize I asked for separate PRs for each action, but that was until I saw how this was implemented in Keystone, so I think this might be an edge case. And technically this can still be broken out into two PRs: one for assign, one for unassign.

Thoughts?

Also pinging @jrperritt to see if he has any objections to this method.

jrperritt:
> Also pinging jrperritt to see if he has any objections to this method.

Nope, SGTM

acceptance/openstack/identity/v3/identity.go
priyankporwal: @dongwoo1005 Is similar being done for other *remove*/*cleanup* kind of functions in the codebase? I worry that an assumption is being made about how this function would be invoked by caller. Perhaps some callers just want to get a regular error like all other functions and deal with it in their layer.

acceptance/openstack/identity/v3/identity.go
```diff
@@ -166,3 +182,71 @@ func DeleteDomain(t *testing.T, client *gophercloud.ServiceClient, domainID stri
...
+	err := roles.UnassignFromUserOnProject(client, role.ID, user.ID, project.ID).ExtractErr()
+	if err != nil {
+		t.Fatalf("Unable to remove role")
+	}
+
+	t.Logf("Removed role %s from user %s on project %s", role.Name, user.Name, project.Name)
+}
+
+// FindRole finds all roles that the current authenticated client has access
+// to and returns the first one found. An error will be returned if the lookup
```
priyankporwal: Should the function name reflect that 'Any' (or 'First' of 'One') role is being returned?

acceptance/openstack/identity/v3/identity.go
```diff
@@ -166,3 +182,71 @@ func DeleteDomain(t *testing.T, client *gophercloud.ServiceClient, domainID stri
...
+		return nil, err
+	}
+
+	allRoles, err := roles.ExtractRoles(allPages)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, r := range allRoles {
+		role = &r
```
priyankporwal: Is allRoles stable, i.e. does the list contain the same ordered set of roles each time this gets called?

acceptance/openstack/identity/v3/identity.go
```diff
@@ -166,3 +182,71 @@ func DeleteDomain(t *testing.T, client *gophercloud.ServiceClient, domainID stri
...
+		break
+	}
+
+	t.Logf("Successfully found a role %s with ID %s", role.Name, role.ID)
+
+	return role, nil
+}
+
+// FindProject finds all projects that the current authenticated client has access
+// to and returns the first one found. An error will be returned if the lookup
```
priyankporwal: Similar comment about the 'Any'/'First'/'One' project in the function name.

acceptance/openstack/identity/v3/identity.go
jtopjian: We're quite lenient with functions used for acceptance testing. As long as it doesn't blow the cloud up, I'm OK with it. :)

priyankporwal: @jtopjian Just curious, is this function in a section that calls out "test purpose only"? It is not obvious that this function is meant for acceptance testing only. Perhaps a comment would do for now.

priyankporwal: Nevermind, I didn't see the 'acceptance' at the very top of the folder structure.

jtopjian:
Not exactly sure why people are approving this PR when I laid out an alternative means of implementing it.

@dongwoo1005 Have you had time to look at the alternative implementation?

dongwoo1005:
@jtopjian I haven't got a time to implement the alternative method yet. I agree with the xor approach as it was the approach used in the python code base. I plan to work on this within the next day or two.

jtopjian:
@dongwoo1005 No problem at all - take all the time you need and let me know if you need help.

openstack/identity/v3/roles/requests.go
```diff
@@ -148,3 +148,74 @@ func ListAssignments(client *gophercloud.ServiceClient, opts ListAssignmentsOpts
...
+	// ProjectID is the ID of a project to assign a role on
+	// Note: exactly one of ProjectID or DomainID must be provided
+	ProjectID string `xor:"DomainID"`
+
+	// DomainID is the ID of a domain to assign a role on
+	// Note: exactly one of ProjectID or DomainID must be provided
+	DomainID string `xor:"ProjectID"`
+}
+
+func toAssignRoleURL(client *gophercloud.ServiceClient, roleID string, opts AssignOpts) (string, error) {
```
jtopjian: If this function is kept, it should be renamed to `toRoleAssignURL`. Trivial change, but it follows the naming pattern throughout Gophercloud.

openstack/identity/v3/roles/requests.go
```diff
@@ -148,3 +148,74 @@ func ListAssignments(client *gophercloud.ServiceClient, opts ListAssignmentsOpts
...
+	var url string
+	if opts.UserID != "" && opts.ProjectID != "" {
+		url = userOnProjectURL(client, opts.ProjectID, opts.UserID, roleID)
+	} else if opts.UserID != "" && opts.DomainID != "" {
+		url = userOnDomainURL(client, opts.DomainID, opts.UserID, roleID)
+	} else if opts.GroupID != "" && opts.ProjectID != "" {
+		url = groupOnProjectURL(client, opts.ProjectID, opts.GroupID, roleID)
+	} else if opts.GroupID != "" && opts.DomainID != "" {
+		url = groupOnDomainURL(client, opts.DomainID, opts.GroupID, roleID)
+	}
```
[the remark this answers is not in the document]
dongwoo1005: I like this idea. 👍

openstack/identity/v3/roles/requests.go
```diff
@@ -148,3 +148,74 @@ func ListAssignments(client *gophercloud.ServiceClient, opts ListAssignmentsOpts
 		return RoleAssignmentPage{pagination.LinkedPageBase{PageResult: r}}
 	})
 }
+
+// AssignOpts provides options to assign a role
+type AssignOpts struct {
```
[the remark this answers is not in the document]
dongwoo1005: I agree that `AssignOpts` may be counter-intuitive to the users calling `Unassign`.
Since it is not a lot of duplicate code and is a better user-facing design, I will make change to the option 1 for now. 
If there is any objection to this, please let me know.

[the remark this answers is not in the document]
jtopjian: > Since it is not a lot of duplicate code and is a better user-facing design, I will make change to the option 1 for now.

This sounds good to me. I've been revisiting this throughout the day and Option 1 seems like the most straightforward solution.

openstack/identity/v3/roles/doc.go
```diff
@@ -58,5 +58,35 @@ Example to List Role Assignments
 	for _, role := range allRoles {
 		fmt.Printf("%+v\n", role)
 	}
+
+Example to Grant a Role
```
jtopjian: How about:

Example to Assign a Role to a User in a Project

openstack/identity/v3/roles/doc.go
```diff
@@ -58,5 +58,35 @@ Example to List Role Assignments
...
+	err := roles.Assign(identityClient, roleID, roles.AssignOpts{
+		UserID:    userID,
+		ProjectID: projectID,
+	}).ExtractErr()
+
+	if err != nil {
+		panic(err)
+	}
+
+Example to Remove a Role
```
jtopjian: Example to Unassign a Role From a User in a Project

openstack/identity/v3/roles/requests.go
```diff
@@ -148,3 +148,115 @@ func ListAssignments(client *gophercloud.ServiceClient, opts ListAssignmentsOpts
...
+	ProjectID string `xor:"DomainID"`
+
+	// DomainID is the ID of a domain to assign a role on
+	// Note: exactly one of ProjectID or DomainID must be provided
+	DomainID string `xor:"ProjectID"`
+}
+
+// UnassignOpts provides options to unassign a role
+type UnassignOpts struct {
+	// UserID is the ID of a user to assign a role
```
jtopjian: `s/assign/unassign/g` in the entire struct.

openstack/identity/v3/roles/results.go
```diff
@@ -188,3 +188,15 @@ func ExtractRoleAssignments(r pagination.Page) ([]RoleAssignment, error) {
 	err := (r.(RoleAssignmentPage)).ExtractInto(&s)
 	return s.RoleAssignments, err
 }
+
+// AssignmentResult represents the result of assign operation.
```
jtopjian: AssignmentResult represents the result of an assign operation.

openstack/identity/v3/roles/results.go
```diff
@@ -188,3 +188,15 @@ func ExtractRoleAssignments(r pagination.Page) ([]RoleAssignment, error) {
...
+// Call ExtractErr method to determine if the request succeeded or failed.
+type AssignmentResult struct {
+	gophercloud.ErrResult
+}
+
+// UnassignmentResult represents the result of unassign operation.
```
jtopjian: UnassignmentResult represents the result of an unassign operation.

dongwoo1005:
@jtopjian Updated 😄

jtopjian:
@dongwoo1005 Nice! Thank you very much for your work and patience with this.

Issue #571, opened 2017-10-24
Identity v3 role C_UD


dklyle:
- [x] Create (#572)
- [x] Delete (#577)
- [x] Update (#578)

Read was already covered by #550 

API Docs:
https://developer.openstack.org/api-ref/identity/v3/#roles

Schema:
https://github.com/openstack/keystone/blob/stable/pike/keystone/assignment/role_backends/sql.py#L196
https://github.com/openstack/keystone/blob/stable/pike/keystone/assignment/role_backends/sql.py#L201

Create:
https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L319

Delete:
https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L383

Update:
https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L367

jtopjian:
:tada: 

@dklyle Thank you very much for the work you've been doing.

PR #572, opened 2017-10-24
Identity v3: role create

dklyle:
Adding create role. This also updates the result structure of `Role` to add an `Extra` field as this patch allowed testing/validation of support for the `Extra` field.

For #571 

code in this PR:

Schema:
https://github.com/openstack/keystone/blob/stable/pike/keystone/assignment/role_backends/sql.py#L196
https://github.com/openstack/keystone/blob/stable/pike/keystone/assignment/role_backends/sql.py#L201

Create:
https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L319


jtopjian:
@dklyle This looks good. Is it ready to be merged?

dklyle:
@jtopjian yes, ready to go from my viewpoint.

jtopjian:
Awesome - thank you for another great PR :)

Issue #573, opened 2017-10-26
Neutron Router: Enable/Disable SNAT


Crapworks:
Hi there,

I learned today that some OpenStack providers have SNAT disabled by default. Of course, I'd like to enable it, so I can actually access the internet from instances without attached floating IPs...

Gophercloud doesn't support this at the moment, because there is no `enable_snat` parameter for the `external_gateway_info` in the router resource. I am not very experienced with GO but as far as I can see this should be a simple fix so I will create a PR for this.

Cheers,
Christian

jtopjian:
Implemented in #574

PR #574, opened 2017-10-26
allow to configure the enable_snat parameter

Crapworks:
For #573

code in this PR:

https://github.com/openstack/neutron/blob/97af09672735fb9d2742e23f3a6e94331d92e3cc/neutron/extensions/l3_ext_gw_mode.py#L32

### DIFF
--- a/openstack/networking/v2/extensions/layer3/routers/results.go
+++ b/openstack/networking/v2/extensions/layer3/routers/results.go
@@ -9,6 +9,7 @@ import (
 // particular network router.
 type GatewayInfo struct {
 	NetworkID        string            `json:"network_id"`
+	EnableSNAT       *bool             `json:"enable_snat,omitempty"`
 	ExternalFixedIPs []ExternalFixedIP `json:"external_fixed_ips,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


openstack/networking/v2/extensions/layer3/routers/results.go
```diff
@@ -9,6 +9,7 @@ import (
 // particular network router.
 type GatewayInfo struct {
 	NetworkID        string            `json:"network_id"`
+	EnableSnat       bool              `json:"enable_snat,omitempty"`
```
jtopjian: Should be `EnableSNAT`.

Crapworks:
@jtopjian Thanks for looking into it and basically writing the PR for me :)

I hope I implemented the changes correctly, especially the naming since the capitalisation was kinda mixed in your patch. Tell me if you need anything else changed.

jtopjian:
> Thanks for looking into it and basically writing the PR for me :)

You did the heavy lifting of finding the missing feature :smile: 

`*bool` and `bool` is one of the more difficult situations when working with Gophercloud.

`EnableSNAT` is the correct form to use.

This looks good to me! Thank you for your work on this.

PR #578, opened 2017-10-27
Identity v3 Role update

dklyle:
For #571 

code in this PR:

Update:
https://github.com/openstack/keystone/blob/83bd595b22944d38eff1cdef77b4c07a75af0fdc/keystone/assignment/controllers.py#L367

Note that the API for roles does not support updating the domain ID per https://developer.openstack.org/api-ref/identity/v3/#update-role


jtopjian:
> Note that the API for roles does not support updating the domain ID per https://developer.openstack.org/api-ref/identity/v3/#update-role

Right - looks like the schema has it locked down to just `name`: https://github.com/openstack/keystone/blob/489e4b1420af26f4d996674723348bb1a67405e9/keystone/assignment/schema.py

Another great PR! Thanks, @dklyle

Issue #579, opened 2017-10-27
add support for sort image new syntax


iaroslavscript:
Hi,
there is a new syntax for sorting images.
It makes it easier to sort by multiple keys.
https://specs.openstack.org/openstack/glance-specs/specs/api/v2/lists-image-api-v2.html
https://github.com/openstack/glance/blob/stable/pike/glance/api/v2/images.py#L683

jtopjian:
Implemented in #580

PR #580, opened 2017-10-27
support for sort new syntax

iaroslavscript:
Hi!

For #579 

code in this PR:
https://github.com/openstack/glance/blob/stable/pike/glance/api/v2/images.py#L683

Links to the docs:
https://specs.openstack.org/openstack/glance-specs/specs/api/v2/lists-image-api-v2.html

### DIFF
--- a/openstack/imageservice/v2/images/requests.go
+++ b/openstack/imageservice/v2/images/requests.go
@@ -45,7 +45,10 @@ type ListOpts struct {
 	// SizeMax filters on the size_max image property.
 	SizeMax int64 `q:"size_max"`
 
-	// Sort will sort the results based on a specified image properties (allow multi field sorting).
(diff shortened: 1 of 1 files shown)


jtopjian:
@iaroslavscript Thanks for submitting this. This patch looks useful and simple, which is always good :)

I recommend adding a little more detail to the field comment. Perhaps something like:

```
// Sort sorts the results using the new style of sorting. See the OpenStack
// Image API reference for the exact syntax.                         
//                                                                           
// Sort cannot be used with the classic sort options (sort_key and sort_dir).                
```

iaroslavscript:
Hi @jtopjian thanks! I updated comments.

jtopjian:
Looks good! Thank you for working on this.

iaroslavscript:
thx :)

Issue #583, opened 2017-11-01
Identity V3: Region CRUD


dklyle:
- [x] List / Get (#584)
- [x] Create (#586)
- [x] Delete (#587)
- [x] Update (#594)

API:
https://developer.openstack.org/api-ref/identity/v3/#regions

Schema:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/backends/sql.py#L34
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/schema.py#L22

List:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L70

Get:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L78

Create:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L54

Update:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L83

Delete:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L92

jtopjian:
:tada: 

Another resource added. Thank you, @dklyle!

PR #584, opened 2017-11-01
Identity v3: Region list / get

dklyle:
For #583

code in this PR:

API:
https://developer.openstack.org/api-ref/identity/v3/#regions

Schema:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/backends/sql.py#L34
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/schema.py#L22

List:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L70

Get:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L78

### DIFF
--- a/acceptance/openstack/identity/v3/regions_test.go
+++ b/acceptance/openstack/identity/v3/regions_test.go
@@ -1,0 +1,61 @@
+// +build acceptance
+
+package v3
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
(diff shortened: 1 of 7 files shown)


jtopjian:
Looks good!

Issue #488, opened 2017-08-16
Discussion: Improvements/Fixes to Grant Access and Share Export Locations Requests


pospispa:
Grant Access and Share Export Locations requests were added in the below PRs:
- Grant Access: https://github.com/gophercloud/gophercloud/pull/346
- Share Export Locations: https://github.com/gophercloud/gophercloud/pull/381

I don't want to clutter discussions in https://github.com/gophercloud/gophercloud/issues/441 with detailed discussions whether the above PRs should be improved/fixed.

***

**Topic:** Newly added fields into results.

According to [List export locations documentation](https://developer.openstack.org/api-ref/shared-file-systems/?expanded=id236-detail) the field `preferred` was added in microversion "2.14" while other fields exist since microversion "2.9". I haven't found any change in [the Manila source code](https://github.com/openstack/manila/blob/stable/newton/manila/api/v2/share_instance_export_locations.py#L41-L50) because of this change except [this one](https://github.com/openstack/manila/blob/stable/newton/manila/api/views/export_locations.py#L79-L83).

Similarly, according to [Grant Access documentation](https://developer.openstack.org/api-ref/shared-file-systems/?expanded=grant-access-detail) the field `access_key` was added in microversion "2.21" while other fields exist since microversion "2.0". I haven't found any change in [the Manila source code](https://github.com/openstack/manila/blob/stable/newton/manila/api/v2/shares.py#L229-L236) because of this change except [this one](https://github.com/openstack/manila/blob/stable/newton/manila/api/views/share_accesses.py#L59-L61) and [this one](https://github.com/openstack/manila/blob/stable/newton/api-ref/source/parameters.yaml#L376-L382).

Even though a field was added in microversion "2.14", respectively in "2.21" the result data structures contain all fields for any microversion as seen [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/v2/shares/results.go#L121-L137) in the `ExportLocation` and [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/v2/shares/results.go#L150-L165) in `AccessRight` respectively.

However, @jtopjian proposed in https://github.com/gophercloud/gophercloud/issues/441#issuecomment-322087992 a different way how to handle extensions to results.

Should current implementation of parsing results of Grant Access and Export Locations requests be reworked?

***

**Topic:** Grant Access top element was renamed.

[...]

Issue #459, opened 2017-08-12
Keystone token expired in the middle of a long run API request


xiangfeiz:
Before starting a PR, please read the [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md).

There are some OpenStack API requests that will invoke other OpenStack API requests to get the job done. For example, a nova volume-detach request will call cinder API to detach the volume. The internal API requests rely on the Keystone token received by the first OpenStack API request from external. The token is valid at the time of the first request from external is processed, but the token may expire in the middle when the internal API requests are processed.

When it happens, the OpenStack objects are in a wrong state. And the external caller won't be able to know it has failed.

There is an config AllowReauth in gophercloud to allow re-authenticate when 401 Unauthorized is received. But it does not help the scenario here because the request from gophercloud to OpenStack returns ok because the token is valid at that time.

I don't know if this can be solved inside OpenStack by switching from token based authentication to trust based authentication (something heat engine uses). But I'd like to propose a solution inside gophercloud.

Add a config to set token expiration time, e.g. 1.5 hours. This should be less than the Keystone token expiration time. The difference is saved for the long run API request to finish all the internal API requests.
Keep track of the time when the keystone token is received.
When setting authentication header, compare the current time with the creation time of the token,  if the duration is larger than the config, request a new token.

xiangfeiz:
The issue reflected in kubernetes: https://github.com/kubernetes/kubernetes/issues/50446

jtopjian:
@xiangfeiz Thanks for reporting this. I'm not familiar with Kubernetes, but I understand what's going on.

`python-keystoneclient` does something like this [already](https://github.com/openstack/python-keystoneclient/blob/1e7543786ec60714a7723eb191bc65c31a42b0aa/keystoneclient/auth/identity/base.py) (search for `reauth`, `min_token`, and `get_access`). 

It's certainly possible for Gophercloud to do this, it's a matter of figuring out the best way to implement it. Thinking out loud:

* Reauthentication is being done [here](https://github.com/gophercloud/gophercloud/blob/50cdddf51c54e8df178af065055cbc5b3cf442e1/provider_client.go#L210)
* A check for early reauthentication would probably best be done somewhere closer to the [start](https://github.com/gophercloud/gophercloud/blob/50cdddf51c54e8df178af065055cbc5b3cf442e1/provider_client.go#L111) of the function. There's probably an element of a bootstrapping/first request scenario to account for.
[...]

PR #585, opened 2017-11-01
Add microversion as a necessary parameter to functions

chestack:
This PR is related to discussion #488.
Since microversion is necessary for GetExportLocations() and GrantAccess(), Is it better to add microversion to the functions?

For #488


openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -79,8 +79,11 @@ func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
 }
 
 // GetExportLocations will get shareID's export locations
-func GetExportLocations(client *gophercloud.ServiceClient, id string) (r GetExportLocationsResult) {
+// client must have Microversion set; minimum supported microversion for Get Export Locations is 2.14
+func GetExportLocations(client *gophercloud.ServiceClient, micoversion string, id string) (r GetExportLocationsResult) {
```
pospispa: Microversion is already a part of the client `struct` that's why I think it's not necessary to another `microversion` parametr to the `GetExportLocations` function.

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -79,8 +79,11 @@ func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
 }
 
 // GetExportLocations will get shareID's export locations
```
pospispa: Would it be enough to improve the description of the GetExportLocations function? Would it be enough to document that client.Microversion must be set and that the value must be "2.14+"?

Note: discussion in https://github.com/gophercloud/gophercloud/issues/459 concluded that gophercloud shouldn't be responsible for unclear OpenStack errors.

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -110,14 +113,17 @@ func (opts GrantAccessOpts) ToGrantAccessMap() (map[string]interface{}, error) {
 
 // GrantAccess will grant access to a Share based on the values in GrantAccessOpts. To extract
 // the GrantAccess object from the response, call the Extract method on the GrantAccessResult.
-func GrantAccess(client *gophercloud.ServiceClient, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
+// client must have Microversion set; minimum supported microversion for Grant Access is 2.7
+func GrantAccess(client *gophercloud.ServiceClient, micoversion string, id string, opts GrantAccessOptsBuilder) (r GrantAccessResult) {
```
pospispa: I have the same comments as for `func GetExportLocations()`.

openstack/sharedfilesystems/v2/shares/requests.go
chestack: @pospispa ,Thanks for your comments.  Agree that it's strange that add microversion parameter into function since microversion is already a part of the client struct. So only improve the description of the function.

openstack/sharedfilesystems/v2/shares/requests.go
chestack: Only improve the description of the function here

chestack:
@pospispa , only improve the function description. please help review.

pospispa:
LGTM.
@jtopjian PTAL

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -79,6 +79,7 @@ func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
 }
 
 // GetExportLocations will get shareID's export locations
+// client must have Microversion set; minimum supported microversion for Get Export Locations is 2.14
```
jtopjian: We'll need some punctuation here. The full comment should now read:

```
GetExportLocations will get shareID's export locations. Client must have 
Microversion set; minimum supported microversion for GetExportLocations 
is 2.14.
```

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -110,6 +111,7 @@ func (opts GrantAccessOpts) ToGrantAccessMap() (map[string]interface{}, error) {
 
 // GrantAccess will grant access to a Share based on the values in GrantAccessOpts. To extract
 // the GrantAccess object from the response, call the Extract method on the GrantAccessResult.
+// client must have Microversion set; minimum supported microversion for Grant Access is 2.7
```
jtopjian: This last line should read:

```
Client must have Microversion set; minimum supported microversion for GrantAccess is 2.7.
```

openstack/sharedfilesystems/v2/shares/requests.go
chestack: Done in next change.

openstack/sharedfilesystems/v2/shares/requests.go
chestack: Done in next change.

jtopjian:
Looks good - thank you!

PR #586, opened 2017-11-01
Identity v3 region create

dklyle:
For #583 

Note region create is unique from most of the rest of the identity resources in that you can directly specify the ID, but providing the ID string is optional.

code in this PR:

API:
https://developer.openstack.org/api-ref/identity/v3/#create-region

Schema:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/backends/sql.py#L161
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/schema.py#L36
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/schema.py#L22

Create:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L54

### DIFF
--- a/acceptance/openstack/identity/v3/identity.go
+++ b/acceptance/openstack/identity/v3/identity.go
@@ -8,6 +8,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/domains"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/groups"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/projects"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/regions"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/roles"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/users"
 )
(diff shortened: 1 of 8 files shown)


dklyle:
I've come across an implementation question. In python-keystoneclient, an `enabled` parameter is used which defaults to True. https://github.com/openstack/python-keystoneclient/blob/master/keystoneclient/v3/regions.py#L84 This value is not a column in the sql table, but instead is included in the `extra` column blob. The fun part is there seems to be no way to change the value of `enabled` or any of the contents of `extra` once the region is created. I'm going to mimic the default behavior of openstackclient and it's use of python-keystoneclient. Please advise if that's not desired. Patch update will follow shortly.

dklyle:
related keystone bug https://bugs.launchpad.net/keystone/+bug/1615076 over a year old.

jtopjian:
Ew. This is a tricky one.

So to confirm: if we only look at the API-side, the region will be disabled unless `enabled` is explicitly passed in the body? `keystoneclient` is doing the user a favor and automatically setting this to `true`?

If so, I feel like the appropriate thing to do is not mimic `keystoneclient`. If `enabled` is not explicitly passed in the `Extra` field, the region will not be enabled. It sounds a little user-unfriendly but here's my reasoning: Gophercloud aims to adhere to the API, warts and all. Gophercloud is not meant to be a user-facing library, it's meant to be the basis of user-facing applications. Therefore, the application could present the user with an `enabled` field and then perform the correct building of `Extra`. Additionally, a function called `CreateRegionLikeItShouldBeCreated` in the `gophercloud/utils` repo could be added which has an explicit enabled field/parameter.

Also, just saw your comment re the bug. To consider this behavior a bug makes sense, IMO. I would imagine the correct fix for the bug would be to make `enabled` and explicit parameter. Once that happens, Gophercloud could add an `Enabled` field. Until then, I think we should proceed without any added intervention.

Let's just make sure this behavior is documented in `doc.go`.

I'm open to debate on this, though.

dklyle:
@jtopjian Your reasoning seems sound. Adding magic always seems like a poor idea. I'm currently trying to verify the linked comment in keystoneclient regarding the role of `enabled`. I can add better documentation in doc.go to instruct the app developer to add `enabled`: true to `Extra`.

dklyle:
Given my last finding, I think this can be reviewed as is. But I'm unsure why the CI is failing.

jrperritt:
@dklyle Travis is failing [at least] due to upstream updates to Go tooling. If you rebase on `master` after #591 is merged, the unit tests should be able to run.

dklyle:
thanks @jrperritt much better

jtopjian:
Looks good!

PR #587, opened 2017-11-01
Identity v3 region delete

dklyle:
For #583 

code in this PR:

Delete:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L92


jtopjian:
Looks like the sample code in `doc.go` is a little off, but I see you've touched most of it up on the Update PR. Definitely fixable post-merge.

PR #589, opened 2017-11-02
fixed bug with endless loop when using delimiter on folded directory

awskii:
For #[588]

### DIFF
--- a/openstack/objectstorage/v1/objects/testing/fixtures.go
+++ b/openstack/objectstorage/v1/objects/testing/fixtures.go
@@ -43,11 +43,18 @@ var ExpectedListInfo = []objects.Object{
 		Name:         "hello",
 		ContentType:  "application/octet-stream",
 	},
+	{
+		Hash:         "d41d8cd98f00b204e9800998ecf8427e",
+		LastModified: time.Date(2016, time.August, 17, 22, 11, 58, 602650000, time.UTC),
+		Bytes:        0,
+		Name:         "directory",
+		ContentType:  "application/directory",
+	},
 }
 
 // ExpectedListNames is the result expected from a call to `List` when just
 // object names are requested.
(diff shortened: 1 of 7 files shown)


openstack/objectstorage/v1/objects/testing/fixtures.go
```diff
@@ -69,16 +76,25 @@ func HandleListObjectsInfoSuccessfully(t *testing.T) {
         "bytes": 14,
         "name": "goodbye",
         "content_type": "application/octet-stream"
-      },
+      },
```
jtopjian: nit: It looks like there's an extra space at the end of this line.

openstack/objectstorage/v1/objects/testing/fixtures.go
```diff
@@ -69,16 +76,25 @@ func HandleListObjectsInfoSuccessfully(t *testing.T) {
...
         "content_type": "application/octet-stream"
-      },
+      }, 
       {
         "hash": "451e372e48e0f6b1114fa0724aa79fa1",
         "last_modified": "2016-08-17T22:11:58.602650",
         "bytes": 14,
         "name": "hello",
         "content_type": "application/octet-stream"
+      },{
```
jtopjian: nit: can you break this onto two lines?

awskii:
Done.

jtopjian:
@awskii I've been able to reproduce this problem outside of a unit test -- thank you for reporting it :)

As you've mentioned, this might be part of a larger issue, which I will investigate. This might take me some time, though, but I'm definitely looking into it.

awskii:
Will PR be merged?

jtopjian:
I need to confirm if this is the appropriate fix or if the root cause is coming from another area of Gophercloud.

jrperritt:
closing in favor of #606

Issue #590, opened 2017-11-02
feature request: octavia service client


jrperritt:
The Neutron Loadbalancers API will be deprecated sometime in the [near?] future, and Octavia is its replacement. I'm told the Octavia API is a superset of the Neutron one (namely, completely Neutron-compatible).

jtopjian:
Implemented in #591

PR #591, opened 2017-11-03
add NewLoadBalancerV2 service client

jrperritt:
For #590

code in this PR:

https://github.com/openstack/openstack-ansible-os_octavia/blob/stable/pike/defaults/main.yml#L122

### DIFF
--- a/internal/testing/pkg.go
+++ b/internal/testing/pkg.go
@@ -1,0 +1,1 @@
+package testing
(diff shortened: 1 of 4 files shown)


jtopjian:
> add go file to internal/testing pkg to appease go1.9 build

I was debugging this for #586 literally right now. :) It would have taken me much longer to have figured this out. I wonder why it only started getting flagged now?

jrperritt:
Yeah, it's frustrating. It's changes to Go's tooling, some of which happen outside the standard library. The next error I get might be due to an update to the `go [fmt | vet]` tool.

jrperritt:
Right, I've verified that in the catalog generated by openstack-ansible.

PR #594, opened 2017-11-03
Identity v3 Region update

dklyle:
Note that ID cannot be updated even though it is settable on create.

There is also currently an issue with updating the Extra field. This I believe is
a bug in keystone. I have posted a bug and patch to fix. https://bugs.launchpad.net/keystone/+bug/1729933

For #583
code in this PR:

API:
https://developer.openstack.org/api-ref/identity/v3/#update-region

Schema:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/backends/sql.py#L175

Update:
https://github.com/openstack/keystone/blob/2764b49bc209245276da31693ef31f1525cf17d8/keystone/catalog/controllers.py#L83

### DIFF
--- a/openstack/identity/v3/regions/doc.go
+++ b/openstack/identity/v3/regions/doc.go
@@ -35,6 +35,19 @@ Example to Create a Region
 	if err != nil {
 		panic(err)
 
+Example to Update a Region
+
+	regionID := "TestRegion"
+
+	updateOpts := regions.UpdateOpts{
+		Description: "Updated Description for region",
+	}
+
+	region, err := regions.Update(identityClient, regionID, updateOpts).Extract()
+	if err != nil {
+		panic(err)
+	}
+
 Example to Delete a Region
 
 	regionID := "TestRegion"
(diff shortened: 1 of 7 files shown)


openstack/identity/v3/regions/doc.go
```diff
@@ -20,5 +20,39 @@ Example to List Regions
...
+		ID:             "TestRegion",
+		Description: "Region for testing"
+		Extra: map[string]interface{}{
+			"email": "testregionsupport@example.com",
+		}
+	}
+
+	region, err := regions.Create(identityClient, createOpts).Extract()
+	if err != nil {
+		panic(err)
```
jtopjian: missing closing brace

openstack/identity/v3/regions/doc.go
```diff
@@ -20,5 +20,39 @@ Example to List Regions
...
+	}
+
+Example to Delete a Region
+
+	regionID := "TestRegion"
+	err := regions.Delete(identityClient, regionID).ExtractErr()
+	if err != nil {
+	}
```
jtopjian: missing `panic(err)`

jtopjian:
Looks good. Thank you, @dklyle!

PR #597, opened 2017-11-07
Remove autocomputed etag value

yanndegat:
The autocomputed etag value should be optionnally set by the end user, not the lib. 
The computed value based on the content of the body is wrong in some usecases

e.g.: for multipart manifests, if set, the value must contain the MD5 checksum of the concatenated ETag values of the object segments.

For #[PUT ISSUE NUMBER HERE]

code in this PR:

[PUT URLS HERE]

### DIFF
--- a/openstack/objectstorage/v1/objects/requests.go
+++ b/openstack/objectstorage/v1/objects/requests.go
@@ -185,8 +185,6 @@ func (opts CreateOpts) ToObjectCreateParams() (io.Reader, map[string]string, str
 	if err != nil {
 		return nil, nil, "", err
 	}
-	localChecksum := fmt.Sprintf("%x", hash.Sum(nil))
-	h["ETag"] = localChecksum
 
 	return buf, h, q.String(), nil
 }


jtopjian:
@yanndegat Thank you for reporting this, however, there is some information missing that is detailed in the PR template:

For #[PUT ISSUE NUMBER HERE]

code in this PR:

[PUT URLS HERE]

Please let me know if you have any questions or help.

jrperritt:
I think in this case the library currently does what most users want. If you don't want to set the ETag, you can pass in your own `CreateOpts` object (possibly by just wrapping the `objects.CreateOpts`) with a custom `ToObjectCreateParams` method to `Create`.

jtopjian:
@jrperritt I've been meaning to look into this one at some point.

multi-part objects might warrant enough of a core/generic case to be included in Gophercloud or at least make ETag computation optional so everyone who wants to do SLO/DLO doesn't have to create their own CreateOpts.

The next steps to investigating this should be to write some example code to do SLO/DLO and see if any problems or barriers occur.

jrperritt:
I still think computing the ETag is what most users want, and that it's not burdensome to pass in custom options. However, I do think it'd be worth adding a new package (or functions) for large object operations.

PR #598, opened 2017-11-07
Fix X-Static-Large-Object json unmarshalling

yanndegat:
X-Static-Large-Object json is a string representation of a bool

For #735

### DIFF
--- a/openstack/objectstorage/v1/objects/results.go
+++ b/openstack/objectstorage/v1/objects/results.go
@@ -141,10 +141,11 @@ func (r *DownloadHeader) UnmarshalJSON(b []byte) error {
 	type tmp DownloadHeader
 	var s struct {
 		tmp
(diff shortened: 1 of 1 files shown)


jtopjian:
@yanndegat Thank you for this PR as well, but you must create issues that describe the reasons for the PR as well as point out areas in the OpenStack code (in this case, Swift) which show that the code changes are valid.

mrhillsman:
@yanndegat do you have an issue associated with this PR? If not, please proceed to close it, thank you for resolving.

jtopjian:
This PR escaped me and I never followed up with @yanndegat on it. It looks like a legitimate issue that warrants some investigation, so I've opened #735 as a reference issue.

mrhillsman:
@jtopjian awesome! Thanks for checking in on it.

jtopjian:
Superseded by #945

Issue #588, opened 2017-11-02
Objects v1: endless loop when perform objects.List on directory with subdirectories


awskii:
Hi.

When you have non empty directory with subdirectories, and specify Delimiter in options, you will have an endless loop because of endless asking server with `&marker=`.  Also we have empty names instead of `$(dirname)/`, e.g.: we have one directory in container on swift server and asking listing of directory and got next:
```
test
test/
```
which reduces by Gophercloud to that:
```
test
''
```

last parameter returned as marker so we have what we have.

I think, problem may be much more deeper than that, but i update tests  and you can see by yourselves. Don't forget to set test timeout.

Also, i provide an easy solution in PR.

jtopjian:
@awskii I've pinpointed the problem and have a working solution here:

https://github.com/jtopjian/gophercloud/commits/objectsv1-subdir

Would you be able to test this and let me know if it works for you?

The above might not be a final version -- I think there's room for some improvement in my implementation, but it _should_ fix the problem.

Details of the problem: Gophercloud is not detecting the existence of a `subdir` entry. When a `subdir` entry exists, it should be used as the paging marker. This is done [here](https://github.com/openstack/python-swiftclient/blob/c50823ebf1fc3d691e1589321fb03a6464604530/swiftclient/client.py#L910-L913) in the Python `swiftclient` package.

awskii:
Ok, i will test it.

awskii:
All my integration tests has passed.

jtopjian:
@awskii That's great news. I updated my branch with a cleaner function to check for the marker. When you have time, can you test once more? If it works, I will open a PR for review.

awskii:
@jtopjian Seems like works fine, nothing has been broken.

jtopjian:
Perfect. Thank you so much for testing.

jrperritt:
closed  by #606

PR #606, opened 2017-11-11
Object Storage v1: Subdir Support

jtopjian:
For #588 
Supersedes #589 

https://github.com/openstack/python-swiftclient/blob/c50823ebf1fc3d691e1589321fb03a6464604530/swiftclient/client.py#L910-L913

### DIFF
--- a/openstack/objectstorage/v1/objects/results.go
+++ b/openstack/objectstorage/v1/objects/results.go
@@ -70,63 +70,7 @@ func (r ObjectPage) IsEmpty() (bool, error) {
 
 // LastMarker returns the last object name in a ListResult.
 func (r ObjectPage) LastMarker() (string, error) {
-	return ExtractLastMarker(r)
-}
-
-// ExtractLastMarker is a function that takes a page of objects and returns the
-// marker for the page. This can either be a subdir or the last object's name.
-func ExtractLastMarker(r pagination.Page) (string, error) {
-	casted := r.(ObjectPage)
-
-	// If a delimiter was requested, check if a subdir exists.
-	queryParams, err := url.ParseQuery(casted.URL.RawQuery)
-	if err != nil {
-		return "", err
-	}
-
-	var delimeter bool
-	if v, ok := queryParams["delimiter"]; ok && len(v) > 0 {
-		delimeter = true
-	}
-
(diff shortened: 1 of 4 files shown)


openstack/objectstorage/v1/objects/results.go
```diff
@@ -66,14 +70,63 @@ func (r ObjectPage) IsEmpty() (bool, error) {
 
 // LastMarker returns the last object name in a ListResult.
 func (r ObjectPage) LastMarker() (string, error) {
-	names, err := ExtractNames(r)
+	return ExtractLastMarker(r)
```
jtopjian: Since this is the only line in the `LastMarker` method, I debated not making a new function at all. Let me know if that's the better way to go.

jtopjian:
@jrperritt ping for review. If this is merge-able, I'd like to request not squashing so @awskii's original solution gets credit.

openstack/objectstorage/v1/objects/results.go
jrperritt: I think a separate function is OK if it'll be used by functions other than this one. Alternatively, if it'll only be used by this `LastMarker`, we could leave the function and lower-case it?

jtopjian: Thanks for pointing this out. I saw a few ways of implementing this and wasn't sure the best method.

To confirm: you're saying leave everything as-is but rename `ExtractLastMarker` to `extractLastMarker`?

Is there any benefit to moving the body of the `ExtractLastMarker` function to the body of `LastMarker`?

jtopjian: Oh nevermind - I read your comment too quickly. Separate function is fine :)

jrperritt: Right, an unexported, separate function is OK.

> Is there any benefit to moving the body of the ExtractLastMarker function to the body of LastMarker?

If `ExtractLastMarker` is only called in one place (`LastMarker`), then we probably don't need it. Up to you.

jtopjian:
@jrperritt ping for review.

jtopjian:
arg. an extraneous file ended up in here.

jtopjian:
OK, there we go. The seemingly simplest things.

Issue #601, opened 2017-11-08
How to get the reason for a failed servers.Create?


sb10:
I currently do something like:

    server, err := servers.Create(p.computeClient, keypairs.CreateOptsExt{
        CreateOptsBuilder: createOpts,
        KeyName:           resources.ResourceName,
    }).Extract()
    if err != nil {
        return
    }
    
    // in a go routine (code simplified)
    for {
        select {
        case <-ticker.C:
            current, err := servers.Get(p.computeClient, server.ID).Extract()
            if err != nil {
                return
            }
            if current.Status == "ACTIVE" {
                // report success
                return
            }
            if current.Status == "ERROR" {
                // report failure
                return
            }
        }
    }

Periodically checking the server Status is what your servers.WaitForStatus() method does. But if the status is ERROR, how do you get the reason for the error?

For example, it may have gone in to error state because "No valid host was found. There are not enough hosts available." (code 500). But how do I get gophercloud to tell me this?

jtopjian:
@sb10 Sorry for the late reply, I somehow missed this one.

I believe this issue is the same (or similar) to https://github.com/gophercloud/gophercloud/issues/243. This is probably the last area of Gophercloud I haven't familiarized myself with, so I can't say for certain if there's a workaround or how to best fix this.

I'm going to close this one in favor of #243, but opening this has renewed my interest, so I'll try to make some time to look into it.

If what you're reporting is altogether different than #243, do let me know and I'll re-open this.

sb10:
No, I think #243 would be an error returned by the api, eg. in my above code it I would capture that error on line 6.

I'm talking about a _successful_ `servers.Get(`) call that that has .Status == "ERROR", but no explanation of the error.

At the bottom of https://developer.openstack.org/api-guide/compute/faults.html you can see an example "faults" section that has the information I'd like to see, though I've no idea how to get the OpenStack API to actually furnish you with a faults section.

jtopjian:
Ah, I see.

When you do `nova --debug show <uuid>`, what's the field that contains the details of the error?

sb10:
The field is called 'fault', here's an example:

> | fault                                | {"message": "Conflict updating instance c2ce4dea-b73f-4d01-8633-2c6032869281. Expected: {'task_state': [u'spawning']}. Actual: {'task_state': None}", "code": 500, "created": "2017-11-11T07:58:39Z"} |

jtopjian:
[...]

PR #607, opened 2017-11-11
Compute v2: Server Fault field

jtopjian:
For #601 

Fault server field: https://github.com/openstack/nova/blob/d79f79abc2c6d457fc88ff3f2e6eaeb478fb5981/nova/api/openstack/compute/views/servers.py#L157-L160

Fault schema: https://github.com/openstack/nova/blob/d79f79abc2c6d457fc88ff3f2e6eaeb478fb5981/nova/api/openstack/compute/views/servers.py#L318-L342


jtopjian:
@sb10 Let me know if this works for you.

Issue #605, opened 2017-11-11
Incomplete Identity v3 services API support


dklyle:
- [x] Create (#610)
- [x] Update (#611)
- [x] List (#612)

Looking at https://developer.openstack.org/api-ref/identity/v3/#service-catalog-and-endpoints

The implementation for `Create` and `Update` services are incomplete in the options that are provided. 
The schema https://github.com/openstack/keystone/blob/master/keystone/catalog/backends/sql.py#L53
supports `id`, `type`, `enabled`, and `extra`. At a minimum, being able to set and update `enabled` is important. The API docs explicitly call out `description` and `name` as well.

Create currently supports setting `type` only
Update currently supports update `type` only.

I'm posting this to ask for suggestions on how to flesh out the rest. The methods `Create` and `Update` already exist and in order to keep backward compatibility we can't just change the method definition. 

Is the answer something as simple creating new methods with similar names?

This is common for a few APIs that have an older implementation. So just looking for a strategy to tackle them.

jtopjian:
Thank you for pointing this out.

Regardless of the reason this happened (simple oversight, `type` being the only settable attribute at the time, etc), IMO, the best solution is to rewrite the function to use the standard `CreateOpts` and `UpdateOpts` pattern. This is a breaking change, but I think it's acceptable here. It should be a one-time thing and doing this will ensure future attributes can be supported without further breakage.

Gophercloud makes no guarantees about breaking changes. An example of a recently merged breaking change is https://github.com/gophercloud/gophercloud/pull/445. 

Of course, it's always best to avoid breaking changes, but it the best long-term solution involves breaking, so be it. The README warns about this and to make sure Gophercloud is vendored. Vendoring is a way to ensure the application can opt-in to this change when it's ready.

> The API docs explicitly call out description and name as well.

From looking a the schema, it seems these fields might be implicitly part of `Extra`?

dklyle:
That's how I interpret it. Once I code it up, I will know for sure. But the python kwargs look like the mechanism `description` and `name` are passed through. Seems strange for keystone to call out optional `extra` fields specifically in the API docs, but that seems like what they've done.

dklyle:
After testing and working on this, `description` and `name` are indeed stored in `extra`. Even more fun is that `name` is a supported filter which I tested and it works.

jtopjian:
@dklyle Thank you for fixing this up :)

PR #610, opened 2017-11-16
Identity v3 fix/update Service define and create

dklyle:
This updates how service creation is done. It is a backward incompatible change. That said, I tested the current implementation and it fails as the body does not contain a service object. So there should not be a lot of users being broken by this change.

I've tried to break up updating Services in logical chunks. This PR got a little more of the bulk so that I could get the tests to pass and try to make it more DRY.

The API doc for this is interesting: 
- https://developer.openstack.org/api-ref/identity/v3/#create-service
There are two named parameters, name and description, that only reside in extra. And enabled gets set to True by default whether passed or not.

For #605

code in this PR:

Schema:
- https://github.com/openstack/keystone/blob/master/keystone/catalog/backends/sql.py#L53
- https://github.com/openstack/keystone/blob/master/keystone/catalog/schema.py#L52
- https://github.com/openstack/keystone/blob/master/keystone/catalog/schema.py#L58

Create:
- https://github.com/openstack/keystone/blob/master/keystone/catalog/controllers.py#L108

### DIFF
--- a/openstack/identity/v3/services/requests.go
+++ b/openstack/identity/v3/services/requests.go
@@ -5,10 +5,52 @@ import (
 	"github.com/gophercloud/gophercloud/pagination"
 )
 
+// CreateOptsBuilder allows extensions to add additional parameters to
+// the Create request.
+type CreateOptsBuilder interface {
+	ToServiceCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts provides options used to create a service.
+type CreateOpts struct {
+	// Type is the type of the service.
+	Type string `json:"type"`
+
+	// Enabled is whether or not the service is enabled.
+	Enabled *bool `json:"enabled,omitempty"`
+
+	// Extra is free-form extra key/value pairs to describe the service.
+	Extra map[string]interface{} `json:"-"`
+}
+
+// ToServiceCreateMap formats a CreateOpts into a create request.
+func (opts CreateOpts) ToServiceCreateMap() (map[string]interface{}, error) {
+	b, err := gophercloud.BuildRequestBody(opts, "service")
+	if err != nil {
+		return nil, err
+	}
+
+	if opts.Extra != nil {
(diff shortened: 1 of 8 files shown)


openstack/identity/v3/services/requests.go
```diff
@@ -5,10 +5,52 @@ import (
...
+	ToServiceCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts provides options used to create a service.
+type CreateOpts struct {
+	// Type is the type of the service.
+	Type string `json:"type"`
+
+	// Enabled is whether or not the service is enabled.
+	Enabled bool `json:"enabled,omitempty"`
```
jtopjian: This should be `*bool`.

It looks like services are enabled by default, so when you pass `Enabled: false`, `omitempty` won't include the field in the body. By using a pointer-to-bool, this causes `omitempty` to only omit the field if it wasn't specified at all.

Perhaps confusingly, `bool` is correct in the result struct.

dklyle: Ah yes, I originally made the same mistake on update, but noticed it there. Thanks. Rolled updated PRs.

jtopjian:
This looks good to me.

@jrperritt This is essentially a rewrite of the services package. Given the missing fields and that @dklyle mentioned the original wasn't working, I don't see a problem with merging this, but wanted to run it by you first.

jtopjian:
@jrperritt When you have a moment, just looking for a yes/no on this change. I've verified everything is correct according to the API, but was wondering if the original implementation was done for a certain reason.

jrperritt:
> but was wondering if the original implementation was done for a certain reason.

Not as far as I know. I suspect we just didn't know there were more fields.

This change seems correct.

jtopjian:
Sorry for the delay - busy week at work. Thank you for looking this over.

@dklyle Thank you for fixing this up.

PR #611, opened 2017-11-16
Identity v3: Services update

dklyle:
This updates how service update is done. It is a backward incompatible change.

For #605 

code in this PR:

Schema:

https://github.com/openstack/keystone/blob/master/keystone/catalog/backends/sql.py#L53
https://github.com/openstack/keystone/blob/master/keystone/catalog/schema.py#L52
https://github.com/openstack/keystone/blob/master/keystone/catalog/schema.py#L65

Update:
- https://github.com/openstack/keystone/blob/master/keystone/catalog/controllers.py#L130


jtopjian:
This looks good, though pending #610

PR #612, opened 2017-11-16
Identity v3: Services list

dklyle:
This updates how service list is done. It is a backward incompatible change in that it removes two paging parameters that don't exist.

For #605 

code in this PR:

Schema:
- https://github.com/openstack/keystone/blob/master/keystone/catalog/backends/sql.py#L53
- https://github.com/openstack/keystone/blob/master/keystone/catalog/schema.py#L52
- https://github.com/openstack/keystone/blob/master/keystone/catalog/schema.py#L65

List:
- https://github.com/openstack/keystone/blob/master/keystone/catalog/controllers.py#L116


jtopjian:
Also looks good.

jtopjian:
Tested and working - thanks!

Issue #614, opened 2017-11-17
Compute v2: Rename IP and RAM quota fields


jtopjian:
During review of #564, I noticed that the IP and RAM fields were named awkwardly. It was out of scope to rename them during that PR, so I'm going to submit a new PR to fix this up.

jtopjian:
Fixed in #615

PR #615, opened 2017-11-17
Quotaset field rename

jtopjian:
For #614


jtopjian:
@jrperritt @dklyle Ping for review

dklyle:
Of course I could be missing something.

jtopjian:
@dklyle Yep, that's included in this PR -- unless I'm missing something, too :)

jrperritt:
@jtopjian Good catch on these

Issue #310, opened 2017-03-27
token renewal is not thread-safe


majewsky:
I can't see anything in the default token reauth functions ([`v2auth`](https://github.com/gophercloud/gophercloud/blob/9a5595b8ffe429439c63781cdc312c254e4ad990/openstack/client.go#L93) and [`v3auth`](https://github.com/gophercloud/gophercloud/blob/9a5595b8ffe429439c63781cdc312c254e4ad990/openstack/client.go#L144)) that protects against multiple threads trying to renew a single client's token at the same time.

I worked around this in my client code by implementing a custom reauth function that [locks a mutex](https://github.com/sapcc/limes/blob/master/pkg/limes/realdriver.go#L81-L82), and this seems to mostly work, but it's not perfect: There can still be duplicate token renewals when multiple threads enter the reauth function at the same time. (When one reauth function call is done, the mutex is unlocked, and the next thread will take the lock and reauth again immediately.)

In any way, it would be nice if gophercloud offers thread-safety by default here, so that ProviderClient instances can be shared between threads without any weird surprises.

armstrongli:
We got the same issue in our production when there're multiple go routines sending requests at the same time. And we have a serious problem that if one of the openstack service is unstable, this client sends infinite re-auth to keystone.

jrperritt:
I'll commit to thinking about this, but not necessarily to doing anything about it. Gophercloud's view on this is that concurrency should be handled in the application.

> And we have a serious problem that if one of the openstack service is unstable, this client sends infinite re-auth to keystone.

@armstrongli A gist of some logs would be helpful. You've claimed that Gophercloud is causing your issues (and it may well be), but you haven't provided any proof.

armstrongli:
@jrperritt There's one case that there're 100(e.g.) go routines do requests, and token gets expired, and go routines start doing reauth. the shared client token is updated by different go routines at different time without any thread safe mechanism. 
Here is one case which causing issues: routine A starts sending service request after `reauth`; routine B start reauth as the previous call failed with 401, and sets the token to be `""`. routine A fails to send service request and gets 401 error.
routine B could encounter the same issue as routine A.

jrperritt:
@armstrongli Yes, I understand what the code does; I asked for the logs that show what you're describing. Either way, as I mentioned above, all that synchronization can and should be handled in your application.

armstrongli:
The logs show that there're a lot of reauth when the token gets expires. There should be only one reauth.

```
...
[...]

PR #616, opened 2017-11-19
add token lock for thread safe re-auth

armstrongli:
For #310

How it works:
 - add lock on token refresh
 - `Authenticaters()` lock the token in memory. It's `ns` lock
 - `reauth()` lock is long lock compared with `Authenticaters()`, it sends request to keystone and unlock it after token got. It's `x * 100ms` lock


jrperritt:
closing this in favor of #636 . @armstrongli thank you for working on this and raising the issue.

Issue #603, opened 2017-11-09
Cannot retry object storage requests with raw body


idrarig:
At this line the `io.Seeker` interface is used (if implemented) to rewind the request body: https://github.com/gophercloud/gophercloud/blob/master/provider_client.go#L219

However, the `io.Reader` returned at https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/objects/requests.go#L191 is not an `io.Seeker`, hence requests retried after re-authentication fail with error code 422 (unprocessable entity).

Here is a quick fix that I'm using:

```
commit d9330d03ac54e4e2b9fbf23b88649dc351fa1a25 (HEAD -> retries)
Author: Nicola Girardi <nicola@aloc.in>
Date:   Thu Nov 9 21:01:16 2017 +0000

    Enable retries for object storage requests

diff --git a/openstack/objectstorage/v1/objects/requests.go b/openstack/objectstorage/v1/objects/requests.go
index f67bfd15..fa883e0a 100644
--- a/openstack/objectstorage/v1/objects/requests.go
+++ b/openstack/objectstorage/v1/objects/requests.go
@@ -187,8 +187,9 @@ func (opts CreateOpts) ToObjectCreateParams() (io.Reader, map[string]string, str
        }
        localChecksum := fmt.Sprintf("%x", hash.Sum(nil))
        h["ETag"] = localChecksum
-
-       return buf, h, q.String(), nil
+       // Wrap the bytes in a bytes.Reader, which implements io.Seeker, so its contents can be read more than once
+       // when being retried by provider_client.go.
+       return bytes.NewReader(buf.Bytes()), h, q.String(), nil
 }
 
 // Create is a function that creates a new object or replaces an existing
```

jtopjian:
@jrperritt If you're familiar with this area, do you mind taking a look? If not, I can make some time to verify.

jrperritt:
@idrarig That change looks good. Feel free to send a PR

idrarig:
Thanks! I'll do that in a week when I'm back... I haven't looked into
testing this though...

PR #617, opened 2017-11-19
Enable retries for object storage requests

idrarig:
For #603

### DIFF
--- a/openstack/objectstorage/v1/objects/requests.go
+++ b/openstack/objectstorage/v1/objects/requests.go
@@ -187,8 +187,9 @@ func (opts CreateOpts) ToObjectCreateParams() (io.Reader, map[string]string, str
 	}
 	localChecksum := fmt.Sprintf("%x", hash.Sum(nil))
 	h["ETag"] = localChecksum
-
-	return buf, h, q.String(), nil
+	// Wrap the bytes in a bytes.Reader, which implements io.Seeker, so its contents can be read more than once
+	// when being retried by provider_client.go.
+	return bytes.NewReader(buf.Bytes()), h, q.String(), nil
 }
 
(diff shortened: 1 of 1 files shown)


jrperritt:
Looking good. Needs at least a unit test; acceptance test would of course be good too.

idrarig:
Hey @jrperritt I was trying to understand a bit more. I see that commit 28256b311 made implementing `io.Seeker` optional for the raw body. But if that's needed to retry after reauthentication, why isn't it simply required to be an `io.ReadSeeker` rather than an `io.Reader`?

jrperritt:
> But if an io.Seeker is needed to retry after reauthentication, why isn't the raw body required to be an io.ReadSeeker rather than an io.Reader?

Good question. Consider the scenario where the user doesn't enable re-authentication.

idrarig:
@jrperritt I don't understand, could you explain more? I'll try to rephrase my question. Given that the point of my PR is making object creation work after reauthentication, the question then is, why fix this particular instance of the problem, and not solve the problem more generally, as in, make all API calls (that use a raw body) work after reauthentication? What is the right thing to do?

gmauleon:
Do you guys still working on that?
I keep having the problem on my side, I'll use @idrarig fix compiled on my side for now.

Thanks!

jtopjian:
I'll try to follow-up with this. It looks like the outstanding item is a unit test which I agree is important for this but could also be a little difficult to put together. I'll give it shot when I have some time.

With regard to the discussion about not implementing `io.Seeker` for all API calls, my best guess is that in cases where reauthentication is disabled (or not requested), a seeker might have some adverse side effects. Either that or it's simply to only use a seeker when absolutely necessary. I could be wrong on both, though.

Issue #613, opened 2017-11-17
BuildQueryString does not build map type data structure


zengchen1024:
For the codes at https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v2/volumes/requests.go#L85-L100. The value of 'r' in the following codes will be wrong.
opt := ListOpts{
    Metadata: map[string]string{"a":"b", }
    Name: "test"
}
r, _ := opt.ToVolumeListQuery()
r is "?name=test" , but the expect value is "?metadata=%7B%27a%27%3A%27b%27%7D&name=test". The reason is BuildQueryString does not build map type data like ListOpts.Metadata.

TommyLike:
This looks like a valid issue. Does it exist in other commands? Cinder have other APIs that support key&value filters.

jtopjian:
I apologize for the late reply. I looked into this the other night but forgot to follow up.

The fact that `BuildQueryString` does not support maps is indeed a valid problem. I'm not sure when I will have time to look into this, but if someone wanted to work on it, I would be happy to review.

It's also probably best to point out where in the Cinder API maps/hashes are supported for list operations and to show a few real URLs as examples.

jtopjian:
Implemented in #619

PR #619, opened 2017-11-20
fix BuildQueryString does not build map type data structure

zengchen1024:
For #613


jtopjian:
@zengchen1024 Thank you for submitting this. There are a few things that need addressed.

First, while I believe that having a map/hash in a query string exists in OpenStack, we need to see an example of it in some OpenStack project's source code. Please provide a link to some such example. This will help us understand the scope of what needs to be implemented.

Second, the `testing` directory already has a `params_test.go` file, so the tests should be put in that file. If possible, it's best to integrate the map tests in the existing tests.

While the code you've provided works, I think it can be cleaned up a bit. To start, I think we can safely assume (at least to start, or at least until proven otherwise), that all query maps will be of `map[string]x`. I'd go as far as to say that they would all be `map[string]string`, but let's defer that for now.

I was able to rewrite the function to this:

```go
          case reflect.Map:
            if v.Type().Key().Kind() == reflect.String {
              var s []string
              for _, k := range v.MapKeys() {
                var value string
                key := k.String()

                switch v.Type().Elem().Kind() {
                case reflect.String:
                  value = v.MapIndex(k).String()
                case reflect.Int:
                  value = strconv.FormatInt(v.MapIndex(k).Int(), 10)
                case reflect.Bool:
                  value = strconv.FormatBool(v.MapIndex(k).Bool())
                }
                opt := fmt.Sprintf("'%s':'%s'", key, value)
                s = append(s, opt)
              }
              params.Add(tags[0], "{"+strings.Join(s, ", ")+"}")
            }
          }
```

If it is safe to assume that all query maps would be `map[string]string`, then we can simplify the above even more by removing the case statements for `Int` and `Bool`.

zengchen1024:
@jtopjian Thanks for your review! One of examples is listing the Cinder volumes (https://github.com/openstack/cinder/blob/master/cinder/api/v2/volumes.py#L83-L85)

jtopjian:
@zengchen1024 Thanks for the link. I searched the `python-cinderclient` package and found these:

* https://github.com/openstack/python-cinderclient/blob/d345ab236ca4a1897612861215ec1ef7bb2c1073/cinderclient/tests/unit/v3/test_shell.py#L124-L127
* https://github.com/openstack/python-cinderclient/blob/fd0a11e7d7bec10a18b8bc0b209e5620fca61590/cinderclient/tests/unit/v3/test_volumes.py#L125-L130

I think the above is good enough to prove that this feature exists.

I'm still wondering if we should only support cases of `map[string]string`. This is for two reasons:

1. As far as I know, all `Metadata` fields in Gophercloud are defined as `map[string]string`.
2. The values of the map are being converted to string, anyway.

@jrperritt What are your thoughts on this implementation as well as the decision to support `map[string]interface{}` or `map[string]string`?

jrperritt:
> What are your thoughts on this implementation as well as the decision to support map[string]interface{} or map[string]string?

Yeah, I think just supporting `map[string]string` is fine.

params.go
```diff
@@ -347,12 +347,21 @@ func BuildQueryString(opts interface{}) (*url.URL, error) {
...
+							}
+							params.Add(tags[0], fmt.Sprintf("{%s}", strings.Join(s, ", ")))
+						}
 					}
 				} else {
 					// Otherwise, the field is not set.
 					if len(tags) == 2 && tags[1] == "required" {
 						// And the field is required. Return an error.
-						return nil, fmt.Errorf("Required query parameter [%s] not set.", f.Name)
+						return &url.URL{}, fmt.Errorf("Required query parameter [%s] not set.", f.Name)
```
jtopjian: This should be `nil` instead of `&url.URL{}`.

zengchen1024: It seems that other methods which invoke BuildQueryString do not check the error first, but call String() of first returned value directly, which will leads to a run time error. see
https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v2/volumes/requests.go#L99

zengchen1024: https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/requests.go#L63

jtopjian: ah, yes, that would definitely case a problem. Good catch.

jtopjian:
This looks good to me. Travis reports it's working, but the openlab tests have a failure. I'm unable to connect to status.openlabtesting.org, so I can't see what the failure is.

jtopjian:
@zengchen1024 Thank you for your work on this.

PR #624, opened 2017-11-22
add simple tenant usage

hydroflame:
continuity of #565, messed up the git tree.


openstack/compute/v2/extensions/simpletenantusage/doc.go
```diff
@@ -0,0 +1,26 @@
+/*
+Package simpletenantusage provides information and interaction with the
+SimpleTenantUsage extension for the OpenStack Compute service.
+
+Example to get usage for an entire region
+	page, err := simpletenantusage.Get(computeClient).AllPages()
+	if err != nil {
+		return err
+	}
+	usage, err := simpletenantusage.ExtractSimpleTenantUsage(page)
```
jtopjian: This should actually be `ExtractSimpleTenantUsages`

openstack/compute/v2/extensions/simpletenantusage/doc.go
```diff
@@ -0,0 +1,26 @@
...
+	if err != nil {
+		return err
+	}
+
+Example to get usage for a particular tenant.
+	page, err := simpletenantusage.GetTenant(computeClient, tenantID).AllPages()
+	if err != nil {
+		return err
+	}
+	usage, err := simpletenantusage.ExtractSimpleTenantUsages(page)
```
jtopjian: This should actually be `ExtractSimpleTenantUsage`

openstack/compute/v2/extensions/simpletenantusage/requests.go
```diff
@@ -0,0 +1,72 @@
...
+		url += query
+	}
+	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
+		return SimpleSingleTenantUsagePage{pagination.SinglePageBase(r)}
+	})
+}
+
+// GetOpts are options for fetching the tenant usage for a Tenant.
+// All int-values are pointers so they can be nil if they are not needed
+// you can use gopercloud.IntToPointer() for convenience
```
jtopjian: The above comment needs to be rewritten to reflect the new `GetOpts` options.

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// TenantUsage is a set of usage information about a tenant over the sampling window
+type TenantUsage struct {
+	// ServerUsages is an array of ServerUsage maps
+	ServerUsages []ServerUsage `json:"server_usages"`
+
+	// Start is the beginning time to calculate usage statistics on compute and storage resources
+	Start gophercloud.JSONRFC3339MilliNoZ `json:"start"`
```
jtopjian: This should be `time.Time`.

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// Stop is the ending time to calculate usage statistics on compute and storage resources
+	Stop gophercloud.JSONRFC3339MilliNoZ `json:"stop"`
```
jtopjian: This should be `time.Time`.

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,130 @@
...
+	TotalMemoryMBUsage float64 `json:"total_memory_mb_usage"`
+
+	// TotalVCPUsUsage multiplies the number of virtual CPUs of the server by hours the server exists, and then adding that all together for each server
+	TotalVCPUsUsage float64 `json:"total_vcpus_usage"`
+}
+
+// ServerUsage is a detailed set of information about a specific instance inside a tenant
+type ServerUsage struct {
+	// EndedAt is the date and time when the server was deleted
+	EndedAt gophercloud.JSONRFC3339MilliNoZ `json:"ended_at"`
```
jtopjian: This should be `time.Time`.

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,130 @@
...
+	LocalGB int `json:"local_gb"`
+
+	// MemoryMB is the memory size of the server (in MB)
+	MemoryMB int `json:"memory_mb"`
+
+	// Name is the name assigned to the server when it was created
+	Name string `json:"name"`
+
+	// StartedAt is the date and time when the server was started
+	StartedAt gophercloud.JSONRFC3339MilliNoZ `json:"started_at"`
```
jtopjian: This should be `time.Time`.

hydroflame:
@jtopjian I have addressed everything except the fixtures, I'm having lots of trouble making either the `nova` or `openstack` binary generate the request I want from our openstack infra. However the code works just fine so I'm assuming that since it works with both real data and fixtures I'm assuming that the fixtures are somewhat representative of the real world. However if you're willing to help with fixtures generation I'm willing to change them.

jtopjian:
@hydroflame I've generated some test fixtures here:

https://gist.github.com/jtopjian/aa773fb2d18185381d0a7128b7b306a8

jrperritt:
I put myself as a reviewer on here just to make sure some things got addressed. I've only skimmed it, but some of the names are off: `Get` looks like it should be `List`, and the options interface ends in `Map` even though it returns a query string. I could be mistaken about those; I'll try to find time to look more closely tomorrow.

jtopjian:
@jrperritt you're right. I've been focused on the larger implementation issues and overlooked the naming. Implementing "simple tenant usage" doesn't seem to be so simple. I appreciate @hydroflame sticking with this.

@hydroflame The time parsing is correct! I've verified that it's working.

I spotted another issue, though, and unfortunately it lead to a series of problems:

`GetOpts` is missing the `Detailed` field. This was included in the previous PR but it has now been removed.

Right now, `GetOpts` is being shared between both `Get` and `GetTenant`. The `Detailed` field is only valid when querying all tenants (`Get` in this case). This means that we now need two separate request structures: one for querying per tenant and one for querying all tenants.

Having the optional time fields in the query is difficult to begin with. Adding an `int` into the mix makes the implementation even more unique. I've given a suggestion of how to implement this toward the end of this comment.

Also, @jrperritt is correct here. There are some naming conventions that are going to need dealt with. `Get` is normally reserved for retrieving a single resource. If it was possible to retrieve one single `ServerUsage` entry, `Get` would make sense.

Given that two distinct set of request options are needed (one with `Detailed` and one without `Detailed`), and since there's been relevant code dropped along the way, I have to suggest opening two separate PRs. This is going to make reviewing and implementing each one more simple (no pun intended).

If there's significant renaming happening with regard to `Get`, it might also be a good idea to look at renaming other parts in general. `SimpleTenantUsage` and `SimpleTenantUsages` matches the API, but those names can lead to a lot of confusion. How about:

* SimpleTenantUsage is changed to:

```
SingleTenantOpts
SingleTenantOptsBuilder
ToSingleTenantQuery
SingleTenant
SingleTenantPage
ExtractSingleTenant
singleTenantURL
```

`SingleTenantOpts` will look like:

```go
type SingleTenantOpts struct {
  // The ending time to calculate usage statistics on compute and storage resources.
  End *time.Time `q:"end"`

  // The beginning time to calculate usage statistics on compute and storage resources.
  Start *time.Time `q:"start"`
}
```

`ToSingleTenantQuery` will look like:

```go
func (opts SingleTenantOpts) ToSingleTenantQuery() (string, error) {
	params := make(url.Values)
	if opts.Start != nil {
		params.Add("start", opts.Start.Format(gophercloud.RFC3339MilliNoZ))
	}

	if opts.End != nil {
		params.Add("end", opts.End.Format(gophercloud.RFC3339MilliNoZ))
	}

	q := &url.URL{RawQuery: params.Encode()}
	return q.String(), nil
}
```

`SingleTenant` will look like:

```go
func SingleTenant(client *gophercloud.ServiceClient, tenantID string, opts SingleTenantOptsBuilder) pagination.Pager {
	url := singleTenantURL(client, tenantID)
	if opts != nil {
		query, err := ToSingleTenantQuery()
		if err != nil {
			return pagination.Pager{Err: err}
		}
		url += query
	}
	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
		return SingleTenantPage{pagination.SinglePageBase(r)}
	})
}
```

* SimpleTenantUsages is changed to:

```
AllTenantsOpts
AllTenantsOptsBuilder
ToAllTenantsQuery
AllTenants
AllTenantsPage
ExtractAllTenants
allTenantsURL
```

`AllTenantsOpts` will look like:

```go
type AllTenantsOpts struct {
  // Detailed will return detailed results if the value is set to 1.
  Detailed int `q:"detailed"`

  // The ending time to calculate usage statistics on compute and storage resources.
  End *time.Time `q:"end"`

  // The beginning time to calculate usage statistics on compute and storage resources.
  Start *time.Time `q:"start"`
}
```

Notice how I've removed the `json` tags here. They're not needed since this information is not going in the body of the request.

`ToAllTenantsQuery` will look like:

```go
func (opts AllTenantsOpts) ToAllTenantsQuery() (string, error) {
  params, err := gophercloud.BuildQueryString(opts)
  if err != nil {
    return "", err
  }

  q := params.Query()

  if opts.Start != nil {
    q.Add("start", opts.Start.Format(gophercloud.RFC3339MilliNoZ))
  }

  if opts.End != nil {
    q.Add("end", opts.End.Format(gophercloud.RFC3339MilliNoZ))
  }

  params = &url.URL{RawQuery: q.Encode()}
  return params.String(), nil
}
```

etc

Each of the above sets would be submitted as a separate PR. I suggest working on one at a time.

hydroflame:
Ok so that explains why details didn't work for me, I think i was using it with the single tenant get and that endpoint doesn't have details, also do you actually mind if I put everything in 1 PR ? I don't mind breaking my own code for now, it's fairly simple, not really a problem for me.

jtopjian:
>  I don't mind breaking my own code for now, it's fairly simple, not really a problem for me.

We have a house rule where PRs must be limited to one "request". I'm pretty flexible with this rule, but since it's now clear that there are two distinct requests (two sets of options, two sets of request functions, two sets of extract functions, etc), it'll have to be broken up into two PRs.

The first PR is going to be larger as it'll include all of the ServerUsage fields and such. But having a PR dedicated to only single tenant usage results and then all tenant usage results is going to help make reviewing this easier.

I do understand that this places more work on you and I apologize for that.

hydroflame:
it looks like its actually gonna be more work for you because the code is very similar for both AllTenant and SingleTenants, but ok

jrperritt:
@hydroflame 

> if there's something you dont like you'll have to see it twice in 2 PR

Not necessarily. You should submit 1 PR, get feedback, let it merge, then submit the other one with lessons-learned from the first.

hydroflame:
so just to be sure, you want me to remove either AllTenants or SingleTenants from this PR, finish that one left then do the other in another PR ?

jtopjian:
> so just to be sure, you want me to remove either AllTenants or SingleTenants from this PR, finish that one left then do the other in another PR ?

Yes, please remove one or the other. We want to focus on one of AllTenants or SingleTenant.

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,138 @@
...
+
+// IsEmpty determines whether or not a SimpleSingleTenantUsagePage is empty.
+func (page SimpleSingleTenantUsagePage) IsEmpty() (bool, error) {
+	ks, err := ExtractSimpleTenantUsage(page)
+	return ks == nil, err
+}
+
+// ExtractSimpleTenantUsage is a function that attempts to interpret any SimpleTenantUsage resource response as a SimpleTenantUsage struct.
+// The difference between ExtractSimpleTenantUsage and ExtractSimpleTenantUsages is that when a tenant ID is provided the JSON is
+// "tenant_usage" (singular) which is a struct, otherwise it is "tenant_usages" (plural) which is an array of structs.
```
jtopjian:
```
// ExtractSingleTenant interprets a SingleTenantPage as a TenantUsage result.
```

openstack/compute/v2/extensions/simpletenantusage/results.go
```diff
@@ -0,0 +1,138 @@
...
+func ExtractSimpleTenantUsage(page pagination.Page) (*TenantUsage, error) {
```
jtopjian: `ExtractSingleTenant`

hydroflame:
@jtopjian addressed everything, I'm ok with the naming convention you want

openstack/compute/v2/extensions/simpletenantusage/doc.go
```diff
@@ -0,0 +1,28 @@
...
+        Start: &start,
+        End: &end,
+    }
+
+    page, err := simpletenantusage.SingleTenant(computeClient, tenantID, singleTenantOpts).AllPages()
+    if err != nil {
+        panic(err)
+    }
+
+    usage, err := simpletenantusage.ExtractSingleTenantUsage(page)
```
jtopjian: This either needs to be `ExtractSingleTenant` or change the function name in `results.go` to `ExtractSingleTenantUsage`. I can go both ways on this.

openstack/compute/v2/extensions/simpletenantusage/doc.go
```diff
@@ -0,0 +1,28 @@
+/*
+Package simpletenantusage provides information and interaction with the
+SimpleTenantUsage extension for the OpenStack Compute service.
+
+Example to Retrieve Usage for a Single Tenant:
+
+	start := time.Date(2017, 01, 21, 10, 4, 20, 0, time.UTC)
+	end := time.Date(2017, 01, 21, 10, 4, 20, 0, time.UTC)
```
jtopjian: minor nit: spacing

hydroflame: I don't know what you mean here
https://play.golang.org/p/CjyvDX94Cv

jtopjian: I apologize. What I meant was that these two lines are indented further in than they should be:

https://github.com/hypermassiveblackhole/gophercloud/blob/4e16d78279dc6aa8f42a662a224a3de5bcfb634f/openstack/compute/v2/extensions/simpletenantusage/doc.go

hydroflame:
@jtopjian @jrperritt status update ?

jrperritt:
I haven't had time to review this again, but I noticed 2 things skimming it:
1. Unless I missed them, there aren't any acceptance tests
2. I'm not a fan of the package name. Understandably it's named after the OpenStack extension, but `simpletenantusage.SingleTenant` stutters and sounds weird to me. I wonder if `usage` wouldn't be more appropriate. It's true that other services could have `usage` packages, but they can by aliased easily enough.

jtopjian:
@jrperritt to confirm, you're talking about renaming the package from `simpletenantusage` to simply `usage`?

jrperritt:
>  you're talking about renaming the package from simpletenantusage to simply usage?

right

jtopjian:
OK - I just wanted to doubly check. I didn't want @hydroflame to accidentally rename something else after all of the other rename work he's done :)

hydroflame:
I don't care, just tell me what you guys want to be accepted, renaming is easy and I want to get rid of our fork at comcast.

jtopjian:
@hydroflame fair enough. Please go ahead and rename from `simpletenantusage` to `usage`.

Also, we are more strictly enforcing acceptance tests for features. The test would go under `acceptance/openstack/compute/v2/usage_test.go`. You can use the other files under that directory as examples to go off of. More information about acceptance tests can be found [here](https://github.com/gophercloud/gophercloud/tree/master/acceptance). Let me know if you need any help with this.

jtopjian:
From what I can see, only `doc.go` needs updated to reflect the new package name. Other than that, this looks good to me.

@jrperritt when you have a moment, can you do a full review?

acceptance/openstack/compute/v2/usage_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	endpointParts := strings.Split(client.Endpoint, "/")
+	tenantID := endpointParts[4]
+
+	page, err := usage.SingleTenant(client, tenantID, nil).AllPages()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	usage, err := usage.ExtractSingleTenant(page)
```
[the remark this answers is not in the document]
hydroflame: Which one do you want changed ?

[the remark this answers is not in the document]
jtopjian: Maybe

```go
tenantUsage, err := usage.ExtractSingleTenant(page)
```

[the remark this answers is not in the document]
jrperritt: Yes, that would be fine

openstack/compute/v2/extensions/usage/doc.go
```diff
@@ -0,0 +1,27 @@
...
+        Start: &start,
+        End: &end,
+    }
+
+    page, err := usage.SingleTenant(computeClient, tenantID, singleTenantOpts).AllPages()
+    if err != nil {
+        panic(err)
+    }
+
+    usage, err := usage.ExtractSingleTenant(page)
```
jrperritt: Still shadowing here

openstack/compute/v2/extensions/usage/requests.go
```diff
@@ -0,0 +1,54 @@
...
+	End *time.Time `q:"end"`
+
+	// The beginning time to calculate usage statistics on compute and storage resources.
+	Start *time.Time `q:"start"`
+}
+
+// SingleTenantOptsBuilder allows extensions to add additional parameters to the
+// SingleTenant request.
+type SingleTenantOptsBuilder interface {
+	ToSingleTenantQuery() (string, error)
```
jrperritt: `ToUsageSingleTenantQuery`

jtopjian:
@jrperritt Can you do a quick review of this one? The two changes look to have been made.

jrperritt:
Thanks for reminding me. LGTM

hydroflame:
Thanks guys

jtopjian:
@hydroflame No problem at all. Thank _you_ for your patience with this one :) If you wanted to submit the follow-up for all tenants, it should go much quicker.

Issue #625, opened 2017-11-24
Compute v2: Flavor Result `is_public` is incorrect


jtopjian:
The current `flavors.Flavor` result has the [`IsPublic`](https://github.com/gophercloud/gophercloud/blob/89881f8ce3ec9fe2aa2a846716dd2315aba0ffb3/openstack/compute/v2/flavors/results.go#L68) field looking for the JSON field `is_public` when it should be `os-flavor-access:is_public`:

https://github.com/openstack/nova/blob/1878039f8fcf757651c7923910a253f187891576/nova/api/openstack/compute/views/flavors.py#L48-L89

jrperritt:
closed by #626

PR #626, opened 2017-11-24
Compute v2: Fix Flavor IsPublic field

jtopjian:
For #625 

https://github.com/openstack/nova/blob/1878039f8fcf757651c7923910a253f187891576/nova/api/openstack/compute/views/flavors.py#L48-L89


jrperritt:
@jtopjian This looks good. 

As an aside, something we may want to do going forward is, in addition to having the acceptance test, check to see that the field is set to what we expect it to be. In cases like this, the test is there, but, as-is, it would've just as well passed when the field was wrong.

jtopjian:
Yeah, I was thinking the same thing when I was looking into this.

jtopjian:
btw, thank you both for reviewing.

Issue #627, opened 2017-11-27
Query parameter 'tenant' is missing from volume&snapshot  API


TommyLike:
'tenant' is required in a multi-tenant environment.

TommyLike:
First commit to gophercloud. I am not sure whether I need update testcases for this change :confused:

jtopjian:
Implemented in #628

PR #628, opened 2017-11-27
Add 'tenant' support in volume&snapshot API

TommyLike:
Query parameter `tenant` is required in a multi-tenants environment. Add this support
in blockstorage V3 APIs.

For #627

Code in Cinder:
Volumes: https://github.com/openstack/cinder/blob/master/cinder/volume/api.py#L581-L595
Snapshots: https://github.com/openstack/cinder/blob/master/cinder/volume/api.py#L638-L647

### DIFF
--- a/openstack/blockstorage/v2/snapshots/requests.go
+++ b/openstack/blockstorage/v2/snapshots/requests.go
@@ -68,6 +68,11 @@ type ListOpts struct {
 	Name     string `q:"name"`
 	Status   string `q:"status"`
 	VolumeID string `q:"volume_id"`
+	// Admin-only option. Set it to true to see all tenant snapshots.
+	AllTenants bool `q:"all_tenants"`
+	// List only snapshots that belongs to one particular tenant.
+	// Setting "AllTenants = true" is required.
+	TenantID string `q:"project_id"`
 }
 
 // ToSnapshotListQuery formats a ListOpts into a query string.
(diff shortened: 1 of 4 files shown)


jtopjian:
@TommyLike Thank you for submitting this.

It looks like Travis is failing due to a `gofmt` issue. Possibly a hanging whitespace at the end of the line? If you run `gofmt` on the changed files, this patch should then pass.

Also, would you mind updating the same files for `blockstorage/v2`? I've confirmed that these query parameters are valid there, too.

Side note: the Cinder code specified is correct, but does not directly mention the `project_id` field. Tracing the function further in Cinder as well as verifying with `python-cinderclient` shows this this change is correct.

TommyLike:
@jtopjian, thanks, it's not easy to mention all of the codes related to the 'project_id'. Glad you find it out finally :smile: . I will try to be more explicit next time.

kiwik:
Hi @TommyLike , you can leave "recheck stable/mitaka" or "recheck stable/pike" in comments to test this patch against OpenStack mitaka or pike release.

jtopjian:
>  it's not easy to mention all of the codes related to the 'project_id'.

Not to mention the mixing of `all_tenants` and `project_id` on the API side. I agree with standardizing on `TenantID` here since "tenant" is used in the `blockstorage/vX` packages.

Thank you for making all changes. This looks good to me. I'm going to follow-up with a commit which re-orders the ListOpts fields. I didn't catch this during the first review and felt it would be too nit-picking to ask for another change.

Issue #629, opened 2017-11-27
Wrong status result in Resize function comment


obedmr:
According to https://developer.openstack.org/api-guide/compute/server_concepts.html
Once resize is finished the status of an instance should be on ``VERIFY_RESIZE `` instead of ``RESIZE_VERIFY ``:
https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/requests.go#L518

jrperritt:
closed by #630

PR #630, opened 2017-11-27
Fix wrong status on Resize function comment

obedmr:
This commit changes the VERIFY status for the resize operation.
The correct VERIFY_RESIZE status was gotten from:
https://developer.openstack.org/api-guide/compute/server_concepts.html

Signed-off-by: Obed N Munoz <obed.n.munoz@intel.com>

For #629

code in this PR:

https://github.com/openstack/nova/blob/master/nova/api/openstack/common.py#L76


jrperritt:
Thank you for submitting the PR, but please do have a look at the style guide: we need references from the source code, not the docs.

obedmr:
thanks @jrperritt, I updated my PR description to the code line .

jrperritt:
@obedmr Thank you for fixing this

Issue #632, opened 2017-11-28
Extend volume is unavailable


TommyLike:
Extend volume action is missing both in blockstorage's V2&V3 APIs.

j-griffith:
Thanks @TommyLike to be clear I don't think gophercloud should try and implement every Cinder feature.  Also, we should avoid anything that isn't clearly supported by all backend devices in my opinion.

Extend makes sense to add, but I also think it's important that we select things wisely and avoid having a ton of micro-versions to maintain and deal with here.

jtopjian:
> Also, we should avoid anything that isn't clearly supported by all backend devices in my opinion.

This is a good point. While I don't want to make a firm decision here, I'm inclined to agree with this solely on the basis that we couldn't confidently test the implementation. It should be easy enough for a vendor to create a Gophercloud-compatible package that implements the third-party driver. 

> Extend makes sense to add, but I also think it's important that we select things wisely and avoid having a ton of micro-versions to maintain and deal with here.

Another good point. I haven't looked into these PRs in detail yet, but my (perhaps too lazy?) strategy was going to be: if it's available in v2, then it's also available in base v3 and shouldn't be an issue with microversions.

jtopjian:
As mentioned in #633, this is already implemented:

https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/extensions/volumeactions/requests.go#L192-L223

And again, I apologize for not catching this sooner.

PR #633, opened 2017-11-28
Supprt extend volume in v2/v3 APIs

TommyLike:
Add extend volume support in v2/v3 APIs

For #632

Action [**os-extend**](https://github.com/openstack/cinder/blob/master/cinder/api/contrib/volume_actions.py#L300) in Cinder

### DIFF
--- a/openstack/blockstorage/v2/volumes/requests.go
+++ b/openstack/blockstorage/v2/volumes/requests.go
@@ -122,6 +122,36 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 	})
 }
 
+// ExtendOptsBuilder allows extensions to add additional parameters to the
+// extend action (os-extend) request.
+type ExtendOptsBuilder interface {
+	ToVolumeExtendMap() (map[string]interface{}, error)
+}
+
+// ExtendOpts holds options for extending volumes. It is passed to the volumes.Extend
+// function.
+type ExtendOpts struct {
+	//New size of the volume in GiB
+	NewSize int `json:"new_size" required:"true"`
+}
+
+func (opts ExtendOpts) ToVolumeExtendMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "os-extend")
+}
+
+// Extend will extend the Volume with specified volume size.
+func Extend(client *gophercloud.ServiceClient, id string, opts ExtendOptsBuilder) (r ActionResult) {
+	b, err := opts.ToVolumeExtendMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Post(actionURL(client, id), b, nil, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+	})
+	return
+}
+
 // UpdateOptsBuilder allows extensions to add additional parameters to the
 // Update request.
 type UpdateOptsBuilder interface {
--- a/openstack/blockstorage/v3/volumes/requests.go
+++ b/openstack/blockstorage/v3/volumes/requests.go
@@ -122,6 +122,36 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 	})
 }
 
+// ExtendOptsBuilder allows extensions to add additional parameters to the
+// extend action (os-extend) request.
+type ExtendOptsBuilder interface {
+	ToVolumeExtendMap() (map[string]interface{}, error)
+}
+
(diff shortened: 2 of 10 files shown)


openstack/blockstorage/v2/volumes/requests.go
```diff
@@ -122,6 +122,36 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 	})
 }
 
+// ExtendOptsBuilder allows extensions to add additional parameters to the
+// extend action (os-extend) request.
+type ExtendOptsBuilder interface {
```
j-griffith: Do we need a separate opts builder for a single argument in this case?  This seems to just add an unnecessary layer.

openstack/blockstorage/v2/volumes/requests.go
```diff
@@ -122,6 +122,36 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
+	//New size of the volume in GiB
+	NewSize int `json:"new_size" required:"true"`
+}
+
+func (opts ExtendOpts) ToVolumeExtendMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "os-extend")
+}
+
+// Extend will extend the Volume with specified volume size.
+func Extend(client *gophercloud.ServiceClient, id string, opts ExtendOptsBuilder) (r ActionResult) {
```
j-griffith: I'd be curious what others think, but it also might be worth considering for cases like this (functions that have only one or two args) to just take a variable directly.  So something like:
    func Extend(client *gophercloud.Serviceclient, id string, newSize int64)

I realize there's a question of consistency, but just throwing it out there as it seems reasonable.

openstack/blockstorage/v3/volumes/requests.go
```diff
@@ -122,6 +122,36 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 	})
 }
```
j-griffith: Same comments as above of course, one other thing to consider is do we need this in V2 or just V3?  I realize there's no shortage of folks that have V2 deployed and will be using V2 but before we start doing 2x changes for every add we might want to think about a strategy/policy here.

I'm hoping to get something moving again on the non-versioned objects approach that would help with this.

openstack/blockstorage/v2/volumes/requests.go
jtopjian: It can seem overly verbose, but it's usually a good idea. At a minimum, for consistency.

openstack/blockstorage/v3/volumes/requests.go
jtopjian: > before we start doing 2x changes for every add we might want to think about a strategy/policy here.

This is the downside of doing the v2/v3 copy when Gophercloud's v2 implementation wasn't complete.

I think it is totally acceptable for someone to submit a v3 patch without the corresponding v2 patch. If this happens, I am happy to do the backporting myself. 

I realize there's a chance that the backported v2 feature may never be used. But if it helps one person down the road, it was worth the few extra steps.

Of course, I completely agree that the situation we're in is not the most efficient. At the moment, I'm personally in a position where manual maintenance is easier for me to handle than digging into a better long-term solution. I hope that changes soon, but this is the best way for me to be able to balance getting people the features they need in Gophercloud.

If you or anyone else is able to get a solution together which resolves this, I'd gladly make it a priority to review and a priority to drown the person in praise :)

j-griffith:
@jtopjian thanks for the input; I certainly wasn't objecting to anything in these PR's just staring some dialogue and wanted to get your opinion so that's great.  I'm good with the changes and will respond appropriately to the review.

openstack/blockstorage/v2/volumes/requests.go
TommyLike: Agree with @jtopjian :smile:

jtopjian:
As I mentioned earlier, I hadn't looked at this PR in detail. When I began to, I realized that this is already implemented:

https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/extensions/volumeactions/requests.go#L192-L223

The `volumeactions` package is agnostic to v2/v3, so it should seamlessly work with either a v2 or v3 client. In fact, I believe that was @j-griffith's idea way back in the `rackspace/gophercloud` repo :)

I do apologize for not catching this right off the bat, though.

I'm going to close this PR, but please let me know if there's something further to discuss here.

Issue #634, opened 2017-11-28
Force delete is not supported in volume V2&V3 APIs


TommyLike:
Action **force_delete**  is not supported in volume V2&V3 APIs.

jtopjian:
Implemented in #635 

@TommyLike Thank you for adding this. :)

PR #635, opened 2017-11-28
Feature/support force delete

TommyLike:
Add **force_delete** action support in volume actions

For #634 

Code in Cinder [admin actions](https://github.com/openstack/cinder/blob/master/cinder/api/contrib/admin_actions.py#L131)

### DIFF
--- a/openstack/blockstorage/extensions/volumeactions/requests.go
+++ b/openstack/blockstorage/extensions/volumeactions/requests.go
@@ -261,3 +261,10 @@ func UploadImage(client *gophercloud.ServiceClient, id string, opts UploadImageO
 	})
 	return
 }
+
+// ForceDelete will delete the volume regardless of state.
+func ForceDelete(client *gophercloud.ServiceClient, id string) (r ForceDeleteResult) {
+	_, r.Err = client.Post(forceDeleteURL(client, id), map[string]interface{}{"os-force_delete": ""}, nil, nil)
+	return
+}
+
--- a/openstack/blockstorage/extensions/volumeactions/requests.go
+++ b/openstack/blockstorage/extensions/volumeactions/requests.go
@@ -267,4 +267,3 @@ func ForceDelete(client *gophercloud.ServiceClient, id string) (r ForceDeleteRes
 	_, r.Err = client.Post(forceDeleteURL(client, id), map[string]interface{}{"os-force_delete": ""}, nil, nil)
 	return
 }
-
--- a/openstack/blockstorage/extensions/volumeactions/results.go
+++ b/openstack/blockstorage/extensions/volumeactions/results.go
@@ -184,3 +184,9 @@ func (r UploadImageResult) Extract() (VolumeImage, error) {
 	err := r.ExtractInto(&s)
 	return s.VolumeImage, err
 }
+
+
+// ForceDeleteResult contains the response body and error from a ForceDelete request.
+type ForceDeleteResult struct {
+	gophercloud.ErrResult
+}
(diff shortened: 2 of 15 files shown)


TommyLike:
I create this branch based on my extend volume branch #633 , it looks like this pr includes all of the changes in these two branches. Have no idea how to handle it correctly :disappointed:

jtopjian:
> I create this branch based on my extend volume branch #633 , it looks like this pr includes all of the changes in these two branches. Have no idea how to handle it correctly 

There are two ways of doing this:

1. If this PR is completely independent of #633, then create a new branch from master.
2. If this PR requires #633, then this PR will wait until #633 is merged. If there were changes required in #633, you will need to rebase.

I'll have time to review both of these either today or tomorrow - thank you for submitting them. :)

jtopjian:
Thank you for raising these points. As I've mentioned, I'm OK with the 2x implementation, but your thoughts and background knowledge on `os-force_delete` are definitely appreciated.

> the other concerns however are around the fact that force-delete rarely works by itself without some additional clean up around attachments.

Would you say that `os-force_delete` is incomplete/dangerous enough that it's advisable to not add it? Or is it an API call that requires some additional API calls to fully work and should be used with care?

TommyLike:
@j-griffith  Cinder have force delete action and force parameter for delete API. But as they both use the same method inside. only using force delete here is better for consistency and misunderstanding.

j-griffith:
@jtopjian 
> Would you say that os-force_delete is incomplete/dangerous enough that it's advisable to not add > it? Or is it an API call that requires some additional API calls to fully work and should be used with > care?

It's a call that may need extra calls to fully work, primarily clean up of attachments, but certainly not anything that should bar us from including it here.  Just a point that there is more to it.

@TommyLike 
> Cinder have force delete action and force parameter for delete API. But as they both use the same > method inside, only using force delete here is better for consistency and avoid confusing.

My point here was that there are actually two force-delete options in the Cinder API:
1. The `cinder force-delete` cmd which is an admin-action extension in API version 2.x
2. The `cinder delete --force` cmd in API version 3.x

What I was suggesting is that rather than implement a `force-delete` function in gophercloud, why not add an option to the existing volumes.delete function and depending on API server do the right thing?

jtopjian:
@j-griffith 

> Just a point that there is more to it.

And it's much appreciated. I hope I have not come off as disagreeing or negating your input today.

@TommyLike It looks like this can be implemented as a version-agnostic "volume action":

https://github.com/gophercloud/gophercloud/tree/master/openstack/blockstorage/extensions/volumeactions

I checked, and _this_ one hasn't been added yet 😉 

With regard to the `v3/volumes/requests/DeleteOpts.Force` feature: that was was added in Cinder 3.23, which is available with the Ocata release. @j-griffith the v3 feature that seemed to be most popular during the copy/paste discussion was multi-attach. What microversion is that available in? If it's within Ocata, that might be a good de-facto max limit to run with for now.

TommyLike:
@j-griffith  hmm, the ``force delete action`` both exists in V2 and V3 while ``force delete option`` only available since 3.23. Can we support microversion in gophercloud?

jtopjian:
> Can we support microversion in gophercloud?

See https://github.com/gophercloud/gophercloud/issues/441 for details on that :)

TommyLike:
@jtopjian you are right, I need to move them to the correct place.

jtopjian:
@TommyLike This looks good to me. Is this ready to be merged?

TommyLike:
@jtopjian I think so :smile:

PR #636, opened 2017-11-29
add UseTokenLock method in ProviderClient to allow safe concurrent access

jrperritt:
For #310 

- adds a `mut` field and `UseTokenLock`, `Token`, and `SetToken` methods to `ProviderClient`
- adds a unit test for concurrent reauth
- removes the `Debug` field from `ProviderClient`, which was a relic and isn't used anywhere

Intended to be backwards-compatible (with the exception of the unused and now-removed `Debug` field).

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -84,9 +84,9 @@ func (client *ProviderClient) AuthenticatedHeaders() map[string]string {
 	return map[string]string{"X-Auth-Token": t}
 }
 
-// UseSafeReauth creates a mutex that is used to allow safe concurrent re-authentication.
+// UseTokenLock creates a mutex that is used to allow safe concurrent access to the auth token.
 // If the application's ProviderClient is not used concurrently, this doesn't need to be called.
-func (client *ProviderClient) UseSafeReauth() {
+func (client *ProviderClient) UseTokenLock() {
 	client.mut = new(sync.RWMutex)
 }
 
--- a/provider_client.go
+++ b/provider_client.go
@@ -75,20 +75,33 @@ type ProviderClient struct {
 // AuthenticatedHeaders returns a map of HTTP headers that are common for all
 // authenticated service requests.
 func (client *ProviderClient) AuthenticatedHeaders() map[string]string {
-	if client.mut != nil {
-		client.mut.RLock()
-		defer client.mut.RUnlock()
-	}
-	if client.TokenID == "" {
+	t := client.Token()
+	if t == "" {
 		return map[string]string{}
 	}
-	return map[string]string{"X-Auth-Token": client.TokenID}
+	return map[string]string{"X-Auth-Token": t}
 }
 
 // UseSafeReauth creates a mutex that is used to allow safe concurrent re-authentication.
 // If the application's ProviderClient is not used concurrently, this doesn't need to be called.
 func (client *ProviderClient) UseSafeReauth() {
 	client.mut = new(sync.RWMutex)
+}
+
+func (client *ProviderClient) Token() string {
+	if client.mut != nil {
+		client.mut.RLock()
+		defer client.mut.RUnlock()
+	}
+	return client.TokenID
+}
+
+func (client *ProviderClient) SetToken(t string) {
+	if client.mut != nil {
+		client.mut.Lock()
+		defer client.mut.Unlock()
+	}
+	client.TokenID = t
 }
 
 // RequestOpts customizes the behavior of the provider.Request() method.
--- a/provider_client.go
+++ b/provider_client.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"sync"
 )
 
 // DefaultUserAgent is the default User-Agent string set in the request header.
@@ -68,16 +69,26 @@ type ProviderClient struct {
 	// authentication functions for different Identity service versions.
 	ReauthFunc func() error
 
-	Debug bool
+	mut *sync.RWMutex
 }
 
 // AuthenticatedHeaders returns a map of HTTP headers that are common for all
 // authenticated service requests.
 func (client *ProviderClient) AuthenticatedHeaders() map[string]string {
+	if client.mut != nil {
+		client.mut.RLock()
+		defer client.mut.RUnlock()
+	}
 	if client.TokenID == "" {
 		return map[string]string{}
 	}
 	return map[string]string{"X-Auth-Token": client.TokenID}
+}
+
+// UseSafeReauth creates a mutex that is used to allow safe concurrent re-authentication.
+// If the application's ProviderClient is not used concurrently, this doesn't need to be called.
+func (client *ProviderClient) UseSafeReauth() {
+	client.mut = new(sync.RWMutex)
 }
 
 // RequestOpts customizes the behavior of the provider.Request() method.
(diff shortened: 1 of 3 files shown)


provider_client.go
```diff
@@ -68,18 +69,28 @@ type ProviderClient struct {
...
+	}
 	if client.TokenID == "" {
 		return map[string]string{}
 	}
 	return map[string]string{"X-Auth-Token": client.TokenID}
 }
 
+// UseLock creates a mutex that is used to allow safe concurrent access to a ProviderClient.
+// If the application's ProviderClient is not used concurrently, this doesn't need to be called.
+func (client *ProviderClient) UseLock() {
```
jrperritt: Maybe there's a better name for this?

jrperritt: I decided to change it to `UseSafeReauth`, but still open to suggestions.

jtopjian:
This looks good to me. Seems like a pretty standard implementation of Go's mutex.

Before I approve the PR, I have one question about the test case: why is there a lock within the reauth function? I removed it and the test still passes. That makes sense to me since the reauth function is  being executed within a lock, so I would imagine that everything within the reauth function would already be locked.

jrperritt:
> why is there a lock within the reauth function? 

Good question. The lock around the reauth function is the newly added `ProviderClient.mut`. The lock in the unit test is on the `info` object, which is written to in the reauth function but also read in the route handler.

jrperritt:
@jtopjian Expanding a bit more, I think you're correct that the `info` lock in the unit test is not necessary if everything is working correctly, but in general, when an object is able to be both written and read concurrently like that one, it'll need a synchronizing mechanism (like a mutex or channel).

jtopjian:
OK, perfect - thanks. I was trying to distinguish whether the `info` lock was a required part of `UseSafeReauth` or if it was a separate lock implementation. If that makes sense.

jrperritt:
Right. It's just there to verify that only one goroutine tries to reauth in the unit test.

provider_client.go
```diff
@@ -210,7 +219,15 @@ func (client *ProviderClient) Request(method, url string, options *RequestOpts)
 			}
 		case http.StatusUnauthorized:
 			if client.ReauthFunc != nil {
-				err = client.ReauthFunc()
+				if client.mut != nil {
+					client.mut.Lock()
+					if curtok := client.TokenID; curtok == prereqtok {
```
armstrongli: the token is one public field and can be touched outside. It would be better to make it private and set up interface for touching token to avoid pollutions.
From this part, the token can be dirty, too, and the reauth may not work.

jrperritt: The token is public for historical reasons (namely, Gophercloud predates `internal` directory magic), but applications need not (and should not) set it. If an application is setting the token on its own, it's the onus of the application to do it safely.

jrperritt: However, your point is valid and is something to consider in the future if we decide to make that backward-incompatible change.

jrperritt: @armstrongli A compromise could be adding public get and set functions for the token, in which contain the `ProviderClient` locking/unlocking. Would that be helpful? Or maybe that's what you're suggesting (with the addition of unexporting the token)?

armstrongli: yes. exactly

jrperritt: Good idea. I've added them.

armstrongli:
@jrperritt code looks good to me. Thank you for your commit

jtopjian:
Agreed. This change looks to be backwards compatible with existing ways of token authentication.

Looks good to me. Again, I'll let you merge when you're ready.

jrperritt:
@jtopjian @armstrongli thank y'all for the reviews. I'll leave this open until the morning-time here in case there's any more commentary. If not, I'll squash the commits and merge.

jrperritt:
@jtopjian @armstrongli I made 2 changes to this PR this morning. The first change is to the method name: `UseSafeReauth` -> `UseTokenLock`. With the addition of the access methods `Token` and `SetToken`, the method is no longer strictly about re-authentication, but rather token access in general. The second change is to call `UseTokenLock` in `gophercloud.NewClient`. With that, users will get the safe token access for free if they create the provider client via `NewClient` or `AuthenticatedClient`. 

Two things of note:
1. calling `UseTokenLock` only initializes the new `ProviderClient.mut` field to allow for locking/unlocking, so adding the call to `NewClient` is still expected to be backwards-compatible,
2. applications wanting to define their own reauth functions will need to be able to set the `ProviderClient.TokenID` field themselves  (as I did in the unit test in this PR), necessitating it remaining exported. This is because the `mut` is locked in order to call the reauth function, so calling `SetToken` inside the reauth function would wait indefinitely.

jtopjian:
Makes sense to me.

armstrongli:
@jrperritt
it's getting more and more similar to the patch i commited, i have to say.
the differences are
* token change interface
* lock location

armstrongli:
the reason i add the lock to the reauth function is that the lock belongs to the client, instead of the request.
the request shouldn't know or sense of the existence of the lock.

jrperritt:
> it's getting more and more similar to the patch i commited, i have to say.
the differences are
    token change interface
    lock location

Whether or not to export the mutex, where the locking happens, and handling multiple goroutines needing to reauth simultaneously is the whole thing. Our PRs differ in all those aspects.

jrperritt:
> the request shouldn't know or sense of the existence of the lock.

The request needs to grab the lock to prevent other requests from unnecessarily jumping in the re-auth function.

armstrongli:
>The request needs to grab the lock to prevent other requests from unnecessarily jumping in the re-auth function.

it's another thing. the reauth function can skip the reauth if the token is already different.
the lock needs be hidden totally in the client itself.

jrperritt:
> the reauth function can skip the reauth if the token is already different.

And how would the reauth function know what the request's original auth token was?

armstrongli:
>And how would the reauth function know what the request's original auth token was?

get the token before calling reauth. the getting token has independent lock, so the token got is thread safe and right.
then get the lock and check token again to see different or not in case of other routine updates the token.  send auth-token if the same, return directly if different.
i can update the path i have and please take a look

armstrongli:
@jrperritt 
Please take a look at the commit I have here to see how to get the original token before reauth, and how to skip redundant reauth.
https://github.com/armstrongli/gophercloud-1/commit/6e7813087c56ffc5c198963242051679c787ff22
```go

	if opts.CanReauth() {
		if client.ReauthFunc == nil {
			client.ReauthFunc = func() error {
				// get the token before reauth
				client.TokenLock.Lock()
				originToken := client.TokenID
				client.TokenLock.Unlock()

				client.TokenLock.Lock()
				defer client.TokenLock.Unlock()
				// get token in case other routine update the token
				newToken := client.TokenID
				if originToken != newToken {
					// skip reauth in case the token has been updated
					return nil
				}

				client.TokenID = ""
				return v3auth(client, endpoint, opts, eo)
			}
		}
	}
```

jrperritt:
@armstrongli Yes, I understand. Two problems immediately jump out at me with that:

1. >  to see how to get the original token before reauth

That's not necessarily the original token. The original token is in the request header, which you don't have access to there.

2. You're exporting the mutex, which we likely won't be doing.

It seems like we'll have to agree to disagree about this, however after this PR lands you're welcome to submit a PR that adds the `if client.ReauthFunc == nil {` bits.

armstrongli:
agree +1

jrperritt:
Thanks all for the lively discussion

Issue #637, opened 2017-12-01
AccTests: Add Test for BlockStorage ForceDelete


jtopjian:
With #635 merged, an acceptance test should be added.

jtopjian:
Implemented in #635

PR #638, opened 2017-12-01
AccTests: BlockStorage v2 ForceDelete

jtopjian:
For #637 

I'm curious if this will pass OpenLab tests. I believe this is an admin-only action. If the test fails, I'll modify it to only run if the username is `admin`.


jtopjian:
Looks like tests are being run as `admin`. That should be fine for now. In the long-term, I think it would be best to run tests as the `demo` (in DevStack terms) user unless admin privs are required. This is on my TODO list.

PR #639, opened 2017-12-01
Refactor volume actions' action URL method

TommyLike:
Replace volume action's xxxxURL method with actionURL

Volume actions' url is always in the format of ``/{project}/volumes/{volume_id}/action``, there is no need
to define new xxxxURL individually.

For: #640 
**side node**: This change may be discussable or already been discussed before :smile: 

Cinder API [Document](https://developer.openstack.org/api-ref/block-storage/v3/#volume-actions-volumes-action)

### DIFF
--- a/openstack/blockstorage/extensions/volumeactions/requests.go
+++ b/openstack/blockstorage/extensions/volumeactions/requests.go
@@ -47,7 +47,7 @@ func Attach(client *gophercloud.ServiceClient, id string, opts AttachOptsBuilder
 		r.Err = err
 		return
 	}
-	_, r.Err = client.Post(attachURL(client, id), b, nil, &gophercloud.RequestOpts{
+	_, r.Err = client.Post(actionURL(client, id), b, nil, &gophercloud.RequestOpts{
 		OkCodes: []int{202},
 	})
 	return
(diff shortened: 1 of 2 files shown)


jtopjian:
@TommyLike Thanks for submitting this. You'll have to open a corresponding issue, though. All PRs must have a matching Issue.

TommyLike:
@jtopjian will do, although this does not look like an issue.

jtopjian:
Agreed. It's not an _issue_, but it's where all non-code review discussion happens. For example:

> I noticed the `volumeactions` URLs can be simplified - is there benefit to doing this?

:smile:

jrperritt:
> although this does not look like an issue.

@TommyLike If you have not already done so, please read the [style guide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md). It is explained there.

TommyLike:
`please read the style guide. It is explained there.`
@jrperritt Checking this again

jtopjian:
I see no issues with doing this. @jrperritt ?

TommyLike:
@jtopjian Do I need to increase the test coverage by another commit?

jrperritt:
> Do I need to increase the test coverage by another commit?

No, this looks fine

Issue #641, opened 2017-12-02
Compute v2: Add ability to attach, detach and show a port for the given server


ozerovandrei:
Nova's os-interface extension provides the ability to manage server's interfaces.
Currently the gophercloud can only be used for list server interfaces action via the **compute/v2/extensions/attachinterfaces** extension.

I intend to make several PRs that add support for the attach, detach and show actions for the os-interface.

Doc-reference: https://developer.openstack.org/api-ref/compute/#port-interfaces-servers-os-interface

Show code-reference:
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/attach_interfaces.py#L75

Attach code-reference: https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/attach_interfaces.py#L104

Detach code-reference:
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/attach_interfaces.py#L161

jtopjian:
@ozerovandrei Thank you for submitting this work. I've left review comments on all three.

Overall, this is _really_ good work and is really appreciated - thank you very much :)

jtopjian:
Thank you for adding support for this :)

PR #642, opened 2017-12-02
Add support to get interface of a server

ozerovandrei:
Add support to get interface of a server through a GET on: /v2.1/{tenant_id}/servers/{server_id}/os-interface/{port_id}

For #641 

code in this PR:

https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/attach_interfaces.py#L75

### DIFF
--- a/openstack/compute/v2/extensions/attachinterfaces/doc.go
+++ b/openstack/compute/v2/extensions/attachinterfaces/doc.go
@@ -18,5 +18,14 @@ Example of Listing a Server's Interfaces
 	for _, interface := range allInterfaces {
 		fmt.Printf("%+v\n", interface)
 	}
+
+Example to Get a Server's Interface
+
+	interfaceID = "0dde1598-b374-474e-986f-5b8dd1df1d4e"
+	serverID := "b07e7a3b-d951-4efc-a4f9-ac9f001afb7f"
+	interface, err := attachinterfaces.Get(computeClient, serverID, interfaceID).Extract()
+	if err != nil {
+		panic(err)
+	}
 */
 package attachinterfaces
--- a/openstack/compute/v2/extensions/attachinterfaces/doc.go
+++ b/openstack/compute/v2/extensions/attachinterfaces/doc.go
@@ -21,9 +21,9 @@ Example of Listing a Server's Interfaces
 
 Example to Get a Server's Interface
 
-	interfaceID = "0dde1598-b374-474e-986f-5b8dd1df1d4e"
+	portID = "0dde1598-b374-474e-986f-5b8dd1df1d4e"
 	serverID := "b07e7a3b-d951-4efc-a4f9-ac9f001afb7f"
-	interface, err := attachinterfaces.Get(computeClient, serverID, interfaceID).Extract()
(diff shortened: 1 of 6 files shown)


openstack/compute/v2/extensions/attachinterfaces/requests.go
```diff
@@ -11,3 +11,11 @@ func List(client *gophercloud.ServiceClient, serverID string) pagination.Pager {
 		return InterfacePage{pagination.SinglePageBase(r)}
 	})
 }
+
+// Get requests details on a single interface attachment by the server and port IDs.
+func Get(client *gophercloud.ServiceClient, serverID, portID string) (r GetResult) {
+	_, r.Err = client.Get(getInterfaceURL(client, serverID, portID), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 203},
```
jtopjian: Are you sure this returns both a 200 and 203?

openstack/compute/v2/extensions/attachinterfaces/doc.go
```diff
@@ -18,5 +18,14 @@ Example of Listing a Server's Interfaces
 	for _, interface := range allInterfaces {
 		fmt.Printf("%+v\n", interface)
 	}
+
+Example to Get a Server's Interface
+
+	interfaceID = "0dde1598-b374-474e-986f-5b8dd1df1d4e"
```
jtopjian: The interface ID seems to be the PortID. Let's change this to `portID`.

openstack/compute/v2/extensions/attachinterfaces/doc.go
```diff
@@ -18,5 +18,14 @@ Example of Listing a Server's Interfaces
...
+	serverID := "b07e7a3b-d951-4efc-a4f9-ac9f001afb7f"
+	interface, err := attachinterfaces.Get(computeClient, serverID, interfaceID).Extract()
```
jtopjian: `s/interfaceID/portID`

openstack/compute/v2/extensions/attachinterfaces/requests.go
ozerovandrei: Not really. Thought that it is necessary, because `Get` for the regular server also contains that code:
https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/requests.go#L305

openstack/compute/v2/extensions/attachinterfaces/doc.go
ozerovandrei: Done: 5300702.

openstack/compute/v2/extensions/attachinterfaces/doc.go
ozerovandrei: Done: 5300702.

openstack/compute/v2/extensions/attachinterfaces/requests.go
jtopjian: That's interesting. I have a feeling the `servers.Get` is incorrect. The string "203" doesn't exist anywhere in Nova. Would you mind removing this?

jtopjian:
Looks good. Thanks!

PR #643, opened 2017-12-02
Add Nova interface-attach support

ozerovandrei:
Add support to attach interface to a server through a POST on: /v2.1/{tenant_id}/servers/{server_id}/os-interface

The same operation with the OpenStack CLI is done with:
  nova interface-attach <server>

For #641 

code in this PR:

https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/attach_interfaces.py#L104

### DIFF
--- a/openstack/compute/v2/extensions/attachinterfaces/requests.go
+++ b/openstack/compute/v2/extensions/attachinterfaces/requests.go
@@ -19,3 +19,47 @@ func Get(client *gophercloud.ServiceClient, serverID, portID string) (r GetResul
 	})
 	return
 }
+
+// CreateOptsBuilder allows extensions to add additional parameters to the
+// Create request.
+type CreateOptsBuilder interface {
+	ToAttachInterfacesCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts specifies parameters of a new interface attachment.
+type CreateOpts struct {
+
+	// PortID is the ID of the port for which you want to create an interface.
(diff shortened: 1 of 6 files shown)


openstack/compute/v2/extensions/attachinterfaces/requests.go
```diff
@@ -11,3 +11,54 @@ func List(client *gophercloud.ServiceClient, serverID string) pagination.Pager {
...
+	// Slice of FixedIPs. If you request a specific FixedIP address without a
+	// NetworkID, the request returns a Bad Request (400) response code.
+	FixedIPs []FixedIP `json:"fixed_ips,omitempty"`
+
+	// A device role tag that can be applied to a network interface when
+	// attaching it to the VM.
+	// The guest OS of a server that has devices tagged in this manner can access
+	// hardware metadata about the tagged devices from the metadata API and on
+	// the config drive, if enabled.
+	Tag string `json:"tag,omitempty"`
```
jtopjian: It looks like `tag` was added in [microversion 2.49](https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/schemas/attach_interfaces.py#L50-L52). We still haven't determined an official way of implementing features introduced in microversions. Because of that, let's remove `Tag` for now.

jtopjian:
This looks good, but will need rebased and merge conflicts resolved.

jtopjian:
Thank you :)

Issue #645, opened 2017-12-02
Deadlock on reauth


mlowery:
Perhaps introduced in #636, I think I'm seeing a deadlock. See [unit test](https://gist.github.com/mlowery/b44c5757485bcecffb9493804d3584e2) that demonstrates the issue (it never returns).

I think it's because during reauth, we're holding the [writer lock](https://github.com/gophercloud/gophercloud/blob/36c5f2c8d78de24fc1325f4b2044b46d4763cb42/provider_client.go#L242) but in `Request`, we ask for [`AuthenticatedHeaders()`](https://github.com/gophercloud/gophercloud/blob/36c5f2c8d78de24fc1325f4b2044b46d4763cb42/provider_client.go#L192) which attempts an `RLock`. This `RLock` during reauth is where it never returns for me.

As far as solutions, what is the reason to hold the write lock during reauth (instead of letting reauth func control it)? Is it because legacy reauth funcs don't know about the lock and you'd like to keep it that way?

During a reauth, why do we need authenticated headers anyway? Again, I assume this is because both reauth and every other request goes through common code.

During a write locked reauth, there is no need for a reader lock, but the code that performs the reauth is the generic `Request` and so cannot assume this right?

jrperritt:
> I think I'm seeing a deadlock ... I think it's because during reauth, we're holding the writer lock but in Request, we ask for AuthenticatedHeaders() which attempts an RLock. This RLock during reauth is where it never returns for me.

Yep, that's exactly what it is.

> what is the reason to hold the write lock during reauth

This is all discussed in #636 . But yes, we don't want to expose the lock right now.

Thank you for reporting this. I've created a PR (#646) to address this.

mlowery:
@jrperritt: In this [file](https://gist.github.com/mlowery/4d84dc0140c346c855ddc79be0e05cab), there are two tests. The first test shows a deadlock on a recursive reauth (by triggering the 401 to start a reauth then returning 401 from that reauth) with just `AllowReauths=true`. (In other words, the lock is already held in response to first 401 and another lock is attempted in response to the second 401 and there is no recursive/reentrant feature to the lock.) Now you can argue that it's my fault for not stopping the recursion. But I don't understand why the out-of-the-box reauth should itself allow a reauth in the first place.

[...]

PR #646, opened 2017-12-02
fix reauth deadlock by not calling Token() during reauth

jrperritt:
For #645

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -73,14 +73,24 @@ type ProviderClient struct {
 
 	mut *sync.RWMutex
 
-	isreauthing bool
+	reauthmut *reauthlock
+}
+
+type reauthlock struct {
+	sync.RWMutex
+	reauthing bool
 }
 
 // AuthenticatedHeaders returns a map of HTTP headers that are common for all
 // authenticated service requests.
 func (client *ProviderClient) AuthenticatedHeaders() (m map[string]string) {
-	if client.isreauthing {
-		return
+	if client.reauthmut != nil {
+		client.reauthmut.RLock()
+		if client.reauthmut.reauthing {
+			client.reauthmut.RUnlock()
+			return
+		}
+		client.reauthmut.RUnlock()
 	}
 	t := client.Token()
 	if t == "" {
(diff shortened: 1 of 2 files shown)


jrperritt:
cc @jtopjian

jtopjian:
I was pretty close to figuring this one out :)

It doesn't look like it's possible to detect the type of lock held? If it was possible, then the rlock could check if the wlock was currently in place, right?

jtopjian:
oh shoot. I now see that an rlock can't be in place when there's an existing wlock. I was going to question this the other day but I assumed that it was possible. Lesson learned.

jrperritt:
> I now see that an rlock can't be in place when there's an existing wlock

Correct. Grabbing the write lock will block all other operations. It was my fault for not putting `AuthenticationHeaders` in the unit test; that would've caught this.

mlowery:
@jrperritt: Not at my desk to validate this claim but my concern with this change is that it introduces a field that isn’t guarded with a mutex in all cases. I think a `go test -race` unit test might be able to prove this. Thoughts?

jrperritt:
> my concern with this change is that it introduces a field that isn’t guarded with a mutex in all cases

Yes, that's true, but I think in practice the worst case is that a goroutine may send an unauthorized request.

jrperritt:
The commit I just pushed show what it would look like race-free.

jrperritt:
I'm going to merge this to [hopefully] fix the bug, but I'll keep thinking about a better way to do concurrent reauth.

mlowery:
@jrperritt: Haven't had a chance to grab the latest yet. Thanks for your prompt response on this!

jrperritt:
@mlowery Of course, thank you for reporting it!

provider_client.go
```diff
@@ -72,14 +72,29 @@ type ProviderClient struct {
...
 }
 
 // AuthenticatedHeaders returns a map of HTTP headers that are common for all
 // authenticated service requests.
-func (client *ProviderClient) AuthenticatedHeaders() map[string]string {
+func (client *ProviderClient) AuthenticatedHeaders() (m map[string]string) {
+	if client.reauthmut != nil {
+		client.reauthmut.RLock()
+		if client.reauthmut.reauthing {
+			client.reauthmut.RUnlock()
```
mlowery: @jrperritt: Perhaps a slightly improved version would be:

```go
if client.reauthmut != nil {
    client.reauthmut.RLock()
    reauthing := client.reauthmut.reauthing
    client.reauthmut.RUnlock()
    if reauthing {
        return
    }
}
t := client.Token()
```

mlowery: (Removes the need for two `RUnlock` calls.)

jrperritt: > Perhaps a slightly improved version would be:

Maybe? That'd work too, but it looks like the same number of lines. Note that only one of the `RUnlock`s is ever called.

Issue #647, opened 2017-12-03
Configuration group time parsing error


eugenetaranov:
Hello,

currently database configuration group creation throws an error failing at timestamp parsing https://github.com/gophercloud/gophercloud/blob/master/openstack/db/v1/configurations/results.go#L12-L13:
`Error: parsing time ""2017-12-03T23:28:09"" as ""2006-01-02T15:04:05Z07:00"": cannot parse """ as "Z07:00"`

Example code:
```
package main

import (
	"fmt"

	"github.com/gophercloud/gophercloud"
	"github.com/gophercloud/gophercloud/openstack"
	"github.com/gophercloud/gophercloud/openstack/db/v1/configurations"
)

func main() {
	opts := gophercloud.AuthOptions{
		IdentityEndpoint: "keystone_endpoint",
		Username:         "username",
		Password:         "password",
		DomainName:       "Default",
		TenantName:       "tenant",
	}

	provider, err := openstack.AuthenticatedClient(opts)
	if err != nil {
		fmt.Println("Provider error:", err)
	}

	client, err := openstack.NewDBV1(provider, gophercloud.EndpointOpts{
		Region: "region",
	})
	if err != nil {
		fmt.Println("Client error:", err)
	}

	createOpts := &configurations.CreateOpts{
		Name:        "test",
		Description: "description",
	}

	datastore := configurations.DatastoreOpts{
		Version: "mysql-5.7",
		Type:    "mysql",
	}
	createOpts.Datastore = &datastore

	values := make(map[string]interface{})
	values["collation_server"] = "latin1_swedish_ci"
	createOpts.Values = values

	cgroup, err := configurations.Create(client, createOpts).Extract()

	if err != nil {
		fmt.Println("Error:", err)
	}

	fmt.Println(cgroup)
}
```

jtopjian:
Implemented in #648

PR #648, opened 2017-12-03
Configuration group time parsing error 

eugenetaranov:
For #647

Fixes time parsing error.

### DIFF
--- a/openstack/db/v1/configurations/results.go
+++ b/openstack/db/v1/configurations/results.go
@@ -1,19 +1,17 @@
 package configurations
 
 import (
-	"time"
-
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
 
 // Config represents a configuration group API resource.
 type Config struct {
-	Created              time.Time `json:"created"`
-	Updated              time.Time `json:"updated"`
-	DatastoreName        string    `json:"datastore_name"`
-	DatastoreVersionID   string    `json:"datastore_version_id"`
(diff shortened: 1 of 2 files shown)


jtopjian:
In order to correctly convert the time that the API server is sending back into a `time.Time` type, you'll need to implement a custom unmarshal method. 

It looks like this is already being done in the `instances` package, so you probably just have to make the appropriate changes to `configurations`:

https://github.com/gophercloud/gophercloud/blob/master/openstack/db/v1/instances/results.go#L71-L88

eugenetaranov:
@jtopjian thank you for advice, makes sense indeed.
Applied, please review

jrperritt:
@eugenetaranov Some feedback on this:
- "fix" is not a helpful title; please change it to be more descriptive. Matching the issue title would be fine.
- There needs to be an accompanying link to the OpenStack source code proving your change is correct

jtopjian:
https://github.com/gophercloud/gophercloud/pull/196 had the fix for the `databases` resource. The updated relevant link is:

https://github.com/openstack/trove/blob/stable/pike/trove/common/base_wsgi.py#L453

Since it's in the base wsgi file, I think it's safe to assume it'll apply to all time fields in the DB service.

It wouldn't be fair of me to make @eugenetaranov hunt for the code when I knew where it was, but @jrperritt is correct: please remember to provide the relevant service-side code when submitting PRs.

eugenetaranov:
thanks @jtopjian, you have saved my day :)

jtopjian:
This looks good to me and I've confirmed it works. I will follow-up with an acceptance test for this.

Issue #650, opened 2017-12-05
Compute v2: attachinterfaces acceptance test


jtopjian:
With #641 merged, we should add an acceptance test for the interface CRUD functionality.

jtopjian:
Implemented in #651

PR #651, opened 2017-12-05
Compute v2: attachinterfaces acceptance test

jtopjian:
For #650


jtopjian:
I don't believe this was run in TOL tests, but I've confirmed it works.

Issue #649, opened 2017-12-05
Add volume type support in blockstorage resource


TommyLike:
- [x] read (#657)
- [x] create (#654)
- [x] update (#656)
- [x] delete (#655)
- [ ] encryption
- [x] extra specs

jtopjian:
There's a group of API calls which manage volume types, so I think it is suitable.

https://developer.openstack.org/api-ref/block-storage/v3/index.html#volume-types-types

jtopjian:
@TommyLike Thank you for your work on this :smile:

TommyLike:
@jtopjian we still missing some APIs regarding volume type resource, here: [volume type extra specs](https://developer.openstack.org/api-ref/block-storage/v3/#create-or-update-extra-specs-for-volume-type) and [encryption](https://developer.openstack.org/api-ref/block-storage/v3/#show-an-encryption-type)

jtopjian:
@TommyLike OK, no problem. Perhaps it would be best to make a checklist of all features to be added?

winiciusallan:
The remaining checkbox was filed with https://github.com/gophercloud/gophercloud/commit/48aa6d531e91ce5c653f514fa6db2c012d431ec8, so I believe we're pretty much done with this. The updated list would now be:

- [x] read (#657)
- [x] create (#654)
- [x] update (#656)
- [x] delete (#655)
- [x] encryption (https://github.com/gophercloud/gophercloud/pull/2642)
- [x] extra specs (https://github.com/gophercloud/gophercloud/pull/2123)

If someone thinks something is still missing, feel free to re-open the issue or create a new one.

PR #653, opened 2017-12-05
Add basic CRUD support in V3 volume types

TommyLike:
Add Create/Delete/List/Update/Get support for volume type in volume V3.

For #649 

Volume types in [V3](https://developer.openstack.org/api-ref/block-storage/v3/#volume-types-types)


jtopjian:
@TommyLike Thank you for working on this. However, a PR needs to be restricted to one request type / route, as detailed in the [styleguide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md).

Please see https://github.com/gophercloud/gophercloud/issues/583 as an example of how implementing full C, R, U, and D for a resource should be handled. Let me know if you have any questions.

PR #655, opened 2017-12-06
Add DELETE support in V3 volume types

TommyLike:
Add Delete support for volume type in volume V3.

For #649 

Volume type delete [V3](https://developer.openstack.org/api-ref/block-storage/v3/#delete-a-volume-type)

PR #656, opened 2017-12-06
Add UPDATE support in V3 volume types

TommyLike:
Add Update support for volume type in volume V3.

For #649 

Volume type update in [V3](https://developer.openstack.org/api-ref/block-storage/v3/#update-a-volume-type)

Volume type update [Code](https://github.com/openstack/cinder/blob/master/cinder/api/contrib/types_manage.py#L101)

### DIFF
--- a/openstack/blockstorage/v3/volumetypes/requests.go
+++ b/openstack/blockstorage/v3/volumetypes/requests.go
@@ -101,3 +101,38 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 		return VolumeTypePage{pagination.LinkedPageBase{PageResult: r}}
 	})
 }
+
+// UpdateOptsBuilder allows extensions to add additional parameters to the
+// Update request.
+type UpdateOptsBuilder interface {
+	ToVolumeTypeUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts contain options for updating an existing Volume Type. This object is passed
+// to the volumetypes.Update function. For more information about the parameters, see
+// the Volume Type object.
+type UpdateOpts struct {
+	Name        string `json:"name,omitempty"`
+	Description string `json:"description,omitempty"`
+	IsPublic    *bool  `json:"is_public,omitempty"`
+}
+
+// ToVolumeUpdateMap assembles a request body based on the contents of an
+// UpdateOpts.
+func (opts UpdateOpts) ToVolumeTypeUpdateMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "volume_type")
+}
+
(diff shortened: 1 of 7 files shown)


TommyLike:
Need to merge delete branch #655 first.

jtopjian:
@TommyLike Looks like there's a `gofmt` error that Travis is flagging:

https://travis-ci.org/gophercloud/gophercloud/jobs/319561696#L885-L886

I'm not sure what editor you use, but it might be helpful to add some type of Go plugin that will automatically call `gofmt` upon saving. The `vim-go` plugin does this for vi/vim.

Otherwise, manually running `gofmt` on `volumetypes_test.go` will resolve this.

TommyLike:
@jtopjian Thanks I use GoLand as my personal IDE, looks like I need to find a way to execute the go fmt command automatically before pushing the changes:)

jtopjian:
@TommyLike Can you link to the Cinder code for this?

TommyLike:
@jtopjian  oops, done@

openstack/blockstorage/v3/volumetypes/requests.go
```diff
@@ -101,3 +101,38 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
+	ToVolumeTypeUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts contain options for updating an existing Volume Type. This object is passed
+// to the volumetypes.Update function. For more information about the parameters, see
+// the Volume Type object.
+type UpdateOpts struct {
+	Name        string `json:"name,omitempty"`
+	Description string `json:"description,omitempty"`
+	IsPublic    bool   `json:"is_public"`
```
jtopjian: This should be `*bool` and the json tag should be `json:"is_public,omitempty"`

`bool`s are tricky to work with. Let's say someone wants to update the name of a public volume type, so they do:

```go
updateOpts := volumetypes.UpdateOpts{
  Name: "new name",
}
```

But the request that would be sent to Cinder would be:

```json
{
  "volume_type": {
    "name": "new name",
    "is_public": false
  }
}
```

You could use `omitempty`, but that means no one could ever change a volume type from public to private because `false` would always be omitted.

Therefore, using `*bool` allows for three values: true, false, and nil. `omitempty` prevents an empty/null field from being sent to Cinder when the value is `nil`.

Search the Gophercloud repo for `iTrue` to see how passing values to bool-pointers is usually handled.

jtopjian:
@TommyLike This looks good - thank you. I'm going to follow-up with a PR to make a change to the Create options: renaming `PublicAccess` to `IsPublic`. The functionality is still the same, but it better conforms to the service-side [code](https://github.com/openstack/cinder/blob/master/cinder/api/contrib/types_manage.py#L63).

Issue #658, opened 2017-12-06
Compute v2: Support OS-EXT-STS attributes


shiftky:
Compute v2 API [/servers/detail](https://developer.openstack.org/api-ref/compute/#list-servers-detailed) supports the extended status information:

```
The extended status information appears in the OS-EXT-STS:vm_state,
OS-EXT-STS:power_state, and OS-EXT-STS:task_state attributes.
```

But currently gophercloud does not support these attributes.
I think these attributes could be useful.

jtopjian:
Implemented in #659

PR #659, opened 2017-12-06
Compute v2: Add the extended status information API

shiftky:
I added the extended status information API as ServerExtendedStatusExt result extension.

For #658 

 * [/nova/api/openstack/compute/extended_status.py](https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/extended_status.py)
 * OpenStack compute API doc
   * [List Server Details](https://developer.openstack.org/api-ref/compute/#list-servers-detailed)
   * [Show Server Details](https://developer.openstack.org/api-ref/compute/#show-server-details)

### DIFF
--- a/openstack/compute/v2/extensions/extendedstatus/results.go
+++ b/openstack/compute/v2/extensions/extendedstatus/results.go
@@ -19,8 +19,8 @@ const (
 	SUSPENDED
 )
 
-func (ps PowerState) String() string {
-	switch ps {
+func (r PowerState) String() string {
+	switch r {
 	case NOSTATE:
 		return "NOSTATE"
 	case RUNNING:
--- a/openstack/compute/v2/extensions/extendedstatus/results.go
+++ b/openstack/compute/v2/extensions/extendedstatus/results.go
@@ -11,8 +11,10 @@ type ServerExtendedStatusExt struct {
 const (
 	NOSTATE = iota
 	RUNNING
+	_UNUSED1
 	PAUSED
 	SHUTDOWN
+	_UNUSED2
 	CRASHED
 	SUSPENDED
 )
@@ -31,6 +33,8 @@ func (ps PowerState) String() string {
 		return "CRASHED"
 	case SUSPENDED:
 		return "SUSPENDED"
+	case _UNUSED1, _UNUSED2:
+		return "_UNUSED"
 	default:
 		return "N/A"
 	}
--- a/openstack/compute/v2/extensions/extendedstatus/results.go
+++ b/openstack/compute/v2/extensions/extendedstatus/results.go
@@ -1,0 +1,37 @@
+package extendedstatus
+
+type PowerState int
+
+type ServerExtendedStatusExt struct {
+	TaskState  string     `json:"OS-EXT-STS:task_state"`
+	VmState    string     `json:"OS-EXT-STS:vm_state"`
+	PowerState PowerState `json:"OS-EXT-STS:power_state"`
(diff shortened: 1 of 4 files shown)


jtopjian:
@shiftky I did a cursory review and this looks good. 

But the PR template explains that you need to provide the URLs to the Nova code which implements this on the service-side. That way we can verify the implementation is correct.

shiftky:
I have upated the PR description with addition.

jtopjian:
@shiftky Thanks.

I was able to track down the power state definitions here: https://github.com/openstack/nova/blob/master/nova/objects/fields.py#L951

It looks like there are some gaps in the sequence (note the `_UNUSED` fields). Because of this, I think the `iota` method here might misalign the actual power state of the server.

shiftky:
@jtopjian Thank you for your review!

> It looks like there are some gaps in the sequence (note the _UNUSED fields). 

I missed that! My code was not aligned properly indeed.

I committed the fix for this problem. I confirmed that it is now able to get the same result as when using OpenStack [`nova`](https://docs.openstack.org/python-novaclient/latest/cli/nova.html#see-also) command.

```
% nova list | grep -v Running
+--------------------------------------+----------------------------+---------+------------+-------------+----------------------------+
| ID                                   | Name                       | Status  | Task State | Power State | Networks                   |
+--------------------------------------+----------------------------+---------+------------+-------------+----------------------------+
| 29102f9e-2abb-4840-a17f-0807dbebfc55 | prd-db017                  | ACTIVE  | -          | NOSTATE     | provider-xxxx=10.xx.xx.xx  |
| 3ff2fa2f-747a-442d-8a83-1aaec77cc89a | web20                      | ERROR   | -          | NOSTATE     |                            |
| 618c69f1-650d-4dd8-b3d2-c4e021dc55ec | stg-web-test               | SHUTOFF | -          | Shutdown    | provider-xxxx=10.xx.xx.xx  |
+--------------------------------------+----------------------------+---------+------------+-------------+----------------------------+

% go run main.go | grep -v RUNNING
+--------------------------------------+----------------------------+---------+-------------+
|                  ID                  |            NAME            | STATUS  | POWER STATE |
+--------------------------------------+----------------------------+---------+-------------+
| 29102f9e-2abb-4840-a17f-0807dbebfc55 | prd-db017                  | ACTIVE  | NOSTATE     |
| 3ff2fa2f-747a-442d-8a83-1aaec77cc89a | web20                      | ERROR   | NOSTATE     |
| 618c69f1-650d-4dd8-b3d2-c4e021dc55ec | stg-web-test               | SHUTOFF | SHUTDOWN    |
+--------------------------------------+----------------------------+---------+-------------+
```

openstack/compute/v2/extensions/extendedstatus/results.go
```diff
@@ -0,0 +1,41 @@
...
+	RUNNING
+	_UNUSED1
+	PAUSED
+	SHUTDOWN
+	_UNUSED2
+	CRASHED
+	SUSPENDED
+)
+
+func (ps PowerState) String() string {
```
jtopjian: I apologize, I should have caught this earlier. According to the [styleguide](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md#naming) we name all receivers `r`. So change `ps` to `r` here.

After that, this PR is good to go.

jtopjian:
@shiftky Looks good! Thank you very much for contributing this and for making the requested changes.

Issue #663, opened 2017-12-08
TestNetworksProviderCRUD run failed


liu-sheng:
When try to run the Acceptance tests of networking v2 extensions API, it raised an error, seems because the *provider.ExtractGet* function is undefined.

related logs:

$ go test -v -tags "fixtures acceptance" github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2/extensions -list "Test"                         
acceptance/openstack/networking/v2/extensions/provider_test.go:29:21: undefined: provider.ExtractGet

jtopjian:
Thank you for catching these. The network and port resources had some extension work done to them a while back and the acceptance tests were never modified to account for it.

jrperritt:
@jtopjian Isn't the TOL testing intended to catch things like this?

jtopjian:
Yes, but I believe only a selective set of tests are being run. The TOL team is probably looking at expanding the set and are catching issues like this.

jrperritt:
> Yes, but I believe only a selective set of tests are being run

Ah, gotcha. Additionally, they may only be running the tests that have changed (or, better, the tests that use a package that's changed), which would be reasonable

PR #664, opened 2017-12-08
Fix the undefined function error of TestNetworksProviderCRUD test

liu-sheng:
When try to run the TestNetworksProviderCRUD test, the test will extract
info of a new created network and then print it, it extracts the network
info by "provider.ExtractGet" function, which is undefined. This change
use
the "Extract" method of returned network object to instead.

For #663


jtopjian:
Looks good to me. Thanks!

PR #667, opened 2017-12-08
Support for setting availability_zone_hints to a router

stone:
For #622


jtopjian:
Looks good to me. Thanks!

Issue #670, opened 2017-12-14
build failed of TestShareTypeExtraSpecs test of shared file system v2 API


liu-sheng:
A syntax error occurred when listing the tests of shared file system v2 API. see:

go test -v -tags "fixtures acceptance" ./acceptance/openstack/sharedfilesystems/v2  -list "Test"

acceptance/openstack/sharedfilesystems/v2/sharetypes_test.go:73:8: unknown field 'Specs' in struct literal of type sharetypes.SetExtraSpecsOpts

jtopjian:
Fixed in #671

PR #671, opened 2017-12-14
Fix a small syntax error of TestShareTypeExtraSpecs test

liu-sheng:
This test try to init a sharetypes.SetExtraSpecsOpts object with a wrong
field name, this change fix it.

For #670


jtopjian:
@liu-sheng Nice catch, thank you. 

This test was originally working, but #345 requested for it to be changed. It was changed in #517 but I missed checking if there was a corresponding acceptance test (they're a bit difficult for me to run, so if OpenLab can run them, that's awesome).

I'll merge this once all checks have fun.

jtopjian:
And there we go.

Issue #676, opened 2017-12-19
Pagination support is missing in volume and snapshot resources.


TommyLike:
Pagniation is supported in blockstorage's [volume ](https://developer.openstack.org/api-ref/block-storage/v3/#list-accessible-volumes-with-details) and [snapshot ](https://developer.openstack.org/api-ref/block-storage/v3/#list-snapshots-and-details) resources.

jtopjian:
Implemented in #678 and #679

Issue #677, opened 2017-12-19
Acceptance testcase is missing from blockstorage V3 volume&snapshot resources


TommyLike:
Acceptance testcase is missing from blockstorage V3 [volume&snapshot resources](https://github.com/gophercloud/gophercloud/tree/master/acceptance/openstack/blockstorage/v3)

jtopjian:
Implemented in #680 and #681 

Thanks, @TommyLike :)

PR #678, opened 2017-12-19
Support pagination in snapshot resources 

TommyLike:
Pagniation is supported in blockstorage's [snapshot ](https://developer.openstack.org/api-ref/block-storage/v3/#list-snapshots-and-details) resource.

Code: [Snapshot](https://github.com/openstack/cinder/blob/master/cinder/api/v3/snapshots.py#L78)
For #676

### DIFF
--- a/acceptance/openstack/blockstorage/v3/snapshots_test.go
+++ b/acceptance/openstack/blockstorage/v3/snapshots_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/snapshots"
+	"github.com/gophercloud/gophercloud/pagination"
 )
 
 func TestSnapshotsList(t *testing.T) {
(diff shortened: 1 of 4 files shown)


jtopjian:
@TommyLike Thank you for putting this together. A few minor comments:

* This should really be two separate PRs: one for snapshots and one for volumes. Both PRs can reference #676. If you were planning on doing #677, it might even make sense to first do #677 (again, two separate PRs) so it would be easy to test the paging functionality outside of the unit tests. But if you weren't going to do #677 (which is fine), let me know and I'll make sure they're added before these paging fixes are merged.

* Small nit-pick about the formatting: the `ListOpts` has a newline between each field, but the fields you added don't. When you look at the file [as a whole](https://github.com/TommyLike/gophercloud/blob/9449bdd576aadd28e94ad996127e962f6314e125/openstack/blockstorage/v3/snapshots/requests.go#L65-L91) it looks inconsistent.

TommyLike:
@jtopjian  will do

jtopjian:
@TommyLike This looks good as well. Can you add an acceptance test similar to the volume one?

jtopjian:
This looks good to me (tests pass, too :). Thank you, @TommyLike :)

PR #685, opened 2017-12-20
Only try to re-auth once

jrperritt:
For #645 

This change uses a stripped `ProviderClient` and `AuthOptions` to make the re-auth request. The intent is to only try to re-authenticate once to avoid deadlock with the `ProviderClient.mutex`. The deadlock can occur if the first re-auth attempt fails; this is due to the recursive use of the `ProviderClient`, which this PR aims to fix.

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -159,9 +159,21 @@ func v2auth(client *gophercloud.ProviderClient, endpoint string, options gopherc
 	}
 
 	if options.AllowReauth {
+		// here we're creating a throw-away client (tac). it's a copy of the user's provider client, but
+		// with the token and reauth func zeroed out. combined with setting `AllowReauth` to `false`,
+		// this should retry authentication only once
+		tac := *client
+		tac.ReauthFunc = nil
+		tac.TokenID = ""
+		tao := options
+		tao.AllowReauth = false
 		client.ReauthFunc = func() error {
-			client.TokenID = ""
-			return v2auth(client, endpoint, options, eo)
+			err := v2auth(&tac, endpoint, tao, eo)
+			if err != nil {
+				return err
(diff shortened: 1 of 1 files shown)


jrperritt:
@mlowery I figured you were busy with other things so I decided to try to fix this. I ran this change over the unit tests from the gist you posted in #645 , and it passed those. I didn't add them to this PR but would like to have them in the repo. Can you submit a PR that adds them?

jrperritt:
> While I haven't reproduced the issue in question in #645, I understand the point of this PR.

Right, nor have I. I think it's a case where it'd be difficult to write a reliable acceptance test, but easy to write a unit test. If @mlowery isn't able to submit those unit tests in the next week, I'll try to write one for this.

> I have a few non-blocking comments:

Thanks, I'll address these

mlowery:
@jrperritt: Thanks for this! Was out of town. Regarding a unit test, I have one which is the same as the second test function [here](https://gist.github.com/mlowery/4d84dc0140c346c855ddc79be0e05cab) but with the last three lines changed as below. Perhaps not very robust but worked for me.

```go
	errCh := make(chan error)
	go func() {
		_, err := cl.Get(cl.ServiceURL("op1"), nil, &gophercloud.RequestOpts{
			OkCodes: []int{200},
		})
		errCh <- err
	}()

	select {
	case err := <-errCh:
		t.Logf("Get returned with: %v", err)
	case <-time.After(2 * time.Second):
		t.Fatalf("assumed infinite loop")
	}
```

jrperritt:
@mlowery Sure, that sounds fine. If you want (and have the time) to submit a PR for the unit test, please go ahead. Otherwise I'll write one up this week.

Issue #540, opened 2017-09-29
Compute v2: Flavor Extra Specs


jtopjian:
Flavors can contain "extra specs", or additional metadata about the flavor. Gophercloud currently does not have any support for this.

- [x] List Extra Specs (#686)
- [x] Get an Extra Spec (#686)
- [x] Create an Extra Spec (#689)
- [x] Update an Extra Spec (#712)
- [x] Delete an Extra Spec (#705)

API reference: 
* https://developer.openstack.org/api-ref/compute/#flavors-extra-specs-flavors-os-flavor-extra-specs

Nova implementation: 

* https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavors_extraspecs.py

Schema: 

* https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/schemas/flavors_extraspecs.py
* https://github.com/openstack/nova/blob/stable/pike/nova/api/validation/parameter_types.py#L333-L341

Extra Specs are defined just as Server Metadata is defined, so they will be a `map[string]string`.

ss22ever:
@jtopjian I would like to take that.

jtopjian:
@ss22ever You are more than welcome to, but I'd like to ask that contributors ensure the amount of work they take on at once can be completed in a timely manner.

ss22ever:
@jtopjian Thanks, just got familiar with the project, trying my best to contribute in a timely manner.Can expect a pr soon.

dklyle:
I haven't seen any progress on this. @ss22ever, if you are still working on this let me know. I've just done the List / Get part for now, but can finish the rest shortly.

jtopjian:
:tada:

PR #686, opened 2017-12-20
Flavor Extra Specs: List / Get

dklyle:
For #540

code in this PR:

API reference:
* List
https://developer.openstack.org/api-ref/compute/#list-extra-specs-for-a-flavor

* Get
https://developer.openstack.org/api-ref/compute/#show-an-extra-spec-for-a-flavor

Nova implementation:
* List
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavors_extraspecs.py#L50
* Get
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavors_extraspecs.py#L98

Schema:
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/schemas/flavors_extraspecs.py

### DIFF
--- a/openstack/compute/v2/flavors/requests.go
+++ b/openstack/compute/v2/flavors/requests.go
@@ -230,3 +230,14 @@ func IDFromName(client *gophercloud.ServiceClient, name string) (string, error) 
 		return "", err
 	}
 }
+
+// ExtraSpecs requests all the extra-specs for the given flavor ID.
+func ListExtraSpecs(client *gophercloud.ServiceClient, flavorID string) (r ListExtraSpecsResult) {
+	_, r.Err = client.Get(extraSpecsListURL(client, flavorID), &r.Body, nil)
+	return
+}
+
+func GetExtraSpec(client *gophercloud.ServiceClient, flavorID string, key string) (r GetExtraSpecResult) {
+	_, r.Err = client.Get(extraSpecsGetURL(client, flavorID, key), &r.Body, nil)
+	return
+}
--- a/openstack/compute/v2/flavors/results.go
+++ b/openstack/compute/v2/flavors/results.go
@@ -182,3 +182,44 @@ type FlavorAccess struct {
 	// TenantID is the unique ID of the tenant.
 	TenantID string `json:"tenant_id"`
 }
+
+// Extract interprets any extraSpecsResult as ExtraSpecs, if possible.
+func (r extraSpecsResult) Extract() (map[string]string, error) {
+	var s struct {
+		ExtraSpecs map[string]string `json:"extra_specs"`
+	}
+	err := r.ExtractInto(&s)
+	return s.ExtraSpecs, err
+}
+
+// extraSpecsResult contains the result of a call for (potentially) multiple
+// key-value pairs. Call its Extract method to interpret it as a
+// map[string]interface.
+type extraSpecsResult struct {
+	gophercloud.Result
+}
+
+// ListExtraSpecsResult contains the result of a Get operation. Call its Extract
+// method to interpret it as a map[string]interface.
+type ListExtraSpecsResult struct {
+	extraSpecsResult
+}
+
+// extraSpecResult contains the result of a call for individual a single
+// key-value pair.
+type extraSpecResult struct {
+	gophercloud.Result
+}
+
+// GetExtraSpecResult contains the result of a Get operation. Call its Extract
+// method to interpret it as a map[string]interface.
+type GetExtraSpecResult struct {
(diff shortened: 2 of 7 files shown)


openstack/compute/v2/flavors/results.go
```diff
@@ -162,3 +162,43 @@ type FlavorAccess struct {
...
+		ExtraSpecs map[string]string `json:"extra_specs"`
+	}
+	err := r.ExtractInto(&s)
+	return s.ExtraSpecs, err
+}
+
+// ExtraSpecsResult contains the result of a call for (potentially) multiple
+// key-value pairs. Call its Extract method to interpret it as a
+// map[string]interface.
+type ExtraSpecsResult struct {
```
jtopjian: if `ExtraSpecsResult` will only be used internally as the basis of other types, let's rename it to `extraSpecsResult`.

openstack/compute/v2/flavors/requests.go
```diff
@@ -200,3 +200,14 @@ func IDFromName(client *gophercloud.ServiceClient, name string) (string, error)
...
 	}
 }
+
+// ExtraSpecs requests all the extra-specs for the given flavor ID.
+func ListExtraSpecs(client *gophercloud.ServiceClient, id string) (r ListExtraSpecsResult) {
+	_, r.Err = client.Get(extraSpecsListURL(client, id), &r.Body, nil)
+	return
+}
+
+func GetExtraSpecs(client *gophercloud.ServiceClient, id string, key string) (r GetExtraSpecsResult) {
```
jtopjian: Maybe change `id` here to `flavorID` so it isn't confused with a supposed ExtraSpecID.

openstack/compute/v2/flavors/results.go
```diff
@@ -162,3 +162,43 @@ type FlavorAccess struct {
...
+
+// ListExtraSpecsResult contains the result of a Get operation. Call its Extract
+// method to interpret it as a map[string]interface.
+type ListExtraSpecsResult struct {
+	ExtraSpecsResult
+}
+
+// ExtraSpecResult contains the result of a call for individual a single
+// key-value pair.
+type ExtraSpecResult struct {
```
jtopjian: Same possible lower-casing here, too.

dklyle:
Access returns a list of maps and extra-specs returns a single map. Not sure if that makes a difference, but thought I'd ask. I thought metadata looked like a closer match and tried mirroring that. But I'm certainly open to changing it.

jtopjian:
> Access returns a list of maps and extra-specs returns a single map. Not sure if that makes a difference, but thought I'd ask. I thought metadata looked like a closer match and tried mirroring that. But I'm certainly open to changing it.

:thinking: 

OK, good point. Let's keep it as-is then.

openstack/compute/v2/flavors/requests.go
jtopjian: Oh, also, should be `GetExtraSpec` singular?

dklyle: Done.

openstack/compute/v2/flavors/results.go
dklyle: Done

openstack/compute/v2/flavors/results.go
dklyle: Done

openstack/compute/v2/flavors/requests.go
```diff
@@ -200,3 +200,14 @@ func IDFromName(client *gophercloud.ServiceClient, name string) (string, error)
...
 	}
 }
+
+// ExtraSpecs requests all the extra-specs for the given flavor ID.
+func ListExtraSpecs(client *gophercloud.ServiceClient, flavorID string) (r ListExtraSpecsResult) {
+	_, r.Err = client.Get(extraSpecsListURL(client, flavorID), &r.Body, nil)
+	return
+}
+
+func GetExtraSpec(client *gophercloud.ServiceClient, flavorID string, key string) (r GetExtraSpecsResult) {
```
jtopjian: Thanks for changing the function name to singular. I think the result type also be singular (It would have to be changed in other places, too) since the "internal" type is `extraSpecResult`?

jtopjian:
@dklyle oops. The AddAccess PR which was just merged caused some file conflicts here. Sorry about that :(

jtopjian:
@dklyle This looks good - thank you for all of the changes.

One thing to change in #689 would be to make sure the `IDFromName` function is at the bottom of the `requests.go` file.

PR #687, opened 2017-12-21
Compute v2: Flavor Access Add

jtopjian:
For #506 

Code and reference is listed in #506

### DIFF
--- a/openstack/compute/v2/flavors/doc.go
+++ b/openstack/compute/v2/flavors/doc.go
@@ -59,5 +59,18 @@ Example to List Flavor Access
 	for _, access := range allAccesses {
 		fmt.Printf("%+v", access)
 	}
+
+Example to Grant Access to a Flavor
+
+	flavorID := "e91758d6-a54a-4778-ad72-0c73a1cb695b"
+
+	accessOpts := flavors.AddAccessOpts{
+		Tenant: "15153a0979884b59b0592248ef947921",
+	}
+
+	accessList, err := flavors.AddAccess(computeClient, flavor.ID, accessOpts).Extract()
+	if err != nil {
+		panic(err)
+	}
 */
 package flavors
--- a/openstack/compute/v2/flavors/requests.go
+++ b/openstack/compute/v2/flavors/requests.go
@@ -162,6 +162,36 @@ func ListAccesses(client *gophercloud.ServiceClient, id string) pagination.Pager
 	})
 }
 
+// AddAccessOptsBuilder allows extensions to add additional parameters to the
+// AddAccess requests.
+type AddAccessOptsBuilder interface {
+	ToAddAccessMap() (map[string]interface{}, error)
+}
+
+// AddAccessOpts represents options for adding access to a flavor.
+type AddAccessOpts struct {
+	// Tenant is the project/tenant ID to grant access.
+	Tenant string `json:"tenant"`
+}
+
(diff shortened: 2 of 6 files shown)


jtopjian:
I forgot I had this ready to go. Rebased and cleaned up.

Ready for review.

jtopjian:
Ah, OpenLab was helpful here:

```
compute.go:277: Attempting to create flavor flavor_o8Ozf
compute.go:293: Successfully created flavor b98d6b36-9772-4cf0-a414-b9e3a452326e
flavors_test.go:137: Unable to add access to flavor: Invalid request due to incorrect syntax or missing required parameters.
compute.go:593: Deleted flavor: b98d6b36-9772-4cf0-a414-b9e3a452326e
```

The test passes for me in Ocata. I wonder if Pike now validates the tenant ID.

jtopjian:
Yep. Hey, look at that. CI in action :smile:

openstack/compute/v2/flavors/doc.go
```diff
@@ -59,5 +59,18 @@ Example to List Flavor Access
 	for _, access := range allAccesses {
 		fmt.Printf("%+v", access)
 	}
+
+Example to Grant Access to a Flavor
+
+	flavorID := "e91758d6-a54a-4778-ad72-0c73a1cb695b"
+
+	accessOpts := flavors.AddAccessOpts{
+		Tenant: "15153a0979884b59b0592248ef947921",
```
dklyle: Since this is the documentation, shouldn't this be an UUID to provide the additional visual prompt as to an appropriate value? I know in the AddAccessOpts it states project/tenant ID, but I think it would be helpful to change here. It's a easy change, but not necessarily enough to hold up the patch on.

dklyle: The rest of the patch looks great.

jtopjian: At least on Ocata, the project and user IDs are these kinds of strings rather than UUIDs:

```
Successfully created project ACPTTESTWZC1JOxR with ID 8befa1ba8c224584bb2b7b55c6e35a24
```

 Has this changed post-Ocata?

dklyle: no, you're right. sorry for the noise.

jtopjian: It's no problem at all.

jtopjian:
@dklyle thank you for the review.

I'm going to go ahead and merge this. It's a pretty standard addition and between @dklyle's review and the OL CI catching an error, I think we're good.

openstack/compute/v2/flavors/requests.go
```diff
@@ -162,6 +162,36 @@ func ListAccesses(client *gophercloud.ServiceClient, id string) pagination.Pager
 	})
 }
 
+// AddAccessOptsBuilder allows extensions to add additional parameters to the
+// AddAccess requests.
+type AddAccessOptsBuilder interface {
+	ToAddAccessMap() (map[string]interface{}, error)
```
jrperritt: @jtopjian I think these always need to be of the form `To<Resource><Action><Return Type>` (i.e. `ToFlavorAddAccessMap`)

jtopjian: Yes, you're right. I've fixed this in #688. Sorry about that.

PR #689, opened 2017-12-21
Flavor Extra Specs Create 

dklyle:
For #540 

code in this PR:

API reference:
https://developer.openstack.org/api-ref/compute/#create-extra-specs-for-a-flavor

Nova implementation:
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavors_extraspecs.py#L61

Schema:
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/schemas/flavors_extraspecs.py#L24

### DIFF
--- a/openstack/compute/v2/flavors/doc.go
+++ b/openstack/compute/v2/flavors/doc.go
@@ -73,7 +73,23 @@ Example to Grant Access to a Flavor
 		panic(err)
 	}
 
+Example to Create Extra Specs for a Flavor
+
+	flavorID := "e91758d6-a54a-4778-ad72-0c73a1cb695b"
+
+	createOpts := flavors.ExtraSpecsOpts{
+		"hw:cpu_policy":        "CPU-POLICY",
+		"hw:cpu_thread_policy": "CPU-THREAD-POLICY",
+	}
+	createdExtraSpecs, err := flavors.CreateExtraSpecs(computeClient, flavorID, createOpts).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Printf("%+v", createdExtraSpecs)
+
 Example to Get Extra Specs for a Flavor
+
 	flavorID := "e91758d6-a54a-4778-ad72-0c73a1cb695b"
 
 	extraSpecs, err := flavors.ListExtraSpecs(computeClient, flavorID).Extract()
@@ -82,5 +98,6 @@ Example to Get Extra Specs for a Flavor
 	}
 
 	fmt.Printf("%+v", extraSpecs)
+
 */
 package flavors
--- a/acceptance/openstack/compute/v2/flavors_test.go
+++ b/acceptance/openstack/compute/v2/flavors_test.go
@@ -154,7 +154,7 @@ func TestFlavorAccessCRUD(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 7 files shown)


openstack/compute/v2/flavors/doc.go
```diff
@@ -72,5 +72,29 @@ Example to Grant Access to a Flavor
 	if err != nil {
 		panic(err)
 	}
+
+Example to Create Extra Specs for a Flavor
```
jtopjian: nit: Newline below this line.

openstack/compute/v2/flavors/doc.go
```diff
@@ -72,5 +72,29 @@ Example to Grant Access to a Flavor
...
+		"hw:cpu_thread_policy": "CPU-THREAD-POLICY",
+	}
+	createdExtraSpecs, err := flavors.CreateExtraSpecs(computeClient, flavorID, createOpts).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Printf("%+v", createdExtraSpecs)
+
+Example to Get Extra Specs for a Flavor
```
jtopjian: nit: Newline below this line.

dklyle: Done.

openstack/compute/v2/flavors/doc.go
dklyle: Done.

dklyle:
I made the suggested changes. The one I find questionable is the `CreateExtraSpecsOptsBuilder`. Mainly because the json construct in `ToExtraSpecsCreateMap` is setting `extra_specs`. This particular API doesn't leave a higher level object to hang additional parameters off of. So the usage of  `CreateExtraSpecsOptsBuilder` would be to put values into ExtraSpecOpts which is only a `map[string]string`. I imagine one could create some special values in the map to add extra parameters, but I'm not sure that would change the structure of the map. As such, I find this addition questionable. Maybe a more elaborate change is required around building the opts. Not sure.

jtopjian:
@dklyle Yeah, the standard comment we leave for the Builder interfaces doesn't make a lot of sense in this case. You're right - there's not much one can do in the way of adding additional parameters here when they can just pass them all in the original map.

The `ToExtraSpecsCreateMap` function can be used for other things beyond adding additional data, though. For example, it can validate the specs being requested so they adhere to in-house rules or something like that. In a way, the interface and function are more of a hook.

dklyle:
Fair enough. That's an explanation I can buy. I think the singular set of a specific key may need the same limit of only 1 item.

jtopjian:
> I think the singular set of a specific key may need the same limit of only 1 item.

It looks like this might be required for the Update call, although it looks like the [function](https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavors_extraspecs.py#L79) will accept an arbitrary amount of specs and as long as the key in the URL matches at least one spec, everything else will pass. But I haven't tested that theory.

If that's true, though, I would side on passing the entire map to Nova and letting it do the validation.

This all looks good, though. Thank you for the updates and definitely thank you for pushing back on suggestions :)

jtopjian:
@dklyle Let's finish off ExtraSpecs before RemoveAccess - that should minimize the amount of toes being stepped on :)

No rush, though.

Issue #690, opened 2017-12-25
Support host aggregates


ildarsv:
There is no support of the Nova host aggregates in the gophercloud.

ildarsv:
#691

jtopjian:
Implemented in #691

PR #691, opened 2017-12-25
Add os-aggregates support

ildarsv:
For #690 

Add support for the Nova host aggregates to list host aggregates.

API reference:
https://developer.openstack.org/api-ref/compute/#list-aggregates

Implementation:
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/aggregates.py#L44

### DIFF
--- a/acceptance/openstack/compute/v2/aggregates_test.go
+++ b/acceptance/openstack/compute/v2/aggregates_test.go
@@ -1,0 +1,32 @@
+// +build acceptance compute aggregates
+
+package v2
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/ildarsv/gophercloud/openstack/compute/v2/extensions/aggregates"
+)
+
+func TestAggregatesList(t *testing.T) {
+	client, err := clients.NewComputeV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create a compute client: %v", err)
+	}
+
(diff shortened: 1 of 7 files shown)


ildarsv:
Hello @jaypipes, @jtopjian ! Please review this stuff, when you have time :)

jtopjian:
@ildarsv Thank you for submitting this. I did a cursory scan and this looks good - thank you for limiting this to only List as well as including documentation.

There are three things missing, though, before I can do a full review:

1. Link to the Nova code that implements aggregates.
2. Unit tests
3. Acceptance test (even if the test returns no actual results)

ildarsv:
@jtopjian Thanks for the quick response! Sorry for the non-observance of rules.
I've added link to the Nova code and required tests. Please review when you have time.

jtopjian:
@ildarsv Thanks!

In general this looks good. I have one item for discussion, though: the list of fields defined here:

https://github.com/openstack/nova/blob/09ed3b5626ac306e59856db89858eb7bee4daa0e/nova/objects/aggregate.py#L242-L250

shows a `metadata` field. I think we should include this in the `results.go` file. It should be a simple `map[string]string` type.

I think it should be safe to leave out the `deleted` field and the timestamp fields, at least for now. There's a comment in the `objects/aggregate.py` file that mentions `delete` and `deleted_at` being deprecated.

Let me know if you have any questions or see a reason not to include `metadata`.

ildarsv:
@jtopjian Good point, thanks! I've added `metadata` field. Yes, let's skip the rest fields.

jtopjian:
This looks good to me. Thank you for working on it :)

Issue #672, opened 2017-12-15
[Enhancement] Add support for subnet pools on OpenStack


louis-gounot:
Hello,

Would it be posible to add subnet pools support for OpenStack networking.

Documentation is available here : 
- https://docs.openstack.org/neutron/pike/admin/config-subnet-pools.html (general presentation)
- https://developer.openstack.org/api-ref/network/v2/index.html#subnet-pools-extension-subnetpools (API spec)

It will also impact subnet management as subnet pool extension allows to create a subnet without providing a CIDR (we only provide a CIDR length or use the default length configured on subnet pool). So CIDR attribute from subnet will move from mandatory to exclusively mandatory with subnet-pool-id/prefixlength.

Regards,

Louis

louis-gounot:
Main use case is for private clouds (on premise cloud) where addressing plan consistency is quite usefull.

Secondary use case is the use of "address scope" extension which allows direct connectivity between tenants without SNAT/FIP NAT when tenants are in the same address scope (a bit like VPC peering on AWS).

jtopjian:
> Would it be posible to add subnet pools support for OpenStack networking.

This is certainly an applicable feature to add. If you would be interested in adding this, take a look at our [contributing guide](https://github.com/gophercloud/gophercloud/blob/master/.github/CONTRIBUTING.md). But if you aren't interested (which is fine), I'll leave this issue opened until the feature is added.

ozerovandrei:
Hi.
I would like to take this issue.

Code references:
https://github.com/openstack/neutron/blob/stable/pike/neutron/db/db_base_plugin_v2.py#L707
https://github.com/openstack/neutron/blob/stable/pike/neutron/api/v2/attributes.py#L214
https://github.com/openstack/neutron/blob/stable/pike/neutron/ipam/subnet_alloc.py#L244

jtopjian:
List has been implemented in #694. I will leave this issue open as a point of reference if someone wants to add Create, Update, and Delete.

ozerovandrei:
@louis-gounot all API requests were implemented.

jtopjian:
:tada: 

@ozerovandrei Thank you for your work on this :smile:

PR #694, opened 2017-12-28
Add Neutron subnetpool list support.

ozerovandrei:
Add support to list Neutron subnetpools through a GET request on `/v2.0/subnetpools`

`openstack subnet pool list`
or
`neutron subnetpool-list`

For #672 

code in this PR:

https://github.com/openstack/neutron/blob/stable/pike/neutron/db/db_base_plugin_v2.py#L707
https://github.com/openstack/neutron/blob/stable/pike/neutron/ipam/subnet_alloc.py#L244

### DIFF
--- a/openstack/networking/v2/extensions/subnetpools/testing/fixtures.go
+++ b/openstack/networking/v2/extensions/subnetpools/testing/fixtures.go
@@ -6,73 +6,73 @@ import (
 
 const SubnetPoolsListResult = `
 {
-	  "subnetpools": [
-	  		{
-	  				"address_scope_id": null,
-	  				"created_at": "2017-12-28T07:21:41Z",
-	  				"default_prefixlen": "8",
-	  				"default_quota": null,
-	  				"description": "IPv4",
-	  				"id": "d43a57fe-3390-4608-b437-b1307b0adb40",
-	  				"ip_version": 4,
-	  				"is_default": false,
-	  				"max_prefixlen": "32",
-	  				"min_prefixlen": "8",
-	  				"name": "MyPoolIpv4",
-	  				"prefixes": [
-	  						"10.10.10.0/24",
-	  						"10.11.11.0/24"
-	  				],
-	  				"project_id": "1e2b9857295a4a3e841809ef492812c5",
-	  				"revision_number": 1,
-	  				"shared": false,
-	  				"tenant_id": "1e2b9857295a4a3e841809ef492812c5",
-	  				"updated_at": "2017-12-28T07:21:41Z"
-	  		},
-	  		{
-	  				"address_scope_id": "0bc38e22-be49-4e67-969e-fec3f36508bd",
-	  				"created_at": "2017-12-28T07:21:34Z",
-	  				"default_prefixlen": "64",
-	  				"default_quota": null,
-	  				"description": "IPv6",
-	  				"id": "832cb7f3-59fe-40cf-8f64-8350ffc03272",
-	  				"ip_version": 6,
-	  				"is_default": true,
-	  				"max_prefixlen": "128",
-	  				"min_prefixlen": "64",
-	  				"name": "MyPoolIpv6",
-	  				"prefixes": [
-	  						"fdf7:b13d:dead:beef::/64",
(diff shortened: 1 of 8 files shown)


jtopjian:
@ozerovandrei This is really nice work - thank you!

One request I have is if you would be able to add an acceptance test? See the `acceptance/openstack/networking/v2/networks_test.go` file (or the subnets file) for an example of how to do a simple `List` test.

jtopjian:
@ozerovandrei Thanks!

I seem to be running into an issue with this. I created a subnet pool like so:

```shell
$ openstack subnet pool create --pool-prefix 192.168.0.0/16 --default-prefix-length 24 foobar
```

And when I run the acceptance test, I get:

```
=== RUN   TestSubnetPoolsList
--- FAIL: TestSubnetPoolsList (1.75s)
        subnetpools_test.go:21: Unable to list subnetpools: json: cannot unmarshal string into Go struct field SubnetPool.default_prefixlen of type int
FAIL
exit status 1
FAIL    github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2/extensions/subnetpools    1.758s
```

If I run `openstack --debug subnet pool list`, I get the following:

```json
{                                                            
    "subnetpools": [                                         
        {                                                    
            "address_scope_id": null,                        
            "created_at": "2017-12-29T16:42:38Z",            
            "default_prefixlen": "24",                       
            "default_quota": null,                           
            "description": "",                               
            "id": "c438ac8c-a9bc-4c25-a235-a212287fe9ce",    
            "ip_version": 4,                                 
            "is_default": false,                             
            "max_prefixlen": "32",                           
            "min_prefixlen": "8",                            
            "name": "foobar",                                
            "prefixes": [                                    
                "192.168.0.0/16"                             
            ],                                               
            "project_id": "47b5e614bb7b4842a685b98cb7048efb",
            "revision_number": 1,                            
            "shared": false,                                 
            "tags": [],                                      
            "tenant_id": "47b5e614bb7b4842a685b98cb7048efb", 
            "updated_at": "2017-12-29T16:42:38Z"             
        }                                                    
    ]                                                        
}                                                            
```

Notice how some of the fields which you've defined as `int` are being returned as strings. Are you seeing similar behavior or is your OpenStack environment returning integers?

If the latter, then a custom `UnmarshalJSON` function will need to be added in order to handle both string and int. There are a few places in Gophercloud which you can use as an example of this such as `openstack/compute/v2/extensions/floatingips/results.go`.

ozerovandrei:
@jtopjian yes, unfortunately I have the same issue inside my environment. But I'm using the `Newton` release for now. So I've thought that it's a some bug in that old release. 
Newest Neutron API claims that `default_prefixlen`, `max_prefixlen` and `min_prefixlen` should be integers: https://developer.openstack.org/api-ref/network/v2/index.html#list-subnet-pools

I will add a custom `UnmarshalJSON` function then.

jtopjian:
> Newest Neutron API claims that default_prefixlen, max_prefixlen and min_prefixlen should be integers: https://developer.openstack.org/api-ref/network/v2/index.html#list-subnet-pools

Unfortunately this is why we can't use the API reference as the only source of truth. The Response Parameter table indeed shows those fields as being `int`, but the Response Example shows them as being strings. The only way to know for sure is to dig into the Neutron code itself + validate with an acceptance test. It's a pain, but it makes sure everything is correct :smile:

openstack/networking/v2/extensions/subnetpools/testing/fixtures.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/subnetpools"
+)
+
+const SubnetPoolsListResult = `
+{
+	  "subnetpools": [
+	  		{
+	  				"address_scope_id": null,
+	  				"created_at": "2017-12-28T07:21:41Z",
+	  				"default_prefixlen": 8,
```
jtopjian: `"8"`

etc

jtopjian:
LGTM! Thank you, @ozerovandrei :)

PR #695, opened 2017-12-28
BlockStorage v3: Rename VolumeType PublicAccess to IsPublic

jtopjian:
This commit renames the VolumeType CreateOpts PublicAccess field
to IsPublic.

For #649 

https://github.com/openstack/cinder/blob/master/cinder/api/contrib/types_manage.py#L63-L64

### DIFF
--- a/acceptance/openstack/blockstorage/v3/volumetypes_test.go
+++ b/acceptance/openstack/blockstorage/v3/volumetypes_test.go
@@ -53,16 +53,17 @@ func TestVolumeTypesCRUD(t *testing.T) {
 	}
 
 	createOpts := volumetypes.CreateOpts{
-		Name:         "create_from_gophercloud",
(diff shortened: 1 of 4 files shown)


jtopjian:
This one is a little confusing because the service looks for two different `is_public` fields:

* Create: https://github.com/openstack/cinder/blob/master/cinder/api/contrib/types_manage.py#L63-L64
* Update: https://github.com/openstack/cinder/blob/master/cinder/api/contrib/types_manage.py#L111

The result returns both of these fields:

* `is_public`
* `os-volume-type-access:is_public`

Both mirror each other and will always have the same setting.

Renaming `PublicAccess` to `IsPublic` for consistency makes sense.

Ping @jrperritt and/or @dklyle for review.

jtopjian:
@jrperritt Thanks!

Issue #696, opened 2017-12-29
Objects v1: containers metadata missing policy name


thiagodasilva:
The container's GetHeader struct is missing the 'X-Storage-Policy' header.

jtopjian:
Implemented in #697

PR #697, opened 2017-12-29
Add X-Storage-Policy to GetHeader struct

thiagodasilva:
The X-Storage-Policy header is missing from the
container's GetHeader struct.

For #696


jtopjian:
@thiagodasilva Thank you for submitting this. Before we can review this, can you post the relevant Swift code which implements this header?

thiagodasilva:
@jtopjian 
Thanks for looking at this, here are the links:

The container servers returns the policy index:
https://github.com/openstack/swift/blob/master/swift/container/server.py#L65

and the proxy adds the policy name to the response:
https://github.com/openstack/swift/blob/master/swift/proxy/controllers/base.py#L1883

jtopjian:
LGTM - thank you :)

PR #701, opened 2018-01-02
Add Neutron subnetpool get support

ozerovandrei:
Add support to show a Neutron subnetpool through a GET request on
/v2.0/subnetpools/{subnetpool_id}.
`openstack subnet pool show`
or
`neutron subnetpool-show`

For #672 

code in this PR:

https://github.com/openstack/neutron/blob/stable/pike/neutron/db/db_base_plugin_v2.py#L707
https://github.com/openstack/neutron/blob/stable/pike/neutron/ipam/subnet_alloc.py#L244


jtopjian:
LGTM - thank you :)

PR #703, opened 2018-01-03
Expose the real error message of BadRequest error

liu-sheng:
This change expose the error message of BadRequest error from openstack
API, this is helpful to find out the reason of request failure.

For #243

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -103,7 +103,11 @@ type ErrDefault503 struct {
 }
 
 func (e ErrDefault400) Error() string {
-	return "Invalid request due to incorrect syntax or missing required parameters."
+	e.DefaultErrString = fmt.Sprintf(
+		"Bad request with: [%s %s], error message: %s",
+		e.Method, e.URL, e.Body,
+	)
+	return e.choseErrString()
 }
 func (e ErrDefault401) Error() string {
 	return "Authentication failed"


jtopjian:
@liu-sheng Thank you so much for looking into this. This is something I've been meaning to look at forever and procrastinated because I thought it would be more difficult. :)

An easy way to test this is to change the `TestServergroupsAffinityPolicy` acceptance test to use a policy of `affinityyyy` (or any other incorrect policy). Doing this yields:

```
=== RUN   TestServergroupsAffinityPolicy
--- FAIL: TestServergroupsAffinityPolicy (1.55s)
        servergroup_test.go:63: Unable to create server group: Bad request with: [POST http://162.246.156.73:8774/v2.1/47b5e614bb7b4842a685b98cb7048efb/os-server-groups], error message: {"badRequest": {"message": "Invalid input for field/attribute 0. Value: affinityy. u'affinityy' is not one of ['anti-affinity', 'affinity']", "code": 400}}
FAIL
```

I'm a little hesitant to use formatting such as brackets/braces in strings because the pattern is rarely ever followed. I wonder if something like this would work:

```go
	e.DefaultErrString = fmt.Sprintf(
		"Error performing %s on %s: %s",
		e.Method, e.URL, e.Body,
	)
```

Another option is to simply do:

```go
	e.DefaultErrString = e.Body
```

But having the method and URL included is helpful.

@jrperritt thoughts?

jtopjian:
@liu-sheng Was there a reason you closed this PR? Short of the error message being rewritten, I think this PR is worthwhile.

liu-sheng:
oh, sorry, my mistake

PR #705, opened 2018-01-04
Flavor Extra Spec Delete

dklyle:
For #540

code in this PR:

API reference:

https://developer.openstack.org/api-ref/compute/#delete-an-extra-spec-for-a-flavor

Nova implementation:

https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavors_extraspecs.py#L111

There is a note in the above that the error _should_ be 204, rather than the 200 that is returned. The value in the API spec is also 200. Testing proved that the return code is 200 rather than 204.


jtopjian:
LGTM - thanks!

PR #706, opened 2018-01-05
Add Neutron subnetpool update support

ozerovandrei:
Add support to update a Neutron subnetpool through a PUT request on /v2.0/subnetpools.
`openstack subnet pool set`
 or
`neutron subnetpool-update`

For [#672](https://github.com/gophercloud/gophercloud/issues/672)

code in this PR:

https://github.com/openstack/neutron/blob/stable/pike/neutron/db/db_base_plugin_v2.py#L1146

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/subnetpools/subnetpools_test.go
+++ b/acceptance/openstack/networking/v2/extensions/subnetpools/subnetpools_test.go
@@ -29,7 +29,7 @@ func TestSubnetPoolsCRUD(t *testing.T) {
 		Name: newName,
 	}
 
-	newSubnetPool, err = subnetpools.Update(client, subnetPool.ID, updateOpts).Extract()
+	newSubnetPool, err := subnetpools.Update(client, subnetPool.ID, updateOpts).Extract()
 	if err != nil {
 		t.Fatalf("Unable to update the subnetpool: %v", err)
 	}
--- a/acceptance/openstack/networking/v2/extensions/subnetpools/subnetpools_test.go
+++ b/acceptance/openstack/networking/v2/extensions/subnetpools/subnetpools_test.go
@@ -23,6 +23,18 @@ func TestSubnetPoolsCRUD(t *testing.T) {
 	}
 
 	tools.PrintResource(t, subnetPool)
+
+	newName := tools.RandomString("TESTACC-", 8)
+	updateOpts := &subnetpools.UpdateOpts{
+		Name: newName,
+	}
+
+	newSubnetPool, err = subnetpools.Update(client, subnetPool.ID, updateOpts).Extract()
+	if err != nil {
+		t.Fatalf("Unable to update the subnetpool: %v", err)
+	}
+
+	tools.PrintResource(t, newSubnetPool)
 }
 
 func TestSubnetPoolsList(t *testing.T) {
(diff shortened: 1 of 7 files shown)


ozerovandrei:
I've removed two garbage dots from the SubnetPools documentation. I've done them accidentally in previous PRs for the [#672](https://github.com/gophercloud/gophercloud/issues/672). Hope that don't need separate PR 😄 .

Also fixture `SubnetPoolUpdateResponse` contains fields `default_prefixlen `, `max_prefixlen `, `min_prefixlen ` in the `int` format rather than `string` like in the `CreateResponse`. I've tested this in my environment and they are really `int`.

jtopjian:
> I've removed two garbage dots from the SubnetPools documentation. I've done them accidentally in previous PRs for the #672. Hope that don't need separate PR

Thank you for fixing these. No need for a new PR. :)

openstack/networking/v2/extensions/subnetpools/requests.go
```diff
@@ -145,3 +145,74 @@ func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r Create
...
+	// For IPv6 subnetpools, default is 64.
+	MinPrefixLen int `json:"min_prefixlen,omitempty"`
+
+	// MaxPrefixLen is the maximum prefix size that can be allocated from the subnetpool.
+	// For IPv4 subnetpools, default is 32.
+	// For IPv6 subnetpools, default is 128.
+	MaxPrefixLen int `json:"max_prefixlen,omitempty"`
+
+	// AddressScopeID is the Neutron address scope to assign to the subnetpool.
+	AddressScopeID string `json:"address_scope_id,omitempty"`
```
jtopjian: We might want to do `*string` here.

openstack/networking/v2/extensions/subnetpools/requests.go
```diff
@@ -145,3 +145,74 @@ func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r Create
...
+
+	// Description is thehuman-readable description for the resource.
+	Description string `json:"description,omitempty"`
+
+	// IsDefault indicates if the subnetpool is default pool or not.
+	IsDefault bool `json:"is_default,omitempty"`
```
jtopjian: We might want to do `*bool` here.

ozerovandrei:
@jtopjian oh, I haven't thought about that.
That means that we also need a `*string` for the `Description` field and a `*int` for the `DefaultQuota ` field.
Because they're also can be empty or zero.

jtopjian:
Yes, I agree. Thank you for catching those :)

kiwik:
recheck

kiwik:
recheck

acceptance/openstack/networking/v2/extensions/subnetpools/subnetpools_test.go
```diff
@@ -23,6 +23,18 @@ func TestSubnetPoolsCRUD(t *testing.T) {
 	}
 
 	tools.PrintResource(t, subnetPool)
+
+	newName := tools.RandomString("TESTACC-", 8)
+	updateOpts := &subnetpools.UpdateOpts{
+		Name: newName,
+	}
+
+	newSubnetPool, err = subnetpools.Update(client, subnetPool.ID, updateOpts).Extract()
```
jtopjian: This should be:

```go
newSubnetPool, err := subnetpools.Update(client, subnetPool.ID, updateOpts).Extract()
```

jtopjian:
@ozerovandrei Thank you for your patience with this PR :)

PR #708, opened 2018-01-08
Compute v2: Update docs for Availability Zones

jtopjian:
Just some minor fixes.

For #320


jtopjian:
Travis is failing on tip but it's yet another timeout failure.

Issue #707, opened 2018-01-07
openstack/orchestration/v1/stacks unit tests are broken


nim-nim:
+ echo 'Testing openstack/orchestration/v1/stacks…'
+ pushd /builddir/build/BUILD/gophercloud-fe4853e064e386919340e25c447d0fed6cd8c6ea/_build/src/github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacks
+ go test -compiler gc -ldflags ' -extldflags '\''-Wl,-z,relro -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '\'''
--- FAIL: TestGetRRFileContents (0.00s)
	convenience.go:236: Failure in environment_test.go, line 184:at [resource_registry] expected "2015-04-30", but got map[string]interface {}{"My::WP::Server":"my_env.yaml", "resources":map[string]interface {}{"my_db_server":map[string]interface {}{"OS::DBInstance":"my_db.yaml"}}}
	convenience.go:236: Failure in environment_test.go, line 184:at  expected map[string]interface {}{"resource_registry":"2015-04-30", "My::WP::Server":"file:///builddir/build/BUILD/gophercloud-fe4853e064e386919340e25c447d0fed6cd8c6ea/_build/src/github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacks/my_env.yaml", "resources":map[string]interface {}{"my_db_server":map[string]interface {}{"OS::DBInstance":"file:///builddir/build/BUILD/gophercloud-fe4853e064e386919340e25c447d0fed6cd8c6ea/_build/src/github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacks/my_db.yaml"}}}, but got <nil>
	convenience.go:35: Failure in environment_test.go, line 184: The structures were different.
--- FAIL: TestTemplateParsing (0.00s)
	convenience.go:236: Failure in template_test.go, line 36:at [parameters][flavor][default] expected 4353, but got 4353
	convenience.go:35: Failure in template_test.go, line 36: The structures were different.
FAIL

jtopjian:
Looks like `stacks` is now passing. Thank you @nim-nim and @ozerovandrei for reporting and fixing.

nim-nim:
Works for me, thanks a lot!

PR #709, opened 2018-01-08
Fix Orchestration TestGetRRFileContents

ozerovandrei:
Fix "Content-Type" in orchestration environment unit tests ("application/jason" -> "application/json").

Fix environments unit test "TestGetRRFileContents" by updating the "expectedParsed" to be a valid representation of a Heat's resource_registry definition.

Also update environment's fileMaps to sucessfully apply the "fixFileRefs" method.

For #707 

A valid representation of a resource registry can be viewed for example in the [docs/heat/environment.html](https://docs.openstack.org/heat/pike/template_guide/environment.html#override-a-resource-with-a-custom-resource)

### DIFF
--- a/openstack/orchestration/v1/stacks/environment_test.go
+++ b/openstack/orchestration/v1/stacks/environment_test.go
@@ -170,13 +170,20 @@ service_db:
 	th.AssertEquals(t, expectedEnvFilesContent, env.Files[fakeEnvURL])
 	th.AssertEquals(t, expectedDBFilesContent, env.Files[fakeDBURL])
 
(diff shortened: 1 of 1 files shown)


kiwik:
recheck

jtopjian:
@jrperritt similar here as in #710. I can't vouch for the validity of the fixture (meaning I'm not sure if the fixture represents a true Heat configuration) but given these are obvious errors, it seems appropriate to merge. If the fixtures are incorrect, that's probably another task in itself.

jrperritt:
> given these are obvious errors, it seems appropriate to merge

Agreed

jtopjian:
@ozerovandrei Thank you for fixing this :)

PR #710, opened 2018-01-08
Fix Orchestration TestTemplateParsing

ozerovandrei:
Fix "Content-Type" in orchestration templates unit tests ("application/jason" -> "application/json").

Set "default" field to be a string in "ValidJSONTemplate", "ValidYAMLTemplate", "InvalidTemplateNoVersion", "InvalidEnvironment", "ValidJSONTemplateParsed" orchestration templates test fixtures.

That automatically fixes the TestTemplateParsing unit test.

For #707 
 
Default field in the flavor is a reference to the default flavor that will be used in the template. It can't be an `int` as it represents the name (or maybe ID) of the flavor.
There are some examples in the official [heat-templates](https://github.com/openstack/heat-templates) repo: [autohealing_group.yaml#L17](https://github.com/openstack/heat-templates/blob/master/hot/autohealing/autohealing_group.yaml#L17).

### DIFF
--- a/openstack/orchestration/v1/stacks/fixtures.go
+++ b/openstack/orchestration/v1/stacks/fixtures.go
@@ -6,7 +6,7 @@ const ValidJSONTemplate = `
   "heat_template_version": "2014-10-16",
   "parameters": {
     "flavor": {
-      "default": 4353,
+      "default": "debian2G",
       "description": "Flavor for the server to be created",
       "hidden": true,
       "type": "string"
@@ -32,7 +32,7 @@ parameters:
   flavor:
     type: string
     description: Flavor for the server to be created
-    default: 4353
+    default: debian2G
     hidden: true
 resources:
   test_server:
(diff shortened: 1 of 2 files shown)


kiwik:
recheck

jtopjian:
Just to narrow the problem down, if you change [this](https://github.com/gophercloud/gophercloud/blob/master/openstack/orchestration/v1/stacks/fixtures.go#L9) to a string and [this](https://github.com/gophercloud/gophercloud/blob/master/openstack/orchestration/v1/stacks/fixtures.go#L183) to a string, then the flavor is parsed correctly. However, I  agree with changing the value globally to something like "debian2G" as it's more clear all around.

@jrperritt I'm not familiar with Heat and can't vouch for the validity of the fixtures themselves, but since this PR is focused on updating something within Gophercloud only, it seems appropriate to merge. Thoughts?

It also looks like Travis is skipping these tests. It'd probably be a good amount of work to move them to `testing`.

jrperritt:
>  I'm not familiar with Heat and can't vouch for the validity of the fixtures themselves

Nor am I

> but since this PR is focused on updating something within Gophercloud only, it seems appropriate to merge. Thoughts?

I agree

jtopjian:
@ozerovandrei Thank you for this one, too!

PR #712, opened 2018-01-10
Flavor Extra Spec Update

dklyle:
For #540

code in this PR:

API reference:
https://developer.openstack.org/api-ref/compute/#update-an-extra-spec-for-a-flavor

Nova implementation:
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/flavors_extraspecs.py#L79

Schema:
https://github.com/openstack/nova/blob/stable/pike/nova/api/openstack/compute/schemas/flavors_extraspecs.py#L24

The API will return an error if more than one key-value pair is passed and also if the key differs from the key parameter on the URL. As the API accepts more than one key-value pair and then properly returns an error, per the discussion on #689, the API is left to do the error handling.

### DIFF
--- a/acceptance/openstack/compute/v2/flavors_test.go
+++ b/acceptance/openstack/compute/v2/flavors_test.go
@@ -181,6 +181,15 @@ func TestFlavorExtraSpecsCRUD(t *testing.T) {
 		t.Fatalf("Unable to delete ExtraSpec: %v\n", err)
 	}
 
+	updateOpts := flavors.ExtraSpecsOpts{
+		"hw:cpu_thread_policy": "CPU-THREAD-POLICY-BETTER",
+	}
+	updatedExtraSpec, err := flavors.UpdateExtraSpec(client, flavor.ID, updateOpts).Extract()
+	if err != nil {
+		t.Fatalf("Unable to update flavor extra_specs: %v", err)
+	}
+	tools.PrintResource(t, updatedExtraSpec)
+
 	allExtraSpecs, err := flavors.ListExtraSpecs(client, flavor.ID).Extract()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


jtopjian:
Looking back at #689, one thing I didn't catch was that the spec key needs to be in the URL. I also tested my theory and Nova is only allowing one spec to pass. Because of these two things, I wonder if we should use the same method that server/instance metadatum creation is doing: https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/requests.go#L646-L684.

The idea here would be to have the function signature as: `flavors.UpdateExtraSpec(client, flavor.ID, updateOpts)`, then internally make sure only one k/v is passed and then craft the URL accordingly.

Another option is to have a signature like ``flavors.UpdateExtraSpec(client, flavor.ID, key, value)` and craft a map internally.

Both would remove the requirement of having to specify the key twice: once in the `updateOpts` and once in the function.

I'm leaning on the former since it follows the usual pattern of updates in Gophercloud.

Thoughts?

dklyle:
I think I misunderstood your original comment about where the error should be handled. I have now updated the patch to limit accepted options to 1 key-pair on the gophercloud side and obtain the key for the URL parameter from that key. This makes more sense. It also allows update to add additional (new) extra specs, better exposing the supported nova API behavior.

dklyle:
This method rather than the second option you presented also allows potential extensibility where hard-coding both parameters would not.

liu-sheng:
recheck

jtopjian:
> I think I misunderstood your original comment about where the error should be handled.

You didn't at all - my original comment was incorrect, so it's on me - sorry about that!

This looks good. Thank you for working on this series :)

liu-sheng:
recheck

Issue #713, opened 2018-01-11
Can't add custom handling for 403 responses


databus23:
I want to add custom handling of nova and neutron responses with status code 403. Thats not possible at the moment because there is no `Err403er` interface defined.

The content of 403 responses is important when the quota subsystem is used. An example quota error returned by nova looks like this:
```
{
    "forbidden": {
        "code": 403,
        "message": "Quota exceeded for cores: Requested 4, but already used 18 of 20 cores"
    }
}
```

I propose to add `ErrDefault403`, `Err403er` types and add a case for that to the error handling in the providerclient's `Request` method.

jrperritt:
closed by #714

PR #714, opened 2018-01-11
Add support for custom 403 error handling

databus23:
For #713

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -72,6 +72,11 @@ type ErrDefault401 struct {
 	ErrUnexpectedResponseCode
 }
 
+// ErrDefault403 is the default error type returned on a 403 HTTP response code.
+type ErrDefault403 struct {
+	ErrUnexpectedResponseCode
+}
+
 // ErrDefault404 is the default error type returned on a 404 HTTP response code.
 type ErrDefault404 struct {
 	ErrUnexpectedResponseCode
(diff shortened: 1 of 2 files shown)


jtopjian:
The only question I have is if a message of "Forbidden" will be sufficient for everyone? Will there ever be a need to display the actual message sent from the server? One of the most common pieces of feedback I get with Terraform is that the 400 error masks the real message.

databus23:
I agree that `Forbidden` is not helpful. We could add something like this. https://github.com/gophercloud/gophercloud/pull/703 where the default errors contains the full response body. Its not great as this means we are returning raw json to the user which has to make sense of it. Solving this properly generically is hard because every service may have a different format for error responses.

I still like the approach using the `ErrorContext` to provide custom parsing of errors on a per service level like in this PR: https://github.com/gophercloud/gophercloud/pull/303

jtopjian:
> Its not great as this means we are returning raw json to the user which has to make sense of it. Solving this properly generically is hard because every service may have a different format for error responses.

Agreed all around.

I'm in favor of passing the full error message to the user, though. IMO, it's better than masking it.

@jrperritt thoughts?

jtopjian:
@databus23 Overall this looks good to me. Since this is a change to "core", I'd like @jrperritt to also give it a review when he has time.

databus23:
@jtopjian Seems like @jrperritt is busy with other things. Any chance we can proceed anyway?

jtopjian:
@databus23 I'm sorry about the delay. If you can, hang in there a little. I like to have @jrperritt sign off on anything that's a "core" change.

Issue #715, opened 2018-01-12
Add support of Compute services list


ildarsv:
There is no support of listing Compute services.
https://developer.openstack.org/api-ref/compute/#list-compute-services

I'm working on it.

jtopjian:
Implemented in #716

PR #716, opened 2018-01-12
Compute service list support

ildarsv:
For #715 

Add support for the running Compute service list.

API reference:
https://developer.openstack.org/api-ref/compute/#list-compute-services

Implementation:
https://github.com/openstack/nova/blob/stable/ocata/nova/api/openstack/compute/services.py#L200

### DIFF
--- a/acceptance/openstack/compute/v2/services_test.go
+++ b/acceptance/openstack/compute/v2/services_test.go
@@ -1,0 +1,32 @@
+// +build acceptance compute services
+
+package v2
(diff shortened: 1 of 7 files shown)


ildarsv:
Hello @jtopjian ! Thank you for the review!
Got you, I've removed `ForcedDown` and added `UpdatedAt`.

jtopjian:
LGTM - nice work on this :smile:

PR #717, opened 2018-01-13
Add get request in subnetpools acceptance tests

ozerovandrei:
It would be better if we will test the `GET` request inside the `TestSubnetPoolsCRUD` acceptance test.

For #672 

Support for `Get` was implemented in PR [#701](https://github.com/gophercloud/gophercloud/pull/701) and for `Update` in [#706](https://github.com/gophercloud/gophercloud/pull/706).


jtopjian:
Good idea. Thanks!

PR #718, opened 2018-01-13
Add Neutron subnetpool delete support

ozerovandrei:
Add a Delete request, unit test and acceptance tests with acceptance
reusable "DeleteSubnetPool" function.

`openstack subnet pool delete`
or
`neutron subnetpool-delete`

For #672 

code in this PR:

https://github.com/openstack/neutron/blob/stable/pike/neutron/db/db_base_plugin_v2.py#L1211


jtopjian:
LGTM!

Issue #721, opened 2018-01-14
Compute v2: Hypervisors API


dstdfx:
I haven't found the following parts of Hypervisors API at Gophercloud: 

- https://developer.openstack.org/api-ref/compute/#show-hypervisor-details 
- https://developer.openstack.org/api-ref/compute/#show-hypervisor-uptime

I'm ready to send few PRs.

jtopjian:
:tada: 

Thank you for your work on this :)

PR #722, opened 2018-01-14
Compute v2: Get Hypervisor Details

dstdfx:
For #721

The same for Openstack CLI:

- `openstack hypervisor show <hypervisor-id>`
- `nova hypervisor-show hypervisor-show <hypervisor-id>` 

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:

API doc:
- [https://developer.openstack.org/api-ref/compute/#show-hypervisor-details](https://developer.openstack.org/api-ref/compute/#show-hypervisor-details)

Source code:
- [https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/hypervisors.py#L302](https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/hypervisors.py#L302)

PR #724, opened 2018-01-15
Compute v2: Get Hypervisor Uptime

dstdfx:
For #721

The same for Openstack CLI:
- `nova hypervisor-uptime <hypervisor-id>`

code in this PR:

API doc: 
- [https://developer.openstack.org/api-ref/compute/#show-hypervisor-uptime](https://developer.openstack.org/api-ref/compute/#show-hypervisor-uptime)

Source code:

- [https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/hypervisors.py#L324](https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/hypervisors.py#L324)


dstdfx:
@jtopjian  I'll add an acceptance test when #722 is merged. I'm gonna reuse this [function](https://github.com/gophercloud/gophercloud/pull/722/commits/16d913dd8a6248bd30dd441ae76a688cb4aaee4f#diff-cf9320b87be12c98238985adf9efbcabR69) in acceptance tests.

jtopjian:
@dstdfx This also looks good, but I suspect I'll run into the same problem as in #722. I think using `strconv` might generally be a better way of converting the int to a string.

dstdfx:
recheck

dstdfx:

Issue #700, opened 2018-01-02
New feature: Zun (Containers Service)


kevin-zhaoshuai:
Current is [Zun](https://wiki.openstack.org/wiki/Zun) a Containers Service in OpenStack.  All supported stable Rest API is [here](https://developer.openstack.org/api-ref/application-container/). Gophercloud doesn't support Zun now.

Zun has two API entrypoint. One is "v1", the other is "experimental". 
The v1 is stable API for container management, also "experimental" API is for Container Capsule, which is mostly like "Pod" in Kubernetes, or "Container Group" in Azure container instance. Link of all info for [experimental API](https://review.openstack.org/#/c/484602/).

Now for zun, we want to join the project "[virtual-kubelet](https://github.com/virtual-kubelet/virtual-kubelet)", connecting OpenStack with Kubernetes. The project needs Zun to have a supported Go client. So now we plan to implement the "experimental" API first, then implement the v1  API.

For experimental API, we want to implement:
capsule/create
capsule/delete
capsule/show
capsule/describe

For v1 API, there are 3 separated part. [Refs](https://developer.openstack.org/api-ref/application-container/)
Manage containers
Manage service
Manage host

ref:
(1) https://docs.openstack.org/zun/latest/contributor/capsule.html
(2) https://github.com/virtual-kubelet/virtual-kubelet/issues/22
(3) https://review.openstack.org/#/c/437759/19/specs/container-composition.rst

jtopjian:
I think adding Zun is certainly valid. @jrperritt ?

One thing to keep in mind is that in-house knowledge of Zun is most likely very limited (personally, I only know Zun exists and nothing more). Therefore, it'll be very important that PRs for Zun strictly follow our style and contributing guide. This will help make code reviews easier. If we have to hunt for the required information ourselves, it could delay the PR by possibly several weeks until we find the time to do so.

jrperritt:
I think it's valid insofar as it looks like an OpenStack service (I've never heard of it), but experimental APIs are probably a non-starter for merging into `master`. 

@kevin-zhaoshuai I can make you a `zun-exp` branch if you like, where you can commit the experimental API until it's stable. We did that a while back for `magnum`, which I think in hindsight was a good decision.

kevin-zhaoshuai:
@jtopjian @jrperritt Thanks a lot!
I think zun-exp is a good solution for raft development, if experimental API is stable, I will move it to stable.
[...]

PR #725, opened 2018-01-17
Add support for Zun experimental API "Capsule"Get

kevin-zhaoshuai:
For #700

code in this PR:

[zun capsule API](https://github.com/openstack/zun/blob/master/zun/api/controllers/experimental/capsules.py)

### DIFF
--- a/acceptance/openstack/container/experimental/capsules_test.go
+++ b/acceptance/openstack/container/experimental/capsules_test.go
@@ -1,0 +1,28 @@
+package experimental
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/openstack/container/experimental/capsules"
+	th "github.com/gophercloud/gophercloud/testhelper"
+)
+
+func TestCapsuleGet(t *testing.T) {
+	client, err := clients.NewContainerExperimentalClient()
+	if err != nil {
+		t.Fatalf("Unable to create an container experimental client: %v", err)
+	}
+	th.AssertNoErr(t, err)
+	capsuleUUID := "e6c913bb-b4e4-409d-8b71-3e029f196458"
+	if capsuleUUID == "" {
+		t.Fatalf("In order to retrieve a capsule, the CapsuleUUID must be set")
+	}
+	capsule, err := capsules.Get(client, capsuleUUID).Extract()
+	// Get a capsule
+
+	th.AssertNoErr(t, err)
+	th.AssertEquals(t, capsule.Status, "Running")
+	th.AssertEquals(t, capsule.MetaName, "template")
+	th.AssertEquals(t, capsule.CPU, float64(2.0))
+}
--- a/openstack/container/experimental/capsules/results.go
+++ b/openstack/container/experimental/capsules/results.go
@@ -84,6 +84,7 @@ type Address struct {
 	Addr             string  `json:"addr"`
 	Port             string  `json:"port"`
 	Version          float64 `json:"version"`
+	SubnetID         string  `json:"subnet_id"`
 }
 
 func (r *Capsule) UnmarshalJSON(b []byte) error {
--- a/openstack/container/experimental/capsules/results.go
+++ b/openstack/container/experimental/capsules/results.go
@@ -1,0 +1,106 @@
+package capsules
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+)
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// Extract is a function that accepts a result and extracts a capsule resource.
+func (r commonResult) Extract() (*Capsule, error) {
+	var s *Capsule
+	err := r.ExtractInto(&s)
+	return s, err
+}
+
+// GetResult represents the result of a get operation.
+type GetResult struct {
+	commonResult
+}
+
+// Represents a Container Orchestration Engine Bay, i.e. a cluster
+type Capsule struct {
+	// UUID for the capsule
+	UUID string `json:"uuid"`
+
+	// ID for the capsule
+	ID int `json:"id"`
+
+	// User ID for the capsule
+	UserID string `json:"user_id"`
+
+	// Project ID for the capsule
+	ProjectID string `json:"project_id"`
+
+	// cpu for the capsule
+	CPU float64 `json:"cpu"`
+
+	// Memory for the capsule
+	Memory string `json:"memory"`
+
+	// The name of the capsule
+	MetaName string `json:"meta_name"`
+
+	// Indicates whether capsule is currently operational. Possible values include:
+	// Running,
+	Status string `json:"status"`
+
+	// The created time of the capsule.
+	CreatedAt time.Time `json:"-"`
+
+	// The updated time of the capsule.
+	UpdatedAt time.Time `json:"-"`
+
+	// Links includes HTTP references to the itself, useful for passing along to
+	// other APIs that might want a server reference.
+	Links []interface{} `json:"links"`
+
+	// The capsule version
+	CapsuleVersion string `json:"capsule_version"`
+
+	// The capsule restart policy
+	RestartPolicy string `json:"restart_policy"`
+
+	// The capsule metadata labels
+	MetaLabels map[string]string `json:"meta_labels"`
+
+	// The list of containers uuids inside capsule.
+	ContainersUUIDs []string `json:"containers_uuids"`
+
+	// The capsule IP addresses
+	Addresses map[string][]Address `json:"addresses"`
+
+	// The capsule volume attached information
+	VolumesInfo map[string][]string `json:"volumes_info"`
+}
+
+type Address struct {
+	PreserveOnDelete bool    `json:"preserve_on_delete"`
+	Addr             string  `json:"addr"`
+	Port             string  `json:"port"`
+	Version          float64 `json:"version"`
+}
+
+func (r *Capsule) UnmarshalJSON(b []byte) error {
+	type tmp Capsule
+	var s struct {
+		tmp
+		CreatedAt gophercloud.JSONRFC3339ZNoT `json:"created_at"`
+		UpdatedAt gophercloud.JSONRFC3339ZNoT `json:"updated_at"`
+	}
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*r = Capsule(s.tmp)
+
+	r.CreatedAt = time.Time(s.CreatedAt)
+	r.UpdatedAt = time.Time(s.UpdatedAt)
+
+	return nil
+}
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -424,3 +424,22 @@ func NewSharedFileSystemV2Client() (*gophercloud.ServiceClient, error) {
 		Region: os.Getenv("OS_REGION_NAME"),
 	})
 }
+
+// NewContainerExperimentalClient returns a *ServiceClient for making calls
+// to the OpenStack Container Experimental API. An error will be returned
+// if authentication or client creation was not possible.
+func NewContainerExperimentalClient() (*gophercloud.ServiceClient, error) {
+	ao, err := openstack.AuthOptionsFromEnv()
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := openstack.AuthenticatedClient(ao)
+	if err != nil {
+		return nil, err
+	}
+
+	return openstack.NewContainerExperimental(client, gophercloud.EndpointOpts{
+		Region: os.Getenv("OS_REGION_NAME"),
+	})
+}
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -394,3 +394,8 @@ func NewLoadBalancerV2(client *gophercloud.ProviderClient, eo gophercloud.Endpoi
 	sc.ResourceBase = sc.Endpoint + "v2.0/"
 	return sc, err
 }
+
+// NewContainerExperimental creates a ServiceClient that may be used with experimental container package
+func NewContainerExperimental(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
+	return initClientOpts(client, eo, "container-experimental")
+}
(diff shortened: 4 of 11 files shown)


kevin-zhaoshuai:
@jtopjian @jrperritt help to review? Thanks

jtopjian:
@kevin-zhaoshuai Thank you for working on this :)

In general, this is a really good start.

I have a few requests for this first PR that should also be done in future PRs:

1. Please use a consistent format. For example, only half of the fields [here](https://github.com/kevin-zhaoshuai/gophercloud/blob/6c58c656cff53fbdfd5658b608da93135e41a0b3/openstack/container/experimental/capsules/results.go#L24-L61) have comments. Please add comments to all of them, even if they are very brief ("Name is the name of the capsule" is fine). 

2. When you link to a file like https://github.com/openstack/zun/blob/master/zun/api/controllers/experimental/capsules.py, please link directly to the function that is being called. For this PR, I am assuming you mean [here](https://github.com/openstack/zun/blob/master/zun/api/controllers/experimental/capsules.py#L252), right? But that doesn't tell us a lot. We need to know where all of [these](https://github.com/kevin-zhaoshuai/gophercloud/blob/6c58c656cff53fbdfd5658b608da93135e41a0b3/openstack/container/experimental/capsules/results.go#L24-L61) fields are defined so we're confident that the struct is correct.  
 
   I dug around the Zun code and found [this](https://github.com/openstack/zun/blob/2ed50d1163121c40bf8cf7a2dc5c6e11ac33ec7a/zun/api/controllers/experimental/views/capsules_view.py) which looks like it matches, but I'm not sure -- can you confirm?

   In general, we need to see where all of the options and result fields are defined. This can be difficult -- digging around Python code isn't fun -- but we need to make sure the requests and responses are correct. Since we're not very familiar with Zun, it'll be even more important to see that information.

3. Can you provide an acceptance test for this? General documentation on acceptance tests can be found [here](https://github.com/gophercloud/gophercloud/tree/master/acceptance). See [here](https://github.com/gophercloud/gophercloud/blob/master/acceptance/clients/clients.go) and [here](https://github.com/gophercloud/gophercloud/tree/master/acceptance/openstack/compute/v2) for the basis of creating tests.

Again, this is a really good start - nice work.

kevin-zhaoshuai:
Thanks @jtopjian for your help.
For  1, I will add more details
For 2,  yeah I mean the function [get_one](https://github.com/openstack/zun/blob/master/zun/api/controllers/experimental/capsules.py#L252).
Here the link that is the field returned:
[url](https://github.com/openstack/zun/blob/2ed50d1163121c40bf8cf7a2dc5c6e11ac33ec7a/zun/api/controllers/experimental/views/capsules_view.py)
_basic_keys = (
    'id',
    'uuid',
    'user_id',
    'project_id',
    'created_at',
    'status',
    'restart_policy',
    'meta_name',
    'meta_labels',
    'containers_uuids',
    'capsule_version',
    'memory',
    'cpu'
)
The field attribute referred from [link](https://github.com/openstack/zun/blob/master/zun/objects/capsule.py). The 'created_at' is the same with other OpenStack resources.
For 3, I will add the acceptance test.

jtopjian:
@kevin-zhaoshuai Thank you for continuing to work on this. It looks like Travis is flagging some `gofmt` errors. One easy way to resolve this is to configure your IDE/editor to run `gofmt` whenever the file is saved. Let me know if you have any questions.

kevin-zhaoshuai:
@jtopjian Thanks! I've added gofmt to IDE and modified the code.

openstack/container/experimental/capsules/results.go
```diff
@@ -0,0 +1,77 @@
...
+	ID int `json:"id"`
+
+	// User ID for the capsule
+	UserID string `json:"user_id"`
+
+	// Project ID for the capsule
+	ProjectID string `json:"project_id"`
+
+	// cpu for the capsule
+	CPU float64 `json:"cpu"`
```
jtopjian: Is it possible for CPU to be a fraction (ie: 1.2 or .075)? If not, then this should be an `int`.

openstack/container/experimental/capsules/results.go
```diff
@@ -0,0 +1,77 @@
...
+	Links []interface{} `json:"links"`
+
+	// The capsule version
+	CapsuleVersion string `json:"capsule_version"`
+
+	// The capsule restart policy
+	RestartPolicy string `json:"restart_policy"`
+
+	// The capsule metadata labels
+	MetaLables map[string]string `json:"meta_labels"`
```
jtopjian: I think this should be `MetaLabels` and not `MetaLables`.

acceptance/openstack/container/experimental/capsules_test.go
```diff
@@ -0,0 +1,28 @@
...
+	th "github.com/gophercloud/gophercloud/testhelper"
+)
+
+func TestCapsuleGet(t *testing.T) {
+	client, err := clients.NewContainerExperimentalClient()
+	if err != nil {
+		t.Fatalf("Unable to create an container experimental client: %v", err)
+	}
+	th.AssertNoErr(t, err)
+	capsuleUUID := "e6c913bb-b4e4-409d-8b71-3e029f196458"
```
jtopjian: This is OK for now, but once `List` is implemented, this hard-coded UUID should be removed.

openstack/container/experimental/capsules/results.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// The list of containers uuids inside capsule.
+	ContainersUUIDs []string `json:"containers_uuids"`
+
+	// The capsule IP addresses
+	Addresses map[string]interface{} `json:"addresses"`
```
jtopjian: Do the keys to `Addresses` change frequently or do they have a common structure to them? In your test fixture, you have:

```json
			"addresses": {
				"b1295212-64e1-471d-aa01-25ff46f9818d": [
					{
						"version": 4,
						"preserve_on_delete": false,
						"addr": "172.24.4.11",
						"port": "8439060f-381a-4386-a518-33d5a4058636"
					}
				]
			},
```

So instead of `map[string]interface{}`, would it be better to have the following?

```go
Addresses map[string]Address `json:"addresses"`
```

And then:

```go
type Address struct {
  PreserveOnDelete bool `json:"preserve_on_delete"`
  Addr string `json:"addr"`
  Port string `json:"port"`
  Version int `json:"version"`
}
```

If so, please link to the appropriate Zun python code where the Address struct is defined. It looks like this feature was recently added.

openstack/container/experimental/capsules/results.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// The name of the capsule
+	Name string `json:"meta_name"`
+
+	// Indicates whether capsule is currently operational. Possible values include:
+	// Running,
+	Status string `json:"status"`
+
+	// The created time of the capsule.
+	Created string `json:"created_at"`
```
jtopjian: This should be:

```go
CreatedAt time.Time `json:"created_at"`
```

If there's a problem converting the timestamp, then it should be:

```go
CreatedAt time.Time `json:"-"`
```

And then implement a custom `UnmarshalJSON` method. See this [FAQ entry](https://github.com/gophercloud/gophercloud/blob/master/FAQ.md#overriding-default-unmarshaljson-method) for an example.

openstack/container/experimental/capsules/results.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// The updated time of the capsule.
+	Updated string `json:"updated_at"`
```
jtopjian: This should be:

```go
UpdatedAt time.Time `json:"updated_at"`
```

openstack/container/experimental/capsules/results.go
```diff
@@ -0,0 +1,77 @@
...
+	ProjectID string `json:"project_id"`
+
+	// cpu for the capsule
+	CPU float64 `json:"cpu"`
+
+	// Memory for the capsule
+	Memory string `json:"memory"`
+
+	// The name of the capsule
+	Name string `json:"meta_name"`
```
jtopjian: This should be:

```go
MetaName string `json:"meta_name"`
```

liu-sheng:
recheck

kevin-zhaoshuai:
@jtopjian Hi guys,
How to deal with the coverage test failed?

jrperritt:
> How to deal with the coverage test failed?

It looks like there are tests, so no need to worry about that

jtopjian:
@kevin-zhaoshuai Thank you again for working on this :)

If you don't mind, I would like to try to set up a Zun environment so I can test this out. It's both for my own curiosity and to have a second verification that this works (the first being yours). Unfortunately it'll take me a few days to have time to do this.

Also a quick note: when making additional commits after a review has started, please add additional commits rather than amending/squashing. It helps us see the changes that are made. Github can squash everything for us once the PR is good to go.

kevin-zhaoshuai:
@jtopjian 
Yes sure no problem. Feel free to ping me if you need any help about Zun.

openstack/container/experimental/capsules/results.go
```diff
@@ -0,0 +1,106 @@
...
+	ContainersUUIDs []string `json:"containers_uuids"`
+
+	// The capsule IP addresses
+	Addresses map[string][]Address `json:"addresses"`
+
+	// The capsule volume attached information
+	VolumesInfo map[string][]string `json:"volumes_info"`
+}
+
+type Address struct {
```
jtopjian: I'm seeing a `subnet_id` field being returned:

```json
    "addresses": {                                                  
        "1883661f-ddb9-4414-81b6-e4e0b880a14f": [                   
            {                                                       
                "addr": "172.24.4.9",                               
                "port": "ba0accd0-f3cd-49af-b745-42b650704344",     
                "preserve_on_delete": false,                        
                "subnet_id": "e26b1199-4a6d-4279-a122-feb72691d8ae",
                "version": 4                                        
            }                                                       
        ]                                                           
    },                                                              
```

openstack/container/experimental/capsules/results.go
```diff
@@ -0,0 +1,106 @@
...
+	MetaLabels map[string]string `json:"meta_labels"`
+
+	// The list of containers uuids inside capsule.
+	ContainersUUIDs []string `json:"containers_uuids"`
+
+	// The capsule IP addresses
+	Addresses map[string][]Address `json:"addresses"`
+
+	// The capsule volume attached information
+	VolumesInfo map[string][]string `json:"volumes_info"`
```
jtopjian: Similar to `Addresses`, this should probably be `map[string][]Volume`. However, the Zun documentation says that volume support is incomplete. If this is true, then `[]string` is fine for now.

jtopjian: Actually, I apologize. The above comment is assuming that `VolumesInfo` is returning an object/hash/struct about Volumes (such as name and ID). However, if it is only returning a list of UUIDs (similar to `ContainerUUIDs`, then ignore the above comment altogether.

openstack/container/experimental/capsules/results.go
kevin-zhaoshuai: Yes, the subnet_id field is recently merged. I will add.

openstack/container/experimental/capsules/results.go
kevin-zhaoshuai: Volume support for Zun doesn't support Cinder multiple Attach. For one-one attach, this works fine. Besides, Zun will add volume multiple attach recently. So the volumes_info data structure will be compatible with volume attach.
The structure of VolumesInfo is:
"volumes_info": {
              volume1_uuid: [container1_uuid, container2_uuid],
              volume2_uuid: [container1_uuid, container3_uuid],
              volume3_uuid: [container3_uuid, container4_uuid]
}
So for the key volume uuid, it will record all the containers that attached to them.

For addresses, there is a new [patch](https://review.openstack.org/542604) added subnet yesterday. I will add corresponding support.

openstack/container/experimental/capsules/results.go
kevin-zhaoshuai: For addresses, there is a new [patch](https://review.openstack.org/542604) added subnet yesterday. I will add corresponding support.

kevin-zhaoshuai:
@jtopjian Thanks for your kindly review.
I'm sorry for the doc about [Capsule quick start](https://docs.openstack.org/zun/latest/contributor/capsule.html). This doc is out-of-date, the template for capsule could not use now. I've submitted a patch to fix that, you can try with the new yaml in [here](https://review.openstack.org/543338).
Also please apply the [patch-in-review](https://review.openstack.org/543380) to Zun since there is a [regression bug](https://bugs.launchpad.net/zun/+bug/1748820) for capsule create action.

For your question 1, 
I think the main reason is Capsule create actually failed. After using the patch and yaml above, I think it will be OK for you to create capsule. 

2. Zun API. 
Zun have two API controller. One is V1, for container create, it is stable now. The other is Experimental, only for Capsule.
Besides, we plan to merge Capsule API to V1 this release(Rocky). Since Experimental API(for Capsule) is urgent to implement the [Virtual-kubelet](https://github.com/virtual-kubelet/virtual-kubelet), so I will add it first, then add they other V1 api for containers.
3.  Use the yaml in [here](https://review.openstack.org/543338) will work.

Thanks again for your corporation and review. Pls let me know if you need any help.:-)

openstack/container/experimental/capsules/results.go
jtopjian: hah - how coincidental 😄 

sounds good.

jtopjian:
> Since Experimental API(for Capsule) is urgent to implement the Virtual-kubelet, so I will add it first, then add they other V1 api for containers.

This sounds good to me.

Once this PR is merged, the other PRs will be merged more quickly. The first PR of a new resource usually takes the longest for both the submitter and reviewer to get familiar with the code.

Once the `subnet_id` field is committed, this PR is good to go.

> Use the yaml in here will work.

It does, yes! Thank you 😄

kevin-zhaoshuai:
@jtopjian  Add the subnet patch now.
Thanks for your review!

jtopjian:
@kevin-zhaoshuai Looks good to me. Thank you for your patience with this.

If there are further upstream changes to the fields, you can submit a small PR and reference #700.

kevin-zhaoshuai:
@jtopjian Big Thanks!

PR #726, opened 2018-01-17
Update provider_client.go

freddebacker:
I ran into an issue where I was using terraform-openstack-provider (which uses gophercloud) through an HTTP proxy. Terraform was creating resources in an openstack cloud. Since creating the resources took some time, the initial response from openstack was "still creating...". Further polling of the resource status resulted in receiving cached copies of "still creating..." until time-out.

One could argue this is an openstack issue not including proper cache-control headers in the response, but being explit on the request shouldn't hurt... Especially because gophercloud is meant to query API's which responses are volatile in nature.

For #727


jrperritt:
closing per #727

Issue #729, opened 2018-01-18
BuildRequestBody and embedded structs


jtopjian:
Per discussion at https://github.com/gophercloud/utils/pull/16, we've identified a bug which prevents certain struct types, such as `time.Time` from working with `BuildRequestBody`.

jtopjian:
Fixed in #730

PR #730, opened 2018-01-18
Prevent Recursive BuildRequestBody

jtopjian:
For #729 

@jrperritt I felt more comfortable adding the check for `"-"` after all of the other validation field checks were done. I wasn't sure if there are cases of something like this floating around:

```go
Username username `json:"-" required:"true"`
```


jtopjian:
@jrperritt Thank you for the quick review. I'll wait until the OpenLab tests are done and then merge.

Issue #732, opened 2018-01-19
Add support of Blockstorage services list


ildarsv:
There is no support of listing Blockstorage services.
https://developer.openstack.org/api-ref/block-storage/v3/index.html#list-all-cinder-services

I'm working on it.

jtopjian:
Implemented in #733

PR #733, opened 2018-01-19
Blockstorage service list support

ildarsv:
For #732 

API reference:
https://developer.openstack.org/api-ref/block-storage/v3/index.html#list-all-cinder-services

Implementation:
https://github.com/openstack/cinder/blob/stable/ocata/cinder/api/contrib/services.py#L46-L102

### DIFF
--- a/acceptance/openstack/blockstorage/extensions/services_test.go
+++ b/acceptance/openstack/blockstorage/extensions/services_test.go
@@ -16,7 +16,7 @@ func TestServicesList(t *testing.T) {
 		t.Fatalf("Unable to create a blockstorage client: %v", err)
 	}
 
-	allPages, err := services.List(blockClient).AllPages()
+	allPages, err := services.List(blockClient, services.ListOpts{}).AllPages()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


jrperritt:
@ildarsv Thank you for the PR, but we need the exact line number(s) of the implementation; just giving the link to an entire file usually isn't enough. The references need to be to the function, the request parameters, and the response parameters.

ildarsv:
@jrperritt I'm sorry, my bad. I've updated the link.

jtopjian:
@ildarsv No problem and thanks :)

Overall this is good, but it looks like there are some `ListOps` for this action:

https://github.com/openstack/cinder/blob/stable/ocata/cinder/api/contrib/services.py#L58-L61

Would you be able to implement that? Using the `ListOpts` from another blockstorage package as an example should be fine - I don't see anything out of the ordinary happening here. Let me know if you need help, though.

ildarsv:
@jtopjian yes, sure! Please review :)

jtopjian:
@ildarsv This looks good to me. Thanks :)

Issue #738, opened 2018-01-27
Compute v2: Host aggregates API 


dstdfx:
I haven't found the following parts of Host aggregates API at Gophercloud:

- [x] [Create Aggregate ](https://developer.openstack.org/api-ref/compute/#create-aggregate)
- [x] [Delete Aggregate](https://developer.openstack.org/api-ref/compute/#delete-aggregate)
- [x] [Show Aggregate Details](https://developer.openstack.org/api-ref/compute/#show-aggregate-details)
- [x] [Update Aggregate](https://developer.openstack.org/api-ref/compute/#update-aggregate)
- [x] [Add Host](https://developer.openstack.org/api-ref/compute/#add-host)
- [x] [Remove Host](https://developer.openstack.org/api-ref/compute/#remove-host)
- [x] [Create Or Update Aggregate Metadata](https://developer.openstack.org/api-ref/compute/#create-or-update-aggregate-metadata)

I'm gonna work on it

jtopjian:
🎉 

@dstdfx Thank you for working on this :)

PR #739, opened 2018-01-27
Compute v2: Create aggregate

dstdfx:
For #738 

code in this PR:

API doc: 
- https://developer.openstack.org/api-ref/compute/#create-aggregate 

Source code:
- https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/aggregates.py#L56

### DIFF
--- a/openstack/compute/v2/extensions/aggregates/doc.go
+++ b/openstack/compute/v2/extensions/aggregates/doc.go
@@ -1,5 +1,5 @@
 /*
-Package aggregates creates aggregates and returns information about the host aggregates in the
+Package aggregates manages information about the host aggregates in the
 OpenStack cloud.
 
 Example of Create an aggregate
--- a/openstack/compute/v2/extensions/aggregates/doc.go
+++ b/openstack/compute/v2/extensions/aggregates/doc.go
@@ -1,6 +1,20 @@
 /*
-Package aggregates returns information about the host aggregates in the
+Package aggregates creates aggregates and returns information about the host aggregates in the
 OpenStack cloud.
+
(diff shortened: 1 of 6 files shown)


dstdfx:
@jtopjian I'll add acceptance tests for this PR and #740 after they are merged. It must be useful to test them together in a single testcase.

dstdfx:
recheck

openstack/compute/v2/extensions/aggregates/requests.go
```diff
@@ -11,3 +11,31 @@ func List(client *gophercloud.ServiceClient) pagination.Pager {
 		return AggregatesPage{pagination.SinglePageBase(r)}
 	})
 }
+
+type CreateOpts struct {
+	// The name of the host aggregate.
+	Name string `json:"name"`
```
jtopjian: `name` is a required field, so this should be:

```go
Name string `json:"name" required:"true"`
```

openstack/compute/v2/extensions/aggregates/doc.go
```diff
@@ -1,7 +1,21 @@
 /*
-Package aggregates returns information about the host aggregates in the
+Package aggregates creates aggregates and returns information about the host aggregates in the
```
jtopjian: Might as well rename this to:

```
Package aggregates manages information about the host aggregates in the
OpenStack cloud
```

dstdfx:
@jtopjian Okay, will keep it in mind.

jtopjian:
@dstdfx Looks good to me. Thank you for working on this. :)

PR #740, opened 2018-01-27
Compute v2: Delete aggregate

dstdfx:
For #738 

code in this PR:

API doc: 
- https://developer.openstack.org/api-ref/compute/#delete-aggregate

Source code: 
- https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/aggregates.py#L124

### DIFF
--- a/openstack/compute/v2/extensions/aggregates/requests.go
+++ b/openstack/compute/v2/extensions/aggregates/requests.go
@@ -46,6 +46,7 @@ func Create(client *gophercloud.ServiceClient, opts CreateOpts) (r CreateResult)
 func Delete(client *gophercloud.ServiceClient, aggregateID int) (r DeleteResult) {
 	v := strconv.Itoa(aggregateID)
 	_, r.Err = client.Delete(aggregatesDeleteURL(client, v), &gophercloud.RequestOpts{
-		OkCodes: []int{200}})
+		OkCodes: []int{200},
+	})
 	return
 }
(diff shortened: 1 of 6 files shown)


jtopjian:
@dstdfx This will need rebased. Looks good to me, otherwise.

dstdfx:
@jtopjian Done :)

jtopjian:
@dstdfx I did a quick test of this and it looks like Nova is returning a 200 code. Can you confirm if that's what you're seeing?

```
=== RUN   TestAggregatesDelete
--- FAIL: TestAggregatesDelete (3.10s)
        aggregates_test.go:21: Expected HTTP response code [202 204] when accessing [DELETE http://localhost:8774/v2.1/6c3c1bb5e47a4c5982b8a87f61c62027/os-aggregates/1], but got 200 instead
```

dstdfx:
@jtopjian You're right. I've changed it.

openstack/compute/v2/extensions/aggregates/requests.go
```diff
@@ -39,3 +41,11 @@ func Create(client *gophercloud.ServiceClient, opts CreateOpts) (r CreateResult)
 	})
 	return
 }
+
+// Delete makes a request against the API to delete an aggregate.
+func Delete(client *gophercloud.ServiceClient, aggregateID int) (r DeleteResult) {
+	v := strconv.Itoa(aggregateID)
+	_, r.Err = client.Delete(aggregatesDeleteURL(client, v), &gophercloud.RequestOpts{
+		OkCodes: []int{200}})
```
jtopjian: This is a very minor nit and I apologize, but we should make this look like the other OkCodes:

```go
_, r.Err = client.Delete(aggregatesDeleteURL(client, v), &gophercloud.RequestOpts{
        OkCodes: []int{200},
})

dstdfx: Wow, I'm surprised that `go fmt` didn't fix it.

jtopjian: Yeah, I was surprised that both `fmt` _and_ `vet` didn't flag it. Thank you for fixing this :)

jtopjian:
@dstdfx Thank you for your work on this. If you can, can you submit a PR with an acceptance test for this?

dstdfx:
@jtopjian Sure, next PR will be an acceptance test for create/delete aggregates.

Issue #742, opened 2018-01-28
cannot trigger a thread-safe token renewal from the API


majewsky:
When I have a `var p *gophercloud.ProviderClient`, I can call `p.ReauthFunc()`, but that is not thread-safe even with `p.UseTokenLock()`.

jrperritt:
closed by #743

PR #743, opened 2018-01-28
add ProviderClient.Reauthenticate() function

majewsky:
If a user wants to do their own HTTP requests and reauthenticate in case of 401 responses, they can already use `ProviderClient.ReauthFunc()`, but that function is not thread-safe. This commit provides a safer alternative by pulling the relevant piece of code out of `ProviderClient.Request()`.

For #742.

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -126,6 +126,36 @@ func (client *ProviderClient) SetToken(t string) {
 	client.TokenID = t
 }
 
+//Reauthenticate calls client.ReauthFunc in a thread-safe way. If this is
+//called because of a 401 response, the caller may pass the previous token. In
+//this case, the reauthentication can be skipped if another thread has already
+//reauthenticated in the meantime. If no previous token is known, an empty
+//string should be passed instead to force unconditional reauthentication.
+func (client *ProviderClient) Reauthenticate(previousToken string) (err error) {
+	if client.ReauthFunc == nil {
+		return nil
+	}
+
+	if client.mut == nil {
+		return client.ReauthFunc()
+	}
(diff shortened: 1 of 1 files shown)


majewsky:
Any opinion? I would like to have this in since I want to rely on this in a library that I'm currently building.

majewsky:
@jrperritt What do you think about this?

jrperritt:
I can take a look at this tomorrow. At first glance, it looks reasonable to me.

jtopjian:
If I understand this correctly, and forgive me if I'm being dense, the idea here is to call `client.Reauthenticate()` and never call `client.ReauthFunc()` directly?

jrperritt:
@jtopjian right, I think the goal is to have thread-safe re-authentication without having to use `ProviderClient.Request` so you could, say, use a different HTTP library.

Otherwise, @majewsky , is there any reason to export the `Reauthenticate` method?

majewsky:
As you said. My usecase is a custom Swift client library: I tried to build it on top of `ProviderClient.Request` at first, but found it too limiting. I tried patching around it for a while (e.g. I set OkCodes to a 500-element slice with codes 100..599 to avoid Gophercloud's error handling and get the bare http.Response in all cases), but in the end, it emerged that the only thing I really want to re-use from Gophercloud is the authentication parts. For reference, this is the interface between my library and Gophercloud: https://github.com/majewsky/schwift/blob/master/gopherschwift/package.go#L69-L105

jrperritt:
OK, thanks. Sorry it took so long for me to get around to this

PR #745, opened 2018-01-30
Acceptance test for create, get and delete aggregate

dstdfx:
For #738 

Related to PRs:
- #739 
- #740 
- #746

### DIFF
--- a/acceptance/openstack/compute/v2/aggregates_test.go
+++ b/acceptance/openstack/compute/v2/aggregates_test.go
@@ -31,24 +31,22 @@ func TestAggregatesList(t *testing.T) {
 	}
 }
 
-func TestAggregatesCreateDelete(t *testing.T) {
+func TestAggregatesCreateGetDelete(t *testing.T) {
 	client, err := clients.NewComputeV2Client()
 	if err != nil {
 		t.Fatalf("Unable to create a compute client: %v", err)
 	}
 
-	opts := aggregates.CreateOpts{
-		Name:             "name",
-		AvailabilityZone: "london",
-	}
-
-	aggregate, err := aggregates.Create(client, opts).Extract()
+	createdAggregate, err := CreateAggregate(t, client)
 	if err != nil {
 		t.Fatalf("Unable to create an aggregate: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


dstdfx:

jtopjian:
@dstdfx Thanks for adding the tests for this. This works, but since you're going to implement the full suite of aggregate support, it would be beneficial in the long-term to make the acceptance tests work like the others. In particular, there should be helper functions for both create and delete.

An easy example to base this off of is "flavors": https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/compute/v2/flavors_test.go#L78-L91

Note how there's a `CreateFlavor` and `DeleteFlavor` helper function. These are defined here:

* https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/compute/v2/compute.go#L140-L163
* https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/compute/v2/compute.go#L585-L594

This way, along with the user-facing support for aggregates, internally Gophercloud also has helper functions which can be used (and re-used) for various acceptance tests. For example, in theory, one could create an acceptance test to create an aggregate and launch a server in that aggregate easily by using the helper functions in `compute.go`.

Let me know if you have any questions or need help with this.

dstdfx:
recheck

dstdfx:
recheck

jtopjian:
recheck

acceptance/openstack/compute/v2/aggregates_test.go
```diff
@@ -30,3 +30,23 @@ func TestAggregatesList(t *testing.T) {
...
+	client, err := clients.NewComputeV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create a compute client: %v", err)
+	}
+
+	createdAggregate, err := CreateAggregate(t, client)
+	if err != nil {
+		t.Fatalf("Unable to create an aggregate: %v", err)
+	}
+	defer DeleteAggregate(t, client, aggregate)
```
jtopjian: This should be 

```go
defer DeleteAggregate(t, client, createdAggregate)
```

jtopjian:
LGTM! Thanks, @dstdfx :)

PR #746, opened 2018-01-30
Compute v2: Get aggregate

dstdfx:
For #738 

code in this PR:

API doc:

- https://developer.openstack.org/api-ref/compute/#show-aggregate-details

Source code: 

- https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/aggregates.py#L88

### DIFF
--- a/openstack/compute/v2/extensions/aggregates/results.go
+++ b/openstack/compute/v2/extensions/aggregates/results.go
@@ -78,16 +78,20 @@ type aggregatesResult struct {
 	gophercloud.Result
 }
 
-type CreateResult struct {
-	aggregatesResult
-}
-
-func (r CreateResult) Extract() (*Aggregate, error) {
+func (r aggregatesResult) Extract() (*Aggregate, error) {
 	var s struct {
 		Aggregate *Aggregate `json:"aggregate"`
 	}
 	err := r.ExtractInto(&s)
 	return s.Aggregate, err
+}
+
+type CreateResult struct {
+	aggregatesResult
+}
+
+type GetResult struct {
+	aggregatesResult
 }
 
 type DeleteResult struct {
--- a/openstack/compute/v2/extensions/aggregates/testing/fixtures.go
+++ b/openstack/compute/v2/extensions/aggregates/testing/fixtures.go
@@ -60,39 +60,66 @@ const AggregateCreateBody = `
 }
 `
 
-// First aggregate from the AggregateListBody
-var FirstFakeAggregate = aggregates.Aggregate{
-	AvailabilityZone: "",
-	Hosts:            []string{},
-	ID:               1,
-	Metadata:         map[string]string{},
-	Name:             "test-aggregate1",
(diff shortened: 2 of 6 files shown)


dstdfx:
@jtopjian If this PR is merged before #745, I'll add an acceptance test there.

openstack/compute/v2/extensions/aggregates/results.go
```diff
@@ -78,18 +78,22 @@ type aggregatesResult struct {
...
 	err := r.ExtractInto(&s)
 	return s.Aggregate, err
 }
 
+type CreateResult struct {
+	aggregatesResult
+}
+
+type GetResult struct {
+	aggregatesResult
```
dklyle: There appear to be some missing fields in the `aggregateResult` struct that can be returned in the body on a GET and POST `deleted_at` and `deleted` are in both. This may have been discussed in the review of POST, but I'll add this comment until I can check.

openstack/compute/v2/extensions/aggregates/testing/fixtures.go
```diff
@@ -60,39 +60,66 @@ const AggregateCreateBody = `
...
-	Metadata:         map[string]string{},
-	Name:             "test-aggregate1",
-	CreatedAt:        time.Date(2017, 12, 22, 10, 12, 6, 0, time.UTC),
-	UpdatedAt:        time.Time{},
+const AggregateGetBody = `
+{
+    "aggregate": {
+            "name": "test-aggregate2",
+            "availability_zone": "test-az",
+            "deleted": false,
```
dklyle: in fact you're including the fields in your output.

openstack/compute/v2/extensions/aggregates/results.go
jtopjian: @dklyle Thank you for pointing that out. The omission of those two fields is on me:

https://github.com/gophercloud/gophercloud/pull/691#issuecomment-354357122

Maybe I misunderstood why those fields are marked as deprecated? As well, I suppose it's beneficial to add them for older clouds.

dklyle: @jtopjian I may have missed the deprecation notice. I don't see them flagged in the api docs. But I am known to miss things.

jtopjian: They're marked as deprecated at the top of https://github.com/openstack/nova/blob/master/nova/objects/aggregate.py#L38 but in hindsight, it's very possible that is meant for some internal processing of those fields.

We can definitely add these fields in. It might be best to do it in a separate PR for cleanliness.

openstack/compute/v2/extensions/aggregates/testing/fixtures.go
dstdfx: This fields are used at #691 as well.

openstack/compute/v2/extensions/aggregates/results.go
dstdfx: @jtopjian Are there any other "requests for changes"? I could add these fields next PR.

PR #747, opened 2018-01-31
Add missing `Deleted` and `DeletedAt` fields

dstdfx:
For #738

### DIFF
--- a/openstack/compute/v2/extensions/aggregates/results.go
+++ b/openstack/compute/v2/extensions/aggregates/results.go
@@ -31,6 +31,14 @@ type Aggregate struct {
 	// The date and time when the resource was updated,
 	// if the resource has not been updated, this field will show as null.
 	UpdatedAt time.Time `json:"-"`
+
+	// The date and time when the resource was deleted,
+	// if the resource has not been deleted yet, this field will be null.
+	DeletedAt time.Time `json:"-"`
+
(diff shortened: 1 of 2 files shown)


dstdfx:
recheck

dstdfx:
@jtopjian Is there some problem with OpenLab?

jtopjian:
@dstdfx Yes, looks like there might be. @liu-sheng ?

jtopjian:
Looks like it was either a temporary issue or it has been fixed :)

I think I understand why these two fields are marked as deprecated in the Nova code: they will never return any other value besides their zero-value. This is because it's not possible to see _all_ aggregates, including those that have been deleted. And if you're only seeing aggregates that aren't deleted, then `deleted` will be false and `deleted_at` won't have a value 😄 

However, Nova _is_ returning these fields:

```json
{
    "aggregates": [
        {
            "availability_zone": null,
            "created_at": "2018-01-31T19:36:59.000000",
            "deleted": false,
            "deleted_at": null,
            "hosts": [],
            "id": 5,
            "metadata": {},
            "name": "foo",
            "updated_at": null,
            "uuid": "7a3cf5e8-0709-41f4-826e-8b2a0e49a4fa"
        }
    ]
}
```

Because of that, I'm OK with adding these fields. It's also possible that prior versions of Nova provided the ability to see old aggregates.

@dstdfx Thank you for adding these.

/cc @dklyle

PR #751, opened 2018-02-02
Compute v2: Add host to aggregate

dstdfx:
For #738 

code in this PR:

API doc: 

- https://developer.openstack.org/api-ref/compute/#add-host

Source code:

- https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/aggregates.py#L141

### DIFF
--- a/acceptance/openstack/compute/v2/aggregates_test.go
+++ b/acceptance/openstack/compute/v2/aggregates_test.go
@@ -90,3 +90,27 @@ func TestAggregatesUpdate(t *testing.T) {
 
 	tools.PrintResource(t, updatedAggregate)
 }
+
+func TestAggregatesAddHost(t *testing.T) {
+	client, err := clients.NewComputeV2Client()
+	if err != nil {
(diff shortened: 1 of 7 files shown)


dstdfx:
@jtopjian This is ready to review.

jtopjian:
@dstdfx The acceptance test is failing for this (note that OpenLab only runs a small subset of tests at this time. It checks the syntax for all tests, though).

The host being added to the aggregate has to be a real compute node. When the acceptance test runs, you're using `new-cmp` which, unless is the name of the hypervisor, will return an error.

The best way around this would be to use the `hypervisors.List` request to get a list of all hypervisors, check if the result length is greater than 0, then use the first result.

dstdfx:
@jtopjian Done

jtopjian:
@dstdfx Thanks. This looks good. The acceptance test is failing, but that's due to the aggregate still containing hosts. It'll pass once `DeleteHost` is implemented.

dstdfx:
@jtopjian I'll implement it next PR.

Issue #752, opened 2018-02-04
Networking v2: Specify a Subnet when Creating a Floating IP


jtopjian:
It's possible to specify a subnet when creating a floating IP:

https://github.com/openstack/neutron-lib/blob/9a2724a951bb45ecde69feccfb9f2de88c1eeb0d/neutron_lib/api/definitions/l3.py#L112

jrperritt:
closed by #753

PR #753, opened 2018-02-04
Networking v2: Create Floating IP with Subnet

jtopjian:
For #752


jtopjian:
@dklyle @jrperritt ping for review

jtopjian:
Thank you both for the quick review :)

PR #754, opened 2018-02-04
Compute v2: Remove host

dstdfx:
For #738 

code in this PR:

API doc: 
- https://developer.openstack.org/api-ref/compute/#remove-host

Source code: 
- https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/aggregates.py#L164


dstdfx:
@jtopjian This is ready too

PR #756, opened 2018-02-07
Compute v2: Create Or Update Aggregate Metadata

dstdfx:
For #738 

code in this PR:

API doc:

- https://developer.openstack.org/api-ref/compute/#create-or-update-aggregate-metadata

Source code: 

- https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/aggregates.py#L186

### DIFF
--- a/acceptance/openstack/compute/v2/aggregates_test.go
+++ b/acceptance/openstack/compute/v2/aggregates_test.go
@@ -147,7 +147,7 @@ func TestAggregatesSetMetadata(t *testing.T) {
 	defer DeleteAggregate(t, client, createdAggregate)
 
 	opts := aggregates.SetMetadataOpts{
-		Metadata: map[string]string{"key": "value"},
+		Metadata: map[string]interface{}{"key": "value"},
 	}
 
 	aggregateWithMetadata, err := aggregates.SetMetadata(client, createdAggregate.ID, opts).Extract()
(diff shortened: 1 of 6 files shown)


jtopjian:
@dstdfx This looks good, but I think we will want `Metadata` to be `map[string]interface{}`.

Metadata for other resources is almost always `map[string]string`, but it looks like this API does not provide an action to delete metadata. Instead, you specify `nil/null/None` as the key value, as defined here:

https://github.com/openstack/nova/blob/09ed3b5626ac306e59856db89858eb7bee4daa0e/nova/objects/aggregate.py#L411

So the opts would look like:

```go
opts = aggregates.SetMetadataOpts{
  Metadata: map[string]interface{}{"key": nil},
}
```

jrperritt:
Just a note: we'll want an acceptance test that updates the metadata to remove a key and verifies that it's been removed.

dstdfx:
@jtopjian @jrperritt Done

jtopjian:
@dstdfx This looks good to me - thanks!

> and verifies that it's been removed.

I plan to add verification to the acceptance tests soon. It should only take a day or two to update them. Since there isn't much precedence for this at the moment, I've been lenient. Once there are enough examples to point to, we can ensure it's added to future tests.

Issue #731, opened 2018-01-19
Network v2: RBAC-policies CRUD


PriyankaJ77:
Reference guide for RBAC: https://github.com/openstack/neutron/blob/master/doc/source/admin/config-rbac.rst

API:
https://developer.openstack.org/api-ref/network/v2/#rbac-policies

Schema:
https://github.com/openstack/neutron/blob/8f79a92e45c84819cfc4d1cda99c9320ae78bbaf/neutron/db/rbac_db_models.py

List:
https://github.com/openstack/neutron/blob/59e2c40f14157e0f2f64c340070cf46758d522e1/neutron/db/rbac_db_mixin.py#L117

Get:
https://github.com/openstack/neutron/blob/59e2c40f14157e0f2f64c340070cf46758d522e1/neutron/db/rbac_db_mixin.py#L112

Create:
https://github.com/openstack/neutron/blob/59e2c40f14157e0f2f64c340070cf46758d522e1/neutron/db/rbac_db_mixin.py#L41

Update:
https://github.com/openstack/neutron/blob/59e2c40f14157e0f2f64c340070cf46758d522e1/neutron/db/rbac_db_mixin.py#L69

Delete:
https://github.com/openstack/neutron/blob/59e2c40f14157e0f2f64c340070cf46758d522e1/neutron/db/rbac_db_mixin.py#L85

PriyankaJ77:
@jtopjian ,@jrperritt could you please review on this issue and let me know if I can continue working on this.

jtopjian:
@PriyankaJ77 Sure, sounds good to me :)

jtopjian:
:tada: 

@PriyankaJ77 @manju754 Thank you both for your work on this!

I think this is complete, but if it's not, please let me know.

PR #755, opened 2018-02-07
Network v2: RBAC-Create

PriyankaJ77:
For #731

Create RBAC policy for given tenant.

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:

API:
https://developer.openstack.org/api-ref/network/v2/#create-rbac-policy

Schema:
https://github.com/openstack/neutron/blob/8f79a92e45c84819cfc4d1cda99c9320ae78bbaf/neutron/db/rbac_db_models.py#L92

Create:
https://github.com/openstack/neutron/blob/59e2c40f14157e0f2f64c340070cf46758d522e1/neutron/db/rbac_db_mixin.py#L41

PR #758, opened 2018-02-13
Network v2: RBAC-LIST

manju754:
For #731 

code in this PR:

API:
https://developer.openstack.org/api-ref/network/v2/#list-rbac-policies

Schema:
https://github.com/openstack/neutron/blob/8f79a92e45c84819cfc4d1cda99c9320ae78bbaf/neutron/db/rbac_db_models.py#L92

List:
https://github.com/openstack/neutron/blob/59e2c40f14157e0f2f64c340070cf46758d522e1/neutron/db/rbac_db_mixin.py#L117


manju754:
@jtopjian @kvrshenoy2 Please review this PR.

jtopjian:
@manju754 Thank you for your work on this. This PR should be rebased with master so that the non-rbac commits are removed. It might also make sense to wait until #755 is merged so those changes can be incorporated here.

manju754:
@jtopjian Can you review this PR. I had re-based the PR with master and resolved all the conflicts.

jtopjian:
@manju754 Thanks for continuing to work on the RBAC resources.

This will need rebased with master as there are some merge conflicts from the previous commit.

I have some questions with regard to some of the ListOpts fields: 

1. Why specify an `ID`? Isn't that the same as doing a `Get`?
2. Why specify both `ProjectID` and `TenantID`? If they are both the same, let's standardize on `ProjectID`.

Also, it might be a good idea to add a unit test for `AllPages`. There have been a few times in the past where `EachPage` works but `AllPages` triggered an error. I think all of those bugs have been worked out, but it's still useful to test `AllPages` just in case.

Other than that, this looks good.

For posterity / others who read this PR: it isn't really possible to reference a single place in the Neutron code which defines valid List fields. Neutron has a set of standard fields (marker, limit, etc). The other fields are the fields/attributes of the resource itself.

manju754:
@jtopjian Can you review this PR. I had re-based the PR with master, resolved all the conflicts and addressed review comments.

jtopjian:
@manju754 All looks good but unfortunately it'll need another rebase from #781. Perhaps you and @PriyankaJ77 can figure out which PRs should be merged in order? I'm just going with what reaches my inbox first 😄

openstack/networking/v2/extensions/rbacpolicies/doc.go
```diff
@@ -27,13 +27,42 @@ Example to Create a RBAC Policy
...
+}
+
+Example to Get RBAC Policy by ID
+
+rbacPolicyID := "94fe107f-da78-4d92-a9d7-5611b06dad8d"
+rbacpolicy, err := rbacpolicies.Get(rbacClient, rbacPolicyID).Extract()
+if err != nil{
+  panic(err)
+}
+fmt.Printf("%+v", rbacpolicy)
```
jtopjian: The formatting will need fixed here - the code should be indented.

manju754: Done

manju754:
@jtopjian Can you review this PR. I had re-based the PR with master, resolved all the conflicts and addressed review comments.

jtopjian:
@manju754 Thanks so much for your patience with this. This looks good to go :)

Issue #723, opened 2018-01-15
Add support for VPNaaS


simonre:
There is currently no support for VPN as a service in gophercloud.

https://developer.openstack.org/api-ref/network/v2/#vpnaas-2-0-unmaintained-vpn-vpnservices-ikepolicies-ipsecpolicies-endpoint-groups-ipsec-site-connections

I'm working on it.

---
- [x] IKE Policies: List
- [x] IKE Policies: Create
- [x] IKE Policies: Get
- [x] IKE Policies: Update
- [x] IKE Policies: Delete
---
- [x] IPSec Policies: List
- [x] IPSec Policies: Create
- [x] IPSec Policies: Get
- [x] IPSec Policies: Update
- [x] IPSec Policies: Delete
---
- [x] IPSec Connections: List
- [x] IPSec Connections: Create
- [x] IPSec Connections: Get
- [x] IPSec Connections: Update
- [x] IPSec Connections: Delete
---
- [x] VPN Endpoints: List
- [x] VPN Endpoints: Create
- [x] VPN Endpoints: Get
- [x] VPN Endpoints: Update
- [x] VPN Endpoints: Delete
---
- [x] VPN Services: List
- [x] VPN Services: Create
- [x] VPN Services: Get
- [x] VPN Services: Update
- [x] VPN Services: Delete

dstdfx:
@simonre This feature mark as UNMAINTAINED, I reckon it makes sense to discuss it with @jtopjian or @jrperritt first

jtopjian:
My understanding is that VPNaaS has gone through some ups and downs. I haven't seen it discussed on the mailing lists lately, but the last I saw, there was interest in reviving it.

The [commit history](https://github.com/openstack/neutron-vpnaas/commits/master) shows activity. Given that it still exists as an OpenStack project and there's recent activity, I think we can include it here.

However, the usual caveats apply for a service like this: in-house (Gophercloud team) knowledge of VPNaaS is probably very little-to-none. If you do work on this, please make sure you follow our style and contributor guides very closely to help us make sure the code is correct.

fatmcgav:
@simonre Have you got any beta code that I can take a look at? As could do with this feature :)

jtopjian:
@simonre I updated this issue with a list of actions to implement. Let me know if something is incorrect.

simonre:
@jtopjian Looks good

jtopjian:
[...]

PR #760, opened 2018-02-13
Vpnaas: Create VPN service

simonre:
For #723

code in this PR:

* https://github.com/openstack/neutron-vpnaas/blob/058469e1b99b647537a5228c6a384d93df5484df/neutron_vpnaas/db/vpn/vpn_db.py#L473
* https://github.com/openstack/neutron-vpnaas/blob/058469e1b99b647537a5228c6a384d93df5484df/neutron_vpnaas/db/vpn/vpn_db.py#L459

API:
https://developer.openstack.org/api-ref/network/v2/#create-vpn-service

### DIFF
--- a/openstack/networking/v2/extensions/vpnaas/services/requests.go
+++ b/openstack/networking/v2/extensions/vpnaas/services/requests.go
@@ -15,7 +15,7 @@ type CreateOpts struct {
 	// and the caller will be the owner.
 	TenantID    	string  `json:"tenant_id,omitempty"`
 	SubnetID    	string 	`json:"subnet_id,omitempty"`
-	RouterID    	string 	`json:"router_id", required:"true"`
+	RouterID    	string 	`json:"router_id",required:"true"`
 	Description 	string  `json:"description,omitempty"`
 	AdminStateUp 	*bool 	`json:"admin_state_up"`
 	ProjectID   	string  `json:"project_id"`
--- a/openstack/networking/v2/extensions/vpnaas/services/requests.go
+++ b/openstack/networking/v2/extensions/vpnaas/services/requests.go
@@ -13,19 +13,19 @@ type CreateOpts struct {
 	// TenantID specifies a tenant to own the VPN service. The caller must have
 	// an admin role in order to set this. Otherwise, this field is left unset
 	// and the caller will be the owner.
-	TenantID     string `json:"tenant_id,omitempty"`
+	TenantID string `json:"tenant_id,omitempty"`
 	//The ID of the subnet
-	SubnetID     string `json:"subnet_id,omitempty"`
+	SubnetID string `json:"subnet_id,omitempty"`
 	//The ID of the router
-	RouterID     string `json:"router_id" required:"true"`
+	RouterID string `json:"router_id" required:"true"`
 	//Human readable description of the service
-	Description  string `json:"description,omitempty"`
+	Description string `json:"description,omitempty"`
 	//The administrative state of the resource, which is up (true) or down (false).
-	AdminStateUp *bool  `json:"admin_state_up"`
+	AdminStateUp *bool `json:"admin_state_up"`
 	//The human readable name of the service
-	Name         string `json:"name,omitempty"`
+	Name string `json:"name,omitempty"`
 	//The ID of the flavor
-	FlavorID     string `json:"flavor_id,omitempty"`
+	FlavorID string `json:"flavor_id,omitempty"`
 }
 
 // ToServiceCreateMap casts a CreateOpts struct to a map.
--- a/openstack/networking/v2/extensions/vpnaas/services/requests.go
+++ b/openstack/networking/v2/extensions/vpnaas/services/requests.go
@@ -20,7 +20,7 @@ type CreateOpts struct {
 	AdminStateUp 	*bool 	`json:"admin_state_up"`
 	ProjectID   	string  `json:"project_id"`
 	Name 			string  `json:"name,omitempty"`
-	FlavorID        string  `json:"flavor_id"`
+	FlavorID        string  `json:"flavor_id,omitempty"`
 }
 
 // ToVPNServiceCreateMap casts a CreateOpts struct to a map.
--- a/openstack/networking/v2/extensions/vpnaas/services/requests.go
+++ b/openstack/networking/v2/extensions/vpnaas/services/requests.go
@@ -5,7 +5,7 @@ import "github.com/gophercloud/gophercloud"
 // CreateOptsBuilder allows extensions to add additional parameters to the
 // Create request.
 type CreateOptsBuilder interface {
-	ToVPNServiceCreateMap() (map[string]interface{}, error)
+	ToServiceCreateMap() (map[string]interface{}, error)
 }
 
 // CreateOpts contains all the values needed to create a new VPN service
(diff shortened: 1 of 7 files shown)


simonre:

openstack/networking/v2/extensions/vpnaas/services/requests.go
```diff
@@ -0,0 +1,41 @@
...
+import "github.com/gophercloud/gophercloud"
+
+// CreateOptsBuilder allows extensions to add additional parameters to the
+// Create request.
+type CreateOptsBuilder interface {
+	ToVPNServiceCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts contains all the values needed to create a new VPN service
+type CreateOpts struct {
```
jtopjian: doc nit: all fields should have a descriptive comment even if it's something simple like:

```
Name is the human-readable name of the VPN Service.
```

openstack/networking/v2/extensions/vpnaas/services/requests.go
```diff
@@ -0,0 +1,41 @@
...
+	// TenantID specifies a tenant to own the VPN service. The caller must have
+	// an admin role in order to set this. Otherwise, this field is left unset
+	// and the caller will be the owner.
+	TenantID     string `json:"tenant_id,omitempty"`
+	SubnetID     string `json:"subnet_id,omitempty"`
+	RouterID     string `json:"router_id" required:"true"`
+	Description  string `json:"description,omitempty"`
+	AdminStateUp *bool  `json:"admin_state_up"`
+	ProjectID    string `json:"project_id"`
```
jtopjian: I don't see `project_id` defined anywhere [here](https://github.com/openstack/neutron-vpnaas/blob/058469e1b99b647537a5228c6a384d93df5484df/neutron_vpnaas/db/vpn/vpn_db.py#L473), so maybe it's an internal neutron field? What happens if `TenantID` and `ProjectID` are different UUIDs?

openstack/networking/v2/extensions/vpnaas/services/requests.go
```diff
@@ -0,0 +1,41 @@
+package services
+
+import "github.com/gophercloud/gophercloud"
+
+// CreateOptsBuilder allows extensions to add additional parameters to the
+// Create request.
+type CreateOptsBuilder interface {
+	ToVPNServiceCreateMap() (map[string]interface{}, error)
```
jtopjian: This should be `ToServiceCreateMap`.

openstack/networking/v2/extensions/vpnaas/services/requests.go
```diff
@@ -0,0 +1,41 @@
...
+	RouterID     string `json:"router_id" required:"true"`
+	Description  string `json:"description,omitempty"`
+	AdminStateUp *bool  `json:"admin_state_up"`
+	ProjectID    string `json:"project_id"`
+	Name         string `json:"name,omitempty"`
+	FlavorID     string `json:"flavor_id,omitempty"`
+}
+
+// ToVPNServiceCreateMap casts a CreateOpts struct to a map.
+func (opts CreateOpts) ToVPNServiceCreateMap() (map[string]interface{}, error) {
```
jtopjian: `ToServiceCreateMap`

openstack/networking/v2/extensions/vpnaas/services/results.go
```diff
@@ -0,0 +1,40 @@
+package services
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// Service is a VPN Service
+type Service struct {
```
jtopjian: doc nit: these fields should have comments to describe what they are.

openstack/networking/v2/extensions/vpnaas/services/results.go
```diff
@@ -0,0 +1,40 @@
...
+	TenantID     string `json:"tenant_id"`
+	SubnetID     string `json:"subnet_id,omitempty"`
```
jtopjian: `omitempty` isn't valid for results. it can safely be removed.

jtopjian:
@simonre Thank you for the updates. Travis is failing due to some `gofmt` issues. These can be fixed by manually running `gofmt` on the changed files or configuring your IDE/editor to run `gofmt` upon save.

The OpenLab failures were a temporary issue with the CI system. I wouldn't worry about those.

jtopjian:
@simonre Thank you again for the updates.

This looks good to me. Unfortunately I can't test this myself since I don't have access to VPNaaS, but from reviewing the Neutron code, this looks correct. In addition, the acceptance test looks correct, too.

Can you verify that this code works? Meaning: have you tested it yourself to create an actual VPN service? If it works for you, I'll go ahead and merge this.

The inline comments/docs could use some formatting, but I don't consider doc updates to be blocking unless there are other changes required. If you plan on continuing to work on the other parts of the VPNaaS, then maybe fix them there. You can use [this](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/subnetpools/requests.go) file for reference (or really any of the `requests.go` and `results.go` files). Note the formatting of the comments for the fields, particularly the space between `//` and the comment as well as a new line to separate the fields.

The reason we want to make sure formatting is correct is because the inline comments turn into the official godoc reference documentation.

simonre:
@jtopjian Thanks for the quick feedback. I just updated the doc.go sample code to code that I can verify worked for me to create an actual VPN service. I also updated the inline formatting in requests.go, is that what you had in mind?

jtopjian:
@simonre 

>  I also updated the inline formatting in requests.go, is that what you had in mind?

Yep - That's exactly what I had in mind. Though the same format should be applied to the `Service` result struct, too. Let me know if you want to fix that now or in another PR.

> I can verify worked for me to create an actual VPN service

Perfect - thank you 😄

simonre:
I'll fix it now. Just wanted to make sure I'm actually doing the right thing before doing it for all files.

jtopjian:
@simonre I apologize - I missed one thing in the unit tests:

There should be a comparison to the struct that is returned from the Create / Extract call:

```go
expected := services.Service{
  RouterID: "66e3b16c-8ce5-40fb-bb49-ab6d8dc3f2aa",
  ...
}

actual, err := services.Create(fake.ServiceClient(), options).Extract()
th.AssertNoErr(t, err)
th.AssertDeepEquals(t, expected, actual)
```

This same pattern will need to be in the other unit tests (where applicable), so best to get in the habit now :)

simonre:
@jtopjian 
`th.AssertDeepEquals(t, expected, actual)`
doesn't seem to work because of the pointer to AdminStateUp. I added individual comparisons like the ones [here](https://github.com/gophercloud/gophercloud/blob/cf2bbaa52ab69d063466a3c46a5918d05d752625/openstack/networking/v2/extensions/fwaas/firewalls/testing/requests_test.go#L224) instead which should do the same.

jtopjian:
@simonre We definitely want to ensure a struct can be compared successfully, and not just individual fields.

Using "networks" as an example, the `AdminStateUp` field for the result struct is `bool` rather than `*bool`:

https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/networks/results.go#L57

However, the CreateOpts `AdminStateUp` should probably remain `*bool` as this is the standard used for the networking services. For example:

https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/networks/requests.go#L69

Unless there's a good reason to keep the result field as `*bool`, let's change it to `bool`. The only reason I can think of using `*bool` is if there are three distinct states for `AdminStateUp` (true/false/nil). Given the prevalence of `AdminStateUp` already in Gophercloud, I can't see VPNaaS doing something different.

simonre:
@jtopjian You're right. I changed it again. I also did the same in the ipsecpolicy branch.

jtopjian:
@simonre This looks good to me. Thank you very much for your help and patience.

Issue #761, opened 2018-02-14
Compute v2: Fix Flavor Builder Names


jtopjian:
I was doing some Terraform work and noticed the flavor extra spec builder names need renamed to follow the correct format.

jtopjian:
Fixed in #762

PR #762, opened 2018-02-14
Compute v2: Fix flavor extra spec builder names

jtopjian:
For #761 

/cc @dklyle (I think Github won't let me request a review because you aren't a collaborator for this repo?)


jtopjian:
> I was trying to remember why I proposed the naming I did. I guess it was an attempt at brevity.

I also missed it during review which is odd because @jrperritt pointed out the same thing with the flavor access PR which was done at the same time. No big deal 😄

jtopjian:
Thank you both for the quick review.

Issue #763, opened 2018-02-14
Networking: Subnetpool id field in the Subnet


ozerovandrei:
Neutron subnetpools support were implemented some time ago: [#672](https://github.com/gophercloud/gophercloud/issues/672).
But the subnet resource doesn't have the subnetpool id in it's CRUD methods in the Gophercloud.
That should be fixed.

[API reference](https://developer.openstack.org/api-ref/network/v2/index.html#subnets)

PR #764, opened 2018-02-14
Add SubnetPoolID into the Subnet struct and GET requests

ozerovandrei:
Add ability to populate subnet's subnetpool_id field from GET requests.
Also add this field to GET and LIST unit tests.

For #763 

code in this PR:

https://github.com/openstack/neutron/blob/stable/pike/neutron/objects/subnet.py#L179


ozerovandrei:
@jtopjian this can be reviewed. I think that it's better to add the SubnetPoolID to create and update requests in different PRs. Also I've checked that in my OpenStack test installation.

jtopjian:
@ozerovandrei Looks good to me. Nice work 😄

PR #765, opened 2018-02-14
Add the SubnetPoolID to the Subnet Create request

ozerovandrei:
Add the "subnetpool_id" field to the subnet creation request with an acceptance test.
Also update the unit test.

For #763 

code in this PR:

Here you can see a subnetpool_id in the subnet collection of a resource map:
https://github.com/openstack/neutron/blob/stable/pike/neutron/api/v2/attributes.py#L149
it's provided into this method:
https://github.com/openstack/neutron/blob/master/neutron/neutron_plugin_base_v2.py#L33

### DIFF
--- a/acceptance/openstack/networking/v2/networking.go
+++ b/acceptance/openstack/networking/v2/networking.go
@@ -188,6 +188,32 @@ func CreateSubnetWithNoGateway(t *testing.T, client *gophercloud.ServiceClient, 
 	return subnet, nil
 }
 
+// CreateSubnetWithSubnetPool will create a subnet associated with the provided subnetpool on the specified Network ID.
+// An error will be returned if the subnet or the subnetpool could not be created.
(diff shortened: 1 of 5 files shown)


ozerovandrei:
This is ready for a review.
It depends on a #764.

jtopjian:
@ozerovandrei This also looks good. Nice work on the acceptance test, too.

There's one issue, though: because I squashed the previous PR, merging this might cause the git history to look strange. Would you mind rebasing this with master so only the one commit is in this PR?

After that, this is good to go.

ozerovandrei:
@jtopjian done.

jtopjian:
recheck

jtopjian:
The openlab failure looks to be an issue with devstack. Since it passed before, it should be fine to merge this.

@ozerovandrei Thank you again for your work on this.

Issue #766, opened 2018-02-15
Networking V2: some subnetpools boolean fields should be pointers in the create request


ozerovandrei:
I've just started working on the [terraform-provider-openstack#242](https://github.com/terraform-providers/terraform-provider-openstack/issues/242) and noticed that we're using a `bool` type for some fields inside the `subnetpools.CreateOpts`:
```
// Shared indicates whether this network is shared across all projects.
Shared bool `json:"shared,omitempty"`

// IsDefault indicates if the subnetpool is default pool or not.
IsDefault bool `json:"is_default,omitempty"`
```
They're should be `*bool` so create request body won't has them by default, because they're optional.
(And `Shared` in-line commentary should be fixed also. It's not a network, it's a subnetpool).

Api reference: [Create SubnetPool](https://developer.openstack.org/api-ref/network/v2/#create-subnet-pool)

jrperritt:
Currently those fields default to `false` (which is the zero-value for booleans). How would this change affect that (i.e. what will be possible with this change that can't currently be done)?

jtopjian:
Perhaps to clarify:

If a user specifies `Shared` as `false`, Gophercloud will never include it in the request. So the question is: what's Neutron's default behavior when `shared` isn't sent in the body? Is it to make the subnet pool not shared? If so, then everything is fine and no change is required.

But if the omission of `shared` in the request body is different than `"shared": false` in the body, then yes, these do need changed to pointers.

The best place to determine this would be in the neutron service code.

side note: handling bools is one area that has always felt more complicated than it needs to be (not that it's _very_ complicated, it just feels like you need to think about it more than you should). I feel there's a valid argument/opinion for making all bools `*bool` even if it's not required sometimes. I'm not saying we should do that for Gophercloud - just saying I sympathize whenever a question of bools comes up. The best solution for this issue is to dig into the code and figure out what the default behavior is.

ozerovandrei:
[...]

PR #767, opened 2018-02-15
Use pointers to some subnetpools CreateOpts fields

ozerovandrei:
We need to use pointers to optional "Shared" and "IsDefault" fields of
the CreateOpts structure.

For #766 

code in this PR:

https://github.com/openstack/neutron/blob/stable/pike/neutron/db/db_base_plugin_v2.py#L1117


ozerovandrei:
recheck

ozerovandrei:
As discussed in the #766 we don't need it.

PR #770, opened 2018-02-19
Network v2: RBAC-GET-BY-ID

manju754:
For #731 

code in this PR:

API:
https://developer.openstack.org/api-ref/network/v2/#list-rbac-policies

Schema:
https://github.com/openstack/neutron/blob/8f79a92e45c84819cfc4d1cda99c9320ae78bbaf/neutron/db/rbac_db_models.py#L92

Get by ID:
https://github.com/openstack/neutron/blob/59e2c40f14157e0f2f64c340070cf46758d522e1/neutron/db/rbac_db_mixin.py#L112


jtopjian:
@manju754 Thanks for your work on this. When you have a moment, can you rebase / squash with the master branch?

I did a cursory look and everything looks good, but I'll do a more detailed check once it's rebased.

One area that might be simplified is removing the Get acceptance test and adding a Get call to the `TestRBACPolicyCreate` test.

manju754:
@jtopjian Can you review this PR. I had re-based the PR with master and resolved all the conflicts  and made necessary changes as suggested.

jtopjian:
@manju754 This looks good to me - nice work :smile: 

Perhaps one thing to add later is an example of `Get` to the `doc.go` file. I don't consider docs to be blocking, so I'm going to merge this now.

PR #771, opened 2018-02-19
Vpnaas: Service Get/List

simonre:
For #723 

code in this PR:

List:
https://github.com/openstack/neutron-vpnaas/blob/058469e1b99b647537a5228c6a384d93df5484df/neutron_vpnaas/db/vpn/vpn_db.py#L531
Get:
https://github.com/openstack/neutron-vpnaas/blob/058469e1b99b647537a5228c6a384d93df5484df/neutron_vpnaas/db/vpn/vpn_db.py#L526

API:
https://developer.openstack.org/api-ref/network/v2/#show-vpn-service-details
https://developer.openstack.org/api-ref/network/v2/#list-vpn-services

### DIFF
--- a/openstack/networking/v2/extensions/vpnaas/services/requests.go
+++ b/openstack/networking/v2/extensions/vpnaas/services/requests.go
@@ -1,6 +1,9 @@
 package services
 
-import "github.com/gophercloud/gophercloud"
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
 
 // CreateOptsBuilder allows extensions to add additional parameters to the
 // Create request.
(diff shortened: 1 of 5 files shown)


simonre:

openstack/networking/v2/extensions/vpnaas/services/requests.go
```diff
@@ -50,3 +53,53 @@ func Create(c *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResul
...
+// List request.
+type ListOptsBuilder interface {
+	ToServiceListQuery() (string, error)
+}
+
+// ListOpts allows the filtering and sorting of paginated collections through
+// the API. Filtering is achieved by passing in struct field values that map to
+// the VPN service attributes you want to see returned.
+type ListOpts struct {
+	ID           string `q:"id"`
```
jtopjian: I think ID is a moot field to have for a List call. I understand it probably works because Neutron-based services can list on any field. IMO let's remove this unless you see a valid use-case to keep it.

openstack/networking/v2/extensions/vpnaas/services/testing/requests_test.go
```diff
@@ -75,3 +76,107 @@ func TestCreate(t *testing.T) {
...
+
+		fmt.Fprintf(w, `
+{
+    "vpnservice": {
+        "router_id": "66e3b16c-8ce5-40fb-bb49-ab6d8dc3f2aa",
+        "status": "PENDING_CREATE",
+        "name": "vpnservice1",
+        "admin_state_up": true,
+        "subnet_id": null,
+        "project_id": "10039663455a446d8ba2cbb058b0f578",
```
jtopjian: If `project_id` is returned in the result, perhaps we should add it to the `Services` result struct?

jtopjian: Actually, maybe hold off on this. See my comment in #784

simonre: I'll add it to the `Service`struct. Do you think it makes sense to add it to the ListOpts as well given that you can list by TenantID already?

jtopjian: Yes, please do.

jtopjian:
@simonre This looks good to me. I wanted to check if you were finished and if this is ready to be merged?

simonre:
@jtopjian Yes, this is finished. I wasn't sure whether I should ping you again.

jtopjian:
@simonre Pinging is always best so I don't accidentally merge something that wasn't ready 😄 

Thanks!

PR #777, opened 2018-02-19
Allow dns_name and dns_domain to be set on networking resources

grahamhayes:
For ##776

code in this PR:

Network `dns_domain` : https://github.com/openstack/neutron/blob/master/neutron/objects/network.py#L216

Port `dns_domain` and `dns_name` : https://github.com/openstack/neutron/blob/master/neutron/objects/ports.py#L286 which links to https://github.com/openstack/neutron/blob/master/neutron/objects/ports.py#L207

Floating IP `dns_domain` and `dns_name` : https://github.com/openstack/neutron/blob/master/neutron/objects/floatingip.py#L33-L34

### DIFF
--- a/openstack/networking/v2/extensions/layer3/floatingips/requests.go
+++ b/openstack/networking/v2/extensions/layer3/floatingips/requests.go
@@ -23,6 +23,8 @@ type ListOpts struct {
 	SortDir           string `q:"sort_dir"`
 	RouterID          string `q:"router_id"`
 	Status            string `q:"status"`
+	DNSDomain         string `q:"dns_domain"`
+	DNSName           string `q:"dns_name"`
 }
 
 // List returns a Pager which allows you to iterate over a collection of
@@ -55,6 +57,8 @@ type CreateOpts struct {
 	FixedIP           string `json:"fixed_ip_address,omitempty"`
 	SubnetID          string `json:"subnet_id,omitempty"`
 	TenantID          string `json:"tenant_id,omitempty"`
+	DNSDomain         string `json:"dns_domain,omitempty"`
+	DNSName           string `json:"dns_name,omitempty"`
 }
 
 // ToFloatingIPCreateMap allows CreateOpts to satisfy the CreateOptsBuilder
@@ -114,7 +118,9 @@ type UpdateOptsBuilder interface {
 // linked to. To associate the floating IP with a new internal port, provide its
 // ID. To disassociate the floating IP from all ports, provide an empty string.
 type UpdateOpts struct {
-	PortID *string `json:"port_id"`
+	PortID    *string `json:"port_id"`
+	DNSDomain string  `json:"dns_domain,omitempty"`
+	DNSName   string  `json:"dns_name,omitempty"`
 }
 
 // ToFloatingIPUpdateMap allows UpdateOpts to satisfy the UpdateOptsBuilder
--- a/openstack/networking/v2/extensions/layer3/floatingips/results.go
+++ b/openstack/networking/v2/extensions/layer3/floatingips/results.go
@@ -39,6 +39,12 @@ type FloatingIP struct {
 
 	// RouterID is the ID of the router used for this floating IP.
 	RouterID string `json:"router_id"`
+
+	// Specifies the DNS domain used for autocreating public DNS entries for this floating IP (overrides port and network DNSDomain settings)
+	DNSDomain string `json:"dns_domain,omitempty"`
+
+	// Specifies the DNS Name used as the first label of autocreated public DNS entries for this floating IP (overrides port DNSName setting)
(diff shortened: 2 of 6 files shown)


jtopjian:
@grahamhayes Thank you for submitting this.

According to https://docs.openstack.org/mitaka/networking-guide/config-dns-int.html, DNS integration is an extension that isn't enabled by default. The extension definition is here:

https://github.com/openstack/neutron/blob/a0e0e8b6686b847a4963a6aa6a3224b5768544e6/neutron/extensions/dns.py

If this was a standard/default extension, it might have been OK to integrate these into the core fields, but since it's something that must be explicitly enabled, we should treat these fields as a proper extension.

We have existing resources that you can use as examples. Two good ones to look at are:

* https://github.com/gophercloud/gophercloud/tree/master/openstack/networking/v2/extensions/portsbinding
* https://github.com/gophercloud/gophercloud/tree/master/openstack/networking/v2/extensions/portsecurity

One notable difference is that this single extension is adding fields for multiple resources. So, for example with "create", instead of having [`CreateOptsExt`](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/portsbinding/requests.go#L7-L24), you'll probably have to do `CreateNetworkOptsExt`, `CreateFloatingIPOptsExt`, and `CreatePortOptsExt`. 

As well, given that there are 3 resources which need fields added to multiple request types (List, Create, Update, and the result structs), this feature set should be broken out into separate PRs. 

I realize this complicates what may have seemed like a simple PR to add some fields, but again, given this is an extension which must be explicitly enabled, it's best to treat it as a proper extension. 

I am happy to help answer questions or assist if you run into any problems.

grahamhayes:
@jtopjian thanks for the review.

While this is an extra extension (due to the way neutron operates, everything is an extension, but you need to enable some of them to have a working install), it is a special one and the neutron objects will actually store the ``dns_name`` and ``dns_domain`` parameters, and display them in the API even if the extension is disabled.

This allows for the data to be present if an admin enables the plugin at a later date.

jtopjian:
@grahamhayes 

> While this is an extra extension (due to the way neutron operates, everything is an extension, but you need to enable some of them to have a working install),

Right - this is why I mentioned that if this was a standard/default extension, it'd be OK to include these fields as "core".

> it is a special one and the neutron objects will actually store the dns_name and dns_domain parameters, and display them in the API even if the extension is disabled.

I'm not seeing this... In my test environment, I have the DNS extension disabled and the following fields are being returned when retrieving a network:

```json
{
    "network": {
        "admin_state_up": true,
        "availability_zone_hints": [],
        "availability_zones": [
            "nova"
        ],
        "created_at": "2018-01-07T23:50:03Z",
        "description": "",
        "id": "db861f74-f386-4de7-ba47-ec5d6b85d768",
        "ipv4_address_scope": null,
        "ipv6_address_scope": null,
        "mtu": 1450,
        "name": "private",
        "port_security_enabled": true,
        "project_id": "5aa69d9135524fddab887759f3b6c92d",
        "revision_number": 4,
        "router:external": false,
        "shared": false,
        "status": "ACTIVE",
        "subnets": [
            "9a00fe2c-2496-4d17-8db5-78cf1a986c52"
        ],
        "tags": [],
        "tenant_id": "5aa69d9135524fddab887759f3b6c92d",
        "updated_at": "2018-01-08T00:02:02Z"
    }
}
```

If this was a special/core/standard extension, wouldn't it return `dns_domain` regardless if there was a value?

Likewise, creating a network with `dns_domain` without the extension disabled returns an error:

```shell
$ neutron net-create --dns-domain example.com foo
neutron CLI is deprecated and will be removed in the future. Use openstack CLI instead.
Unrecognized attribute(s) 'dns_domain'
```

I get what you're saying about this being an important extension, but I do believe that breaking it out into a dedicated extension package is the right way to go.

grahamhayes:
Yeah - this was an error on my part, I disabled the plugin but didn't actually restart neutron.

On a clean install, seeing ^ errors. 

I will pull it out into an extension, but it will be a week or two, will have to wait until after the PTG.

Thanks for double checking :)

jtopjian:
No problem at all and no rush. Coincidentally I've submitted a PR which will look very similar:

https://github.com/gophercloud/gophercloud/pull/787

which you can use as a template. The one question we need to agree on is the name of the struct to use. It might sound a little bike sheddy but since it's the first occurrence, it'll set precedence. Should be sorted out by the time you're ready to work on this.

As well, if you find you aren't able to get back to this, I'm happy to put a PR together. I think it would be a very worthwhile feature to have. From a Terraform point of view, it'll allow the creation of a custom domain/zone with the DNS resources and creation of networks and ports which can easily leverage that domain.

Issue #779, opened 2018-02-20
Image Service v2: Add Missing Fields to ListOpts


jtopjian:
Following-up on an old Terraform feature request, there are a few fields missing from ListOpts:

1. It's possible to list by `container_format` and `disk_format`:

https://github.com/openstack/glance/blob/05e9bdb656d9c120ed3cd6ffc8ae7dbf5614b5e4/glance/db/simple/api.py#L324-L325

2. `Tag` should actually be `Tags` as multiple tags are allowed:

https://github.com/openstack/glance/blob/05e9bdb656d9c120ed3cd6ffc8ae7dbf5614b5e4/glance/db/simple/api.py#L340

https://developer.openstack.org/api-ref/image/v2/index.html#images

> To filter using image tags, use the filter tag (note the singular). To filter on multiple tags, include each tag separately in the query. For example, to find images with the tag ready, include tag=ready in your query string. To find images tagged with ready and approved, include tag=ready&tag=approved in your query string. (Note that only images containing both tags will be included in the response.)

3. It's possible to search by `created_at` and `updated_at`, but they can also have filter operations applied to them:

See "Time Comparison Filters" here: https://developer.openstack.org/api-ref/image/v2/index.html#images

https://github.com/openstack/glance/blob/9eddf85d8ab457d2eabb21597bd3e5b8d33f0086/glance/common/utils.py#L696-L723

https://github.com/openstack/glance/blob/05e9bdb656d9c120ed3cd6ffc8ae7dbf5614b5e4/glance/db/simple/api.py#L317-L323

jrperritt:
closed by #780

PR #780, opened 2018-02-20
Image Service v2: Add Missing Fields to ListOpts

jtopjian:
For #779 

The first commit is the most notable and deserves a good review. I don't mind changing the names of the new types or consts:

https://github.com/gophercloud/gophercloud/commit/b364a87bb579aa578e8ae95721546e3acaba191e

The second commit, changing `Tag` to `Tags` will be a breaking change for some. I thought about keeping `Tag` for backwards compatibility, but that felt awkward.

The third commit is pretty standard.

/cc @dklyle

### DIFF
--- a/acceptance/openstack/imageservice/v2/images_test.go
+++ b/acceptance/openstack/imageservice/v2/images_test.go
@@ -134,7 +134,7 @@ func TestImagesListByDate(t *testing.T) {
 	}
 }
 
-func TestImagesFilterByTags(t *testing.T) {
+func TestImagesFilter(t *testing.T) {
 	client, err := clients.NewImageServiceV2Client()
 	if err != nil {
 		t.Fatalf("Unable to create an image service client: %v", err)
@@ -148,7 +148,9 @@ func TestImagesFilterByTags(t *testing.T) {
 	defer DeleteImage(t, client, image)
 
 	listOpts := images.ListOpts{
-		Tags: []string{"foo", "bar"},
+		Tags:            []string{"foo", "bar"},
+		ContainerFormat: "bare",
+		DiskFormat:      "qcow2",
 	}
 
 	allPages, err := images.List(client, listOpts).AllPages()
--- a/acceptance/openstack/imageservice/v2/images_test.go
+++ b/acceptance/openstack/imageservice/v2/images_test.go
@@ -133,3 +133,35 @@ func TestImagesListByDate(t *testing.T) {
 		t.Fatalf("Expected 0 images, got %d", len(allImages))
 	}
 }
+
+func TestImagesFilterByTags(t *testing.T) {
+	client, err := clients.NewImageServiceV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create an image service client: %v", err)
+	}
+
+	image, err := CreateEmptyImage(t, client)
+	if err != nil {
+		t.Fatalf("Unable to create empty image: %v", err)
+	}
+
+	defer DeleteImage(t, client, image)
+
+	listOpts := images.ListOpts{
+		Tags: []string{"foo", "bar"},
+	}
+
+	allPages, err := images.List(client, listOpts).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to retrieve all images: %v", err)
+	}
+
+	allImages, err := images.ExtractImages(allPages)
+	if err != nil {
+		t.Fatalf("Unable to extract images: %v", err)
+	}
+
+	if len(allImages) == 0 {
+		t.Fatalf("Query resulted in no results")
+	}
+}
--- a/acceptance/openstack/imageservice/v2/images_test.go
+++ b/acceptance/openstack/imageservice/v2/images_test.go
@@ -4,6 +4,7 @@ package v2
 
 import (
 	"testing"
+	"time"
 
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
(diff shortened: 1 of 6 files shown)


acceptance/openstack/imageservice/v2/images_test.go
```diff
@@ -148,7 +148,9 @@ func TestImagesFilterByTags(t *testing.T) {
 	defer DeleteImage(t, client, image)
 
 	listOpts := images.ListOpts{
-		Tags: []string{"foo", "bar"},
+		Tags:            []string{"foo", "bar"},
+		ContainerFormat: "bare",
```
dklyle: Per https://developer.openstack.org/api-ref/image/v2/#show-images `the in operator` both these look like they could be more complicated, but this should work as is for single value matching.

jtopjian: Right. I figured the easiest thing might be to keep these as string as to allow people to pass in non-single-value strings.

openstack/imageservice/v2/images/types.go
```diff
@@ -77,3 +81,22 @@ const (
...
+	FilterLT  ImageDateFilter = "lt"
+	FilterLTE ImageDateFilter = "lte"
+	FilterNEQ ImageDateFilter = "neq"
+	FilterNE  ImageDateFilter = "eq"
+)
+
+// ImageDateQuery represents a date field to be used for listing images.
+type ImageDateQuery struct {
+	Date   time.Time
+	Filter ImageDateFilter
```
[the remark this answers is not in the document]
jtopjian: If a date is specified but no filter is given, then the query performs a "date not equal to" filter? That's kind of weird. I don't mind adding a comment saying as much.

jtopjian:
@dklyle Thanks for the review. I also see that `ID` is still missing, which might seem counter-intuitive for a List call, but the user can specify a string of `in:uuid,uuid` if they want.

Also, the `Status` can also have `in` operations, so maybe that should be changed to a string as well.

I'll add two commits for this shortly.

openstack/imageservice/v2/images/types.go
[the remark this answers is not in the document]
jtopjian: oh nevermind - that's a typo in my code. It should be `FilterEQ`. I deleted the wrong character when I lazily copied the line :smile:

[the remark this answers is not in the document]
dklyle: Even stranger is that FilterNE is date equals FilterNEQ is date not equals. Per the code above.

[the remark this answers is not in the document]
dklyle: Ok. That makes a lot more sense.

jtopjian:
OK, I've added a bunch of small commits which address the above-mentioned items. Please definitely nit pick this PR. This feature set is something I've put off for a while knowing the Glance API is a little weird and will probably require a few iterations.

jtopjian:
> I'm glad handling the in business was so straightforward

I agree. It's a little lazy, but it works. I think creating a proper schema for it would be overly complicated.

Thank you for your help and review on this. I'll let @jrperritt to do a sign-off of the breaking changes + new type when he has some time.

acceptance/openstack/imageservice/v2/images_test.go
```diff
@@ -78,3 +79,91 @@ func TestImagesCreateDestroyEmptyImage(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Unable to extract images: %v", err)
+	}
+
+	if len(allImages) > 0 {
+		t.Fatalf("Expected 0 images, got %d", len(allImages))
+	}
+}
+
+func TestImagesFilter(t *testing.T) {
```
jrperritt: Is this testing the filter? It looks like it's just creating an image and listing all images.

openstack/imageservice/v2/images/requests.go
```diff
@@ -37,7 +48,9 @@ type ListOpts struct {
 	Owner string `q:"owner"`
 
 	// Status filters on the status of the image.
-	Status ImageStatus `q:"status"`
+	// Multiple statuses can be specified by constructing a string
+	// such as "in:saving,queued".
+	Status string `q:"status"`
```
jrperritt: It seems like users could still create their own `ImageStatus` by wrapping a string, right? Or would this not work:
```go
imageStatus := images.ImageStatus("in:saving,queued")
```
?

openstack/imageservice/v2/images/requests.go
```diff
@@ -56,12 +69,52 @@ type ListOpts struct {
 
 	// SortDir will sort the list results either ascending or decending.
 	SortDir string `q:"sort_dir"`
-	Tag     string `q:"tag"`
+
+	// Tags filters on specific image tags.
+	Tags []string `q:"tag"`
+
+	// CreatedAt filters images based on their creation date.
+	CreatedAt *ImageDateQuery
```
jrperritt: I wonder if these should be named differently since `CreatedAt` is used elsewhere as a `time.Time`. Would something like `CreatedAtQuery` be any better?

acceptance/openstack/imageservice/v2/images_test.go
jtopjian: This test is using `ListOpts` to "filter" the image by certain attributes (lines 150-154).

openstack/imageservice/v2/images/requests.go
jtopjian: Ah, good call. I'll change this.

openstack/imageservice/v2/images/requests.go
jtopjian: Makes sense to me. `CreatedAtQuery` it is.

jtopjian:
@jrperritt ping for review

Issue #343, opened 2017-05-18
Keystone v3 is deprecating region in favor or region_id


arnaudmorin:
Since keystone 3.2, as seen in keystone documentation (look for requests/response example):
https://developer.openstack.org/api-ref/identity/v3/

the field region is deprecated in favor of region_id.
Gophercloud is still using region.

Some Public Cloud keystone deployment (OVH) already dropped this parameter and only give region_id, and as a result is gophercloud keystonev3 is broken.

We should at least patch this file: 
https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/tokens/results.go#L15
and replace region by region_id.

Best is to handle both cases.

What cores dev think about that?

dims:
Hey @lbragstad, what do you think of this?

lbragstad:
The version should be discoverable using `GET /v3` [0]. That should help determine which representation of region to use. Are you seeing specific issues or confusing using ``region`` or ``region_id`` with the endpoints resource?

[0] https://github.com/openstack/keystone/blob/017b3477fdc3922f7a0ea0db420f3ed1df55b386/keystone/version/controllers.py#L144

jrperritt:
closed by #783

PR #783, opened 2018-02-21
Identity v3: Add support for RegionID in token catalog endpoints

jtopjian:
For #343 

I was looking at older issues and came across #343. After digging around in the Keystone code, I determined two things:

1. The catalog endpoints in a token contain _both_ region and region_id:

   https://github.com/openstack/keystone/blob/master/keystone/catalog/backends/sql.py#L360
   
   I double confirmed this by modifying the above line to read `regionnnn` and, sure enough, the token response contained the malformed name. Secondly, our test fixtures already had both `region` and `region_id`.

2. "region" is still properly handled as of the current Keystone master branch. #343 mentions that OVH has dropped the `region` parameter. If this is true, then:
   
   a. Perhaps it was only temporary or else I think we would have heard from more users by now.
   b. They are deviating from "vanilla" openstack which means the problem isn't Gophercloud's to solve.

_However_, I've patched the endpoint search code to account for a missing region field. I figured it was easy to do, future-proofs a possible removal of `region` (there are comments in the Keystone code which mention `region` is for backwards compatibility), and it would be a pain for apps which want to support multiple clouds where one cloud is using `region_id` exclusively and it's only a few extra characters for Gophercloud core to enable this.

/cc @dklyle 

marking as `[wip]` until openlab clears this change.

### DIFF
--- a/openstack/testing/endpoint_location_test.go
+++ b/openstack/testing/endpoint_location_test.go
@@ -178,6 +178,30 @@ var catalog3 = tokens3.ServiceCatalog{
 				},
 			},
 		},
+		tokens3.CatalogEntry{
+			Type: "someother",
+			Name: "someother",
+			Endpoints: []tokens3.Endpoint{
+				tokens3.Endpoint{
+					ID:        "1",
+					Region:    "someother",
+					Interface: "public",
+					URL:       "https://public.correct.com/",
+				},
+				tokens3.Endpoint{
+					ID:        "2",
+					RegionID:  "someother",
+					Interface: "admin",
+					URL:       "https://admin.correct.com/",
+				},
+				tokens3.Endpoint{
+					ID:        "3",
+					RegionID:  "someother",
+					Interface: "internal",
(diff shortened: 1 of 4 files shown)


openstack/testing/endpoint_location_test.go
```diff
@@ -178,6 +178,30 @@ var catalog3 = tokens3.ServiceCatalog{
 				},
 			},
 		},
+		tokens3.CatalogEntry{
+			Type: "someother",
+			Name: "someother",
+			Endpoints: []tokens3.Endpoint{
+				tokens3.Endpoint{
+					ID:        "1",
+					Region:    "someother",
```
jtopjian: Note that `Region` here is intentional to test that both `Region` and `RegionID` are working.

jtopjian:
@jrperritt @dklyle This is ready for review at your conveniences.

jtopjian:
@dklyle Thanks for your review on this, too.

PR #784, opened 2018-02-21
Vpnaas: Update Service

simonre:
For #723 

code in this PR:

https://github.com/openstack/neutron-vpnaas/blob/058469e1b99b647537a5228c6a384d93df5484df/neutron_vpnaas/db/vpn/vpn_db.py#L502

API:

https://developer.openstack.org/api-ref/network/v2/#update-vpn-service

### DIFF
--- a/openstack/networking/v2/extensions/vpnaas/services/requests.go
+++ b/openstack/networking/v2/extensions/vpnaas/services/requests.go
@@ -61,6 +61,42 @@ func Delete(c *gophercloud.ServiceClient, id string) (r DeleteResult) {
 	return
 }
 
+// UpdateOptsBuilder allows extensions to add additional parameters to the
+// Update request.
+type UpdateOptsBuilder interface {
+	ToServiceUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts contains the values used when updating a VPN service
+type UpdateOpts struct {
+	// Name is the human readable name of the service.
+	Name *string `json:"name,omitempty"`
+
+	// Description is the human readable description of the service.
+	Description *string `json:"description,omitempty"`
+
+	// AdminStateUp is the administrative state of the resource, which is up (true) or down (false).
+	AdminStateUp *bool `json:"admin_state_up,omitempty"`
+}
+
+// ToServiceUpdateMap casts aa UodateOpts struct to a map.
+func (opts UpdateOpts) ToServiceUpdateMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "vpnservice")
+}
+
+// Update allows VPN services to be updated.
+func Update(c *gophercloud.ServiceClient, id string, opts UpdateOptsBuilder) (r UpdateResult) {
+	b, err := opts.ToServiceUpdateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = c.Put(resourceURL(c, id), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	return
+}
+
 // ListOptsBuilder allows extensions to add additional parameters to the
 // List request.
 type ListOptsBuilder interface {
--- a/openstack/networking/v2/extensions/vpnaas/services/testing/requests_test.go
+++ b/openstack/networking/v2/extensions/vpnaas/services/testing/requests_test.go
@@ -194,7 +194,74 @@ func TestDelete(t *testing.T) {
 		th.TestHeader(t, r, "X-Auth-Token", fake.TokenID)
 		w.WriteHeader(http.StatusNoContent)
 	})
-
 	res := services.Delete(fake.ServiceClient(), "5c561d9d-eaea-45f6-ae3e-08d1a7080828")
 	th.AssertNoErr(t, res.Err)
+}
+
+func TestUpdate(t *testing.T) {
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+
+	th.Mux.HandleFunc("/v2.0/vpn/vpnservices/5c561d9d-eaea-45f6-ae3e-08d1a7080828", func(w http.ResponseWriter, r *http.Request) {
+
+		th.TestMethod(t, r, "PUT")
+		th.TestHeader(t, r, "X-Auth-Token", fake.TokenID)
+		th.TestHeader(t, r, "Content-Type", "application/json")
+		th.TestHeader(t, r, "Accept", "application/json")
+		th.TestJSONRequest(t, r, `
+{
+    "vpnservice":{
+        "name": "updatedname",
+        "description": "updated service",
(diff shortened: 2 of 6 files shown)


simonre:

openstack/networking/v2/extensions/vpnaas/services/requests.go
```diff
@@ -57,3 +57,39 @@ func Delete(c *gophercloud.ServiceClient, id string) (r DeleteResult) {
...
+// UpdateOptsBuilder allows extensions to add additional parameters to the
+// Update request.
+type UpdateOptsBuilder interface {
+	ToServiceUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts contains the values used when updating a VPN service
+type UpdateOpts struct {
+	// Name is the human readable name of the service.
+	Name string `json:"name,omitempty"`
```
jtopjian: Is it possible to _not_ have a Name? If so, let's say a user originally set a name but then wants to remove the name. By having the field configured like the above, that's not possible because an empty string is `string`'s zero-value and with `omitempty`, it won't be sent.

So we'll need to do:

```go
Name *string `json:"name,omitempty"`
```

That way a pointer to an empty string will be sent and successfully clear the name.

openstack/networking/v2/extensions/vpnaas/services/requests.go
```diff
@@ -57,3 +57,39 @@ func Delete(c *gophercloud.ServiceClient, id string) (r DeleteResult) {
...
+
+	// Description is the human readable description of the service.
+	Description string `json:"description,omitempty"`
```
jtopjian: Same as above.

openstack/networking/v2/extensions/vpnaas/services/testing/requests_test.go
```diff
@@ -85,7 +85,71 @@ func TestDelete(t *testing.T) {
...
+
+		fmt.Fprintf(w, `
+{
+    "vpnservice": {
+        "router_id": "66e3b16c-8ce5-40fb-bb49-ab6d8dc3f2aa",
+        "status": "PENDING_CREATE",
+        "name": "updatedname",
+        "admin_state_up": false,
+        "subnet_id": null,
+        "tenant_id": "10039663455a446d8ba2cbb058b0f578",
```
jtopjian: This is odd. So in this test fixture `project_id` is not here... 

Ignoring the API docs, because they can sometimes have old or incorrect information, what are you seeing in your working environment? You can view the raw API request/response by calling neutron with `neutron --debug`.

jtopjian:
@simonre Also, this file is a good resource to determine which fields can be used in both Create (`allow_post`) and Update (`allow_put`).

openstack/networking/v2/extensions/vpnaas/services/testing/requests_test.go
simonre: The API is returning 'project_id' in the response in my working environment when running 
`neutron --debug vpn-service-list`. Should I include it as a field in the result then?

jtopjian: Yes please. We always prefer actual API output over API references.

simonre: Ok. I was under the impression that I should only include the fields reference in the source code [here](https://github.com/openstack/neutron-vpnaas/blob/058469e1b99b647537a5228c6a384d93df5484df/neutron_vpnaas/db/vpn/vpn_models.py).

jtopjian: Yeah, sometimes there's a lot of mystery and investigation to this.

You're absolutely right: if the service code is showing fields, that's confident confirmation that the fields are valid.

If the API references are showing fields that the aren't in the code, then there are two possibilities:

1. The API reference is wrong (it happens frequently enough)
2. The field is being added _somewhere_ else in the service code. This happens a lot with Neutron, specifically with `project_id` and `tenant_id`. So the next source of confident information is the API interaction between you and your OpenStack environment. 

For fields like `project_id`, `created_at`, or some other "meta" field, we can be fairly confident that they're _somewhere_ in Neutron and as long as you're seeing them in your API interaction, that's good enough for me. But let's say your API interaction was also showing something like `router_id`. Then we'll really want to dig into the Neutron code (by either grepping the entire code for `router_id` or using the github search box) and figuring out where it's coming from. That's ugly work, but it makes sure Gophercloud is correct.

With regard to test fixtures, please always use fixtures generated yourself rather than fixtures from the API references. We (gophercloud) agreed to enforce this a while back but I occasionally forget to check. Using "real" fixtures helps us make sure the data is correct.

Let me know if all of that makes sense.

simonre: Yes, that does make a lot of sense. I'll keep that in mind from now on. Thanks for the explanation!

simonre:
@jtopjian I put the state of this back to [WIP] because I noticed a problem with writing acceptance tests for this. A VPN service can only be updated after its state has been set to `ACTIVE`. The state of a VPN Service is only set to `ACTIVE` after an IPSec Site connection for it has been created. This means that there are two alternatives for this pull request:
1. Merge without an acceptance test
2. Create files for all other resources for VPNaas first, then create an IPSec policy, an IKE policy and an IPSec site connection inside the acceptance test and test the update method after waiting for the `ACTIVE` state of the VPN service. 
Let me know if this didn't make sense or I overlooked another way of acceptance testing this.

jtopjian:
@simonre Good catch. Let's go with Option 2.

I would prefer to merge the VPN suite in order of functionality. If it's not possible to Update a connection entirely through Gophercloud, let's hold off until it is.

jtopjian:
@simonre Makes sense to me - thank you for the detailed explanation. This sounds like a pretty complex test fixture. I'm sure there's _some_ way to do it, but I'm not immediately sure how. 

> I can create the two sides of the connection without a problem entirely through gophercloud

This is good enough for me. You've done some great work with the other acceptance tests and if this is the only test omitted in the entire VPNaaS suite, I think that is acceptable. :)

The pieces will be in place for either you or someone else to add this in the future.

simonre:
> if this is the only test omitted

It would actually be one of two tests omitted in the suite. The other one would be for the update function for ipsec site connections. Updating a site connection has the exact same issue(Can't be updated while the state is `PENDING_CREATE`). 

One other solution I could think of would be to somehow get the connection state to either `DOWN` or `ERROR` as according to the documentation those might work with the update function. Problem is again that I am not sure how to provoke those states intentionally.

For now I'll omit both tests if that's ok.

jtopjian:
> For now I'll omit both tests if that's ok.

Yes, that sounds good to me.

simonre:

jtopjian:
@simonre Looks good to me!

PR #787, opened 2018-02-22
Networking v2: Port Security Create

jtopjian:
For #298 

Continuing with catching up on old work.

This is a rather large PR since it handles Create for both networks and ports, but since they're identical in implementation, I added them both here. The bulk of the additions are tests, though.

One thing to confirm with @jrperritt is the name `PortCreateOptsExt` and `NetworkCreateOptsExt`. I waffled on using either that form or `CreatePortOptsExt` and `CreateNetworkOptsExt`. This is the first occurrence of an extension being able to support extending multiple `CreateOpts`, so the name is new.

/cc @dklyle

### DIFF
--- a/openstack/networking/v2/extensions/portsecurity/requests.go
+++ b/openstack/networking/v2/extensions/portsecurity/requests.go
@@ -1,0 +1,55 @@
+package portsecurity
+
+import (
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
+)
+
+// PortCreateOptsExt adds port security options to the base ports.CreateOpts.
+type PortCreateOptsExt struct {
+	ports.CreateOptsBuilder
+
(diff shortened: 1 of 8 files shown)


openstack/networking/v2/extensions/portsecurity/requests.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/ports"
+)
+
+// PortCreateOptsExt adds port security options to the base ports.CreateOpts.
+type PortCreateOptsExt struct {
+	ports.CreateOptsBuilder
+
+	// PortSecurityEnabled toggles port security on a port.
+	PortSecurityEnabled *bool `json:"port_security_enabled,omitempty"`
```
jtopjian: The reason I chose `*bool` here is because when the port security extension is enabled, port security is turned on by default. `*bool` will prevent accidentally turning off port security.

Issue #788, opened 2018-02-23
Networking v2: Fix up subnetpools


jtopjian:
During implementation of https://github.com/terraform-providers/terraform-provider-openstack/pull/243, we discovered that it is not possible to query by Prefixes.

The fields which Neutron allows for listing is dynamic, based on the fields of the resource itself. I guess a limitation of this is fields which are arrays.

Doing further testing, I found that it is also not possible to list by `created_at` or `updated_at`.

And in reviewing the code, I found a few other areas I missed during the original code review:

* The `bool` fields in `ListOpts` need to be `*bool` to allow values of `false` to be sent.
* The result struct's `CreatedAt` and `UpdatedAt` need converted to `time.Time`. This is probably the first occurrence I've seen where the timestamp doesn't need a special conversion.

jrperritt:
closed by #789

PR #789, opened 2018-02-23
Networking v2: Fix subnetpools

jtopjian:
For #788 

The following fixes have been applied:

* Removal of Prefixes from ListOpts
* Changing ListOpts Shared to *bool
* Changing ListOpts IsDefault to *bool
* Changing the result CreatedAt to time.Time
* Changing the result UpdatedAt to time.Time

/cc @dklyle 

I've tested these fixes pretty extensively to make sure they're correct. 

I'm not a fan of including `ID` and `tenant_id` in `ListOpts`, but since they were already merged and since they _do_ work, I'm going to leave them.

If someone has time to do a review of this, it would help move the Terraform PRs along.


jtopjian:
Thanks, guys!

PR #793, opened 2018-02-27
 Vpnaas: Update IKE policy

simonre:

For #723 

code in this PR:

https://github.com/openstack/neutron-vpnaas/blob/058469e1b99b647537a5228c6a384d93df5484df/neutron_vpnaas/db/vpn/vpn_db.py#L338

### DIFF
--- a/openstack/networking/v2/extensions/vpnaas/ikepolicies/requests.go
+++ b/openstack/networking/v2/extensions/vpnaas/ikepolicies/requests.go
@@ -180,12 +180,14 @@ type LifetimeUpdateOpts struct {
 
 // UpdateOpts contains the values used when updating an IKE policy
 type UpdateOpts struct {
-	Description         *string             `json:"description,omitempty"`
-	Name                *string             `json:"name,omitempty"`
-	AuthAlgorithm       AuthAlgorithm       `json:"auth_algorithm,omitempty"`
-	EncryptionAlgorithm EncryptionAlgorithm `json:"encryption_algorithm,omitempty"`
-	PFS                 PFS                 `json:"pfs,omitempty"`
-	Lifetime            *LifetimeUpdateOpts `json:"lifetime,omitempty"`
+	Description           *string               `json:"description,omitempty"`
+	Name                  *string               `json:"name,omitempty"`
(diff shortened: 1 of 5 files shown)


simonre:

openstack/networking/v2/extensions/vpnaas/ikepolicies/requests.go
```diff
@@ -105,3 +105,42 @@ func Create(c *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResul
...
+	ToPolicyUpdateMap() (map[string]interface{}, error)
+}
+
+type LifetimeUpdateOpts struct {
+	Units Unit `json:"units,omitempty"`
+	Value int  `json:"value,omitempty"`
+}
+
+// UpdateOpts contains the values used when updating an IKE policy
+type UpdateOpts struct {
```
jtopjian: It looks like `Phase1Negotiation` can be updated: https://github.com/openstack/neutron-vpnaas/blob/master/neutron_vpnaas/extensions/vpnaas.py#L418

simonre:
@jtopjian I added the field and I also noticed that you can update the IKE version as well so I added that, too. 

In cases like this where it's just a one or two line change, would you mind if I just rebased without you explicitly asking for it? 
Also: Yes, both this and the "Update IPSec Policy" branch can be merged now. The problem with the "Update service" branch is that `Service` has a field `Status` which has to be a certain value (`ACTIVE`) for openstack to allow an update of its values. This field however does not change from `PENDING_CREATE` to `ACTIVE` until an IPSec Site Connection for the VPN service has been created. IPSec and IKE policies do not have this problem and can be updated immediately upon creation.

jtopjian:
@simonre Rebasing in certain cases (like this) is fine, so you're more than welcome to do it in preparation.

The house rule about not rebasing or amending commits is for someone to make a dedicated commit of the requested changes so the reviewer can distinguish those changes from the original.

jtopjian:
@simonre Looks like this one didn't make this round of merges before requiring another rebase :(

simonre:
@jtopjian I think this is ready now.

Issue #511, opened 2017-09-05
Neutron Ports: Add support for `extra_dhcp_opts`


fatmcgav:
As requested in https://github.com/terraform-providers/terraform-provider-openstack/issues/99, the [Ports](https://github.com/gophercloud/gophercloud/tree/master/openstack/networking/v2/ports) package should support the `extra_dhcp_opts` param. 

Further details on API available here: https://developer.openstack.org/api-ref/networking/v2/index.html?expanded=create-port-detail

jtopjian:
Additional work required: Locate where `extra_dhcp_opts` is defined in Neutron.

Bonus: add `extra_dhcp_opts` to an existing [Port acceptance test](https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/networking/v2/ports_test.go) and verify it works.

jtopjian:
@ozerovandrei If you don't hear back from @bugosiphionah in the next day, then I would say you can proceed with working on this.

ozerovandrei:
Ok. I created some PRs. I've decided to implement it in a separate extension for the networking v2 package.

jtopjian:
Implemented in #804 

@ozerovandrei Thank you for your work with this :smile:

PR #804, opened 2018-03-04
Add networking port extra DHCP opts Update call

ozerovandrei:
Add an Update request to the networking v2 extra DHCP options extension.
Add unit and acceptance tests.

For #511

code in this PR:

https://github.com/openstack/neutron-lib/blob/stable/pike/neutron_lib/api/definitions/extra_dhcp_opt.py#L25
https://github.com/openstack/neutron/blob/stable/pike/neutron/agent/linux/dhcp.py#L954
https://github.com/openstack/neutron/blob/stable/pike/neutron/db/extra_dhcp_opt/models.py#L20
https://github.com/openstack/neutron/blob/stable/pike/neutron/objects/ports.py#L266
https://github.com/openstack/neutron/blob/stable/pike/neutron/objects/ports.py#L320


jtopjian:
@ozerovandrei This all looks good to me.

There's one issue that I didn't catch yesterday: `[]ExtraDHCPOpts` should actually be `[]ExtraDHCPOpt` since each element of the slice represents a single DHCP option.

Rather than you make the change and then have to do a bunch of rebasing across all 3 PRs, I'm going to merge this PR (which contains all of the work) and then follow-up with my own commit to fix this. It's my fault I didn't catch the name and you shouldn't have to pay for it :smile:

PR #808, opened 2018-03-06
Networking v2: Port Security Update

jtopjian:
For #298 

/cc @dklyle


jtopjian:
Ping for review

This is the last of #298. 

Also of note, it looks like OpenLab is running all Port acceptance tests, so a successful test run implies the acceptance test here worked.

I'm slowly making my way to do some acceptance refactoring which will do some asserts on the tests as well as reach out to the OpenLab team to see what needs done to get more tests running.

Issue #206, opened 2016-11-25
Attach security group to VM - EOF error despite successful request handling at backend


pschollbach:
**Source:**
`/compute/v2/extensions/secgroups/requests.go`

**Method:**
`AddServer()`

**Problem:**
Method returns EOF error despite the request being perfectly handled by Openstack backend. The attachment of security group succeeded (verified with OS CLI). Gophercloud should return `nil` instead.

**Analysis:**
The attachment (request) is done according
http://developer.openstack.org/api-ref/compute/?expanded=add-security-group-to-a-server-addsecuritygroup-action-detail
The http response code is `202` but there is absolutely no body returned (empty body). It looks like gophercloud would have expected `{}` as body (according test cases).
This will fail in `/github.com/gophercloud/gophercloud/provider_client.go`:
```code
func (client *ProviderClient) Request(method, url string, options *RequestOpts) (*http.Response, error) {
...
	// Parse the response body as JSON, if requested to do so.
	if options.JSONResponse != nil {
		defer resp.Body.Close()
		if err := json.NewDecoder(resp.Body).Decode(options.JSONResponse); err != nil {

--> seems the json decoder returns EOF error, which is then forwarded to gophercloud consumers
```

**Minor comment:**
Shouldn't OpenStack return `204 No contents` in first place? But then, `defaultOkCodes` for `method == "POST"` would need 204 too.

**Question:**
We can provide a PR if preferred ways of fixing this are indicated. Would some "empty body check" within `provider_client.go` might have side effects for other requests / response scenarios?! Any other ideas? Thanks!

jtopjian:
@pschollbach I'm not sure how this issue ran under the radar for so long - I apologize about that. 

I've seen EOF reported from time to time, but I've never been able to reproduce the issue myself. Do you have a specific scenario that would confidently trigger an EOF? Additionally, since this is an older issue, has the problem resolved on its own (whether by a different patch or a newer version of Go)?

pschollbach:
@jtopjian 
Don't mind the delay because likely you were busy with much higher-prioritized tasks.

Anyway, I have moved on to another company (completely different projects) and therefore have no access to this test environment + code base anymore. I'm afraid we need to rely on my rather "shaky" memories. :-/

I think it was:
```
POST /servers/{server_id}/action 

Adds a security group to a server.

Specify the addSecurityGroup action in the request body.
Normal response codes: 202
```
(from OS API spec.)

Furthemore the API specification states:
```
[...]

PR #815, opened 2018-03-09
Compute v2: Fix EOF errors in compute secgroups

jtopjian:
For #206 

This commit modifies the `AddServer` and `RemoveServer` actions so a response body is not parsed.

I'll admit I haven't done an exhaustive review of the code, but from what I can tell, the lack of a return statement here:

https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/security_groups.py#L441-L456

Is probably good enough.

Note how other functions are returning some object, such as here:

https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/security_groups.py#L340

Also note that the explicit EOF check in the acceptance test was added by me (which I'm removing in this PR). This is because I _swear_ I was seeing intermittent EOFs and not steady ones. But it's been so long since I was working on that, I can't be sure if my testing was correct.

/cc @dklyle

### DIFF
--- a/acceptance/openstack/compute/v2/compute.go
+++ b/acceptance/openstack/compute/v2/compute.go
@@ -681,7 +681,15 @@ func DeleteServer(t *testing.T, client *gophercloud.ServiceClient, server *serve
 		t.Fatalf("Unable to delete server %s: %s", server.ID, err)
 	}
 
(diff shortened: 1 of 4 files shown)


colin-nolan:
I've confirmed that, with our OpenStack system, this fix solves the EOF problem we were encountering.

Issue #818, opened 2018-03-12
Acceptance Test Updates


jtopjian:
This issue is to house a set of acceptance test updates I intend to implement.

This includes:

* Add the ability to view the OpenStack API JSON requests and results when testing
* Perform assertions on the results of resources
* Flag the ability to enable and disable certain tests from running (eg: admin-only tests, service-specific tests)

jtopjian:
Thank you for researching this - it's  good checklist to have. 

I think we can either disregard the following or make them a very low priority because they're either deprecated or an edge-case:

* blockstorage/noauth
* blockstorage/v1
* networking/v2/extensions/lbaas
* networking/v2/extensions/lbaas_v2

There are some that are kind of "heavy" and it might be better to have a dedicated OpenLab test for them (so only the single project will be run):

* containerinfra/v1
* db/v1

Depending on how much longer the default test suite runs with sharedfilesystems enabled, we might also want to have a separate test for that, too (consider that it will be run for _every_ PR).

The others seem like they can be enabled at some point:

* clustering/v1
* messaging/v2
* workflow/v2

I know the clustering and messaging tests don't take a lot of time to run because I've run their tests lots of times locally. The workflow tests might be more of a pilot to start.

We do have some control over what OpenLab runs during testing by editing the various zuul files, so some local experimentation is possible without requiring assistance from OpenLab.

PR #820, opened 2018-03-13
Acc Compute v2: Updating the Compute v2 Acceptance Tests

jtopjian:
For #818

### DIFF
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -43,9 +43,6 @@ type AcceptanceTestChoices struct {
 
 	// DBDatastoreTypeID is the datastore type version for DB tests.
 	DBDatastoreVersion string
-
-	// LiveMigrate indicates ability to run multi-node migration tests
-	LiveMigrate bool
 }
 
 // AcceptanceTestChoicesFromEnv populates a ComputeChoices struct from environment variables.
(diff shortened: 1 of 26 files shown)


jtopjian:
This is ready for review.

There's a lot going on here, but the changes generally fall under the following categories:

1. Addition of `AssertEquals` in various places. There aren't checks for _everything_ but it's a good start and definitely enough to have more confidence in the acceptance testing.

2. Addition of `RequiresX` convenience functions: https://github.com/jtopjian/gophercloud/blob/947904130a9773785edd1b726a727839cd2bf27f/acceptance/clients/conditions.go

    These are pretty handy and add some nice-looking declarations to the tests: https://github.com/jtopjian/gophercloud/blob/947904130a9773785edd1b726a727839cd2bf27f/acceptance/openstack/compute/v2/migrate_test.go#L30

    These functions are key to getting the entire acceptance suite into OpenLab.

3. General cleanup where it was obvious to do so.

4. Removal of a verbose error messages: https://github.com/gophercloud/gophercloud/pull/820/commits/947904130a9773785edd1b726a727839cd2bf27f

    This was intentionally done as the last commit so I could easily nix it if requested. On one hand, the verbose error messages make the acceptance tests useful reference examples. On the other hand, omitting them makes the tests smaller and easier to write (thus more appealing to contributors).

This PR also includes #819 since I found it handy for my own debugging here.

Happy to make any changes as requested.

/cc @dklyle

jtopjian:
Merge conflict has been resolved.

I should have been more clear with my previous comment: I don't expect anyone to review each and every change here. This is a bulk update to the compute acceptance tests and if I was only adding a bunch of `AssertEquals` statements, I'd be comfortable merging them on my own.

But I'd like to get feedback / signoffs on:

1. The checks added to [conditions.go](https://github.com/jtopjian/gophercloud/blob/ccd20d93b2497848bd9424343f1e901af101bd5f/acceptance/clients/conditions.go)
2. The removal of all of the `t.Logf` statements.

If both of those are acceptable, I'm going to go through the rest of the tests and make similar changes in bulk.

If those items shouldn't be added or need changed, I'll make them here and then proceed with the other tests in a similar fashion.

Issue #823, opened 2018-03-13
Add support for Senlin clustering


JackKuei:
Currently there is no support for senlin clustering in gophercloud.

[https://developer.openstack.org/api-ref/clustering/](url)
---
- [ ] Profiles: List
- [ ] Profiles: Create
- [ ] Profiles: Get
- [ ] Profiles: Update
- [ ] Profiles: Delete
- [ ] Profiles: Validate
---
- [ ] PolicyTypes: List
- [ ] PolicyTypes: Get
---
- [ ] Policies: Get
- [ ] Policies: Create
- [ ] Policies: Get
- [ ] Policies: Update
- [ ] Policies: Delete
- [ ] Policies: Validate
---
- [ ] Clusters: List
- [ ] Clusters: Create
- [ ] Clusters: Get
- [ ] Clusters: Update
- [ ] Clusters: Delete
- [ ] Clusters: Resize
- [ ] Clusters: ScaleIn
- [ ] Clusters: ScaleOut
- [ ] Clusters: AddNode
- [ ] Clusters: RemoveNode
- [ ] Clusters: ReplaceNode
- [ ] Clusters: AttachPolicy
- [ ] Clusters: DetachPolicy
- [ ] Clusters: UpdatePolicy
- [ ] Clusters: CollectAttributes
- [ ] Clusters: CheckHealth
- [ ] Clusters: Recover
- [ ] Clusters: Ops
---
- [ ] ClusterPolicies: List
- [ ] ClusterPolicies: Get
---
- [ ] Nodes: List
- [ ] Nodes: Create
- [ ] Nodes: Adopt
- [ ] Nodes: AdoptPreview
- [ ] Nodes: Get
- [ ] Nodes: Update
- [ ] Nodes: Delete
- [ ] Nodes: CheckHealth
- [ ] Nodes: Recover
- [ ] Nodes: Ops
---
- [ ] Receivers: List
- [ ] Receivers: Create
- [ ] Receivers: Get
- [ ] Receivers: Update
- [ ] Receivers: Delete
- [ ] Receivers: Notify
---
- [ ] Events: List
- [ ] Events: Get
---
- [ ] Webhooks: Trigger
---
- [ ] Actions:  List
- [ ] Actions:  Get
---
- [ ] Services:  List

jtopjian:
@JackKuei I've lost track of which PRs are the next in queue. Can you let me know which ones are ready for review?

JackKuei:
[...]

PR #824, opened 2018-03-14
Senlin: Webhook Trigger

JackKuei:
Add webhooks trigger for senlin clustering

code in this PR:

[[webhooks:trigger]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/webhooks.py#L28)

### DIFF
--- a/openstack/clustering/v1/webhooks/doc.go
+++ b/openstack/clustering/v1/webhooks/doc.go
@@ -1,0 +1,11 @@
+/*
+Package webhooks enables triggers an action represented by a webhook from the OpenStack
+Clustering Service.
+
+Example to Trigger webhook action
+	result := webhooks.Trigger(webhookClient, "my-webhook")
+	if result.Err != nil {
+		panic(result.Err)
+	}
+*/
+package webhooks
--- a/openstack/clustering/v1/webhooks/doc.go
+++ b/openstack/clustering/v1/webhooks/doc.go
@@ -1,11 +1,12 @@
 /*
-Package webhooks enables triggers an action represented by a webhook from the OpenStack
-Clustering Service.
+Package webhooks provides the ability to trigger an action represented by a webhook from the OpenStack Clustering
+Service.
 
 Example to Trigger webhook action
-	result := webhooks.Trigger(webhookClient, "my-webhook")
-	if result.Err != nil {
-		panic(result.Err)
+
+	result, err := webhooks.Trigger(webhookClient, "f93f83f6-762b-41b6-b757-80507834d394", nil).Extract()
+	if err != nil {
+		panic(err)
 	}
 */
 package webhooks
--- a/openstack/clustering/v1/webhooks/doc.go
+++ b/openstack/clustering/v1/webhooks/doc.go
@@ -4,11 +4,12 @@ Service.
 
 Example to Trigger webhook action
 
-	result, err := webhooks.Trigger(webhookClient,
-									"f93f83f6-762b-41b6-b757-80507834d394",
-									webhooks.TriggerOpts{V: "1"}).Extract()
+	result, err := webhooks.Trigger(serviceClient(), "f93f83f6-762b-41b6-b757-80507834d394", webhooks.TriggerOpts{V: "1"}).Extract()
 	if err != nil {
 		panic(err)
 	}
+
+	fmt.Println("result", result)
+
 */
 package webhooks
--- a/openstack/clustering/v1/webhooks/doc.go
+++ b/openstack/clustering/v1/webhooks/doc.go
@@ -4,7 +4,9 @@ Service.
 
 Example to Trigger webhook action
 
-	result, err := webhooks.Trigger(webhookClient, "f93f83f6-762b-41b6-b757-80507834d394", nil).Extract()
+	result, err := webhooks.Trigger(webhookClient,
+									"f93f83f6-762b-41b6-b757-80507834d394",
+									webhooks.TriggerOpts{V: "1"}).Extract()
 	if err != nil {
 		panic(err)
 	}
--- a/openstack/clustering/v1/webhooks/requests.go
+++ b/openstack/clustering/v1/webhooks/requests.go
@@ -1,0 +1,43 @@
+package webhooks
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// TriggerOpts Webhooks request parameters
+type TriggerOpts struct {
+	// V (Optional)	query	string	The webhook implementation version requested.
+	V string `q:"V, omitempty"`
+
+	// params (Optional)	query	object	The query string that forms the inputs to use for the targeted action.
+	Params string `q:"params, omitempty"`
+}
+
+// TriggerOptsBuilder Query string builder interface for webhooks
+type TriggerOptsBuilder interface {
+	ToTriggerQuery() (string, error)
+}
+
+// Query string builder for webhooks
+func (opts TriggerOpts) ToTriggerQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
+// Trigger an action represented by a webhook.
+func Trigger(client *gophercloud.ServiceClient, id string, opts TriggerOptsBuilder) (r TriggerResult) {
+	url := triggerURL(client, id)
+	if opts != nil {
+		query, err := opts.ToTriggerQuery()
+		if err != nil {
+			r.Err = err
+			return
+		}
+		url += query
+	}
+
+	_, r.Err = client.Post(url, nil, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 201, 202},
+	})
+	return
+}
--- a/openstack/clustering/v1/webhooks/requests.go
+++ b/openstack/clustering/v1/webhooks/requests.go
@@ -1,23 +1,33 @@
 package webhooks
 
 import (
+	"net/url"
+
 	"github.com/gophercloud/gophercloud"
+	"golang.org/x/crypto/openpgp/errors"
 )
 
 // TriggerOpts represents options used for triggering an action
 type TriggerOpts struct {
-	V      string `q:"V"`
-	Params string `q:"params, omitempty"`
+	V      string `q:"V,required"`
+	Params map[string]string
 }
 
 // TriggerOptsBuilder Query string builder interface for webhooks
 type TriggerOptsBuilder interface {
-	ToTriggerQuery() (string, error)
+	ToWebhookTriggerQuery() (string, error)
 }
 
 // Query string builder for webhooks
-func (opts TriggerOpts) ToTriggerQuery() (string, error) {
+func (opts TriggerOpts) ToWebhookTriggerQuery() (string, error) {
 	q, err := gophercloud.BuildQueryString(opts)
+	params := q.Query()
+
+	for k, v := range opts.Params {
+		params.Add(k, v)
+	}
+
+	q = &url.URL{RawQuery: params.Encode()}
 	return q.String(), err
 }
 
@@ -25,12 +35,15 @@ func (opts TriggerOpts) ToTriggerQuery() (string, error) {
 func Trigger(client *gophercloud.ServiceClient, id string, opts TriggerOptsBuilder) (r TriggerResult) {
 	url := triggerURL(client, id)
 	if opts != nil {
-		query, err := opts.ToTriggerQuery()
+		query, err := opts.ToWebhookTriggerQuery()
 		if err != nil {
 			r.Err = err
 			return
 		}
 		url += query
+	} else {
+		r.Err = errors.InvalidArgumentError("Must contain V for TriggerOpt")
+		return
 	}
 
 	_, r.Err = client.Post(url, nil, &r.Body, &gophercloud.RequestOpts{
--- a/openstack/clustering/v1/webhooks/requests.go
+++ b/openstack/clustering/v1/webhooks/requests.go
@@ -4,12 +4,9 @@ import (
 	"github.com/gophercloud/gophercloud"
 )
 
-// TriggerOpts Webhooks request parameters
+// TriggerOpts represents options used for triggering an action
 type TriggerOpts struct {
-	// V (Optional)	query	string	The webhook implementation version requested.
-	V string `q:"V, omitempty"`
-
-	// params (Optional)	query	object	The query string that forms the inputs to use for the targeted action.
+	V      string `q:"V"`
 	Params string `q:"params, omitempty"`
 }
 
--- a/openstack/clustering/v1/webhooks/results.go
+++ b/openstack/clustering/v1/webhooks/results.go
@@ -8,24 +8,15 @@ type commonResult struct {
 	gophercloud.Result
 }
 
-// GetResult is the response of a Get operations. Call its Extract method to
-// interpret it as a Version.
-type GetResult struct {
-	commonResult
-}
-
 type TriggerResult struct {
 	commonResult
 }
 
-// Webhook represents a detailed webhook
-type Webhook struct {
-	Action string `json:"action"`
-}
-
 // Extract retrieves the response action
 func (r commonResult) Extract() (string, error) {
-	var s *Webhook
+	var s struct {
+		Action string `json:"action"`
+	}
 	err := r.ExtractInto(&s)
 	return s.Action, err
 }
--- a/openstack/clustering/v1/webhooks/results.go
+++ b/openstack/clustering/v1/webhooks/results.go
@@ -1,0 +1,31 @@
+package webhooks
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// GetResult is the response of a Get operations. Call its Extract method to
+// interpret it as a Version.
+type GetResult struct {
+	commonResult
+}
+
+type TriggerResult struct {
+	commonResult
+}
+
+// Webhook represents a detailed webhook
+type Webhook struct {
+	Action string `json:"action"`
+}
+
+// Extract retrieves the response action
+func (r commonResult) Extract() (string, error) {
+	var s *Webhook
+	err := r.ExtractInto(&s)
+	return s.Action, err
+}
--- a/openstack/clustering/v1/webhooks/testing/requests_test.go
+++ b/openstack/clustering/v1/webhooks/testing/requests_test.go
@@ -29,25 +29,29 @@ func TestWebhookTrigger(t *testing.T) {
 			}`)
 	})
 
-	result, err := webhooks.Trigger(fake.ServiceClient(),
-		"f93f83f6-762b-41b6-b757-80507834d394",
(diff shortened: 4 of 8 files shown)


JackKuei:
Requesting code review.

openstack/clustering/v1/webhooks/doc.go
```diff
@@ -0,0 +1,11 @@
+/*
+Package webhooks enables triggers an action represented by a webhook from the OpenStack
+Clustering Service.
+
+Example to Trigger webhook action
+	result := webhooks.Trigger(webhookClient, "my-webhook")
+	if result.Err != nil {
+		panic(result.Err)
+	}
```
jtopjian: Since you're defining `TriggerOpts` in the code, wouldn't the above example look more like:

```go
result := webhooks.Trigger(clusterClient, "webhook-uuid", nil)
```

Additionally, `result.Err` is correct, but I think it would be preferable to use the more common form:

```go
result, err := webhooks.Trigger(clusterClient, "webhook-uuid", nil).Extract()
if err != nil {
  panic(err)
}
```

openstack/clustering/v1/webhooks/requests.go
```diff
@@ -0,0 +1,43 @@
+package webhooks
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// TriggerOpts Webhooks request parameters
+type TriggerOpts struct {
+	// V (Optional)	query	string	The webhook implementation version requested.
+	V string `q:"V, omitempty"`
```
jtopjian: While the API documentation shows `V` as a valid query parameter, I don't see it defined in the actual Python code:

https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/webhooks.py#L28

Do you know if V might have been an old parameter?

openstack/clustering/v1/webhooks/requests.go
```diff
@@ -0,0 +1,43 @@
...
+
+	// params (Optional)	query	object	The query string that forms the inputs to use for the targeted action.
```
jtopjian: The above comment looks like it was copied and pasted straight from the API docs (with odd spacing included). Inline documentation is turned into official reference documentation, so we want to make sure it's readable. For the above comment, how about something like:

```
Params defines the input to use for the targeted action.
```

openstack/clustering/v1/webhooks/requests.go
```diff
@@ -0,0 +1,43 @@
...
+	Params string `q:"params, omitempty"`
```
jtopjian: The `q` tag is internal to Gophercloud and doesn't have an `omitempty` modifier. However, I believe `params` is actually in the body of the request as seen here:

https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/webhooks.py#L29-L30

So I believe this should be:

```go
Params string `json:"params,omitempty"`
```

But still, I don't think `string` is the correct data type here. Looking at the unit tests for webhooks (https://github.com/openstack/senlin/blob/9ce18436035fade18a29e9b47f9cf7389c55da81/senlin/tests/unit/objects/requests/test_webhooks.py), it looks as though `params` is a `map[string]string`.

So perhaps:

```go
Params map[string]string `json:"params,omitempty"`
```

openstack/clustering/v1/webhooks/requests.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+// TriggerOptsBuilder Query string builder interface for webhooks
+type TriggerOptsBuilder interface {
+	ToTriggerQuery() (string, error)
```
jtopjian: `Query` is only for actions that utilize a URL query string. Since we're doing a POST, the Gophercloud convention is to use `Map` in the name. I think a better name for this function is:

```go
ToWebhookTriggerMap() (map[string]interface, err)
```

openstack/clustering/v1/webhooks/results.go
```diff
@@ -0,0 +1,31 @@
...
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// GetResult is the response of a Get operations. Call its Extract method to
+// interpret it as a Version.
+type GetResult struct {
+	commonResult
+}
```
jtopjian: I think we can safely remove `GetResult` here since this PR is only defining a `Trigger` action.

openstack/clustering/v1/webhooks/results.go
```diff
@@ -0,0 +1,31 @@
...
+
+type TriggerResult struct {
+	commonResult
+}
+
+// Webhook represents a detailed webhook
+type Webhook struct {
+	Action string `json:"action"`
+}
```
jtopjian: The result of the Trigger action is defined here:

https://github.com/openstack/senlin/blob/c0ef921768046a8795aa1f231ec6fafd5e67f3ea/senlin/engine/service.py#L2518

```
return {'action': action_id}
```

Since "webhook" isn't in the above body, I don't think we should call this struct `Webhook`. Since the result of this action is only returning one field, we should be able to get away with removing this struct entirely and changing the `Extract` method to look like:

```go
func (r commonResult) Extract() (string, err) {
  var s struct {
    Action string `json:"action"`
  }
  err := r.ExtractInto(&s)
  return s.Action, err
}
```

JackKuei:
@jtopjian Thanks for the review. Will start going thru and address area of concerns. We have implemented and currently testing autoscaling in house using senlin and zaqar.

jtopjian:
> We have implemented and currently testing autoscaling in house using senlin and zaqar.

Sounds good. It's good to know you've got a working implementation 😄 I think the bulk of the changes that are going to happen are:

1. Restructuring to be consistent with other area of Gophercloud

2. Identifying ares that might not be utilized in your environment and ensuring they work. For example, it's quite possible `params` and `V` aren't being used in any actions. We'll want to make sure all of Senlin is in a fully functional state.

openstack/clustering/v1/webhooks/doc.go
```diff
@@ -0,0 +1,11 @@
+/*
+Package webhooks enables triggers an action represented by a webhook from the OpenStack
+Clustering Service.
```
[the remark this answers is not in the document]
JackKuei: Sounds good

openstack/clustering/v1/webhooks/doc.go
```diff
@@ -0,0 +1,11 @@
...
+
+Example to Trigger webhook action
```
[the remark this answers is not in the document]
JackKuei: yes

openstack/clustering/v1/webhooks/requests.go
```diff
@@ -0,0 +1,43 @@
+package webhooks
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// TriggerOpts Webhooks request parameters
```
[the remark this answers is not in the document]
JackKuei: Sounds good

openstack/clustering/v1/webhooks/doc.go
JackKuei: Agree, changed to the second form

openstack/clustering/v1/webhooks/requests.go
JackKuei: Removed the comment. Found out that having comment and blank lines prevents go tool to properly format

openstack/clustering/v1/webhooks/results.go
JackKuei: Removed

JackKuei:
Yeah...definitely. Currently it is working for our use cases only would definitely need to polish up to accommodate other use cases and also a good doc. :)

openstack/clustering/v1/webhooks/results.go
JackKuei: Yes, your suggestion is better and cleaner

openstack/clustering/v1/webhooks/requests.go
JackKuei: Need to look into this one. Doc specified query string.

openstack/clustering/v1/webhooks/requests.go
JackKuei: Need to look into this one. Doc specified query string.

openstack/clustering/v1/webhooks/requests.go
JackKuei: Contacted senlin forum.

JackKuei:
@jtopjian For the Acceptance Test, is it better to provide at the end for AutoScaling as oppose to Acceptance Test for each PR since will exercise multiple PR?

jtopjian:
@JackKuei No, it's better to provide an acceptance test with each PR in order to validate that the code works.

There are cases where functionality depends on a previous feature. For example, if someone submitted the ability to Update a resource, it would be tough to test before Create was supported. In that case, Create should be submitted and merged first.

If this is the situation with the Trigger PR, then we could/should wait to merge this until it can be leveraged within Gophercloud by way of other Senlin/Clustering resources.

If the test will leave "hanging" resources after the test (for example, creating a resource and not being able to delete it), that's also acceptable. We would rather see half-implemented tests than no tests at all. In addition, the test environments are all disposable so leftover/hanging resources are fine.

Let me know if that helps clarify things.

openstack/clustering/v1/webhooks/requests.go
jtopjian: Yes! Absolutely code/implementation over docs :smile: 

API docs are a good second reference but we can't use them as the primary reference since there have been numerous times when the docs did not match up with what was really happening in the code. (I've mentioned elsewhere that this is not criticism against people who write docs - having docs go out of sync with the code happens all the time)

I've also mentioned elsewhere that this makes implementing features in Gophercloud difficult -- there's no denying that one. I'm happy to help with this area whenever I can.

We use the service-side/api code as the primary reference but if that proves overly difficult, then the Python client code can also be used (in this case, `python-senlinclient`). As a last resort, running commands on the command-line with `--debug` set can also be helpful.

jtopjian:
@JackKuei I found `V`!

https://github.com/openstack/senlin/blob/master/senlin/api/middleware/webhook.py#L86-L87

But it seems totally nonsensical. The code is immediately `pop`'ing it. `V` doesn't seem to be used anywhere at all.

As well, it looks like `params` is indeed part of the request URL. It looks like API interaction with webhook triggers is done entirely through this middleware. All other webhook interaction is done internally. So the middleware will take the URL parameters and convert it to an actual JSON body.

So it looks like the docs are correct - just missed the fact that the implementation is a middleware piece.

With that said, though, `params` should be something other than a `string`. Having the user specify something like:

```go
x := TriggerOpts{
  Params: "foo=bar&bar=baz",
}
```

Isn't the most friendly thing. Making it a `map[string]string` might clash with something we did for Cinder. Let me think about the best way to handle this and I'll post some sample code.

openstack/clustering/v1/webhooks/requests.go
jtopjian: Per discovery of where triggers are actually processed, this is definitely a Query, but the name of the function still needs 4 parts to conform with the rest of Gophercloud. Let's go with `ToWebhookTriggerQuery`

jtopjian:
@JackKuei I think the following should work:

```go
type TriggerOpts struct {
  V      string `q:"V,required"`
  Params map[string]string
}

...

func (opts TriggerOpts) ToWebhookTriggerQuery() (string, error) {
  params := q.Query()

  for k, v := range opts.Params {
    params.Add(k, v)
  }

  q = &url.URL{RawQuery: params.Encode()}
  return q.String(), err
}
```

And the following unit test should ensure it works:

```go
func TestWebhookParams(t *testing.T) {
  to := webhooks.TriggerOpts{
    V: "1",
    Params: map[string]string{
      "foo": "bar",
      "bar": "baz",
    },
  }

  expected := "?V=1&bar=baz&foo=bar"
  actual, err := to.ToWebhookTriggerQuery()
  th.AssertNoErr(t, err)
  th.AssertEquals(t, actual, expected)
}
```

Thoughts?

animationzl:
recheck

JackKuei:
Spoke to the senlin folks. Here is the link. https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/middleware/webhook.py#L69. Here is their response "We intentionally asked for a 'v=1' query string to make the protocol extensible in future"

JackKuei:
from the code think V is optional. https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/middleware/webhook.py#L86

jtopjian:
> from the code think V is optional. 

I don't think it is. The next line of code issues a `return None` if `V` doesn't exist.

JackKuei:
Hmm....currently we don't supply V and is functional. Let me double check.

JackKuei:
Here is the wireshark trace. Talking to senlin devs to see if the senlin code appends those query strings.
POST /v1/webhooks/157bf6ad/trigger HTTP/1.1
Host: dev.senlin.cloud.blizzard.net:8778
User-Agent: gophercloud/2.0.0
Connection: close
Content-Length: 0
Accept: application/json
X-Auth-Token: gAAAAABau8rr11DxUgcUNVx_fj2ss_Za_FnCGqZh8sUWHY97Arq1KOZup57-tHy24TySa6PhbIiVg0aRUj4cPGWz4ubqQBBDuvdgT9tWdripkazP2i2tMCMy4IzbaSROlfwetXtKlFETJxVakByr604cyMT8pk0pdFSPn4k8iAlI6zWk1qfEbKc2JsMWUT6naECnuKtT3L-ToUdzL_lAvqijKloSZA_U5OwUyu5DzAPF7mcUEa1AQpY
Accept-Encoding: gzip

HTTP/1.1 202 Accepted
Location: http://dev.senlin.cloud.blizzard.net:8778/v1/actions/40ae502a-0889-4eb3-98ee-3142fa51877b
Openstack-Api-Version: clustering 1.0
Vary: OpenStack-API-Version
Content-Length: 50
X-Openstack-Request-Id: req-f442fa4d-088a-4815-93dc-2c012b8f92bb
Date: Wed, 28 Mar 2018 17:06:17 GMT
Connection: close

{"action": "40ae502a-0889-4eb3-98ee-3142fa51877b"}

jtopjian:
@JackKuei That looks good. If omitting `V` is working for you, then it's safe to consider it optional. I'm just misunderstanding the code, then.

JackKuei:
actually i'm confused as well that's why I'm posting a question for senlin devs since it shouldn't have worked. Awaiting for their response.

JackKuei:
For the acceptance test, since webhook trigger requires a receiver to be created, should i just create a code stub with actual function calls commented out until receiver code is PRed?

jtopjian:
Are there plans on adding support for receivers later? If so, I am OK postponing an acceptance test until then. But as mentioned above, it's better if PRs are merged in an order which will allow acceptance tests to be created rather than waiting on other resources.

JackKuei:
yes at the minimum will support autoscaling. but the goal is to have at least 80% of senlin covered and shooting for full coverage

JackKuei:
ok, will add the acceptance test with unsupported function calls commented out

jtopjian:
@JackKuei If you're implementing changes, don't forget about this: https://github.com/gophercloud/gophercloud/pull/824#issuecomment-376771718

If `V` isn't required, then just remove the `required` tag. But the Params field should be modified accordingly.

JackKuei:
yes...working on it now. Just heard from the senlin devs, V is required. They will update the code as well as the doc

JackKuei:
fyi V is required: https://review.openstack.org/#/c/557798/

jtopjian:
@JackKuei Looks like another `gofmt` error here, too.

This is looking good! Let me know when it's ready for another review.

JackKuei:
Miss understood the error earlier. Will make a fix.

jtopjian:
@JackKuei Per discussion here (https://github.com/gophercloud/gophercloud/pull/860#issuecomment-377455699) we've opted to merge #860 first. 

I can still review this at your request, but there will be some merge conflicts that need resolved.

JackKuei:
Fixed merge

openstack/clustering/v1/webhooks/doc.go
```diff
@@ -0,0 +1,14 @@
+/*
+Package webhooks provides the ability to trigger an action represented by a webhook from the OpenStack Clustering
+Service.
+
+Example to Trigger webhook action
+
+	result, err := webhooks.Trigger(webhookClient,
+									"f93f83f6-762b-41b6-b757-80507834d394",
+									webhooks.TriggerOpts{V: "1"}).Extract()
```
jtopjian: Looks like this line needs fixed up

openstack/clustering/v1/webhooks/testing/requests_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+		fmt.Fprintf(w, `
+			{
+				"action": "290c44fa-c60f-4d75-a0eb-87433ba982a3"
+			}`)
+	})
+
+	result, err := webhooks.Trigger(fake.ServiceClient(),
+		"f93f83f6-762b-41b6-b757-80507834d394",
+		webhooks.TriggerOpts{V: "1", Params: map[string]string{"foo": "bar", "bar": "baz"}}).Extract()
```
jtopjian: nit: can this line be cleaned up?

```go
webhookID := "..."
triggerOpts := webhooks.TriggerOpts{
  V: 1,
  ...,
}
result, err := webhooks.Trigger(fake.ServiceClient(), webhookID, triggerOpts).Extract()
```

openstack/clustering/v1/webhooks/testing/requests_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+		fmt.Fprintf(w, `
+			{
+				"action": 123
+			}`)
+	})
+
+	_, err := webhooks.Trigger(fake.ServiceClient(),
+		"f93f83f6-762b-41b6-b757-80507834d394",
+		webhooks.TriggerOpts{V: "1"}).Extract()
```
jtopjian: nit: can this line be cleaned up as well?

openstack/clustering/v1/webhooks/testing/requests_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+		fmt.Fprintf(w, `
+			{
+				"action": "290c44fa-c60f-4d75-a0eb-87433ba982a3"
+			}`)
+	})
+
+	_, err := webhooks.Trigger(fake.ServiceClient(),
+		"f93f83f6-762b-41b6-b757-80507834d394",
+		webhooks.TriggerOpts{}).Extract()
```
jtopjian: nit: can this line be cleaned up as well?

JackKuei:
Fixed formatting for triggerOpts

jtopjian:
@JackKuei I apologize for the delay on this -- I missed your comment :(

This looks good to me. Thank you for your work and patience with this one

JackKuei:
Thank you for being patient and your guidance!

Issue #827, opened 2018-03-14
add 'OS-SCH-HNT:scheduler_hints' to volume CreateOpts


zengchen1024:
'OS-SCH-HNT:scheduler_hints' is an optional field of volume creating API[1][2], but it is not supported in current volume CreateOpts[3][4].

[1] https://developer.openstack.org/api-ref/block-storage/v2/#volumes-volumes
[2] https://developer.openstack.org/api-ref/block-storage/v3/#volumes-volumes
[3] https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v2/volumes/requests.go#L17-L41
[4] https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v3/volumes/requests.go#L17-L41

jtopjian:
@zengchen1024 This would be a great feature to add.

It looks like scheduler hints is an API extension:

https://github.com/openstack/cinder/blob/a15459a27777e98599d70392191b0a6f0288ff34/cinder/api/contrib/scheduler_hints.py

Unless I'm mistaken, this piece is extending the volume request:

https://github.com/openstack/cinder/blob/a15459a27777e98599d70392191b0a6f0288ff34/cinder/api/contrib/scheduler_hints.py#L38-L44

So the implementation in Gophercloud should go here: https://github.com/gophercloud/gophercloud/tree/master/openstack/blockstorage/extensions

The good news is that both v2 and v3 will be able to use this extension, so there's no need to add a v2 and v3 specific scheduler hints implementation.

I think the Nova/Compute scheduler hint extension is a good example to go off of for implementing this:

https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/schedulerhints/requests.go

Let me know if you have any questions or need any help. :)

zengchen1024:
@jtopjian Thanks for your reviews. Your comments are very useful. I will update the pr. Thanks again!

jimmyxian:
@jtopjian @zengchen1024 I have submit a PR to solve this issue. Can you help review again. Thanks

zhaoshan86:
I need this feature too，thanks for your PR @jimmyxian ，and please review the RP @jtopjian . Thank you!

PR #828, opened 2018-03-14
add 'OS-SCH-HNT:scheduler_hints' to volume CreateOpts

zengchen1024:
For #827 

The relevant codes in Openstack/Cinder
https://github.com/openstack/cinder/blob/master/cinder/scheduler/filters/affinity_filter.py#L39-L67

https://github.com/openstack/cinder/blob/master/cinder/scheduler/filters/affinity_filter.py#L75-L101

https://github.com/openstack/cinder/blob/master/cinder/scheduler/filters/instance_locality_filter.py#L76-L120


jtopjian:
Support for this was added in #1241

Issue #830, opened 2018-03-14
Add cascade option to loadbalancers.Delete


jrperritt:
Octavia load balancers have the option of "cascading" a delete, which deletes all a load balancer's children (listener, monitor, etc) as well. The option is not available for Neutron.

jrperritt:
closed by #831

PR #831, opened 2018-03-14
add cascade option to loadbalancers.Delete

jrperritt:
For #830 

code in this PR:

https://github.com/openstack/octavia/blob/2d206289d005737d87f7712ba443a345d5547b18/octavia/api/v2/controllers/load_balancer.py#L451

### DIFF
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -482,3 +482,22 @@ func configureDebug(client *gophercloud.ProviderClient) *gophercloud.ProviderCli
 
 	return client
 }
+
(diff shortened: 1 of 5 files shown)


jtopjian:
side comment: It's my understanding that both Neutron LBaaS and Octavia implement the LBaaS v2 API, but it seems like the two are diverging. It might be smart for us to be mindful that there might come a time when octavia and `lbaas_v2` should be two separate packages.

jrperritt:
I agree. It's my understanding that Octavia is _the_ load balancer service going forward, and all load balancer operations in Neutron are being/have been deprecated.

Issue #832, opened 2018-03-15
Octavia: support L7 policy/rule operations


lingxiankong:
Please refer to: https://docs.openstack.org/octavia/latest/user/guides/l7.html

---
- [x] l7 policy create
- [x] l7 policy list
- [x] l7 policy get
- [x] l7 policy update
- [x] l7 policy delete

---
- [x] l7 rule create
- [x] l7 rule list
- [x] l7 rule get
- [x] l7 rule update
- [x] l7 rule delete

jtopjian:
@lingxiankong Thanks for noting this :)

I see the L7 stuff defined here:

https://github.com/openstack/octavia/blob/master/octavia/api/v2/controllers/l7rule.py
https://github.com/openstack/octavia/blob/master/octavia/api/v2/controllers/l7policy.py

But I also see it in the Neutron LBaaS code:

https://github.com/openstack/neutron-lbaas/search?q=l7

Does this mean that L7 is supported in both Octavia and the "deprecated" Neutron LBaaS? Or does Octavia use the `neutron-lbaas` library? Put another way: Does one need to have Octavia deployed to use L7?

The reason I ask is because, given my comment earlier today (https://github.com/gophercloud/gophercloud/pull/831#issuecomment-373191686), I'm wondering if:

1. Similar to LBs, pools, monitors, etc, L7 is supported by both Neutron LBaaS and Octavia. In which case this implementation can go under the current `lbaas_v2` package with no special conditions/checks.
2. We should add it to `lbaas_v2` but with the same kind of service check implemented in #831.
3. Copy/paste `lbaas_v2` to something like `lbaas_v2_octavia` or `lbaas_v2_neutron` or similar and add the L7 stuff to whichever is deemed the current LBaaS v2 API implementation.

I apologize if I'm making this more difficult than it really is.

lingxiankong:
hi, @jtopjian 

AFAIK, the octavia l7 functionality is totally back-compatible with neutron-lbaas, octavia just has more features than the deprecated neutron-lbaas. I was a developer in octavia, btw.

lingxiankong:
@jtopjian if you don't mind, i can hand up to help with this issue.

jtopjian:
> if you don't mind, i can hand up to help with this issue.

We'd love to get your help on this, whether answering questions, implementing the code, or both 😄 

> AFAIK, the octavia l7 functionality is totally back-compatible with neutron-lbaas, 

[...]

PR #834, opened 2018-03-16
LBaaS v2 l7 policy support [Part 1]: create l7policy

lingxiankong:
For #832

L7policy functionality in Octavia is backward compatible with
Neutron-LBaaS.

Octavia L7 policy create API:
https://developer.openstack.org/api-ref/load-balancer/v2/index.html#create-an-l7-policy

### DIFF
--- a/acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
+++ b/acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
@@ -101,7 +101,7 @@ func TestLoadbalancersCRUD(t *testing.T) {
 	tools.PrintResource(t, newListener)
 
 	// L7 policy
-	_, err := CreateL7Policy(t, client, listener, lb)
+	_, err = CreateL7Policy(t, lbClient, listener, lb)
 	if err != nil {
 		t.Fatalf("Unable to create l7 policy: %v", err)
 	}
--- a/acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
+++ b/acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
@@ -100,6 +100,12 @@ func TestLoadbalancersCRUD(t *testing.T) {
 
 	tools.PrintResource(t, newListener)
 
+	// L7 policy
+	policy, err := CreateL7Policy(t, client, listener, lb)
+	if err != nil {
+		t.Fatalf("Unable to create l7 policy: %v", err)
+	}
+
 	// Pool
 	pool, err := CreatePool(t, lbClient, lb)
 	if err != nil {
--- a/acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
+++ b/acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
@@ -101,7 +101,7 @@ func TestLoadbalancersCRUD(t *testing.T) {
 	tools.PrintResource(t, newListener)
 
 	// L7 policy
-	policy, err := CreateL7Policy(t, client, listener, lb)
+	_, err := CreateL7Policy(t, client, listener, lb)
 	if err != nil {
 		t.Fatalf("Unable to create l7 policy: %v", err)
 	}
--- a/openstack/loadbalancer/v2/l7policies/results.go
+++ b/openstack/loadbalancer/v2/l7policies/results.go
@@ -17,7 +17,7 @@ type L7Policy struct {
 	ListenerID string `json:"listener_id"`
 
 	// The L7 policy action. One of REDIRECT_TO_POOL, REDIRECT_TO_URL, or REJECT.
-	Action Action `json:"action"`
+	Action string `json:"action"`
 
 	// The position of this policy on the listener.
 	Position int32 `json:"position"`
@@ -50,10 +50,10 @@ type Rule struct {
 	ID string `json:"id"`
 
 	// The L7 rule type. One of COOKIE, FILE_TYPE, HEADER, HOST_NAME, or PATH.
-	RuleType RuleType `json:"type"`
+	RuleType string `json:"type"`
 
 	// The comparison type for the L7 rule. One of CONTAINS, ENDS_WITH, EQUAL_TO, REGEX, or STARTS_WITH.
-	CompareType CompareType `json:"compare_type"`
+	CompareType string `json:"compare_type"`
 
 	// The value to use for the comparison. For example, the file type to compare.
 	Value string `json:"value"`
--- a/openstack/loadbalancer/v2/l7policies/results.go
+++ b/openstack/loadbalancer/v2/l7policies/results.go
@@ -1,0 +1,93 @@
+package l7policies
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// L7Policy is a collection of L7 rules associated with a Listener, and which
+// may also have an association to a back-end pool.
+type L7Policy struct {
+	// The unique ID for the L7 policy.
+	ID string `json:"id"`
+
+	// Name of the L7 policy.
+	Name string `json:"name"`
+
+	// The ID of the listener.
+	ListenerID string `json:"listener_id"`
+
+	// The L7 policy action. One of REDIRECT_TO_POOL, REDIRECT_TO_URL, or REJECT.
+	Action Action `json:"action"`
+
+	// The position of this policy on the listener.
+	Position int32 `json:"position"`
+
+	// A human-readable description for the resource.
+	Description string `json:"description"`
+
+	// TenantID is the UUID of the project who owns the L7 policy in neutron-lbaas.
+	// Only administrative users can specify a project UUID other than their own.
+	TenantID string `json:"tenant_id"`
+
+	// Requests matching this policy will be redirected to the pool with this ID.
+	// Only valid if action is REDIRECT_TO_POOL.
+	RedirectPoolID string `json:"redirect_pool_id"`
+
+	// Requests matching this policy will be redirected to this URL.
+	// Only valid if action is REDIRECT_TO_URL.
+	RedirectURL string `json:"redirect_url"`
+
+	// The administrative state of the L7 policy, which is up (true) or down (false).
+	AdminStateUp bool `json:"admin_state_up"`
+
+	// Rules are List of associated L7 rule IDs.
+	Rules []Rule `json:"rules"`
+}
+
+// Rule represents layer 7 load balancing rule.
+type Rule struct {
+	// The unique ID for the L7 rule.
+	ID string `json:"id"`
+
+	// The L7 rule type. One of COOKIE, FILE_TYPE, HEADER, HOST_NAME, or PATH.
+	RuleType RuleType `json:"type"`
+
+	// The comparison type for the L7 rule. One of CONTAINS, ENDS_WITH, EQUAL_TO, REGEX, or STARTS_WITH.
+	CompareType CompareType `json:"compare_type"`
+
+	// The value to use for the comparison. For example, the file type to compare.
+	Value string `json:"value"`
+
+	// TenantID is the UUID of the project who owns the rule in neutron-lbaas.
+	// Only administrative users can specify a project UUID other than their own.
+	TenantID string `json:"tenant_id"`
+
+	// The key to use for the comparison. For example, the name of the cookie to evaluate.
+	Key string `json:"key"`
+
+	// When true the logic of the rule is inverted. For example, with invert true,
+	// equal to would become not equal to. Default is false.
+	Invert bool `json:"invert"`
+
+	// The administrative state of the L7 rule, which is up (true) or down (false).
+	AdminStateUp bool `json:"admin_state_up"`
+}
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// Extract is a function that accepts a result and extracts a l7policy.
+func (r commonResult) Extract() (*L7Policy, error) {
+	var s struct {
+		L7Policy *L7Policy `json:"l7policy"`
+	}
+	err := r.ExtractInto(&s)
+	return s.L7Policy, err
+}
+
+// CreateResult represents the result of a Create operation. Call its Extract
+// method to interpret the result as a L7Policy.
+type CreateResult struct {
+	commonResult
+}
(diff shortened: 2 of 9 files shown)


openstack/networking/v2/extensions/lbaas_v2/l7policies/requests.go
```diff
@@ -0,0 +1,84 @@
...
+	// Only administrative users can specify a project UUID other than their own.
+	TenantID string `json:"tenant_id,omitempty"`
+
+	// ProjectID is the UUID of the project who owns the L7 policy in octavia.
+	// Only administrative users can specify a project UUID other than their own.
+	ProjectID string `json:"project_id,omitempty"`
+
+	// Requests matching this policy will be redirected to the pool with this ID.
+	// Only valid if action is REDIRECT_TO_POOL.
+	RedirectPoolID string `json:"redirect_pool_id,omitempty" xor:"RedirectURL"`
```
jtopjian: Just to confirm: `xor` will require _one_ of the two fields to be set. Is it possible to create an L7 policy without setting either of them? If so, then we need to remove the `xor`. Is it possible that the "reject" action will allow both to be unspecified (https://github.com/openstack/neutron-lbaas/blob/e4aba85427f696a6224b8ed6392e0ba16d6f98cf/neutron_lbaas/drivers/octavia/driver.py#L542) ?

These URLs might also be useful for how the Neutron logic works.

* https://github.com/openstack/neutron-lbaas/blob/eb5877fb004b06c912aa221225ea261773a48aa9/neutron_lbaas/db/loadbalancer/loadbalancer_dbv2.py#L429
* https://github.com/openstack/neutron-lbaas/blob/e4aba85427f696a6224b8ed6392e0ba16d6f98cf/neutron_lbaas/drivers/octavia/driver.py#L533

openstack/networking/v2/extensions/lbaas_v2/l7policies/testing/fixtures.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/l7policies"
+	th "github.com/gophercloud/gophercloud/testhelper"
+	"github.com/gophercloud/gophercloud/testhelper/client"
+)
+
+// SingleL7PolicyBody is the canned body of a Get request on an existing l7policy.
+const SingleL7PolicyBody = `
+{
+	"l7policy": {
```
jtopjian: It looks like this test fixture was copied from the API reference:

https://developer.openstack.org/api-ref/load-balancer/v2/index.html#create-an-l7-policy

We prefer that all test fixtures are generated in a working OpenStack environment. We've found too many cases where the API docs' fixtures do not match what OpenStack/LBaaS really responds with.

jtopjian:
> don't be surprised that acceptance test is not included, i will add the the following PR

Sorry - I read the wrong comment in my inbox and didn't see this. I added a note about this in my review. It's totally fine to include partially working acceptance tests. We would prefer to have partial tests rather than no tests.

openstack/networking/v2/extensions/lbaas_v2/l7policies/requests.go
```diff
@@ -0,0 +1,84 @@
...
+	Name string `json:"name,omitempty"`
+
+	// The ID of the listener.
+	ListenerID string `json:"listener_id" required:"true"`
+
+	// The L7 policy action. One of REDIRECT_TO_POOL, REDIRECT_TO_URL, or REJECT.
+	Action Action `json:"action" required:"true"`
+
+	// The position of this policy on the listener.
+	Position int32 `json:"position,omitempty"`
```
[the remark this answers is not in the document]
lingxiankong: i changed that in my local branch but forgot to submit for some reason

openstack/networking/v2/extensions/lbaas_v2/l7policies/results.go
```diff
@@ -0,0 +1,108 @@
...
+	// The value to use for the comparison. For example, the file type to compare.
+	Value string `json:"value"`
+
+	// TenantID is the UUID of the project who owns the rule in neutron-lbaas.
+	// Only administrative users can specify a project UUID other than their own.
+	TenantID string `json:"tenant_id"`
+
+	// ProjectID is the UUID of the project who owns the rule in octavia.
+	// Only administrative users can specify a project UUID other than their own.
+	ProjectID string `json:"project_id"`
```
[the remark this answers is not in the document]
lingxiankong: For neutron-lbaas, it's 'tenant_id', for octavia, it's 'project_id', see here: https://github.com/openstack/octavia/blob/54a4cf00cf304b108013ca4487c138c13f30983d/octavia/api/v2/types/l7policy.py#L83

openstack/networking/v2/extensions/lbaas_v2/l7policies/requests.go
lingxiankong: you are right, these params don't need to be set if action is

openstack/networking/v2/extensions/lbaas_v2/l7policies/requests.go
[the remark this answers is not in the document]
jtopjian: Just to confirm: let's leave this as `int32`. That'll prevent out-of-bounds integers from being passed.

openstack/networking/v2/extensions/lbaas_v2/l7policies/testing/fixtures.go
lingxiankong: I installed neutron-lbaas lib in my octavia devstack environment, the sample has been changed according to my test, just keep the uuids that are referenced in all the following PRs.

jtopjian:
@lingxiankong One more thing about PRs: after a code review has started, can you not squash/amend commits? I realize we have a good amount of internal/house-rules for working with reviews and I apologize about that. However, with all of those rules out of the way, the rest of the PRs will go quite quick :)

This all looks good to me. The only thing left here is the question I have for @jrperritt.

lingxiankong:
@jtopjian, sorry the inconvenience. I am still getting familiar with github review, the journey is a little bit different with gerrit :-( i will find some docs to read

jtopjian:
@lingxiankong There's nothing to apologize for. We have a set of house rules that catch all new contributors off-guard.

Coincidentally I worked on this earlier today which might be useful:

https://github.com/jtopjian/gophercloud/tree/docs-tutorial/docs/contributor-tutorial

If you have any comments on this, I've opened #837 to discuss it.

lingxiankong:
@jtopjian given you still mark this PR 'Changes requested', just to confirm, i don't need to make any changes to this patch before @jrperritt's reply, right?

jtopjian:
@lingxiankong correct - Depending on the answer to the question I left, there might be further changes. But at the moment this all looks good :)

lingxiankong:
hi, any chance we can make this PR a little progress?

jrperritt:
> hi, any chance we can make this PR a little progress?

Please be patient

openstack/networking/v2/extensions/lbaas_v2/l7policies/results.go
```diff
@@ -0,0 +1,108 @@
...
+	Rules []Rule `json:"rules"`
+}
+
+// Rule represents layer 7 load balancing rule.
+type Rule struct {
+	// The unique ID for the L7 rule.
+	ID string `json:"id"`
+
+	// The L7 rule type. One of COOKIE, FILE_TYPE, HEADER, HOST_NAME, or PATH.
+	RuleType RuleType `json:"type"`
```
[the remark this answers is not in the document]
jrperritt: That's correct: strict types for requests, basic types for responses.

jrperritt:
I don't think we want to be adding support for a deprecated service API. Something like this should go under the recently-decided `loadbalancer` top-level package.

jtopjian:
> I don't think we want to be adding support for a deprecated service API. Something like this should go under the recently-decided loadbalancer top-level package.

Even if the deprecated LBaaS v2 service supports L7 policies? Wouldn't this be the same as someone adding a missing feature to, say, Keystone v2?

jrperritt:
> Even if the deprecated LBaaS v2 service supports L7 policies? 

Yes

> Wouldn't this be the same as someone adding a missing feature to, say, Keystone v2?

No. Neutron users would still be able to use L7 operations in the new package. Additionally, moving it will avoid having the exact same code in 2 places.

jtopjian:
hm... I might be overlooking something.

With the new load balancer service, was your plan to copy `lbaas_v2` or move it? A copy would also have the problem of duplicate code.

I guess I saw the new load balancer service differently: while the vast majority of the code would be identical to `lbaas_v2`, there would be slight changes over time (Cascading Delete is one example -- the proper way of handling that would be to have a standard `Delete` function with `DeleteOpts`, right?) as well as all new features going in load balancer. `lbaas_v2` would, eventually, sit idle as it's been deemed complete and no further work will ever be done.

jrperritt:
> With the new load balancer service, was your plan to copy lbaas_v2 or move it?

Copy, but freeze additions to `lbaas_v2`

> A copy would also have the problem of duplicate code.

Somewhat, but only the new package would get new features.

> Cascading Delete is one example -- the proper way of handling that would be to have a standard Delete function with DeleteOpts, right?

Yes, `CascadingDelete` will be either deprecated or removed.

> I guess I saw the new load balancer service differently: while the vast majority of the code would be identical to lbaas_v2, there would be slight changes over time ... as well as all new features going in load balancer. lbaas_v2 would, eventually, sit idle as it's been deemed complete and no further work will ever be done.

I think that's largely correct (so much so that I'm not sure where we diverge on this (if at all)). Since the Octavia API is a superset of Neutron's LB API, calling the Octavia `loadbalancer` operations with a `networking` client  should work just fine.

jtopjian:
OK, sounds good. I think our difference of approach here is that I'm making a big assumption that the loadbalancer/Octavia service will eventually somehow cause the networking/Neutron client to no longer be able to use the load balancer resources. ie: we add L7 to loadbalancer only and at some point networking/Neutron couldn't use it. It's a big assumption that may never happen, so I'm good with defering this to the new loadbalancer service.

So the next question is: who's going to do that copy? :)

lingxiankong:
@jrperritt @jtopjian so, do you mean we need to create a new folder (copy everything from current `lbaas_v2`) and add new (octavia) features there? So, the users who are using neutron-lbaas can still use `lbaas_v2` and the octavia users are recommended to use functions in the new folder. Am I understanding correctly?

jtopjian:
@lingxiankong Yes, that's the plan. :)

Once @jrperritt gets the new structure in place, your patches should be easy to reapply there.

I do apologize for the delay and inconvenience.

lingxiankong:
@jtopjian @jrperritt not intend to push you, but in case this pr is buried in your todo list, creating `loadbalancer ` folder is still in progress, right?

lingxiankong:
Considering we already created the loadbalancer folder, and all my changes will happen there, I still use a single commit as a start

lingxiankong:
@jtopjian @jrperritt this pr is ready for review, thanks for taking care of the new folder creation.

openstack/loadbalancer/v2/l7policies/results.go
```diff
@@ -0,0 +1,93 @@
...
+	ID string `json:"id"`
+
+	// Name of the L7 policy.
+	Name string `json:"name"`
+
+	// The ID of the listener.
+	ListenerID string `json:"listener_id"`
+
+	// The L7 policy action. One of REDIRECT_TO_POOL, REDIRECT_TO_URL, or REJECT.
+	Action Action `json:"action"`
```
jtopjian: Let's change this to `string`. Per the comment in the prior PR, custom types are only for requests. Results only use basic types (except if it's a struct).

openstack/loadbalancer/v2/l7policies/results.go
```diff
@@ -0,0 +1,93 @@
...
+	Rules []Rule `json:"rules"`
+}
+
+// Rule represents layer 7 load balancing rule.
+type Rule struct {
+	// The unique ID for the L7 rule.
+	ID string `json:"id"`
+
+	// The L7 rule type. One of COOKIE, FILE_TYPE, HEADER, HOST_NAME, or PATH.
+	RuleType RuleType `json:"type"`
```
jtopjian: Let's change this to a string

openstack/loadbalancer/v2/l7policies/results.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// The comparison type for the L7 rule. One of CONTAINS, ENDS_WITH, EQUAL_TO, REGEX, or STARTS_WITH.
+	CompareType CompareType `json:"compare_type"`
```
jtopjian: Let's change this to a string.

acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
```diff
@@ -100,6 +100,12 @@ func TestLoadbalancersCRUD(t *testing.T) {
 
 	tools.PrintResource(t, newListener)
 
+	// L7 policy
+	policy, err := CreateL7Policy(t, client, listener, lb)
```
jtopjian: `s/client/lbClient/`

Also, since `policy` isn't being used elsewhere, Go is going to complain. So you can either change this line to:

```go
_, err := CreateL7Policy(t, client, listener, lb)
```

Or add something like this after the err check:

```go
tools.PrintResource(t, policy)
```

lingxiankong:
@jtopjian new commit was submitted to solve your comments.

acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
```diff
@@ -101,7 +101,7 @@ func TestLoadbalancersCRUD(t *testing.T) {
 	tools.PrintResource(t, newListener)
 
 	// L7 policy
-	policy, err := CreateL7Policy(t, client, listener, lb)
+	_, err := CreateL7Policy(t, client, listener, lb)
```
jtopjian: I'm so sorry, this line should be:

```
_, err = CreateL7Policy(t, lbClient, listener, lb)
```

Note the omission of the `:` since there are no new variables being declared.

But also note that `client` needs changed to `lbClient` as that was changed during the move to the `octavia` directory.

lingxiankong: It should be me to say sorry for my carelessness, thanks for the review!

jtopjian:
@lingxiankong This looks good to me! Thank you for your work and patience with this.

lingxiankong:
thanks, i will start to rebase and change the other ones

PR #842, opened 2018-03-22
 Vpnaas: List IPSec site connections

simonre:
For #723 

code in this PR:

https://github.com/openstack/neutron-vpnaas/blob/058469e1b99b647537a5228c6a384d93df5484df/neutron_vpnaas/db/vpn/vpn_db.py#L270

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/vpnaas/siteconnection_test.go
+++ b/acceptance/openstack/networking/v2/extensions/vpnaas/siteconnection_test.go
@@ -14,6 +14,27 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/vpnaas/siteconnections"
 )
 
+func TestConnectionList(t *testing.T) {
+	client, err := clients.NewNetworkV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create a network client: %v", err)
+	}
+
+	allPages, err := siteconnections.List(client, nil).AllPages()
+	if err != nil {
+		t.Fatalf("Unable to list IPSec site connections: %v", err)
+	}
+
+	allConnections, err := siteconnections.ExtractConnections(allPages)
+	if err != nil {
+		t.Fatalf("Unable to extract IPSec site connections: %v", err)
+	}
+
+	for _, connection := range allConnections {
+		tools.PrintResource(t, connection)
+	}
+}
+
 func TestConnectionCRUD(t *testing.T) {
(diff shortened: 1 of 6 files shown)


simonre:
@jtopjian I think this is ready for review. I'm not sure what's up with that unit test failing but it doesn't seem to be a problem with my code I think. I cannot reproduce the unit test failing on my system either.

jtopjian:
@simonre This looks good to me. The test failure is indeed strange. I think OpenLab is having some problems right now. I'd like to leave this PR opened for another day or two so we can use it as a trigger for false test failures. Let me know if you have questions.

jtopjian:
recheck

jtopjian:
OK, well that's good news: unit tests are passing now, but acceptance is failing. 

@liu-sheng @mrhillsman I'm seeing OpenLab failures across all PRs. I'm not sure if it's a temporary issue that will resolve on its own or more of a deeper problem?

mrhillsman:
+ @animationzl @kiwik

animationzl:
recheck

animationzl:
OpenLab failure is fixed now, for the reason you can refer to this issue
https://github.com/theopenlab/openlab-zuul-jobs/issues/95

Feel sorry for the late recover, will make OpenLab more stable and responsive.

jtopjian:
@animationzl 

> Feel sorry for the late recover, will make OpenLab more stable and responsive.

Quite the opposite! A few days of downtime, especially over the weekend, is a speedy recovery and totally acceptable in my opinion!

Question: What's the best method of reporting problems? Would it be helpful to open issues at https://github.com/theopenlab/openlab-zuul-jobs?

mrhillsman:
@jtopjian for now the best way imo is #askopenlab on freenode

jtopjian:
@mrhillsman Sounds good! I can't guarantee hanging around in the channel since IRC isn't a tool I frequently use, but I can certainly post troublesome issues we're seeing in the future.

mrhillsman:
@jtopjian we are looking into some alternatives as well and connecting all the things; you can also email us - info@openlabtesting.org which sends notifications - i can create a filter for when you message and maybe add [CI ISSUE] in the subject line i think may work well for now ?

jtopjian:
@mrhillsman Email also works. If you find that particular workflow isn't effective, just let me know the next best way. I realize all things about OpenLab are new, so I have no problem being part of pilots and trials. :)

jtopjian:
@simonre Thank you again for this!

Issue #836, opened 2018-03-17
Zaqar: Support Messaging operations


problemv:
API Documentation: https://developer.openstack.org/api-ref/message/

- [ ] Queues
- [ ] Messages
- [ ] Claims
- [ ] Flavors
- [ ] Pools
- [ ] Subscriptions

problemv:
I will be submitting a few PRs to add all of this functionality within the following week. This will coincide with a separate pull to add Senlin support so gophercloud can support Autoscaling.

jtopjian:
@problemv This sounds good to me.

It looks like someone is already working on Senlin: https://github.com/gophercloud/gophercloud/issues/823. Are you two collaborating on this? Just want to make sure duplicate Senlin PRs aren't made.

Also, if you haven't, I recommend reading our [CONTRIBUTING](https://github.com/gophercloud/gophercloud/blob/master/.github/CONTRIBUTING.md) and [STYLE](https://github.com/gophercloud/gophercloud/blob/master/STYLEGUIDE.md) guides. Please let me know if you have any questions about the text - we're always trying to clarify it to make it easier for new contributors.

And definitely let me know if you have any questions about implementation and such.

problemv:
@jtopjian Thanks for the quick reply.

Yeah I am working with the developer who will be adding the Senlin patch.

jtopjian:
@problemv sounds good!

One more quick note:

> I will be submitting a few PRs to add all of this functionality within the following week.

Just to give you a realistic expectation of a timeline, it's probably going to take 3-4 weeks to get all of Zaqar/Messaging merged.

I usually have time to do one round of code reviews per day and the time it takes for change requests and rebasing/merge resolutions naturally add up. 

#723 is a really good example of a fast-moving series of PRs.

PR #846, opened 2018-03-27
Messaging Queue Create

problemv:
For #836 

code in this PR:

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/queues.py#L76

### DIFF
--- a/openstack/messaging/v2/queues/doc.go
+++ b/openstack/messaging/v2/queues/doc.go
@@ -7,13 +7,13 @@ Lists all queues and creates, shows information for updates, deletes, and action
 Example to Create a Queue
 
 	createOpts := queues.CreateOpts{
-		QueueName:					"My_Queue",
-		MaxMessagesPostSize: 		262143,
-		DefaultMessageTTL: 			3700,
-		DefaultMessageDelay: 		25,
-		DeadLetterQueueMessageTTL: 	3500,
-		MaxClaimCount: 				10,
-		Extra: 						map[string]interface{}{"description": "Test queue."},
+		QueueName:                  "My_Queue",
+		MaxMessagesPostSize:        262143,
+		DefaultMessageTTL:          3700,
+		DefaultMessageDelay:        25,
+		DeadLetterQueueMessageTTL:  3500,
+		MaxClaimCount:              10,
+		Extra:                      map[string]interface{}{"description": "Test queue."},
 	}
 
 	err := queues.Create(client, clientID, createOpts).ExtractErr()
--- a/openstack/messaging/v2/queues/doc.go
+++ b/openstack/messaging/v2/queues/doc.go
@@ -16,7 +16,7 @@ Example to Create a Queue
 		Extra:                      map[string]interface{}{"description": "Test queue."},
 	}
 
-	err := queues.Create(client, clientID, createOpts).ExtractErr()
+	err := queues.Create(client, createOpts).ExtractErr()
 	if err != nil {
 		panic(err)
 	}
--- a/openstack/messaging/v2/queues/doc.go
+++ b/openstack/messaging/v2/queues/doc.go
@@ -1,21 +1,22 @@
 /*
 Package queues provides information and interaction with the queues through
-the OpenStack Messaging(Zaqar) service.
+the OpenStack Messaging (Zaqar) service.
 
 Lists all queues and creates, shows information for updates, deletes, and actions on a queue.
 
 Example to Create a Queue
 
 	createOpts := queues.CreateOpts{
-		MaxMessagesPostSize: 262143,
-		DefaultMessageTTL: 3700,
-		DefaultMessageDelay: 25,
-		DeadLetterQueueMessageTTL: 3500,
-		MaxClaimCount: 10,
-		Description: "Test queue.",
+		QueueName:					"My_Queue",
+		MaxMessagesPostSize: 		262143,
+		DefaultMessageTTL: 			3700,
+		DefaultMessageDelay: 		25,
+		DeadLetterQueueMessageTTL: 	3500,
+		MaxClaimCount: 				10,
+		Extra: 						map[string]interface{}{"description": "Test queue."},
 	}
 
-	err := queues.Create(client, "My Queue", clientID, createOpts).ExtractErr()
+	err := queues.Create(client, clientID, createOpts).ExtractErr()
 	if err != nil {
 		panic(err)
 	}
--- a/openstack/messaging/v2/queues/doc.go
+++ b/openstack/messaging/v2/queues/doc.go
@@ -1,0 +1,23 @@
+/*
+Package queues provides information and interaction with the queues through
+the OpenStack Messaging(Zaqar) service.
+
+Lists all queues and creates, shows information for updates, deletes, and actions on a queue.
+
+Example to Create a Queue
+
+	createOpts := queues.CreateOpts{
+		MaxMessagesPostSize: 262143,
+		DefaultMessageTTL: 3700,
+		DefaultMessageDelay: 25,
+		DeadLetterQueueMessageTTL: 3500,
+		MaxClaimCount: 10,
+		Description: "Test queue.",
+	}
+
+	err := queues.Create(client, "My Queue", clientID, createOpts).ExtractErr()
+	if err != nil {
+		panic(err)
+	}
+*/
+package queues
--- a/openstack/messaging/v2/queues/requests.go
+++ b/openstack/messaging/v2/queues/requests.go
@@ -12,12 +12,15 @@ type CreateOptsBuilder interface {
 
 // CreateOpts specifies the queue creation parameters.
 type CreateOpts struct {
-	//The target incoming messages will be moved to when a message can’t
+	// The name of the queue to create.
+	QueueName string `json:"queue_name" required:"true"`
+
+	// The target incoming messages will be moved to when a message can’t
 	// processed successfully after meet the max claim count is met.
 	DeadLetterQueue string `json:"_dead_letter_queue,omitempty"`
 
 	// The new TTL setting for messages when moved to dead letter queue.
-	DeadLetterQueueMessageTTL int `json:"_dead_letter_queue_messages_ttl,omitempty"`
+	DeadLetterQueueMessagesTTL int `json:"_dead_letter_queue_messages_ttl,omitempty"`
 
 	// The delay of messages defined for a queue. When the messages send to
 	// the queue, it will be delayed for some times and means it can not be
@@ -39,22 +42,36 @@ type CreateOpts struct {
 	// for any messages posted to the queue.
 	MaxMessagesPostSize int `json:"_max_messages_post_size,omitempty"`
 
-	// Description of the queue.
-	Description string `json:"description,omitempty"`
+	// Extra is free-form extra key/value pairs to describe the user.
+	Extra map[string]interface{} `json:"-"`
 }
 
 // ToQueueCreateMap constructs a request body from CreateOpts.
 func (opts CreateOpts) ToQueueCreateMap() (map[string]interface{}, error) {
-	return gophercloud.BuildRequestBody(opts, "")
+	b, err := gophercloud.BuildRequestBody(opts, "")
+	if err != nil {
+		return nil, err
+	}
+
+	if opts.Extra != nil {
+		for key, value := range opts.Extra {
+			b[key] = value
+		}
+
+	}
+	return b, nil
 }
 
 // Create requests the creation of a new queue.
-func Create(client *gophercloud.ServiceClient, queueName string, clientID string, opts CreateOptsBuilder) (r CreateResult) {
+func Create(client *gophercloud.ServiceClient, clientID string, opts CreateOptsBuilder) (r CreateResult) {
 	b, err := opts.ToQueueCreateMap()
 	if err != nil {
 		r.Err = err
 		return
 	}
+
+	queueName := b["queue_name"].(string)
+	delete(b, "queue_name")
 
 	_, r.Err = client.Put(createURL(client, queueName), b, r.Body, &gophercloud.RequestOpts{
 		OkCodes:     []int{201, 204},
--- a/openstack/messaging/v2/queues/requests.go
+++ b/openstack/messaging/v2/queues/requests.go
@@ -63,7 +63,7 @@ func (opts CreateOpts) ToQueueCreateMap() (map[string]interface{}, error) {
 }
 
 // Create requests the creation of a new queue.
-func Create(client *gophercloud.ServiceClient, clientID string, opts CreateOptsBuilder) (r CreateResult) {
+func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResult) {
 	b, err := opts.ToQueueCreateMap()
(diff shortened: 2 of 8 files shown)


jtopjian:
@problemv I think the test failures are ignorable here. There _might_ be something wrong with a recent concurrency test -- it looks to be randomly failing from time to time. 

As well, the coverage decrease of 0.004% is absolutely ignorable. :)

problemv:
@jtopjian I have acceptance tests written for this as well, however, they are dependent on multiple functions being available. Is it alright to hold those tests until the end?

jtopjian:
@problemv It's possible - depends on the situation.

This PR is for Queue creation, so the acceptance test would create a queue. However, since there isn't support to delete a queue, we won't be able to clean up after the test. That's OK -- it's only a temporary issue. The next PRs you submit can amend the test until there's a fully functional test that creates, reads, updates, and deletes.

However, if you can't create a queue unless there are other Zaqar resources created first, we should merge those first. This way we can bundle acceptance tests with all PRs. If this is the situation you're in, I'm OK with continuing with this PR but then if you can, switch to a resource we can begin actively testing.

If you're in a situation different than the two above, can you let me know more details?

problemv:
@jtopjian Ideally it would save me having to rewrite the tests to wait until all CRUD operations are merged to add the acceptance tests. I CAN rewrite them it will just take some extra time. 

Your call I am fine either way.

jtopjian:
@problemv Fair enough. Let's do this: I'm open to working on the Queue PRs without adding acceptance tests until the end.

If it turns out that there were significant changes to the Queue implementation that ended up causing you to rewrite your tests anyways, then for the rest of the resources, let's include tests in each PR.

However, if it worked out that either no changes or minor changes were needed, we can keep doing that.

Just a heads up: please make sure the tests fit the patterns seen in the `acceptance` directory (ex: helper functions to `Create` (with randomized info where applicable) and `Delete`). The `compute` acceptance tests were recently refactored, so those would be the best examples to go off of.

jtopjian:
@problemv I noticed you're opening other PRs for Messaging -- is this ready for review? Just a reminder to explicitly request a review or else we'll still consider it a WIP :)

problemv:
@jtopjian, yes those PRs are ready to review. Sorry I didn't realize, I will request one on all future PRs.

openstack/messaging/v2/queues/requests.go
```diff
@@ -0,0 +1,65 @@
...
+
+	// The max number the message can be claimed.
+	MaxClaimCount int `json:"_max_claim_count,omitempty"`
+
+	// The max post size of messages defined for a queue, which will effect
+	// for any messages posted to the queue.
+	MaxMessagesPostSize int `json:"_max_messages_post_size,omitempty"`
+
+	// Description of the queue.
+	Description string `json:"description,omitempty"`
```
jtopjian: I can't find where `Description` is defined in the Messaging service. I searched _all_ of the code and I can only find "description" being used for error messages or for rendering API documentation.

I have a theory about this: From looking through the Zaqar code and commit messages, _all_ of these fields are Metadata, but the ones prefixed with `_` are reserved/special metadata. 

I'm thinking that the Create call will accept _any_ field. You can verify this by adding any number of additional fields to this struct such as:

```go
Foo string `json:"foo,omitempty"`
```

And try to create a queue with `Foo` set. My guess is that "foo" will be added as some metadata of the queue.

If this turns out to be true, we should create a master field called `Metadata` (but keep the reserved metadata fields as-is) and implement it similar to Keystone's "Extra" field:

https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/users/requests.go#L103
https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/users/requests.go#L119-L125

openstack/messaging/v2/queues/requests.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// ToQueueCreateMap constructs a request body from CreateOpts.
+func (opts CreateOpts) ToQueueCreateMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "")
+}
+
+// Create requests the creation of a new queue.
+func Create(client *gophercloud.ServiceClient, queueName string, clientID string, opts CreateOptsBuilder) (r CreateResult) {
```
jtopjian: I'm not a fan of this function signature -- there's a lot going on.

An alternative might be to add `QueueName` and `ClientID` to the `CreateOpts` struct. Then below, you could do:

```go
b, err := opts.ToQueueCreateMap()
...

clientID := b["client_id"]
delete(b, "client_id")
queueName := b["queue_name"]
delete(b, "queue_name")

...
```

Alternatively, if `clientID` is going to be a staple in all headers, then maybe only remove `queueName` from the signature.

Having `QueueName` in the `CreateOpts` makes sense since that follows patterns seen throughout the rest of Gophercloud. For example, when creating a user, the username is part of CreateOpts. I realize that the queue name is part of the request URL here, but as shown above, we can work around that.

The reason I'm focusing on this is because if you look at all `Create` function signatures in Gophercloud, the vast majority only have 2 parameters: client and createopts. A handful have a third parameter for a parent resource ID. Unless we _really_ need to, I'd like to keep this close to form.

Thoughts?

openstack/messaging/v2/queues/urls.go
```diff
@@ -0,0 +1,10 @@
+package queues
+
+import "github.com/gophercloud/gophercloud"
+
+var apiVersion = "v2"
+var apiName = "queues"
```
jtopjian: nit: these could probably be constants.

openstack/messaging/v2/queues/doc.go
```diff
@@ -0,0 +1,23 @@
+/*
+Package queues provides information and interaction with the queues through
+the OpenStack Messaging(Zaqar) service.
```
jtopjian: nit: space between Messaging and Zaqar.

openstack/messaging/v2/queues/doc.go
```diff
@@ -0,0 +1,23 @@
...
+Example to Create a Queue
+
+	createOpts := queues.CreateOpts{
+		MaxMessagesPostSize: 262143,
+		DefaultMessageTTL: 3700,
+		DefaultMessageDelay: 25,
+		DeadLetterQueueMessageTTL: 3500,
+		MaxClaimCount: 10,
+		Description: "Test queue.",
+	}
```
jtopjian: nit: align the fields here

problemv:
@jtopjian Yeah, the Client-ID header is required for almost all requests to Zaqar.

> In Zaqar, this is used to avoid echoing a sender's messages back to the same instance, and may be logged by the server to provide diagnostics and statistics. Should be generated once and persisted between restarts of the client to avoid the race condition where a client sends a message, restarts with a different ID, and then receives its own message, echoed back.

Also thank you so much for the timely review I will get the changes done asap. :)

openstack/messaging/v2/queues/requests.go
problemv: I agree with this 100%. 

ClientID is a required header for (almost) all requests so it makes sense to leave it in the function signature IMO but I will remove queueName. 

Unless you feel really strongly about moving clientID into createOpts I would prefer to leave it as (almost) all of the other functions will have clientID in the signature.

jtopjian: > ClientID is a required header for (almost) all requests so it makes sense to leave it in the function signature IMO but I will remove queueName.

This sounds good to me. If client ID will be used for most/all other functions, having it consistently in the function signature is a good idea.

problemv:
@jtopjian This is ready to be reviewed again. 

Seems like CI sometimes fails out on me for random reasons unrelated to my patches.

jtopjian:
> Seems like CI sometimes fails out on me for random reasons unrelated to my patches.

If it's the unit tests that are failing, then it's #855 

Acceptance test failures are coincidentally ignorable because 1) there's no acc tests yet and 2) only a subset of acc tests are run at the moment :)

openstack/messaging/v2/queues/doc.go
```diff
@@ -0,0 +1,24 @@
...
+
+	createOpts := queues.CreateOpts{
+		QueueName:					"My_Queue",
+		MaxMessagesPostSize: 		262143,
+		DefaultMessageTTL: 			3700,
+		DefaultMessageDelay: 		25,
+		DeadLetterQueueMessageTTL: 	3500,
+		MaxClaimCount: 				10,
+		Extra: 						map[string]interface{}{"description": "Test queue."},
+	}
```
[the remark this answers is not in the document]
jtopjian: It's a common issue with Go and editors. `gofmt` won't format blocks of comments or strings (such as JSON fixtures). So if your editor expands/shrinks tabs/spaces, it can really alter how things look :)

problemv:
I would be alright with doing that for the client-id. 

Just keep in mind that if you post messages with one client-id you will need to use a different one to consume those messages from the queue and by embedding the client-id in the service client I don't know how easy it would be to swap back and fourth for creation/consumption of messages.

jtopjian:
Ah, I see. So in theory an application could be acting on behalf of multiple client-ids but utilizing the same underlying service client?

The idea of storing the client-id within the service client would work best if most of the use-cases would involve a one-to-one service client to client id relationship.

problemv:
Yeah, but it wouldn't be out of the question to have someone utilizing gophercloud instantiate a new client with a different client-id if they need to create/consume messages with a different client-id. 

In theory someone shouldn't be creating and consuming messages with the same client anyway as that client-id is supposed to be for distinguishing between clients anyway.

jtopjian:
@problemv I'll defer to your expertise here. If you think having the ability to specify a client-id per function is the safest/best approach, I'm totally fine with that.

problemv:
@jtopjian I thought about this for a while and discussed with my colleagues. I think doing it they way you suggest as a arbitrary field within the `ServiceClient` is the best option.

I don't want to break convention with my create function if I can help it and doing it this way would clean up the function signature for everything which I really like. 

However, I am not 100% sure how to accomplish this so any direction would be helpful. :)

jtopjian:
@problemv sounds good. The first step in doing this will be to get @jrperritt's sign-off on the modification to the `ServiceClient` struct. I'm also defering to his expertise in design. It's always possible adding this to `ServiceClient` is a bad idea.

One way in which this is a possible bad idea is if someone leans heavily on changing the client-id in the client and then performs some concurrent actions.

@jrperritt is a bit busy at the moment. If we don't hear back from him soon we can proceed with having client-id as a function parameter and always refactor later.

Also, it looks like #860 beat everyone as the first commit so this will need rebased to handle the merge conflict.

jtopjian:
@problemv Quick update on this: we're still discussing the best way forward with the client-id. I'm trying to get a Zaqar environment up and running to verify the implementation but time is not being friendly to me. If I can't get something running soon, we'll proceed with just having client-id in the function parameters.

problemv:
Awesome thank you!

problemv:
@jtopjian Not sure why the openlab checks don't like me :(.

Anyway this and all of the other PRs are rebased and ready to go. I will be adding the acceptance tests PR shortly.

jtopjian:
@problemv Thanks for fixing up the merge conflict. This all looks good. 

I've confirmed that client-id is necessary for all Zaqar/Messaging requests, so we'll want to have a convenient way of handling it. Just trying to sort out whether we want to do that now or later. Either way, once we do that, these PRs will start flowing through. I apologize for this hold up.

problemv:
@jtopjian No problem at all, sorry if sorting out that merge conflict spammed you.

jtopjian:
@problemv OK! Here's the plan we have:

1. The `gophercloud.ServiceClient` will have a new field of `MoreHeaders` added to it.
2. The `NewMessagingV2` function will have a third parameter of `clientID` which it will add to the `MoreHeaders` field.
3. The internal `Request` function will add any headers found in `MoreHeaders` to the outgoing request.

This way, when someone creates a new message client, they only ever have to specify a client-id once and everything else is automatically taken care of.

Work on the service client will be done shortly. In the meantime, if you wanted to start removing the `clientID` parameter from all of the functions, that should get everything in place.

problemv:
@jtopjian Awesome! I will get that change in asap.

jtopjian:
@problemv https://github.com/gophercloud/gophercloud/pull/898 was merged today. The last piece of this PR would be to modify the `NewMessagingV2` function to accept a clientID as a parameter and to add it to the service client before returning it. Let me know if you need help with this.

jtopjian:
@problemv at long last I think this is good to merge :smile: 

Can you confirm that the new client-id implementation is working with acceptance tests?

jtopjian:
@problemv I cherry-picked the Acceptance Test PR, made modifications where appropriate, and this all looks good:

```
$ OS_DEBUG=1 gophercloudtest TestCRUDQueues messaging/v2                                        
~/go/src/github.com/gophercloud/gophercloud ~/go/src/github.com/gophercloud/gophercloud         
=== RUN   TestCRUDQueues                                                                        
2018/04/07 21:08:23 [DEBUG] OpenStack Request URL: PUT http://zaqar:8888/v2/queues/ACPTTESTuLBFH
2018/04/07 21:08:23 [DEBUG] OpenStack request Headers:                                          
Accept: application/json                                                                        
Client-Id: 1f9aa72b-f084-483c-98ee-fbcd914b2d20                                                 
User-Agent: gophercloud/2.0.0                                                                   
X-Auth-Token: ***                                                                               
2018/04/07 21:08:23 [DEBUG] OpenStack Request Body: {                                           
  "_dead_letter_queue_messages_ttl": 3500,                                                      
  "_default_message_delay": 25,                                                                 
  "_default_message_ttl": 3700,                                                                 
  "_max_claim_count": 10,                                                                       
  "_max_messages_post_size": 262143,                                                            
  "description": "Test queue for Gophercloud acceptance tests."                                 
}                                                                                               
2018/04/07 21:08:24 [DEBUG] OpenStack Response Code: 201                                        
2018/04/07 21:08:24 [DEBUG] OpenStack Response Headers:                                         
Content-Length: 0                                                                               
Content-Type: application/json; charset=UTF-8                                                   
Date: Sat, 07 Apr 2018 21:08:23 GMT                                                             
Location: http://zaqar:8888/v2/queues/ACPTTESTuLBFH                                             
Server: Apache                                                                                  
2018/04/07 21:08:24 [DEBUG] Unable to parse OpenStack JSON: unexpected end of JSON input        
--- PASS: TestCRUDQueues (1.53s)                                                                
        messaging.go:14: Attempting to create Queue: ACPTTESTuLBFH                              
PASS                                                                                            
ok      github.com/gophercloud/gophercloud/acceptance/openstack/messaging/v2    1.533s          
~/go/src/github.com/gophercloud/gophercloud                                                     
                                                                                                
root@admin:226:25:~/go/src/github.com/gophercloud/gophercloud                                   
$ openstack messaging queue list                                                                          

+---------------+                                                                               
| Name          |                                                                               
+---------------+                                                                                                                                                         
| ACPTTESTuLBFH |                                                                               
+---------------+                                                                               
```

Thank you again for your help and patience with this PR!

problemv:
Awesome, I just tested it as well. LGTM. Thanks!

PR #848, opened 2018-03-27
Messaging Queue List

problemv:
For #836

code in this PR:

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/queues.py#L253

### DIFF
--- a/openstack/messaging/v2/queues/results.go
+++ b/openstack/messaging/v2/queues/results.go
@@ -2,6 +2,7 @@ package queues
 
 import (
 	"encoding/json"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/internal"
 	"github.com/gophercloud/gophercloud/pagination"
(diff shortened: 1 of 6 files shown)


openstack/messaging/v2/queues/results.go
```diff
@@ -0,0 +1,67 @@
...
+type QueueDetails struct {
+	DeadLetterQueue           string `json:"_dead_letter_queue"`
+	DeadLetterQueueMessageTTL int    `json:"_dead_letter_queue_messages_ttl"`
+	DefaultMessageDelay       int    `json:"_default_message_delay"`
+	DefaultMessageTTL         int    `json:"_default_message_ttl"`
+	MaxClaimCount             int    `json:"_max_claim_count"`
+	MaxMessagesPostSize       int    `json:"_max_messages_post_size"`
+	Description               string `json:"description"`
+	Flavor                    string `json:"flavor"`
+}
```
jtopjian: Since you can create a Queue with arbitrary key/value pairs, those extra metadata pairs will be returned here, too. So we need to extract all of them into a "bucket" field. There's an internal function called [`RemainingKeys`](https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/users/results.go#L72) that can handle this. Let me know if you need help implementing it.

With that in mind, I think we should remove `Description`. I think Flavor is used in other areas of Zaqar, so it should be fine to keep as a "standard" field.

problemv: Done, I also expanded the unit tests to test to test this.

jtopjian:
@problemv I rebased this locally and tried running some acceptance tests on it.

I'm getting an error when using the `AllPages` method of retrieval rather than `EachPage`:

```
convenience.go:35: Failure in queue_test.go, line 69: unexpected error "Get /v2/queues?detailed=true&marker=ACPTTESTuLBFH&limit=10: unsupported protocol scheme \"\""
```

You should be able to reproduce this by adding the following to the List acceptance test:

```go
allPages, err := queues.List(client, listOpts).AllPages()
th.AssertNoErr(t, err)
allQueues, err := queues.ExtractQueues(allPages)
```

jtopjian:
@problemv Figured it out -- fortunately there was precedence for this :)

https://github.com/gophercloud/gophercloud/blob/master/openstack/imageservice/v2/images/results.go#L174-L190

```go
// NextPageURL uses the response's embedded link reference to navigate to the
// next page of results.
func (r QueuePage) NextPageURL() (string, error) {
  var s struct {
    Links []gophercloud.Link `json:"links"`
  }
  err := r.ExtractInto(&s)
  if err != nil {
    return "", err
  }

  next, err := gophercloud.ExtractNextURL(s.Links)
  if err != nil {
    return "", err
  } 
  
  return nextPageURL(r.URL.String(), next)
}
```

and then implement `nextPageURL` in `urls.go` as defined [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/imageservice/v2/images/urls.go#L40-L51)

Also, note how I renamed the receiver from `page` to `r`. That's a style conflict we caught the other day. It occurs throughout Gophercloud but the majority of implementations are `r` and I'll be cleaning that up globally later.

problemv:
@jtopjian Updated and ready to go, I had to modify the unit tests let me know what you think about the change I am not 100% sure if its the best way to do it.

PR #849, opened 2018-03-27
Messaging Queue Update

problemv:
For #836 

code in this PR:

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/queues.py#L125


openstack/messaging/v2/queues/requests.go
```diff
@@ -0,0 +1,158 @@
...
+type UpdateOptsBuilder interface {
+	ToQueueUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts is an array of UpdateQueueBody.
+type UpdateOpts []UpdateQueueBody
+
+// UpdateOpts implements UpdateOpts.
+type UpdateQueueBody struct {
+	Op    string      `json:"op" required:"true"`
```
jtopjian: There are probably a finite amount of ops, right? I did a quick search and I think they are "add", "remove", and "replace":

https://github.com/openstack/zaqar/blob/83300343d23051db31702d1cadbf9e250c37ef24/zaqar/transport/wsgi/v2_0/queues.py#L221-L244

It would be best to have a custom type of UpdateOp and then have each type as a const. Something like:

```go
type UpdateOp string
const (
  ReplaceOp UpdateOp = "replace"
  AddOp UpdateOp = "add"
  RemoveOp UpdateOp = "remove"
)
```

problemv:
@jtopjian Rebased and ready to go as well.

problemv:
@jtopjian Alright I think I did that correctly, tested it on a live env and it is working. 

Thanks again for all of your hard work on this, much appreciated.

jtopjian:
@problemv Looks good to me. Likewise, thank you for your patience with working through these PRs :)

openstack/messaging/v2/queues/doc.go
```diff
@@ -11,7 +11,8 @@ Example to List Queues
 	}
 
 	pager := queues.List(client, listOpts)
-	err = pager.EachPage(func(page pagination.Page) (bool, error) {
+
+    err = pager.EachPage(func(page pagination.Page) (bool, error) {
```
jtopjian: Minor formatting on the doc here. It's safe to fix in a future PR, though.

PR #851, opened 2018-03-27
Messaging Queue Delete

problemv:
For #836 

code in this PR:

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/queues.py#L109

### DIFF
--- a/openstack/messaging/v2/queues/requests.go
+++ b/openstack/messaging/v2/queues/requests.go
@@ -173,3 +173,11 @@ func Update(client *gophercloud.ServiceClient, queueName string, opts UpdateOpts
 	})
 	return
 }
+
+// Get requests details on a single queue, by name.
+func Get(client *gophercloud.ServiceClient, queueName string) (r GetResult) {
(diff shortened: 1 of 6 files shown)


openstack/messaging/v2/queues/requests.go
```diff
@@ -148,3 +148,12 @@ func Get(client *gophercloud.ServiceClient, queueName string, clientID string) (
 		MoreHeaders: map[string]string{"Client-ID": clientID}})
 	return
 }
+
+// Delete deletes the specified queue.
+func Delete(client *gophercloud.ServiceClient, queueName string, clientID string) (r DeleteResult) {
+	_, r.Err = client.Request("DELETE", deleteURL(client, queueName), &gophercloud.RequestOpts{
```
jtopjian: Quick note on this: it should be possible to do `client.Delete` instead of `client.Request` and still pass the `MoreHeaders` option. But let me know if that didn't work for you.

problemv:

Issue #852, opened 2018-03-28
Create Dedicated Load Balancer packages


jtopjian:
Currently, the `LoadBalancerV2` client is piggybacking off of the Neutron `lbaas_v2` packages. Per discussions in #832 and #834, we want to create dedicated packages for Octavia/LBaaS.

lingxiankong:
Thanks for writing this issue!

jrperritt:
closed by #854

PR #854, opened 2018-03-28
Octavia

jtopjian:
For #852


jtopjian:
AFAICT, this works. I ran the acceptance tests on vexxhost and everything worked. There's an issue with the `TestLoadbalancersCRUD` test where the call to cascading delete needs to wait until everything is deleted before proceeding to delete the network and subnet, but that's not a blocking/breaking issue.

lingxiankong:
seems the go fmt failed

lingxiankong:
looks good to me now, thanks!

jtopjian:
@lingxiankong thanks!

@jrperritt when you have a moment, can you do a quick review? I think this is good to go.

openstack/loadbalancer/v2/loadbalancers/requests.go
```diff
@@ -0,0 +1,198 @@
...
+func Delete(c *gophercloud.ServiceClient, id string) (r DeleteResult) {
+	_, r.Err = c.Delete(resourceURL(c, id), nil)
+	return
+}
+
+// CascadingDelete is like `Delete`, but will also delete any of the load balancer's
+// children (listener, monitor, etc).
+// NOTE: This function will only work with Octavia load balancers; Neutron does not
+// support this.
+func CascadingDelete(c *gophercloud.ServiceClient, id string) (r DeleteResult) {
```
jrperritt: Did we want to instead add a `DeleteOptsBuilder` to `Delete`?

jtopjian: Good point. Yes. Better to do it now.

The recent set of commits have this added, though I had to do a force-push due to the merge conflict.

jtopjian:
@jrperritt ping for review

Issue #234, opened 2017-01-18
volume quota api


sb10:
I am doing:

    q, err := quotasets.Get(p.computeClient, os.Getenv("OS_TENANT_ID")).Extract()

and get a value like:

> &{ID:9d1c1c5a5107425e8a0c9febbfcd2152 FixedIps:-1 FloatingIps:10 InjectedFileContentBytes:10240 InjectedFilePathBytes:255 InjectedFiles:5 KeyPairs:0 MetadataItems:128 Ram:65536 SecurityGroupRules:20 SecurityGroups:10 Cores:96 Instances:64}

Using the OpenStack dashboard I can see that I have a quota on the number and total size of volumes I can create. How do I get this information via the gophercloud SDK?

(BTW, I'll note that the FloatingIps value is wrong; I'm allowed 15 and have used 2)

jtopjian:
> How do I get this information via the gophercloud SDK?

Volume quotas are managed by the [Block Storage API](http://developer.openstack.org/api-ref/block-storage/v1/index.html#quota-sets-extension-os-quota-sets). This area isn't implemented in Gophercloud yet.

At the moment, each OpenStack project is responsible for managing its own resources' quotas. There have been several topics over the years to house them under one API.

Horizon is doing this by making calls to each OpenStack service's quota queries and aggregating them in a central dashboard. The unified `openstack` command is beginning to do the same.

> (BTW, I'll note that the FloatingIps value is wrong; I'm allowed 15 and have used 2)

That's a bit strange. Do you see a value of 15 if you do something like:

```shell
$ openstack --debug quota show <project>
$ nova --debug quota show
```

sb10:
The openstack command doesn't work for me (just hangs), and nova quota-show gives me 10. I'll follow up with our admin here to figure that one out.

Should this issue be left open as a "feature request", or is this basically never going to happen?

jtopjian:
> Should this issue be left open as a "feature request", or is this basically never going to happen?

I think you can leave it open. I'm sure Block Storage Quota Sets will eventually get added. I would say the goal of Gophercloud is to support all OpenStack APIs. :)

sb10:
How about neturon quotas? I realised that my quota of 15 floating ips can be found using `neutron quota-show`. I do need to find a user's port quota which is also only listed under neutron; is that retrievable with gophercloud?

jtopjian:
[...]

PR #859, opened 2018-03-30
[wip] Implement volume quota api

waynr:
For #234

I largely copied this code from openstack/compute/v2/extensions/quotasets and
then refactored the tests toward a more data-driven approach. I'd like to do
something similar with openstack/compute/v2/extensions/quotasets, but I thought
I'd see how the maintainers like the current approach.

I haven't taken the time to try this out against an actual openstack
environment but will probably do so tomorrow and remove the "[wip]" marker once
I've done so.

The PR template asked for a list of links to source code that support this, but
all I've been looking at is the API docs:


waynr:
TODO:
* [ ] Test this code against an openstack environment
* [ ] Fix the doc.go files.
* [ ] Try to get feedback soonish.

jtopjian:
@waynr Thank you for this :smile: 

It might be helpful to review this contributor tutorial I've recently made: https://github.com/jtopjian/gophercloud/tree/docs-tutorial/docs/contributor-tutorial

It'll help with some of the details needed to get this merged.

Notably, this will need to be broken up into a few separate PRs: one for each action. In this case, there will probably be 3: get, detailed, and update.

Additionally the section on Code Hunting will help provide some additional information we need to get this merged.

These items are in-house rules to Gophercloud, which is why I've written the tutorial -- all new contributors hit them, so my hope is to make the learning curve easier.

Do let me know if you have any questions, though. I'd like to see this feature added so I'm available for help :)

PR #860, opened 2018-03-30
Clustering policytype list

dkt26111:
For #823 

code in this PR:

API doc:
https://developer.openstack.org/api-ref/clustering/#list-policy-types

API code:
https://github.com/openstack/senlin/blob/master/senlin/api/openstack/v1/policy_types.py#L30

### DIFF
--- a/openstack/clustering/v1/policytypes/results.go
+++ b/openstack/clustering/v1/policytypes/results.go
@@ -6,13 +6,13 @@ import (
 
 // PolicyType represents a clustering policy type in the Openstack cloud
 type PolicyType struct {
-	Name          string `json:"name"`
-	Version       string `json:"version"`
+	Name          string                         `json:"name"`
+	Version       string                         `json:"version"`
 	SupportStatus map[string][]SupportStatusType `json:"support_status,omitempty"`
 }
 
 // SupportStatus represents the support status information for a clustering policy type
-type SupportStatusType struct{
+type SupportStatusType struct {
 	Status string `json:"status"`
 	Since  string `json:"since"`
 }
--- a/openstack/clustering/v1/policytypes/results.go
+++ b/openstack/clustering/v1/policytypes/results.go
@@ -8,10 +8,10 @@ import (
 type PolicyType struct {
 	Name          string                         `json:"name"`
 	Version       string                         `json:"version"`
-	SupportStatus map[string][]SupportStatusType `json:"support_status,omitempty"`
+	SupportStatus map[string][]SupportStatusType `json:"support_status"`
 }
 
-// SupportStatus represents the support status information for a clustering policy type
+// SupportStatusType represents the support status information for a clustering policy type
 type SupportStatusType struct {
 	Status string `json:"status"`
 	Since  string `json:"since"`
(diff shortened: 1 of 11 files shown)


dkt26111:
@jtopjian: Requesting code review.  I'm working with JackKuei on implementing the senlin APIs.

openstack/clustering/v1/policytypes/results.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+// PolicyType represents a clustering policy type in the Openstack cloud
+type PolicyType struct {
+	Name          string                         `json:"name"`
+	Version       string                         `json:"version"`
+	SupportStatus map[string][]SupportStatusType `json:"support_status,omitempty"`
+}
+
+// SupportStatus represents the support status information for a clustering policy type
```
jtopjian: SupportStatusType?

openstack/clustering/v1/policytypes/urls.go
```diff
@@ -0,0 +1,10 @@
+package policytypes
+
+import "github.com/gophercloud/gophercloud"
+
+var apiVersion = "v1"
+var apiName = "policy-types"
```
jtopjian: nit: these could be constants:

```go
const (
  apiVersion = "v1"
  apiName = "policy-types"
)
```

openstack/clustering/v1/policytypes/results.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// PolicyType represents a clustering policy type in the Openstack cloud
+type PolicyType struct {
+	Name          string                         `json:"name"`
+	Version       string                         `json:"version"`
+	SupportStatus map[string][]SupportStatusType `json:"support_status,omitempty"`
```
jtopjian: nit: `omitempty` isn't really needed here - it's only useful in requests.

jtopjian:
@dkt26111 whoops - actually, my mistake. I was looking at policy list and not policy type list :)

openstack/clustering/v1/policytypes/results.go
dkt26111: Yes, you are correct.  I will fix this.

openstack/clustering/v1/policytypes/urls.go
dkt26111: Agree.  I'll fix it.

openstack/clustering/v1/policytypes/results.go
dkt26111: I'll fix that too.

dkt26111:
If everything looks good to you in the latest commits, you can go ahead and merge it.  I will work with Jackkuei to rebase #824 .

jtopjian:
@dkt26111 All looks good! Really nice work on this.

A note on microversions: we haven't figured out a good way of handling them in Gophercloud yet. Basic microversion stuff like what you're testing here is totally fine.

However if there will be times when, for example, a version 1.5 request or response body is different than a 1.0 request/body (meaning the struct fields are different and/or incompatible), this is the area we're still looking into. We can discuss different options if we get into that situation with Senlin.

PR #861, opened 2018-03-30
Clustering PolicyType Get implementation

dkt26111:
For #823 

code in this PR:

API doc:
https://developer.openstack.org/api-ref/clustering/#show-policy-type-details

API code:
https://github.com/openstack/senlin/blob/master/senlin/api/openstack/v1/policy_types.py#L43

### DIFF
--- a/openstack/clustering/v1/policytypes/results.go
+++ b/openstack/clustering/v1/policytypes/results.go
@@ -41,49 +41,8 @@ func (page PolicyTypePage) IsEmpty() (bool, error) {
 // PolicyTypeDetail represents the detailed policy type information for a clustering policy type
 type PolicyTypeDetail struct {
 	Name          string                         `json:"name"`
-	Schema        SchemaType                     `json:"schema"`
+	Schema        map[string]interface{}         `json:"schema"`
 	SupportStatus map[string][]SupportStatusType `json:"support_status,omitempty"`
-}
-
-// SchemaType represents the schema of a clustering policy type
-type SchemaType struct {
-	// senlin.policy.affinity:
-	AvailabilityZone   map[string]interface{} `json:"availability_zone,omitempty"`
-	EnableDrsExtension map[string]interface{} `json:"enable_drs_extension,omitempty"`
-	Servergroup        map[string]interface{} `json:"servergroup,omitempty"`
-
-	// senlin.policy.batch:
-	MaxBatchSize map[string]interface{} `json:"max_batch_size,omitempty"`
-	MinInService map[string]interface{} `json:"min_in_service,omitempty"`
-	PauseTime    map[string]interface{} `json:"pause_time,omitempty"`
-
-	// senlin.policy.health:
-	Detection map[string]interface{} `json:"detection,omitempty"`
-	Recovery  map[string]interface{} `json:"recovery,omitempty"`
-
-	// senlin.policy.scaling:
-	Adjustment map[string]interface{} `json:"adjustment,omitempty"`
-	Event      map[string]interface{} `json:"event,omitempty"`
-
-	// senlin.policy.region_placement:
-	Regions map[string]interface{} `json:"regions,omitempty"`
-
-	// senlin.policy.loadbalance:
-	HealthMonitor   map[string]interface{} `json:"health_monitor,omitempty"`
-	LbStatusTimeout map[string]interface{} `json:"lb_status_timeout,omitempty"`
-	Loadbalancer    map[string]interface{} `json:"loadbalancer,omitempty"`
-	Pool            map[string]interface{} `json:"pool,omitempty"`
-	Vip             map[string]interface{} `json:"vip,omitempty"`
-
-	// senlin.policy.deletion:
-	Criteria              map[string]interface{} `json:"criteria,omitempty"`
-	DestroyAfterDeletion  map[string]interface{} `json:"destroy_after_deletion,omitempty"`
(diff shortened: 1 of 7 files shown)


dkt26111:
@jtopjian Requesting code review.  This is continuation of the clustering policytype API implementation.

openstack/clustering/v1/policytypes/results.go
```diff
@@ -36,3 +37,68 @@ func (page PolicyTypePage) IsEmpty() (bool, error) {
...
+
+// PolicyTypeDetail represents the detailed policy type information for a clustering policy type
+type PolicyTypeDetail struct {
+	Name          string                         `json:"name"`
+	Schema        SchemaType                     `json:"schema"`
+	SupportStatus map[string][]SupportStatusType `json:"support_status,omitempty"`
+}
+
+// SchemaType represents the schema of a clustering policy type
+type SchemaType struct {
```
jtopjian: For each set of policy type fields, will a result only ever have one set returned or is it possible for a policy type to have values for, for example, affinity, batch, and health?

If a result will only ever have `senlin.policy.affinity` set, then we're leaving a lot of unused fields here. In addition, if Senlin adds more policies types in later releases, then they wouldn't be parsable here until Gophercloud was updated. Finally, how would a user know what fields they should be looking at if the bulk of them are empty?
If this is all true, I wonder if we should just make `Schema` a generic `map[string]interface{}`.

openstack/clustering/v1/policytypes/testing/fixtures.go
```diff
@@ -69,6 +71,67 @@ const PolicyTypeBody = `
...
+        "support_status": {
+            "1.0": [
+                {
+                    "status": "SUPPORTED",
+                    "since": "2016.10"
+                }
+            ]
+        }
+    }
+}
```
jtopjian: Side note: normally we don't like to use test fixtures pulled from the API docs. This is for two reasons:

1. It might be an old/stale response (we've seen this with some of the networking API fixtures)
2. It might not exercise all fields we want to test

Admittedly this is something we (I) often forget to verify, though. Since you folks are more familiar with Senlin/Clustering than me, I'll defer to your judgement on whether the API doc fixtures are usable or if you should use fixtures generated from your environment (usually done by using the `--debug` flag with the command-line tools).

openstack/clustering/v1/policytypes/results.go
dkt26111: Only one set will ever be returned.  I've had the same discussion with @problemv and @jackkuei, and we couldn't come to a clear consensus. Our thinking was that the structs make it easier for a developer using Gophercloud to access the policy type fields.

Usually it is rare that Senlin adds a new policy type.  However, a more frequent case is that the policy types are extended (i.e. from version 1.0 to 1.1) with the addition of new fields.  In that case your concern is valid that an update to Gophercloud is required before the new policy type would be usable.

I also prefer making `Schema` a generic `map[string]interface{}` as long as you are OK with the Gophercloud users having to iterate thru the `map[string]interface{}` to retrieve each field.

openstack/clustering/v1/policytypes/testing/fixtures.go
dkt26111: I think the API doc is up-to-date, but I will double-check that.

openstack/clustering/v1/policytypes/results.go
jtopjian: > Our thinking was that the structs make it easier for a developer using Gophercloud to access the policy type fields.

This is true. However, in this case I think there's too much ambiguity with all of the fields to make this struct easily usable.

If the policy types are standard (meaning they're built-in to Senlin and users can't create their own), one other option is to make distinct structs for each policy schema type and then have individual `ExtractXXXX` functions where XXXX is something like `ExtractAffinityPolicySchema`. 

I think the decision here should be based on how often a user is going to consume a Policy Type Schema. If parsing the policy is an important action in Senlin, then the Extract method is the way to go (but also keep in mind that new fields will be omitted until they are added to the struct). But if looking at the policy schema is something which will only be done in niche cases, using a `map[string]interface{}` should be fine -- at least to start with.

dkt26111: Ok, thanks for the clarification.  Inspecting the policy type schema is definitely a rare case, so I will go with the `map[string]interface{}` implementation in this case.

dkt26111:
recheck

jtopjian:
@dkt26111 OpenLab acceptance test failures can be ignored here. OpenLab is only running a small portion of the overall Gophercloud tests and Senlin/Clustering isn't part of that set.

So any failures that occur are transient issues with OpenLab itself :)

dkt26111:
I see.  Are there any efforts to add Senlin acceptance tests to the OpenLab environment?  Or should we work with the OpenLab team to add that?

jtopjian:
We (Gophercloud) will work with OpenLab to get them integrated at some point (OpenLab might have to create a configuration set for Senlin). 

I'm currently working on rewriting / cleaning up our current acceptance test library to get them in better shape for OpenLab testing. The format that you're using here (based on the recent changes to the compute acceptance tests) is perfect -- that'll save me from rewriting them later 😄

dkt26111:
Ok, I will leave it to your team to add the Senlin acceptance to OpenLab ;)

When you get a chance, can you review my latest changes?  It fixes the Schema type.

jtopjian:
LGTM!

Issue #866, opened 2018-04-02
Identity V3: add user to group


zhangzujian:
API reference:
[Identity API v3 - Add user to group](https://developer.openstack.org/api-ref/identity/v3/#add-user-to-group)

Python source code:
https://github.com/openstack/keystone/blob/a6adc72e3e7ff4ff0bff0702e69ebd8f697d6261/keystone/identity/routers.py#L49
https://github.com/openstack/keystone/blob/a6adc72e3e7ff4ff0bff0702e69ebd8f697d6261/keystone/identity/core.py#L1248
https://github.com/openstack/python-keystoneclient/blob/40642d597deecce4dd81428449353bbd0c2ad5be/keystoneclient/v3/users.py#L228

jtopjian:
Implemented in #867

PR #867, opened 2018-04-02
Identity V3: add user to group

zhangzujian:
For #866

API reference:
[Identity API v3 - Add user to group](https://developer.openstack.org/api-ref/identity/v3/#add-user-to-group)

Python source code:
https://github.com/openstack/keystone/blob/a6adc72e3e7ff4ff0bff0702e69ebd8f697d6261/keystone/identity/core.py#L1248
https://github.com/openstack/python-keystoneclient/blob/40642d597deecce4dd81428449353bbd0c2ad5be/keystoneclient/v3/users.py#L228

### DIFF
--- a/acceptance/openstack/identity/v3/users_test.go
+++ b/acceptance/openstack/identity/v3/users_test.go
@@ -156,6 +156,54 @@ func TestUsersListGroups(t *testing.T) {
 	}
 }
 
+func TestUserAddToGroup(t *testing.T) {
+	client, err := clients.NewIdentityV3Client()
+	if err != nil {
+		t.Fatalf("Unable to obtain an identity client: %v", err)
+	}
+
+	createOpts := users.CreateOpts{
+		Password: "foobar",
+		DomainID: "default",
+		Options: map[users.Option]interface{}{
+			users.IgnorePasswordExpiry: true,
+			users.MultiFactorAuthRules: []interface{}{
+				[]string{"password", "totp"},
+				[]string{"password", "custom-auth-method"},
+			},
+		},
+		Extra: map[string]interface{}{
+			"email": "jsmith@example.com",
+		},
+	}
+
+	user, err := CreateUser(t, client, &createOpts)
+	if err != nil {
+		t.Fatalf("Unable to create user: %v", err)
+	}
+	defer DeleteUser(t, client, user.ID)
+
+	tools.PrintResource(t, user)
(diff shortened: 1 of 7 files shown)


zhangzujian:
Ready for review.

zhangzujian:
Ready for review.

jtopjian:
@zhangzujian The acceptance test is failing with an "index out of range" error.

After creating the user, how about creating a temporary group and then add the user to that group? This will work around the assumption that the testing environment has pre-existing groups.

zhangzujian:
@jtopjian 

I copied some code from _TestUsersListInGroup()_, so is it possible the function has the same problem?

Why the accepetance testing in openlab passed? How can I check out the error you mentioned?

jtopjian:
@zhangzujian 

> I copied some code from TestUsersListInGroup(), so is it possible the function has the same problem?

No, it wouldn't have the same problem. The core issue is that you were doing a list of all groups and then picking the first group returned (element 0). In theory this would work _if_ the test environment had pre-existing groups. But since a clean OpenStack environment has no groups by default, element 0 is out of range (since it's an empty list).

> Why the accepetance testing in openlab passed? How can I check out the error you mentioned?

The OpenLab tests are only running a subset of tests right now. You can see the results of the tests here:

http://logs.openlabtesting.org/logs/67/867/913377a985f908cf0027aae1d85faee36cc3946b/check/gophercloud-acceptance-test/b2cdca6/test_results.html

For tests that are not run in OpenLab, I manually verify them myself in a similar test environment. You should be able to do the same if you have access to an OpenStack environment (such as devstack or similar). The README page here has details on how to run the tests:

https://github.com/gophercloud/gophercloud/tree/master/acceptance

Let me know if that helps :)

zhangzujian:
@jtopjian 

> Class:github.com/gophercloud/gophercloud/acceptance/openstack/identity/v3
> Test Name:TestUserAddToGroup
> Running Time:0.25
> State:Passed

I have already checked the report and it shows the test passed. I'm really confused.

zhangzujian:
@jtopjian  Checks passed, please review.

jtopjian:
@zhangzujian This looks good to me. Is this ready to be merged?

zhangzujian:
@jtopjian  Ready for merge.

jtopjian:
Oh, just saw your comment now.

Thank you for making the updates here - nice work!

PR #870, opened 2018-04-03
Clustering Policy List implementation

dkt26111:
For #823 

code in this PR:

API doc:
https://developer.openstack.org/api-ref/clustering/#list-policies

API code:
https://github.com/openstack/senlin/blob/master/senlin/api/openstack/v1/policies.py#L34


dkt26111:
@jtopjian  Requesting code review.  Thanks.

openstack/clustering/v1/policies/requests.go
```diff
@@ -0,0 +1,56 @@
...
+	Limit int `q:"limit"`
+
+	// Marker and Limit control paging. Marker instructs List where to start listing from.
+	Marker string `q:"marker"`
+
+	// Sorts the response by one or more attribute and optional sort direction combinations.
+	Sort string `q:"sort"`
+
+	// GlobalProject indicates whether to include resources for all projects or resources for the current project
+	GlobalProject bool `q:"global_project"`
```
jtopjian: `bool` is going to cause the list query to send `?global_project=false` whenever it isn't explicitly set to `true` in the `listOpts`. Is this OK? If it's best to only send this field in the URL if it's explicitly set to either `true` or `false`, change this to `*bool`.

openstack/clustering/v1/policies/results.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// Policy represents a clustering policy in the Openstack cloud
+type Policy struct {
+	CreatedAt   *time.Time             `json:"-"`
+	Data        map[string]interface{} `json:"data"`
+	DomainUUID  string                 `json:"domain"`
```
jtopjian: Let's change this to just `Domain` to match the JSON field. Same with the other UUID fields here.

openstack/clustering/v1/policies/requests.go
dkt26111: Yea, it's better to only send it when it is explicitly set by the user.  I'll change it to *bool.

openstack/clustering/v1/policies/results.go
dkt26111: Np.  I'll fix that.

dkt26111:
> With regard to the `*time.Time` fields, let's change them to a standard `time.Time`
I'll fix that.

dkt26111:
> If so, should `Cooldown` and `Level` also be added or are those not API-facing fields?

Actually looks like those fields as well as the `Metadata` field are never returned to the user.. The stack trace for policy list call looks like this:

https://github.com/openstack/senlin/blob/master/senlin/objects/policy.py#L104
https://github.com/openstack/senlin/blob/master/senlin/engine/service.py#L536
https://github.com/openstack/senlin/blob/master/senlin/objects/policy.py#L92
https://github.com/openstack/senlin/blob/master/senlin/objects/base.py#L46

The `to_dict` function does not return `Cooldown`, `Level` or `Metadata`.  So I'll remove `Metadata`

dkt26111:
No offense taking at all.  I definitely appreciate you digging into the Senlin code and verifying the validity of the API.  In fact, you were able to uncover the webhook bug for the V parameter which helped us fix the Senlin code.   So definitely keep pointing out any inconsistencies that you see so that we can keep improving both Gophercloud and Senlin :)

openstack/clustering/v1/policies/results.go
```diff
@@ -0,0 +1,77 @@
...
+	type tmp Policy
+	var s struct {
+		tmp
+		CreatedAt gophercloud.JSONRFC3339MilliNoZ `json:"created_at,omitempty"`
+		UpdatedAt gophercloud.JSONRFC3339MilliNoZ `json:"updated_at,omitempty"`
+	}
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		fmt.Println("Error Unmarshal")
+		fmt.Printf("Detail Unmarshal Error: %v", err)
```
jtopjian: Let's remove these two lines

dkt26111: Done

jtopjian:
LGTM - thank you :)

PR #872, opened 2018-04-03
Senlin: Clusters Create

JackKuei:

code in this PR:

[[clusters:create]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/clusters.py#L68)


JackKuei:
Depends on Profile Create #871

JackKuei:
recheck

openstack/clustering/v1/clusters/results.go
```diff
@@ -0,0 +1,127 @@
...
+	ProfileName     string                 `json:"profile_name"`
+	Project         string                 `json:"project"`
+	Status          string                 `json:"status"`
+	StatusReason    string                 `json:"status_reason"`
+	Timeout         int                    `json:"timeout"`
+	UpdatedAt       time.Time              `json:"-"`
+	User            string                 `json:"user"`
+}
+
+func ExtractActionFromLocation(location string) (string, error) {
```
jtopjian: I understand that this function is convenient to have, but since it doesn't directly involve the request being sent or the response being parsed, let's remove it. This kind of function could be added directly to the application which is importing Gophercloud.

Let me know if I'm wrong, though, and it'll have a larger purpose later on.

openstack/clustering/v1/clusters/results.go
```diff
@@ -0,0 +1,127 @@
...
+	Config          map[string]interface{} `json:"config"`
+	CreatedAt       time.Time              `json:"-"`
+	Data            map[string]interface{} `json:"data"`
+	Dependents      map[string]interface{} `json:"dependents"`
+	DesiredCapacity int                    `json:"desired_capacity"`
+	Domain          string                 `json:"domain"`
+	ID              string                 `json:"id"`
+	InitAt          time.Time              `json:"-"`
+	MaxSize         int                    `json:"max_size"`
+	Metadata        map[string]interface{} `json:"metadata"`
```
jtopjian: The Senlin code shows this as a JSON field:

https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/objects/cluster.py#L45

Can you add some metadata to the unit tests to ensure it's parsed correctly?

openstack/clustering/v1/clusters/requests.go
```diff
@@ -0,0 +1,48 @@
...
+
+// CreateOptsBuilder Builder.
+type CreateOptsBuilder interface {
+	ToClusterCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts params
+type CreateOpts struct {
+	Name            string                 `json:"name" required:"true"`
+	DesiredCapacity int                    `json:"desired_capacity" required:"false"`
```
jtopjian: `required:"false"` can be removed here.

openstack/clustering/v1/clusters/requests.go
```diff
@@ -0,0 +1,48 @@
...
+	ProfileID       string                 `json:"profile_id" required:"true"`
+	MinSize         int                    `json:"min_size,omitempty"`
```
jtopjian: Is `0` a valid value to pass? If so, let's change this to `*int`. 

`0` is `int`'s zero-value and combined with `omitempty` means it'll never be sent in the request.

openstack/clustering/v1/clusters/requests.go
```diff
@@ -0,0 +1,48 @@
...
+	Timeout         int                    `json:"timeout,omitempty"`
+	MaxSize         int                    `json:"max_size,omitempty"`
+	Metadata        map[string]interface{} `json:"metadata,omitempty"`
```
jtopjian: `Metadata` in other services is normally `map[string]string`. If you were to have something like:

```
metadata["foo"] = 12
metadata["bar"] = map[string]string{"some": "value"}
```

Senlin will respect the types?

openstack/clustering/v1/clusters/results.go
```diff
@@ -0,0 +1,127 @@
+package clusters
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+
+	"reflect"
```
jtopjian: nit: have reflect bundled with the other standard library packages.

openstack/clustering/v1/clusters/results.go
JackKuei: Don't have a strong opinion on where this function should live. But ExtractingAction is going to be used frequently in order to query the status.

openstack/clustering/v1/clusters/requests.go
JackKuei: Good catch

openstack/clustering/v1/clusters/requests.go
JackKuei: Done

openstack/clustering/v1/clusters/results.go
JackKuei: Done

openstack/clustering/v1/clusters/requests.go
JackKuei: The doc specified ```A collection of key-value pairs associated with an object.``` Senlin allows bool, string, float values for metadata values. Added unit/accep tests to test metadata

openstack/clustering/v1/clusters/results.go
JackKuei: Done

openstack/clustering/v1/clusters/results.go
jtopjian: Understood. I'm guessing the reason for this is to poll the Actions API to determine when a certain action has finished?

If so, and given the function in #873 was very similar, it might make sense to add a function (or functions) to https://github.com/gophercloud/utils which will parse the action ID (returned from any resource) and poll until a given status.

I wrote something like this a while ago and have been meaning to add it to the official utils repo:

* https://github.com/jtopjian/gophertools/blob/master/common.go
* https://github.com/jtopjian/gophertools/blob/master/openstack/compute/v2/tools/servers.go

The utils repo is meant to provide convenience functions which Gophercloud doesn't require to function but would otherwise be repeated in almost every situation.

Issue #875, opened 2018-04-03
Service Client: implement HEAD method


zhangzujian:
Some Keystone APIs use HEAD method.

jtopjian:
Implemented in #876

PR #876, opened 2018-04-03
Service Client: add HEAD method

zhangzujian:
For #875

### DIFF
--- a/openstack/identity/v3/tokens/requests.go
+++ b/openstack/identity/v3/tokens/requests.go
@@ -133,7 +133,7 @@ func Get(c *gophercloud.ServiceClient, token string) (r GetResult) {
 
 // Validate determines if a specified token is valid or not.
 func Validate(c *gophercloud.ServiceClient, token string) (bool, error) {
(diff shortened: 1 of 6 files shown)


jtopjian:
@jrperritt Thoughts on this?

There are currently a handful of places in Gophercloud that are issuing a `HEAD` by doing `client.Request("HEAD")`, so this would be a more convenient form.

zhangzujian:
Ready for review.

zhangzujian:
recheck

jtopjian:
recheck

jrperritt:
I think this is OK, but we'd want to go through and change all the `HEAD` calls to use this new function and make sure all the tests still pass.

jtopjian:
@zhangzujian per @jrperritt's comment, would you be able to replace all existing occurrences of `client.Request("HEAD", ...` with `client.Head(...`

This way we can see if anything would break in CI?

zhangzujian:
@jtopjian Please review.

jtopjian:
Looks good to me - thank you!

PR #879, opened 2018-04-03
Blockstorage v3 quota-set support - part 3, Get Details

waynr:
For #234
Pending #878

\## API:

\## Schema:

```
```

list of quotas.

\## Update:

\## Delete:

### DIFF
--- a/openstack/blockstorage/extensions/quotasets/doc.go
+++ b/openstack/blockstorage/extensions/quotasets/doc.go
@@ -10,5 +10,13 @@ Example to Get a Quota Set
 
 	fmt.Printf("%+v\n", quotaset)
 
+Example to Get a Detailed Quota Set
+
+	quotaset, err := quotasets.GetDetail(blockStorageClient, "project-id").Extract()
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Printf("%+v\n", quotaset)
 */
 package quotasets
--- a/openstack/blockstorage/extensions/quotasets/doc.go
+++ b/openstack/blockstorage/extensions/quotasets/doc.go
@@ -10,9 +10,9 @@ Example to Get a Quota Set
 
 	fmt.Printf("%+v\n", quotaset)
 
-Example to Get a Detailed Quota Set
+Example to Get Quota Set Usage
 
-	quotaset, err := quotasets.GetDetail(blockStorageClient, "project-id").Extract()
+	quotaset, err := quotasets.GetUsage(blockStorageClient, "project-id").Extract()
 	if err != nil {
 		panic(err)
 	}
--- a/openstack/blockstorage/extensions/quotasets/requests.go
+++ b/openstack/blockstorage/extensions/quotasets/requests.go
@@ -1,6 +1,8 @@
 package quotasets
 
 import (
+	"fmt"
+
 	"github.com/gophercloud/gophercloud"
 )
 
@@ -18,6 +20,7 @@ func GetDefaults(client *gophercloud.ServiceClient, projectID string) (r GetResu
 
 // GetUsage returns detailed public data about a previously created QuotaSet.
 func GetUsage(client *gophercloud.ServiceClient, projectID string) (r GetUsageResult) {
-	_, r.Err = client.Get(getUsageURL(client, projectID), &r.Body, nil)
+	u := fmt.Sprintf("%s?usage=true", getURL(client, projectID))
+	_, r.Err = client.Get(u, &r.Body, nil)
 	return
 }
--- a/openstack/blockstorage/extensions/quotasets/requests.go
+++ b/openstack/blockstorage/extensions/quotasets/requests.go
@@ -15,3 +15,9 @@ func GetDefaults(client *gophercloud.ServiceClient, projectID string) (r GetResu
 	_, r.Err = client.Get(getDefaultsURL(client, projectID), &r.Body, nil)
 	return
 }
+
+// GetDetail returns detailed public data about a previously created QuotaSet.
+func GetDetail(client *gophercloud.ServiceClient, projectID string) (r GetDetailResult) {
+	_, r.Err = client.Get(getDetailURL(client, projectID), &r.Body, nil)
+	return
+}
--- a/openstack/blockstorage/extensions/quotasets/requests.go
+++ b/openstack/blockstorage/extensions/quotasets/requests.go
@@ -16,8 +16,8 @@ func GetDefaults(client *gophercloud.ServiceClient, projectID string) (r GetResu
 	return
 }
 
-// GetDetail returns detailed public data about a previously created QuotaSet.
-func GetDetail(client *gophercloud.ServiceClient, projectID string) (r GetDetailResult) {
-	_, r.Err = client.Get(getDetailURL(client, projectID), &r.Body, nil)
+// GetUsage returns detailed public data about a previously created QuotaSet.
+func GetUsage(client *gophercloud.ServiceClient, projectID string) (r GetUsageResult) {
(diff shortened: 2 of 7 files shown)


waynr:
@jtopjian @jrperritt this should be ready for review.

openstack/blockstorage/extensions/quotasets/requests.go
```diff
@@ -15,3 +15,9 @@ func GetDefaults(client *gophercloud.ServiceClient, projectID string) (r GetResu
 	_, r.Err = client.Get(getDefaultsURL(client, projectID), &r.Body, nil)
 	return
 }
+
+// GetDetail returns detailed public data about a previously created QuotaSet.
+func GetDetail(client *gophercloud.ServiceClient, projectID string) (r GetDetailResult) {
```
jtopjian: I think we should call this `GetUsage`. I understand that this API call returns more than what's "in use" and so "detail" certainly makes sense. But the word "detail" isn't found in the API docs or in the cinder code when referencing this call. However, the API docs, URL, and code all mention "usage".

As well, if this is the only API call that returns what's in use, then someone searching for how to return the quota usage will more easily discover this.

Please let me know if you feel differently, though.

openstack/blockstorage/extensions/quotasets/results.go
```diff
@@ -33,6 +33,66 @@ type QuotaSet struct {
 	BackupGigabytes int `json:"backup_gigabytes"`
 }
 
+// QuotaDetailSet represents details of both operational limits of block
+// storage resources and the current usage of those resources.
+type QuotaDetailSet struct {
```
jtopjian: Similar to the `GetUsage` function name, I think this should be called `QuotaSetUsage` since it's returning a set of quota items and their usage.

openstack/blockstorage/extensions/quotasets/results.go
```diff
@@ -33,6 +33,66 @@ type QuotaSet struct {
...
+	// BackupGigabytes is the size (GB) usage information of backup for this
+	// project, including in_use, limit, reserved and allocated attributes.
+	// Note: allocated attribute is available only when nested quota is
+	// enabled.
+	BackupGigabytes QuotaDetail `json:"backup_gigabytes"`
+}
+
+// QuotaDetail is a set of details about a single operational limit that allows
+// for control of block storage usage.
+type QuotaDetail struct {
```
jtopjian: And finally, this might work better as `QuotaUsage` since it's describing the usage of an individual quota item.

Again, I do realize that this also contains "limit" which isn't something "in use", but the other items describe the consumption/usage of the resource.

openstack/blockstorage/extensions/quotasets/results.go
```diff
@@ -33,6 +33,66 @@ type QuotaSet struct {
...
+	// InUse is the current number of provisioned resources of the given type.
+	InUse int `json:"in_use"`
+
+	// Allocated is the current number of resources of a given type allocated
+	// for use.  It is only available when nested quota is enabled. It tells
+	// how
```
jtopjian: This looks to be cut off.

openstack/blockstorage/extensions/quotasets/results.go
```diff
@@ -72,3 +132,23 @@ func (r quotaResult) Extract() (*QuotaSet, error) {
...
+// method to interpret it as a QuotaSet.
+type GetDetailResult struct {
+	quotaDetailResult
+}
+
+// Extract is a method that attempts to interpret any QuotaDetailSet
+// resource response as a set of QuotaDetailSet structs.
+func (r quotaDetailResult) Extract() (QuotaDetailSet, error) {
+	var s struct {
+		QuotaData QuotaDetailSet `json:"quota_set"`
```
jtopjian:
```go
QuotaSetUsage QuotaSetUsage `json:"quota_set"`
```

openstack/blockstorage/extensions/quotasets/urls.go
```diff
@@ -11,3 +11,7 @@ func getURL(c *gophercloud.ServiceClient, projectID string) string {
 func getDefaultsURL(c *gophercloud.ServiceClient, projectID string) string {
 	return c.ServiceURL(resourcePath, projectID, "defaults")
 }
+
+func getDetailURL(c *gophercloud.ServiceClient, projectID string) string {
+	return c.ServiceURL(resourcePath, projectID+"?usage=true")
```
jtopjian: Instead of adding the query here, it might be cleaner to add it in the function itself such as what's done here:

https://github.com/gophercloud/gophercloud/blob/de4b788e8f384b917d879d1d788765f1c752d3e9/openstack/networking/v2/extensions/lbaas_v2/loadbalancers/requests.go#L189-L190

openstack/blockstorage/extensions/quotasets/requests.go
waynr: My only concern about this is the lack of consistency with [the compute quotaset](https://github.com/gophercloud/gophercloud/blob/e7e2296ede45795fce803ac2906a442acb16f1c1/openstack/compute/v2/extensions/quotasets/requests.go#L15), which uses `Detail` rather than `Usage`. That being said, I don't feel too attached and will update per your comments; if you do decide consistency is a virtue we can discard those changes before merge.

waynr:
@jtopjian I think this is ready for review again, thanks for the feedback! Also let me know if you think it's better to be consistent with the compute quotaset gophercloud api, it's not problem for me to revert the changes i made.

openstack/blockstorage/extensions/quotasets/doc.go
```diff
@@ -10,5 +10,13 @@ Example to Get a Quota Set
 
 	fmt.Printf("%+v\n", quotaset)
 
+Example to Get a Detailed Quota Set
```
jtopjian: Example to Get Quota Usage

openstack/blockstorage/extensions/quotasets/doc.go
```diff
@@ -10,5 +10,13 @@ Example to Get a Quota Set
...
+
+	quotaset, err := quotasets.GetDetail(blockStorageClient, "project-id").Extract()
```
jtopjian: `quotasets.GetUsage(...`

openstack/blockstorage/extensions/quotasets/results.go
```diff
@@ -72,3 +131,23 @@ func (r quotaResult) Extract() (*QuotaSet, error) {
...
+// method to interpret it as a QuotaSet.
+type GetUsageResult struct {
+	quotaUsageResult
+}
+
+// Extract is a method that attempts to interpret any QuotaUsageSet resource
+// response as a set of QuotaUsageSet structs.
+func (r quotaUsageResult) Extract() (QuotaUsageSet, error) {
+	var s struct {
+		QuotaData QuotaUsageSet `json:"quota_set"`
```
jtopjian:
```go
QuotaUsageSet QuotaUsageSet `json:"quota_set"`
```

openstack/blockstorage/extensions/quotasets/doc.go
waynr: @jtopjian done

openstack/blockstorage/extensions/quotasets/doc.go
waynr: @jtopjian done

openstack/blockstorage/extensions/quotasets/results.go
waynr: @jtopjian done

waynr:
@jtopjian I think this is ready for review again.

waynr:
Whoops, accidentally clicked "close and comment"

jtopjian:
LGTM - nice work!

PR #880, opened 2018-04-03
Blockstorage v3 quota-set support - part 4, Update

waynr:
For #234
Pending #879

\## API:

\## Schema:

```
```

list of quotas.

\## Update:

\## Delete:

### DIFF
--- a/acceptance/openstack/blockstorage/v3/quotaset_test.go
+++ b/acceptance/openstack/blockstorage/v3/quotaset_test.go
@@ -40,6 +40,51 @@ func TestQuotasetGetUsage(t *testing.T) {
 	tools.PrintResource(t, quotaSetUsage)
 }
 
+var UpdateQuotaOpts = quotasets.UpdateOpts{
+	Volumes:            gophercloud.IntToPointer(100),
+	Snapshots:          gophercloud.IntToPointer(200),
+	Gigabytes:          gophercloud.IntToPointer(300),
+	PerVolumeGigabytes: gophercloud.IntToPointer(50),
+	Backups:            gophercloud.IntToPointer(2),
+	BackupGigabytes:    gophercloud.IntToPointer(300),
+}
+
+var UpdatedQuotas = quotasets.QuotaSet{
+	Volumes:            100,
+	Snapshots:          200,
+	Gigabytes:          300,
+	PerVolumeGigabytes: 50,
+	Backups:            2,
+	BackupGigabytes:    300,
+}
+
+func TestQuotasetUpdate(t *testing.T) {
+	client, projectID := getClientAndProject(t)
+
+	// save original quotas
+	orig, err := quotasets.Get(client, projectID).Extract()
+	th.AssertNoErr(t, err)
+
+	defer func() {
+		restore := quotasets.UpdateOpts{}
+		FillUpdateOptsFromQuotaSet(*orig, &restore)
+
+		_, err = quotasets.Update(client, projectID, restore).Extract()
+		th.AssertNoErr(t, err)
+	}()
+
+	// test Update
+	resultQuotas, err := quotasets.Update(client, projectID, UpdateQuotaOpts).Extract()
+	th.AssertNoErr(t, err)
+	th.AssertEquals(t, UpdatedQuotas, *resultQuotas)
+
+	// We dont know the default quotas, so just check if the quotas are not the
+	// same as before
+	newQuotas, err := quotasets.Get(client, projectID).Extract()
+	th.AssertNoErr(t, err)
+	th.AssertEquals(t, resultQuotas.Volumes, newQuotas.Volumes)
+}
+
 // getClientAndProject reduces boilerplate by returning a new blockstorage v3
 // ServiceClient and a project ID obtained from the OS_PROJECT_NAME envvar.
 func getClientAndProject(t *testing.T) (*gophercloud.ServiceClient, string) {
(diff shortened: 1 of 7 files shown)


waynr:
@jtopjian @jrperritt I think this PR is ready for review!

acceptance/openstack/blockstorage/v3/quotaset_test.go
```diff
@@ -40,6 +41,58 @@ func TestQuotasetGetUsage(t *testing.T) {
...
+	// same as before
+	newQuotas, err := quotasets.Get(client, projectID).Extract()
+	th.AssertNoErr(t, err)
+	if newQuotas.Volumes != resultQuotas.Volumes {
+		t.Fatalf(
+			fmt.Sprintf("Failed to update quotas\n\texpected: %d\t actual: %d",
+				resultQuotas.Volumes,
+				newQuotas.Volumes,
+			),
+		)
```
jtopjian: It might be easier to structure this like:

```go
err := fmt.Errorf("Failed to ...")
t.Fatal(err)
```

But in this case, you can simplify it down to:

```go
newQuotas, err := quotasets.Get(client, projectID).Extract()
th.AssertNoErr(t, err)
th.AssertEquals(t, resultQuotas.Volumes, newQuota.Volumes)
```

`th.AssertEquals` will error and print out an almost identical statement when the two items differ.

acceptance/openstack/blockstorage/v3/quotaset_test.go
```diff
@@ -50,3 +103,11 @@ func getClientAndProject(t *testing.T) (*gophercloud.ServiceClient, string) {
 	th.AssertNoErr(t, err)
 	return client, projectID
 }
+func FillUpdateOptsFromQuotaSet(src quotasets.QuotaSet, dest *quotasets.UpdateOpts) {
```
jtopjian: nit: newline above this. I'm surprised `gofmt` didn't catch that.

openstack/blockstorage/extensions/quotasets/requests.go
```diff
@@ -24,3 +24,61 @@ func GetUsage(client *gophercloud.ServiceClient, projectID string) (r GetUsageRe
...
+	// Force will update the quotaset even if the quota has already been used
+	// and the reserved quota exceeds the new quota.
+	Force bool `json:"force,omitempty"`
+}
+
+// UpdateOptsBuilder enables extensins to add parameters to the update request.
+type UpdateOptsBuilder interface {
+	// Extra specific name to prevent collisions with interfaces for other quotas
+	// (e.g. neutron)
+	ToBlockStorageQuotaUpdateMap() (map[string]interface{}, error)
```
jtopjian: I realize this was copied from the `compute/extensions/v2/quotasets` package, so I'm OK with this.

For posterity or the chance anyone digs this conversation up: I suppose this naming would be useful for someone who wants to create an all-encompassing quota struct that includes compute, storage, network, etc.

openstack/blockstorage/extensions/quotasets/testing/requests_test.go
```diff
@@ -10,13 +11,20 @@ import (
 
 var emptyQuotaSet = quotasets.QuotaSet{}
 var emptyQuotaUsageSet = quotasets.QuotaUsageSet{}
+var emptyUpdateOpts = quotasets.UpdateOpts{}
 
-func testSuccessTestCase(t *testing.T, httpMethod, uriPath, jsonBody string, uriQueryParams map[string]string, expectedQuotaSet quotasets.QuotaSet, expectedQuotaUsageSet quotasets.QuotaUsageSet) error {
+func testSuccessTestCase(t *testing.T, httpMethod, uriPath, jsonBody string, uriQueryParams map[string]string, expectedQuotaSet quotasets.QuotaSet, expectedQuotaUsageSet quotasets.QuotaUsageSet, updateOpts quotasets.UpdateOpts) error {
```
jtopjian: This might be getting a little unwieldy. I wonder if it would be more readable if each test case had its own function? I'm fine with either direction.

openstack/blockstorage/extensions/quotasets/requests.go
```diff
@@ -24,3 +24,61 @@ func GetUsage(client *gophercloud.ServiceClient, projectID string) (r GetUsageRe
 	_, r.Err = client.Get(u, &r.Body, nil)
 	return
 }
+
+// Updates the quotas for the given projectID and returns the new QuotaSet.
+func Update(client *gophercloud.ServiceClient, projectID string, opts UpdateOptsBuilder) (res UpdateResult) {
```
jrperritt: Again, the variable names need to adhere to the style guide: s/res/r

jtopjian: Thanks for catching that -- I missed it.

Self-note that this needs updated here, too: https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/quotasets/requests.go#L22

openstack/blockstorage/extensions/quotasets/testing/requests_test.go
waynr: I prefer this approach as it is (in my opinion) significantly better than the approach in the compute quotasets that I originally tried to copy. I found myself unable to really understand what was going on since it makes use of a nonsensical, inconsistent naming of fixture variables despite the ostensibly easier-to-understand breakdown of test cases into individual functions. I recall being convinced a couple weeks ago that the compute quotaset unit tests are not testing what the function names are claiming that they are testing because of erroneous re-use of fixture variables between test cases.

I could see breaking this function down into smaller functions and pushing the logic for which type of test case function needs to be invoked into `testSuccessTestCases`, but if you could go either way then I'd rather focus on higher priority work.

waynr:
I think this is ready for review again!

openstack/blockstorage/extensions/quotasets/testing/requests_test.go
jrperritt: I agree with @jtopjian : at the very least, this function has too many parameters. I'm also finding it difficult to track the fields of `successTestCases`. I agree the naming in the compute `quotasets` package could be improved, but I'd prefer if we kept the same testing format as in other packages.

waynr:
this is ready for review again

waynr:
I don't know what `NODE_FAILURE` means and none of the test result links take me anywhere meaningful.

jtopjian:
recheck

jtopjian:
OpenLab might be having some issues at the moment. I wouldn't worry too much about it - I've already verified the acceptance test previously.

animationzl:
recheck

PR #881, opened 2018-04-03
Blockstorage v3 quota-set support - part 5, Delete

waynr:
For #234
Pending #880

\## API:

\## Schema:

```
```

list of quotas.

\## Update:

\## Delete:


jtopjian:
@waynr If you'd like, I can take over this PR to finish this feature set.

jtopjian:
I've merged this in #1124. Thank you for your work with this!

PR #882, opened 2018-04-04
Identity V3: remove user from group

zhangzujian:
For #874 

API reference:
[Identity API v3 - Remove user from group](https://developer.openstack.org/api-ref/identity/v3/#remove-user-from-group)

Python source code:
https://github.com/openstack/keystone/blob/a6adc72e3e7ff4ff0bff0702e69ebd8f697d6261/keystone/identity/core.py#L1273
https://github.com/openstack/python-keystoneclient/blob/40642d597deecce4dd81428449353bbd0c2ad5be/keystoneclient/v3/users.py#L266

### DIFF
--- a/acceptance/openstack/identity/v3/users_test.go
+++ b/acceptance/openstack/identity/v3/users_test.go
@@ -196,7 +196,7 @@ func TestUsersListGroups(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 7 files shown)


zhangzujian:
recheck

zhangzujian:
Ready for review.

PR #887, opened 2018-04-04
Identity V3: Check Whether User Belongs to Group

zhangzujian:
For #886

API reference:
[Identity API v3 - Check whether user belongs to group](https://developer.openstack.org/api-ref/identity/v3/#check-whether-user-belongs-to-group)

Python source code:
https://github.com/openstack/keystone/blob/a6adc72e3e7ff4ff0bff0702e69ebd8f697d6261/keystone/identity/core.py#L1361
https://github.com/openstack/python-keystoneclient/blob/40642d597deecce4dd81428449353bbd0c2ad5be/keystoneclient/v3/users.py#L247

### DIFF
--- a/openstack/identity/v3/users/requests.go
+++ b/openstack/identity/v3/users/requests.go
@@ -1,6 +1,8 @@
 package users
 
 import (
+	"net/http"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/groups"
 	"github.com/gophercloud/gophercloud/openstack/identity/v3/projects"
@@ -266,12 +268,19 @@ func AddToGroup(client *gophercloud.ServiceClient, groupID, userID string) (r Ad
 	return
 }
 
-// CheckInGroup checks whether a user belongs to a group.
-func CheckInGroup(client *gophercloud.ServiceClient, groupID, userID string) (r CheckInGroupResult) {
-	url := checkInGroupURL(client, groupID, userID)
-	_, r.Err = client.Request("HEAD", url, &gophercloud.RequestOpts{
-		OkCodes: []int{204},
+// IsMemberOfGroup checks whether a user belongs to a group.
+func IsMemberOfGroup(client *gophercloud.ServiceClient, groupID, userID string) (r IsMemberOfGroupResult) {
+	url := isMemberOfGroupURL(client, groupID, userID)
+	var response *http.Response
+	response, r.Err = client.Request("HEAD", url, &gophercloud.RequestOpts{
+		OkCodes: []int{204, 404},
 	})
+	if r.Err == nil {
+		if (*response).StatusCode == 204 {
+			r.IsMember = true
+		}
+	}
+
 	return
 }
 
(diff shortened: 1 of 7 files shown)


zhangzujian:
Ready for review.

zhangzujian:
Ready for review.

openstack/identity/v3/users/results.go
```diff
@@ -116,6 +116,13 @@ type AddToGroupResult struct {
 	gophercloud.ErrResult
 }
 
+// IsMemberOfGroupResult is the response from a IsMemberOfGroup operation. Call its
+// ExtractErr method to determine if the request succeeded or failed.
+type IsMemberOfGroupResult struct {
+	IsMember bool
+	gophercloud.ErrResult
```
jtopjian: Let's use `gophercloud.Result` instead.

`ErrResult` is used exclusively with `ExtractErr()` and since that's no longer being done here, it would break the pattern.

openstack/identity/v3/users/results.go
```diff
@@ -116,6 +116,13 @@ type AddToGroupResult struct {
 	gophercloud.ErrResult
 }
 
+// IsMemberOfGroupResult is the response from a IsMemberOfGroup operation. Call its
+// ExtractErr method to determine if the request succeeded or failed.
```
jtopjian:
```go
// IsMemberOfGroupResult is the response from an IsMemberOfGroup operation. Call its
// Extract method to determine if the user belongs in the group or not.
```

zhangzujian:
@jtopjian  Ready for review.

jtopjian:
@zhangzujian This looks good to me.

@jrperritt Do you think the implementation here is OK? Is something similar implemented elsewhere in Gophercloud that I might have overlooked?

openstack/identity/v3/users/results.go
```diff
@@ -116,6 +116,13 @@ type AddToGroupResult struct {
 	gophercloud.ErrResult
 }
 
+// IsMemberOfGroupResult is the response from a IsMemberOfGroup operation. Call its
+// Extract method to determine if the request succeeded or failed.
+type IsMemberOfGroupResult struct {
+	IsMember bool
```
jrperritt: `isMember`

openstack/identity/v3/users/requests.go
```diff
@@ -266,6 +268,22 @@ func AddToGroup(client *gophercloud.ServiceClient, groupID, userID string) (r Ad
...
 }
 
+// IsMemberOfGroup checks whether a user belongs to a group.
+func IsMemberOfGroup(client *gophercloud.ServiceClient, groupID, userID string) (r IsMemberOfGroupResult) {
+	url := isMemberOfGroupURL(client, groupID, userID)
+	var response *http.Response
+	response, r.Err = client.Request("HEAD", url, &gophercloud.RequestOpts{
+		OkCodes: []int{204, 404},
+	})
+	if r.Err == nil {
```
jrperritt:
```go
&& response != nil
```

zhangzujian:
@jtopjian Ready for review.

PR #888, opened 2018-04-04
Senlin: Profiles List

JackKuei:

code in this PR:

[[profiles:list]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/profiles.py#L75)

### DIFF
--- a/openstack/clustering/v1/profiles/requests.go
+++ b/openstack/clustering/v1/profiles/requests.go
@@ -75,7 +75,7 @@ func (opts ListOpts) ToProfileListQuery() (string, error) {
 	return q.String(), err
 }
 
-// ListDetail instructs OpenStack to provide a list of profiles.
+// List instructs OpenStack to provide a list of profiles.
 func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pager {
 	url := listURL(client)
 	if opts != nil {
--- a/openstack/clustering/v1/profiles/requests.go
+++ b/openstack/clustering/v1/profiles/requests.go
@@ -61,7 +61,7 @@ type ListOptsBuilder interface {
 
 // ListOpts params
 type ListOpts struct {
-	GlobalProject bool   `q:"global_project"`
+	GlobalProject *bool  `q:"global_project"`
 	Limit         int    `q:"limit"`
 	Marker        string `q:"marker"`
 	Name          string `q:"name"`
--- a/openstack/clustering/v1/profiles/requests.go
+++ b/openstack/clustering/v1/profiles/requests.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
 )
 
 // CreateOptsBuilder for options used for creating a profile.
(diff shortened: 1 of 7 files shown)


openstack/clustering/v1/profiles/results.go
```diff
@@ -29,9 +29,18 @@ func (r commonResult) Extract() (*Profile, error) {
...
+		Profiles []Profile `json:"profiles"`
+	}
+	err := (r.(ProfilePage)).ExtractInto(&s)
+	return s.Profiles, err
+}
+
 type Spec struct {
 	Type       string                 `json:"type"`
-	Version    string                 `json:"version"`
+	Version    interface{}            `json:"version"`
```
jtopjian: I think this should be left as a string. If `Version` requires some smoothing over, then it should go in the `UnmarshalJSON` method similar to here:

https://github.com/gophercloud/gophercloud/blob/master/openstack/clustering/v1/policies/results.go

JackKuei: Not sure whats the best way to handle this one. Might talk to upstream about change it to a string. For policies, json is returning as "1.0" but profiles, json is returning as 1.0. Also when golang converts from float to string it becomes "1". Also hopefully profile spec version doesn't allow "1.0.1" in the future.

JackKuei: Created a bug for senlin to have json always returns string. But might be good idea for older senlin versions.

JackKuei:
Updated Spec Version handling. Created a bug for senlin to have json always output string for version. Good idea to keep converting from float to string?  But json's version 1.0 ==> "1" to golang.

openstack/clustering/v1/profiles/results.go
```diff
@@ -49,10 +59,27 @@ type Profile struct {
...
+}
+
+// IsEmpty determines if a ProfilePage contains any results.
+func (page ProfilePage) IsEmpty() (bool, error) {
+	profiles, err := ExtractProfiles(page)
+	return len(profiles) == 0, err
+}
+
 func (r *Profile) UnmarshalJSON(b []byte) error {
+	type tmpSpec Spec
```
jtopjian: This technically works, but for consistency, `Spec` should have its own `UnmarshalJSON` method. This is important for two reasons: if a future modification to `Spec` requires further unmarshalling, then the method will already be in place, and it prevents others from copying the style which might lead to bugs down the road.

jtopjian:
> Created a bug for senlin to have json always output string for version. 

Nice. I think that's the best long-term solution.

> Good idea to keep converting from float to string?

Yes, I think string is the best option given the number of variations. 

>  If so, json's version 1.0 ==> "1" for golang struct.

I think it would be OK to do a some kind of check for `1.0` and implement maybe `fmt.Sprintf` to get the decimal to show.

openstack/clustering/v1/profiles/requests.go
```diff
@@ -40,3 +41,40 @@ func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r Create
...
 }
+
+// ListOptsBuilder Builder.
+type ListOptsBuilder interface {
+	ToProfileListQuery() (string, error)
+}
+
+// ListOpts params
+type ListOpts struct {
+	GlobalProject bool   `q:"global_project"`
```
jtopjian: One additional comment: similar to #889, if `false` is a valid value to pass, then this should be `*bool`.

Issue #891, opened 2018-04-04
Fix Go Tip Formatting Failures


jtopjian:
Per https://travis-ci.org/gophercloud/gophercloud/jobs/362340957, Travis is giving errors about formatting with go tip.

jrperritt:
closed by #892

PR #892, opened 2018-04-04
Fix go tip gofmt errors

jtopjian:
For #891

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ install:
 - go get github.com/mattn/goveralls
 - go get golang.org/x/tools/cmd/goimports
 go:
-- 1.8
+- 1.10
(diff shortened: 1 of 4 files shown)


jtopjian:
You're kidding me...

jtopjian:
@jrperritt Ping for review. 

The first issue was that, beginning this afternoon, go tip started flagging two files as having a bad format. Upon changing them, the Go 1.8 Travis check flagged those same files for the same reason.

So I modified the `script/format` script to ignore those two files as well as left it open for adding more files in the future in case this keeps happening.

Then I decided to bump the minimum tested Go version in Travis to 1.10.

jtopjian:
@jrperritt Thanks for the quick review on this one :)

PR #895, opened 2018-04-05
Clustering Policy Create Implementation

dkt26111:
For #823 

code in this PR:

API doc:
https://developer.openstack.org/api-ref/clustering/#create-policy

API code:
https://github.com/openstack/senlin/blob/master/senlin/api/openstack/v1/policies.py#L57

### DIFF
--- a/openstack/clustering/v1/policies/requests.go
+++ b/openstack/clustering/v1/policies/requests.go
@@ -96,12 +96,12 @@ func (v *Version) UnmarshalJSON(b []byte) error {
 	}
 
 	if f == 1 {
-		blah := fmt.Sprintf("%.1f", f)
-		*v = Version(blah)
+		version := fmt.Sprintf("%.1f", f)
+		*v = Version(version)
 		return nil
 	} else {
-		blah := strconv.FormatFloat(f, 'f', -1, 64)
-		*v = Version(blah)
+		version := strconv.FormatFloat(f, 'f', -1, 64)
+		*v = Version(version)
 		return nil
 	}
 }
--- a/openstack/clustering/v1/policies/requests.go
+++ b/openstack/clustering/v1/policies/requests.go
@@ -1,10 +1,6 @@
 package policies
 
 import (
-	"encoding/json"
-	"fmt"
-	"strconv"
-
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
@@ -65,13 +61,6 @@ type CreateOpts struct {
 	Spec Spec   `json:"spec"`
 }
 
-type Spec struct {
-	Description string                 `json:"description"`
-	Properties  map[string]interface{} `json:"properties"`
-	Type        string                 `json:"type"`
-	Version     Version                `json:"version"`
-}
-
 // ToPolicyCreateMap formats a CreateOpts into a body map.
 func (opts CreateOpts) ToPolicyCreateMap() (map[string]interface{}, error) {
 	b, err := gophercloud.BuildRequestBody(opts, "")
@@ -80,30 +69,6 @@ func (opts CreateOpts) ToPolicyCreateMap() (map[string]interface{}, error) {
 	}
 
 	return map[string]interface{}{"policy": b}, nil
-}
-
-type Version string
-
-// custom unmarshal function to handle Version returned as either float64 or string
-func (v *Version) UnmarshalJSON(b []byte) error {
-	if b[0] == '"' {
-		return json.Unmarshal(b, (*string)(v))
-	}
-
-	var f float64
-	if err := json.Unmarshal(b, &f); err != nil {
-		return err
-	}
-
-	if f == 1 {
-		version := fmt.Sprintf("%.1f", f)
-		*v = Version(version)
-		return nil
-	} else {
-		version := strconv.FormatFloat(f, 'f', -1, 64)
-		*v = Version(version)
-		return nil
-	}
 }
 
 // Create makes a request against the API to create a policy
--- a/openstack/clustering/v1/policies/requests.go
+++ b/openstack/clustering/v1/policies/requests.go
@@ -54,3 +54,46 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 		return p
 	})
 }
+
+// CreateOpts params
+type CreateOpts struct {
+	Name string `json:"name"`
+	Spec Spec   `json:"-"`
+}
+
+type Spec struct {
+	Description string
+	Properties  map[string]interface{}
+	Type        string
+	Version     string
+}
+
+// ToPolicyCreateMap formats a CreateOpts into a body map.
+func (opts CreateOpts) ToPolicyCreateMap() (map[string]interface{}, error) {
+	b, err := gophercloud.BuildRequestBody(opts, "")
+	if err != nil {
+		return nil, err
+	}
+
+	spec := make(map[string]interface{})
+	spec["description"] = opts.Spec.Description
+	spec["type"] = opts.Spec.Type
+	spec["version"] = opts.Spec.Version
+	spec["properties"] = opts.Spec.Properties
+	b["spec"] = spec
+
+	return map[string]interface{}{"policy": b}, nil
+}
+
+// Create makes a request against the API to create a policy
+func Create(client *gophercloud.ServiceClient, opts CreateOpts) (r CreateResult) {
+	b, err := opts.ToPolicyCreateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Post(policyCreateURL(client), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{201},
+	})
+	return
+}
--- a/openstack/clustering/v1/policies/requests.go
+++ b/openstack/clustering/v1/policies/requests.go
@@ -1,6 +1,10 @@
 package policies
 
 import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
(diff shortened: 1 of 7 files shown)


dkt26111:
@jtopjian Another PR ready for review.  Thanks!

openstack/clustering/v1/policies/requests.go
```diff
@@ -54,3 +54,46 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 		return p
 	})
 }
+
+// CreateOpts params
+type CreateOpts struct {
+	Name string `json:"name"`
+	Spec Spec   `json:"-"`
```
jtopjian: 1.

```
Spec Spec   `json:"spec"`
```

openstack/clustering/v1/policies/requests.go
```diff
@@ -54,3 +54,46 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
+}
+
+type Spec struct {
+	Description string
+	Properties  map[string]interface{}
+	Type        string
+	Version     string
+}
```
jtopjian: 2. Add `json` tags to each of these fields

openstack/clustering/v1/policies/requests.go
```diff
@@ -54,3 +54,46 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
+	if err != nil {
+		return nil, err
+	}
+
+	spec := make(map[string]interface{})
+	spec["description"] = opts.Spec.Description
+	spec["type"] = opts.Spec.Type
+	spec["version"] = opts.Spec.Version
+	spec["properties"] = opts.Spec.Properties
+	b["spec"] = spec
```
jtopjian: 3. Comment this entire block out.

With 1, 2, and 3 in place, everything marshals correctly for me. Are there occurrences when that won't happen and the above method is needed?

openstack/clustering/v1/policies/results.go
```diff
@@ -72,3 +74,62 @@ func (r *Policy) UnmarshalJSON(b []byte) error {
...
+	}
+	t, err := time.Parse(RFC3339WithZ, s)
+	if err != nil {
+		return err
+	}
+	*jt = JSONRFC3339WithZ(t)
+	return nil
+}
+
+func (r *PolicyFromCreate) UnmarshalJSON(b []byte) error {
```
jtopjian: Instead of doing this, I came up with the following:

```go
func (r *Policy) UnmarshalJSON(b []byte) error {
  type tmp Policy
  var s struct {
    tmp
    CreatedAt string `json:"created_at,omitempty"`
    UpdatedAt string `json:"updated_at,omitempty"`
  }
  err := json.Unmarshal(b, &s)
  if err != nil {
    return err
  }
  *r = Policy(s.tmp)

  if s.CreatedAt != "" {
    r.CreatedAt, err = time.Parse(gophercloud.RFC3339MilliNoZ, s.CreatedAt)
    if err != nil {
      r.CreatedAt, err = time.Parse(RFC3339WithZ, s.CreatedAt)
      if err != nil {
        return err
      }
    }
  }

  if s.UpdatedAt != "" {
    r.UpdatedAt, err = time.Parse(gophercloud.RFC3339MilliNoZ, s.UpdatedAt)
    if err != nil {
      r.UpdatedAt, err = time.Parse(RFC3339WithZ, s.UpdatedAt)
      if err != nil {
        return err
      }
    }
  }

  return nil
}
```

This allows the same `Policy` result to be used, but will parse two different time formats.

@jrperritt is there a cleaner way of doing this?

openstack/clustering/v1/policies/requests.go
```diff
@@ -54,3 +54,46 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
 	})
 }
+
+// CreateOpts params
+type CreateOpts struct {
+	Name string `json:"name"`
+	Spec Spec   `json:"-"`
+}
+
+type Spec struct {
```
jtopjian: With this `Spec` type created, does it make sense to change the `Policy` struct's `Spec` field to be of type `Spec` rather than `map[string]interface{}`? Will a Spec ever return more fields than Description, Properties, Type, and Version?

openstack/clustering/v1/policies/requests.go
dkt26111: Done

openstack/clustering/v1/policies/requests.go
dkt26111: Done

openstack/clustering/v1/policies/requests.go
dkt26111: I changed it and It works correctly.

openstack/clustering/v1/policies/results.go
dkt26111: Done.

openstack/clustering/v1/policies/requests.go
dkt26111: Done.  The same Spec struct should be used for the request and return.

dkt26111:
@jtopjian This is ready for review.  I made the requested fix and also added a custom unmarshal for the version field since it could be returned as either a string or number.

openstack/clustering/v1/policies/requests.go
```diff
@@ -54,3 +58,63 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
 	})
 }
+
+// CreateOpts params
+type CreateOpts struct {
+	Name string `json:"name"`
+	Spec Spec   `json:"spec"`
+}
+
+type Spec struct {
```
jtopjian: Even though `Spec` is used in requests, because it's used in responses _and_ has a custom unmarshaller, it should go in `results.go`. That's more of an internal organization style: as soon as it's part of a result, it goes in `results.go`.

openstack/clustering/v1/policies/requests.go
```diff
@@ -54,3 +58,63 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
+	Description string                 `json:"description"`
+	Properties  map[string]interface{} `json:"properties"`
+	Type        string                 `json:"type"`
+	Version     Version                `json:"version"`
```
jtopjian: Let's change the type back to `string`.

openstack/clustering/v1/policies/requests.go
```diff
@@ -54,3 +58,63 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
+func (opts CreateOpts) ToPolicyCreateMap() (map[string]interface{}, error) {
+	b, err := gophercloud.BuildRequestBody(opts, "")
+	if err != nil {
+		return nil, err
+	}
+
+	return map[string]interface{}{"policy": b}, nil
+}
+
+type Version string
```
jtopjian: This can be removed

openstack/clustering/v1/policies/requests.go
```diff
@@ -54,3 +58,63 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
+
+// custom unmarshal function to handle Version returned as either float64 or string
+func (v *Version) UnmarshalJSON(b []byte) error {
```
jtopjian: This can be removed

openstack/clustering/v1/policies/results.go
```diff
@@ -54,21 +56,57 @@ func (r PolicyPage) LastMarker() (string, error) {
 	return policies[len(policies)-1].ID, nil
 }
 
+const RFC3339WithZ = "2006-01-02T15:04:05Z"
+
 func (r *Policy) UnmarshalJSON(b []byte) error {
 	type tmp Policy
 	var s struct {
```
jtopjian: We can handle the unmarshalling of `Version` here. Add it below `UpdatedAt` as:

```go
Version interface{} `json:"version"`
```

And then below the `s.UpdatedAt != ""`  block, you can add logic similar to here:

https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/floatingips/results.go#L42-L47

openstack/clustering/v1/policies/results.go
```diff
@@ -16,12 +16,14 @@ type Policy struct {
...
 	Name      string                 `json:"name"`
 	Project   string                 `json:"project"`
-	Spec      map[string]interface{} `json:"spec"`
+	Spec      Spec                   `json:"spec"`
 	Type      string                 `json:"type"`
 	UpdatedAt time.Time              `json:"-"`
 	User      string                 `json:"user"`
 }
 
+type PolicyFromCreate Policy
```
jtopjian: Now that `Policy` can handle formats returned both in requests and responses, `PolicyFromCreate` should no longer needed. If it still is, let me know. The goal is to ultimately remove this type.

openstack/clustering/v1/policies/requests.go
dkt26111: Done

openstack/clustering/v1/policies/requests.go
dkt26111: I couldn't get it to work with `Version` set to type `string` so I ended up changing it to `interface{}`.  See my comments below.

openstack/clustering/v1/policies/requests.go
dkt26111: Done

openstack/clustering/v1/policies/requests.go
dkt26111: Done

openstack/clustering/v1/policies/results.go
dkt26111: `Version` is nested inside the `Spec` struct so I couldn't define it after `UpdatedAt` inside `s`.
So I ended up defining `Version` as `interface{}` inside `Spec`.  Do you have any better idea?

openstack/clustering/v1/policies/results.go
dkt26111: Done

openstack/clustering/v1/policies/results.go
```diff
@@ -91,6 +98,17 @@ func (r *Policy) UnmarshalJSON(b []byte) error {
 		}
 	}
 
+	switch t := s.tmp.Spec.Version.(type) {
+	case float64:
+		if t == 1 {
+			r.Spec.Version = fmt.Sprintf("%.1f", t)
```
dkt26111: I need this logic here to make sure that a version == 1 is returned as string "1.0"

openstack/clustering/v1/policies/results.go
jtopjian: argh. That's my mistake. I got the two structs (Policy and Spec) mixed up.

We will need a second `UnmarshalJSON` method, but on `Spec` instead of `Version` as the initial implementation.

dkt26111:
@jtopjian : Ok, this is ready for another round of reviews ;)

jtopjian:
@dkt26111 This looks good to me - thank you very much for your patience with this one. I apologize for the confusion with the unmarshal methods.

PR #896, opened 2018-04-06
Add Scope to AuthOptions

jtopjian:
This commit adds a Scope field to the base AuthOptions struct. This
allows an explicit scope to be passed to AuthOptions which will
enable detailed scoping configurations.

For #252

### DIFF
--- a/auth_options.go
+++ b/auth_options.go
@@ -81,6 +81,17 @@ type AuthOptions struct {
 	// TokenID allows users to authenticate (possibly as another user) with an
 	// authentication token ID.
 	TokenID string `json:"-"`
+
+	// Scope determines the scoping of the authentication request.
+	Scope *AuthScope `json:"-"`
+}
+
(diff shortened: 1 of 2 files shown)


auth_options.go
```diff
@@ -81,6 +81,17 @@ type AuthOptions struct {
 	// TokenID allows users to authenticate (possibly as another user) with an
 	// authentication token ID.
 	TokenID string `json:"-"`
+
+	// Scope determines the scoping of the authentication request.
+	Scope *AuthScope `json:"-"`
+}
+
+// AuthScope allows a created token to be limited to a specific domain or project.
+type AuthScope struct {
```
jtopjian: I wasn't sure to call this `Scope` or `AuthScope` -- the latter being more descriptive for the package as a whole. But I opted to call it `Scope` in the `AuthOptions` struct since "auth" is implied there. I can keep it this way, change both to `Scope`.

jtopjian:
@jrperritt Thanks!

PR #905, opened 2018-04-08
LBaaS v2 l7 policy support - part 5: update l7policy

lingxiankong:
For #832

Neutron-LBaaS l7 policy update API implementation:
https://github.com/openstack/neutron-lbaas/blob/ac720b2a49720fb99e4189f93d5a83cfb295ccb3/neutron_lbaas/services/loadbalancer/plugin.py#L931

Octavia l7 policy update API implementation:
https://github.com/openstack/octavia/blob/06bf5c58d5845f684fcaf933605ed112586eefc3/octavia/api/v2/controllers/l7policy.py#L204

### DIFF
--- a/openstack/loadbalancer/v2/l7policies/requests.go
+++ b/openstack/loadbalancer/v2/l7policies/requests.go
@@ -143,3 +143,47 @@ func Delete(c *gophercloud.ServiceClient, id string) (r DeleteResult) {
 	_, r.Err = c.Delete(resourceURL(c, id), nil)
 	return
 }
+
+// UpdateOptsBuilder allows extensions to add additional parameters to the
+// Update request.
+type UpdateOptsBuilder interface {
+	ToL7PolicyUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts is the common options struct used in this package's Update
+// operation.
+type UpdateOpts struct {
+	// Name of the L7 policy.
+	Name string `json:"name,omitempty"`
+
+	// The L7 policy action. One of REDIRECT_TO_POOL, REDIRECT_TO_URL, or REJECT.
+	Action Action `json:"action,omitempty"`
+
+	// The position of this policy on the listener.
+	Position int32 `json:"position,omitempty"`
+
+	// A human-readable description for the resource.
+	Description string `json:"description,omitempty"`
+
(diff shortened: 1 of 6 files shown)


lingxiankong:
@jtopjian @jrperritt hi, update l7 policy is ready to review :-)

openstack/loadbalancer/v2/l7policies/requests.go
```diff
@@ -143,3 +143,47 @@ func Delete(c *gophercloud.ServiceClient, id string) (r DeleteResult) {
...
+// Update request.
+type UpdateOptsBuilder interface {
+	ToL7PolicyUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts is the common options struct used in this package's Update
+// operation.
+type UpdateOpts struct {
+	// Name of the L7 policy.
+	Name string `json:"name,omitempty"`
```
jtopjian: Is it possible to remove a name? If so, this should be `*string`. That will allow a value of `""` to be sent.

openstack/loadbalancer/v2/l7policies/requests.go
```diff
@@ -143,3 +143,47 @@ func Delete(c *gophercloud.ServiceClient, id string) (r DeleteResult) {
...
+
+	// The L7 policy action. One of REDIRECT_TO_POOL, REDIRECT_TO_URL, or REJECT.
+	Action Action `json:"action,omitempty"`
+
+	// The position of this policy on the listener.
+	Position int32 `json:"position,omitempty"`
+
+	// A human-readable description for the resource.
+	Description string `json:"description,omitempty"`
```
jtopjian: Is it possible to remove a description? If so, this should be `*string`.

openstack/loadbalancer/v2/l7policies/requests.go
lingxiankong: That's a valid question, @jtopjian I tested in my octavia environment, an empty string for both name and description are allowed to be specified to update l7 policy, i will update the PR

lingxiankong:
@jtopjian PR updated

jtopjian:
LGTM - thank you :)

The example in `doc.go` needs to be updated to reflect the pointer-to-string, but I'll take care of that.

Issue #906, opened 2018-04-08
Image Service v2: Incorrect Unmarshaling of Empty Images


jtopjian:
Unmarshaling empty images is currently broke because the method is returning early when `size_bytes` is set to null.

This prevents the rest of the unmarshaling from proceeding - notably handling of properties.

https://github.com/gophercloud/gophercloud/blob/8873eb140010928dfbfb1e4a7f9cadab94de3ce2/openstack/imageservice/v2/images/results.go#L105

PR #908, opened 2018-04-08
Image Service v2: Unmarshal Fix and Acc Test Updates

jtopjian:
For #818 
For #906


jtopjian:
@jrperritt Ping for review

PR #909, opened 2018-04-08
testing coverage increase

jtopjian:
Just playing with coveralls. Will open an issue if this proves worthwhile.


jtopjian:
As discovered in #912, there are some packages that are being excluded for coverage. This makes this PR rather moot since there might be other major areas that need better coverage.

Issue #910, opened 2018-04-08
Invalid required tags


jtopjian:
While poking around test coverage, I spotted two places where the "required" tag is used incorrectly. Both occurrences are with header opts. No query opts are being used incorrectly (or at all, for that matter).

jrperritt:
Reposting from the PR to here (which is probably the more correct place for discussion):

I'm fine with either being the format. The `tag,required` is more in agreement with the `json` package (and probably looks better), but then either the `required` label is position-dependent or you iterate over the comma-split slice looking for "required".

jtopjian:
IMO, I'm leaning on the non-comma required field for two reasons:

1. It's more commonly used right now so it'll be a larger change than the opposite.
2. It's more apparent that it's a tag specific go Gophercloud.

jrperritt:
Sounds good

jrperritt:
closed by #913

PR #911, opened 2018-04-08
Fix bad required tags

jtopjian:
For #910


jtopjian:
These are header opts :)

https://github.com/gophercloud/gophercloud/blob/master/params.go#L443

jrperritt:
OK, then the treatment of required fields between headers and query parameters should be reconciled (i.e. they should match).

jtopjian:
Sounds good to me. I was going to reconcile the behavior but saw that the current format (`tag,required`) was being tested in the unit test, so I figured that's how it was intended.

jrperritt:
I'm fine with either being the format. The `tag,required` is more in agreement with the `json` package (and probably looks better), but then either the `required` label is position-dependent or you iterate over the comma-split slice looking for "required".

PR #917, opened 2018-04-09
Clustering Policy Delete Implementation

dkt26111:
For #823

code in this PR:

API doc:
https://developer.openstack.org/api-ref/clustering/#delete-policy

API code:
https://github.com/openstack/senlin/blob/master/senlin/api/openstack/v1/policies.py#L89
https://github.com/openstack/senlin/blob/master/senlin/engine/service.py#L637

### DIFF
--- a/openstack/clustering/v1/policies/requests.go
+++ b/openstack/clustering/v1/policies/requests.go
@@ -1,9 +1,10 @@
 package policies
 
 import (
+	"net/http"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
-	"net/http"
 )
 
 // ListOptsBuilder Builder.
--- a/openstack/clustering/v1/policies/requests.go
+++ b/openstack/clustering/v1/policies/requests.go
@@ -87,7 +87,7 @@ func Create(client *gophercloud.ServiceClient, opts CreateOpts) (r CreateResult)
 }
 
 // Create makes a request against the API to delete a policy
-func Delete(client *gophercloud.ServiceClient, policyID string) (r PolicyDeleteResult) {
+func Delete(client *gophercloud.ServiceClient, policyID string) (r DeleteResult) {
 	var result *http.Response
 	result, r.Err = client.Delete(policyDeleteURL(client, policyID), &gophercloud.RequestOpts{
 		OkCodes: []int{204},
--- a/openstack/clustering/v1/policies/requests.go
+++ b/openstack/clustering/v1/policies/requests.go
@@ -3,6 +3,7 @@ package policies
 import (
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
+	"net/http"
 )
 
 // ListOptsBuilder Builder.
@@ -81,5 +82,15 @@ func Create(client *gophercloud.ServiceClient, opts CreateOpts) (r CreateResult)
 	_, r.Err = client.Post(policyCreateURL(client), b, &r.Body, &gophercloud.RequestOpts{
 		OkCodes: []int{201},
 	})
+	return
+}
+
+// Create makes a request against the API to delete a policy
+func Delete(client *gophercloud.ServiceClient, policyID string) (r PolicyDeleteResult) {
+	var result *http.Response
+	result, r.Err = client.Delete(policyDeleteURL(client, policyID), &gophercloud.RequestOpts{
+		OkCodes: []int{204},
+	})
+	r.Header = result.Header
 	return
 }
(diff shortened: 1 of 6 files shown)


dkt26111:
@jtopjian This one is already ready for review.  It adds the delete policy on top of the create policy PR.

openstack/clustering/v1/policies/results.go
```diff
@@ -54,21 +56,72 @@ func (r PolicyPage) LastMarker() (string, error) {
...
+	err := r.ExtractInto(&s)
+
+	return s.Policy, err
+}
+
+type CreateResult struct {
+	policyResult
+}
+
+type PolicyDeleteResult struct {
```
jtopjian: Let's call this `DeleteResult` since that's the standard name for results.

openstack/clustering/v1/policies/results.go
```diff
@@ -54,21 +56,72 @@ func (r PolicyPage) LastMarker() (string, error) {
...
+	gophercloud.HeaderResult
+}
+
+// DeleteResult contains the delete information from a delete policy request
+type DeleteResult struct {
```
jtopjian: Let's call this `DeleteHeader`. There's precedence for this name here: https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/containers/results.go#L294-L301

openstack/clustering/v1/policies/requests.go
```diff
@@ -118,3 +119,13 @@ func Create(client *gophercloud.ServiceClient, opts CreateOpts) (r CreateResult)
 	})
 	return
 }
+
+// Create makes a request against the API to delete a policy
+func Delete(client *gophercloud.ServiceClient, policyID string) (r PolicyDeleteResult) {
```
jtopjian: This function _should_ be able to be simplified a little:

https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/containers/requests.go#L115-L119

However, keep the `OkCodes` part. 

I haven't tested this myself, just going off of similar existing functions.

openstack/clustering/v1/policies/requests.go
```diff
@@ -54,3 +60,73 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
+		return
+	}
+	_, r.Err = client.Post(policyCreateURL(client), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{201},
+	})
+	return
+}
+
+// Create makes a request against the API to delete a policy
+func Delete(client *gophercloud.ServiceClient, policyID string) (r PolicyDeleteResult) {
```
jtopjian: https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/containers/requests.go#L115-L119

dkt26111: I'm trying it with the code you referenced, but the header is not populated.  It seems like you have to manually assign the r.Header like they do here:
https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/containers/requests.go#L187

jtopjian: That's interesting... I wonder if there's a bug in the containers Delete function then.

dkt26111:
@jtopjian Looks like it is showing unrelated changes from master after I did a rebase.  Let me know if this is too confusing to look at.  In that case I can close this one and create a new PR with just the senlin changes.

jtopjian:
@dkt26111 It's not so much the review but how Github will handle the merge.

I think I'm able to get a clean PR by doing the following:

```shell
$ git checkout master
$ git pull
$ git checkout clustering_policies_delete
$ git checkout -b clustering_policies_delete-tmp 
$ git rebase -i master

select the following commits:

* be669cad
* 9c27a830
* 4a7b108d
* be669cad

$ git rebase --continue
```

I think the result of that looks correct.

dkt26111:
Those are the same steps that I followed when I was rebasing.  But somehow it's showing the other changes as well in GitHub

openstack/clustering/v1/policies/requests.go
```diff
@@ -83,3 +85,13 @@ func Create(client *gophercloud.ServiceClient, opts CreateOpts) (r CreateResult)
 	})
 	return
 }
+
+// Create makes a request against the API to delete a policy
+func Delete(client *gophercloud.ServiceClient, policyID string) (r DeleteResult) {
```
dkt26111: I left the Delete function as is because the test case `th.AssertEquals(t, PolicyDeleteRequestID, actual.RequestID)`  kept failing when I tried it the simplified way.

jtopjian: That sounds good. I'll have to double-check the object storage implementation when I get a chance.

jtopjian:
@dkt26111 This looks good to me. Thank you again for your work with this :)

PR #922, opened 2018-04-10
Messaging Queue Acceptance Tests

problemv:
For #836

code in this PR:

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/queues.py#L43

### DIFF
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -500,6 +500,27 @@ func NewClusteringV1Client() (*gophercloud.ServiceClient, error) {
 	})
 }
 
+// NewMessagingV2Client returns a *ServiceClient for making calls
+// to the OpenStack Messaging (Zaqar) v2 API. An error will be returned
+// if authentication or client creation was not possible.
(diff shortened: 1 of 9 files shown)


jtopjian:
@problemv One additional change to the client function is to add:

```go
client = configureDebug(client)
```

See the other client functions and how it's used there.

This allows you to run the acceptance tests with `OS_DEBUG=1` set and you'll see the full API traffic. It's really helpful for debugging.

problemv:
@jtopjian I also cleaned up the commits for all of these, sorry I let that get out of hand. Too used to working in Gerrit... hopefully this new approach makes things easier for you.

jtopjian:
> FYI the Senlin client is missing it too.

Yeah, this feature was added around the same time Zaqar and Senlin PRs started, so it's understandable that they're missing. Not a big problem.

> I also cleaned up the commits for all of these, sorry I let that get out of hand

No problem at all :)

problemv:
@jtopjian this is ready for review again.

PR #924, opened 2018-04-11
Octavia l7 rule support - part 1: create

lingxiankong:
For #832

Octavia l7 rule create API implementation:
https://github.com/openstack/octavia/blob/06bf5c58d5845f684fcaf933605ed112586eefc3/octavia/api/v2/controllers/l7rule.py#L146

https://github.com/openstack/octavia/blob/86812e1309feb1bbfbc5448c12b80602d2d0a141/octavia/api/v1/types/l7rule.py#L31


lingxiankong:
@jtopjian @jrperritt hi this PR is ready for review, thanks!

lingxiankong:
yeah, I agree with you. Actually, I also noticed that, but just not sure if it's appropriate to fix all the related stuff in this series of patches. We should use project_id instead of tenant_id in `loadbalancer` repo. I will update this patch first and use 'project_id' in all following pathces, for all previous 'tenant_id', we'd better fix those in a separate pull request, what do you think?

jtopjian:
Ah, if LBaaS/Octavia has standardized on `project_id` throughout, then let's definitely rip out `tenant_id`. And yes, let's save that for a PR on its own.

Please proceed with the two changes to the result structs, though -- that way at least we have support for the populated field. Having extraneous fields can wait.

lingxiankong:
@jtopjian just double check, do you mean we will go with supporting both 'ProjectID' and 'TenantID' for now? So the code consumer can choose either one as needed?

lingxiankong:
think again, I changed to only support `project_id`. I think that's consistent with the main purpose of `loadbalancer` folder.

lingxiankong:
@jtopjian new patch submitted and CI passed.

jtopjian:
LGTM! Thank you very much :)

PR #930, opened 2018-04-13
Zun: fix Fatalf(): format %v reads arg #1, but call has only 0 args

jtopjian:
For #700

### DIFF
--- a/acceptance/openstack/identity/v3/endpoint_test.go
+++ b/acceptance/openstack/identity/v3/endpoint_test.go
@@ -15,7 +15,7 @@ import (
 func TestEndpointsList(t *testing.T) {
 	client, err := clients.NewIdentityV3Client()
 	if err != nil {
-		t.Fatalf("Unable to obtain an identity client: %v")
+		t.Fatalf("Unable to obtain an identity client: %v", err)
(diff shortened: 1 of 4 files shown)


kevin-zhaoshuai:
@jtopjian Another interesting thing I also meet is the same error with Travis-ci we meet here.
https://travis-ci.org/gophercloud/gophercloud/jobs/365920694
The gofmt check error.

jtopjian:
Ah, yes. That was fixed in https://github.com/gophercloud/gophercloud/pull/892. Unfortunately #892 also fixed some VPNaaS code that had the same problem, so I can't easily cherry-pick the commits of that PR.

Fortunately that's just a Travis failure. We should be able to still leverage OpenLab testing, though.

The OpenLab tests are showing some other failures that I'll try to fix.

jtopjian:
OK, that's going to be difficult, too, since the errors being reported were fixed in https://github.com/gophercloud/gophercloud/pull/820 which has 27 commits :)

I think it's safe to say that Option 2 described in #700 isn't going to be practical.

jtopjian:
Closing this in favor of #944

PR #933, opened 2018-04-16
Octavia l7 rule support - part 4: delete

lingxiankong:
For #832

Octavia l7 rule Delete API implementation:

- https://github.com/openstack/octavia/blob/69a45c254be854dbdbff946dbe2564711cdecec3/octavia/api/v2/controllers/l7rule.py#L228


lingxiankong:
@jtopjian @jrperritt sorry to bother you again(but i believe the l7 stuff of Octavia will finish soon) :-)

PR #936, opened 2018-04-16
Octavia l7 rule support - part 5(final): update

lingxiankong:
For #832

Octavia l7 rule PUT API implementation:

- https://github.com/openstack/octavia/blob/69a45c254be854dbdbff946dbe2564711cdecec3/octavia/api/v2/controllers/l7rule.py#L189

### DIFF
--- a/openstack/loadbalancer/v2/l7policies/requests.go
+++ b/openstack/loadbalancer/v2/l7policies/requests.go
@@ -279,3 +279,42 @@ func DeleteRule(c *gophercloud.ServiceClient, policyID string, ruleID string) (r
 	_, r.Err = c.Delete(ruleResourceURL(c, policyID, ruleID), nil)
 	return
 }
+
+// UpdateRuleOptsBuilder allows to add additional parameters to the PUT request.
+type UpdateRuleOptsBuilder interface {
+	ToRuleUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateRuleOpts is the common options struct used in this package's Update
+// operation.
+type UpdateRuleOpts struct {
+	// The L7 rule type. One of COOKIE, FILE_TYPE, HEADER, HOST_NAME, or PATH.
+	RuleType RuleType `json:"type,omitempty"`
+
+	// The comparison type for the L7 rule. One of CONTAINS, ENDS_WITH, EQUAL_TO, REGEX, or STARTS_WITH.
+	CompareType CompareType `json:"compare_type,omitempty"`
+
+	// The value to use for the comparison. For example, the file type to compare.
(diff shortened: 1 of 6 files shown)


lingxiankong:
@jtopjian @jrperritt the last one patch for review

jtopjian:
@lingxiankong It looks like `AdminStateUp` is a valid field:

https://github.com/openstack/octavia/blob/69a45c254be854dbdbff946dbe2564711cdecec3/octavia/api/v2/types/l7rule.py#L83

I reviewed https://github.com/gophercloud/gophercloud/pull/924 wondering how we missed that one earlier and saw that you linked to the `v1/types/l7rule.py` file. :)

https://github.com/openstack/octavia/blob/86812e1309feb1bbfbc5448c12b80602d2d0a141/octavia/api/v1/types/l7rule.py#L31

I think the addition of this field in both `CreateOpts` and `UpdateOpts` is small enough to add to this PR without causing too much complication. Or you can create a new PR for it -- your choice.

The field definition would be the same for both `CreateOpts` and `UpdateOpts`:

```go
AdminStateUp *bool `json:"admin_state_up,omitempty"`
```

Other than that, this looks good to me :)

jtopjian:
@lingxiankong Again, commented too early - sorry. It looks like `AdminStateUp` is missing for both Policy and Rule.

It's best to make a separate PR at this point. You can use #832 as the referenced issue -- no need to make a new issue.

The PR will simply include the missing `AdminStateUp` struct for the 4 structs.

openstack/loadbalancer/v2/l7policies/requests.go
```diff
@@ -279,3 +279,42 @@ func DeleteRule(c *gophercloud.ServiceClient, policyID string, ruleID string) (r
...
+}
+
+// ToRuleUpdateMap builds a request body from UpdateRuleOpts.
+func (opts UpdateRuleOpts) ToRuleUpdateMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "rule")
+}
+
+// UpdateRule allows Rule to be updated.
+func UpdateRule(c *gophercloud.ServiceClient, policyID string, ruleID string, opts UpdateRuleOptsBuilder) (r UpdateRuleResult) {
+	b, _ := opts.ToRuleUpdateMap()
```
jtopjian: I was about to merge this and caught this. I see it's also done with the `Update` function which I didn't catch earlier either.

Granted it's really hard to trigger an error with this, we should still check for an error regardless.

```go
b, err := opts.ToRuleUpdateMap()
if err != nil {
  r.Err = err
  return
}
```

jtopjian:
LGTM - thank you for the update :)

Issue #937, opened 2018-04-17
[Octavia] Can not list pools by listener_id


lingxiankong:
Octavia does not support to filter pools by listener_id.
```bash
$ http get "$PREFIX/pools?listener_id=a9dc29e2-4208-4bd8-8d3a-9bb9a4087d93" X-Auth-Token:$token
{
    "debuginfo": null,
    "faultcode": "Client",
    "faultstring": "One or more arguments are either duplicate or invalid"
}
```

jtopjian:
Fixed in #938

PR #938, opened 2018-04-17
Remove 'listener_id' from Pool ListOpts

lingxiankong:
For #937

Octavia does not support to filter pools by listener_id.

Pool database model definiation in Octavia:
https://github.com/openstack/octavia/blob/be505cdfba0543dc6d9722f5c8779e34de84fe57/octavia/db/models.py#L260

### DIFF
--- a/openstack/loadbalancer/v2/pools/requests.go
+++ b/openstack/loadbalancer/v2/pools/requests.go
@@ -25,7 +25,6 @@ type ListOpts struct {
 	Name           string `q:"name"`
 	ID             string `q:"id"`
 	LoadbalancerID string `q:"loadbalancer_id"`
-	ListenerID     string `q:"listener_id"`
 	Limit          int    `q:"limit"`
 	Marker         string `q:"marker"`
 	SortKey        string `q:"sort_key"`


jtopjian:
@lingxiankong Neutron listopts drive me crazy. I've rarely found clear code that defines what fields are possible to list with, so I've more or less taken the approach that if it's a valid field, it can be used to list resources unless proven otherwise.

The error you showed in #937 is good enough for me to merge this.

However, I'm wondering if you know of anywhere in the Octavia code that might define what is a legal list field?

jtopjian:
@lingxiankong Actually I might have commented too soon. The code reference you gave looks to be the right spot.

I started by searching for the error message you gave and found this:

https://github.com/openstack/octavia/blob/30bafb11d7eff23d826b39534e31acc249bfa1f3/octavia/common/exceptions.py#L89

which lead to this:

https://github.com/openstack/octavia/blob/649b33d2474b75a350c81b5538228826280424ef/octavia/api/common/pagination.py#L230

Which lead to this:

https://github.com/openstack/octavia/blob/649b33d2474b75a350c81b5538228826280424ef/octavia/api/common/pagination.py#L224-L225

So if my understanding is correct, then these are the valid fields to list with?

https://github.com/openstack/octavia/blob/be505cdfba0543dc6d9722f5c8779e34de84fe57/octavia/db/models.py#L269-L296

lingxiankong:
@jtopjian The octavia resource filtering functionality is implemented based on database models, so unfortunately, there is no doc about that. For pools, basically, the filtering fields are defined here: https://github.com/openstack/octavia/blob/be505cdfba0543dc6d9722f5c8779e34de84fe57/octavia/common/data_models.py#L249 but octaiva will do some transformation to a couple of fields such like 'enabled', 'loadbalancer_id', 'health_monitor', etc, and the code is spread in different places.

jtopjian:
@lingxiankong Thank you for the background info :)

This looks good to me.

Issue #931, opened 2018-04-13
Networking v2: Subnet CIDR is not a required field


jtopjian:
As reported in https://github.com/terraform-providers/terraform-provider-openstack/issues/282, specifying a CIDR is optional when using subnet pools:

https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/subnets/requests.go#L83

jrperritt:
closed by #940

jrperritt:
really closed by #943

PR #940, opened 2018-04-17
Networking v2: CIDR is not required for subnets

jtopjian:
For #931 

https://github.com/openstack/neutron-lib/blob/3ee0386c314b409c3f5969fdf0b463efa6cca660/neutron_lib/api/definitions/subnet.py#L58-L63


jtopjian:
recheck

jtopjian:
Thank you :)

PR #943, opened 2018-04-17
Networking v2: omitempty when CIDR not specified

jtopjian:
For #931 

Of course I should learn that there are no easy fixes :)

The only notable change here is:

https://github.com/gophercloud/gophercloud/compare/master...jtopjian:networkingv2-subnet-cidr-omitempty?expand=1#diff-204ba5cafb491fe49ec624dbd245dce3

The rest of the changes are supporting tests.


animationzl:
recheck

jrperritt:
No worries

jtopjian:
Thank you :)

Issue #947, opened 2018-04-19
Malformed v3 Endpoint


jtopjian:
As reported in https://github.com/terraform-providers/terraform-provider-openstack/issues/239, when a catalog is publishing an Identity v2 endpoint and the user creates a v3 client, the endpoint will be created as:

```
http://example.com:5000/v2.0/v3
```

I've confirmed this behavior in an environment publishing a v2.0 endpoint.

PR #948, opened 2018-04-19
Ensure IdentityV3 clients have the right endpoint

jtopjian:
This commit ensures IdentityV3 clients have the right endpoint
upon creation. This handles both versionless endpoints and
v2.0 endpoints being returned when doing an endpoint search.

For #947

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -2,10 +2,7 @@ package openstack
 
 import (
 	"fmt"
-	"net/url"
 	"reflect"
-	"regexp"
-	"strings"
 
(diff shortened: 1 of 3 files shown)


jtopjian:
@jrperritt Ping for review.

jtopjian:
@fatmcgav Thanks for the review.

I'm going to go ahead and merge this. We're up to 3 confirmed cases of this being a bug in Terraform.

I have a feeling that once we start looking at endpoint discovery for individual services, this might get further refactored to account for services other than identity.

PR #950, opened 2018-04-19
Add Create for messages

problemv:
For #836 

code in this PR:

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/messages.py#L170

### DIFF
--- a/acceptance/openstack/messaging/v2/message_test.go
+++ b/acceptance/openstack/messaging/v2/message_test.go
@@ -1,0 +1,23 @@
+// +build acceptance messaging messages
+
+package v2
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+)
+
+func TestCreateMessages(t *testing.T) {
+	clientID := "3381af92-2b9e-11e3-b191-71861300734c"
+
+	client, err := clients.NewMessagingV2Client(clientID)
+	if err != nil {
+		t.Fatalf("Unable to create a messaging service client: %v", err)
+	}
+
+	createdQueueName, err := CreateQueue(t, client)
+	defer DeleteQueue(t, client, createdQueueName)
+
+	CreateMessage(t, client, createdQueueName)
+}
--- a/openstack/messaging/v2/messages/requests.go
+++ b/openstack/messaging/v2/messages/requests.go
@@ -1,0 +1,58 @@
+package messages
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// CreateOptsBuilder Builder.
+type CreateOptsBuilder interface {
+	ToMessageCreateMap() (map[string]interface{}, error)
+}
+
+// BatchCreateOpts is an array of CreateOpts.
+type BatchCreateOpts []CreateOpts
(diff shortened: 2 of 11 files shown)


problemv:
@jtopjian Time to start messages :). Thanks for your help through all of this I really appreciate it.

acceptance/openstack/messaging/v2/message_test.go
```diff
@@ -0,0 +1,21 @@
+package v2
```
jtopjian: One thing I totally missed earlier is that a build tag and blank line need to be added at the top of acceptance tests:

Something like:

```go
// +build acceptance messaging messages

package v2
```

If you could update the `queue_test.go` file, it would save me a PR.

The `messaging.go` file does not need this.

The build tag is to exclude the acceptance tests from running during a normal `go test`. The script that Travis runs automatically excludes acceptance tests, but any other person who runs `go test ./...` would trigger the acceptance tests.

I apologize about not catching that before.

openstack/messaging/v2/messages/requests.go
```diff
@@ -0,0 +1,54 @@
...
+	Delay int `json:"delay,omitempty"`
+
+	// Body specifies an arbitrary document that constitutes the body of the message being sent.
+	Body map[string]interface{} `json:"body" required:"true"`
+}
+
+// CreateOpts params to be used with Create.
+type CreateOpts struct {
+	// QueueName specifies the name of the queue to create messages on.
+	QueueName string `json:"queue_name" required:"true"`
```
jtopjian: Totally up to you on this one:

Since there's a hard relationship between a message and a queue, _and_ the queue name is part of the URL, the queue name could go in the function signature:

```go
Create(client, queueName, createOpts)
```

The main reason I didn't like this for Queues was because there wasn't such a relationship.

If you wanted to make queue name part of the function, then the only remaining field in `CreateOpts` would be `Messages`. In that case, you could change this to the Batch pattern like in the queues update.

openstack/messaging/v2/messages/testing/requests_test.go
```diff
@@ -0,0 +1,41 @@
...
+				Body: map[string]interface{}{
+					"event":         "BackupProgress",
+					"current_bytes": "0",
+					"total_bytes":   "99614720",
+				},
+			},
+		},
+	}
+
+	blah := messages.Create(fake.ServiceClient(), createOpts)
```
jtopjian: `s/blah/result/` :)

Or you could do:

```go
actual, err := messages.Create(...).Extract()
```

problemv: I don't know how that got in there.... sorry about that.

problemv:
@jtopjian Changes implemented, I also cleaned up some comment issues that I saw in queue update.

openstack/messaging/v2/messages/testing/requests_test.go
jtopjian: No problem at all 😄

PR #951, opened 2018-04-19
Clustering Policy Validate Implementation

JackKuei:

code in this PR:

[[policies:validate]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/policies.py#L97)

### DIFF
--- a/openstack/clustering/v1/policies/results.go
+++ b/openstack/clustering/v1/policies/results.go
@@ -136,9 +136,6 @@ func (r policyResult) Extract() (*Policy, error) {
 		Policy *Policy `json:"policy"`
 	}
 	err := r.ExtractInto(&s)
-	if err != nil {
-		s.Policy = &Policy{}
-	}
 
 	return s.Policy, err
 }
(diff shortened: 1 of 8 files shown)


openstack/clustering/v1/policies/results.go
```diff
@@ -131,11 +131,14 @@ type policyResult struct {
...
 
-func (r CreateResult) Extract() (*Policy, error) {
+func (r policyResult) Extract() (*Policy, error) {
 	var s struct {
 		Policy *Policy `json:"policy"`
 	}
 	err := r.ExtractInto(&s)
+	if err != nil {
+		s.Policy = &Policy{}
+	}
```
jtopjian: Let's remove this block. Returning `nil` as a value for `*Policy` is fine.

jtopjian:
@JackKuei don't worry about the OpenLab failures. It looks like there's an upstream issue. As well, Senlin isn't being actively tested (though I have put a request in for some point in the future).

jtopjian:
LGTM - thank you!

PR #952, opened 2018-04-20
Clustering Policy Update Implementation

JackKuei:

code in this PR:

[[policies:update]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/policies.py#L74)

### DIFF
--- a/openstack/clustering/v1/policies/requests.go
+++ b/openstack/clustering/v1/policies/requests.go
@@ -90,13 +90,52 @@ func Create(client *gophercloud.ServiceClient, opts CreateOpts) (r CreateResult)
 	return
 }
 
-// Create makes a request against the API to delete a policy
+// Delete makes a request against the API to delete a policy
 func Delete(client *gophercloud.ServiceClient, policyID string) (r DeleteResult) {
 	var result *http.Response
 	result, r.Err = client.Delete(policyDeleteURL(client, policyID), &gophercloud.RequestOpts{
 		OkCodes: []int{204},
 	})
-	r.Header = result.Header
+	if r.Err == nil {
+		r.Header = result.Header
+	}
+	return
+}
+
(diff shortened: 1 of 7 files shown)


JackKuei:
recheck

jtopjian:
@JackKuei This will need rebased from the Validate merge.

openstack/clustering/v1/policies/requests.go
```diff
@@ -86,13 +86,52 @@ func Create(client *gophercloud.ServiceClient, opts CreateOpts) (r CreateResult)
...
+}
+
+// UpdateOptsBuilder builder
+type UpdateOptsBuilder interface {
+	ToPolicyUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts params
+type UpdateOpts struct {
+	Policy map[string]interface{} `json:"-"`
```
jtopjian: I don't think `Policy` is a valid field. Looking at the Senlin code, a Policy Update request is changed into a `PolicyUpdateRequestBody` object which only has a "name" field:

https://github.com/openstack/senlin/blob/a0f875b711d794629581c9643a8bc9594d1db4b3/senlin/objects/requests/policies.py#L58

In the API docs where it has "policy" as an object, I think it means the JSON body:

```json
"policy": {
  "name": "new-name"
}
```

which is taken care of on line 114.

According to this, too, "name" is the only field able to be updated: https://github.com/openstack/senlin/blob/a7fea71b9ee6de096349e1322400dbf84c075634/senlin/engine/service.py#L627-L638

It looks like the full policy (with the new name) is returned, though, so no changes needed there.

jtopjian:
LGTM - thank you :)

PR #953, opened 2018-04-20
Clustering v1: Fixing acceptance tests

jtopjian:
For #823


jtopjian:
I'm going to merge this even though OpenLab failed. I was watching the tests in action and they all passed. The error happened after the tests actually ran.

PR #944, opened 2018-04-18
Zun merge

jtopjian:
For #700 

Supersedes #929

PR #955, opened 2018-04-20
Add Zun Capsule List Operation

kevin-zhaoshuai:
For #700 

Depends-on: #944 

Reference:
https://github.com/openstack/zun/blob/master/zun/api/controllers/v1/capsules.py#L85

### DIFF
--- a/openstack/container/v1/capsules/requests.go
+++ b/openstack/container/v1/capsules/requests.go
@@ -2,6 +2,7 @@ package capsules
 
 import (
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
 )
 
 // CreateOptsBuilder is the interface options structs have to satisfy in order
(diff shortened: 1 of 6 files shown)


openstack/container/v1/capsules/requests.go
```diff
@@ -2,12 +2,51 @@ package capsules
...
 		OkCodes: []int{200, 203},
 	})
 	return
 }
+
+// ListOpts allows the filtering and sorting of paginated collections through
+// the API. Filtering is achieved by passing in struct field values that map to
+// the server attributes you want to see returned. Marker and Limit are used
+// for pagination.
+type ListOpts struct {
```
hongbin: It looks list supports an option 'all_projects'. It is of boolean type and (in default policy) only admin can set it to True.

kevin-zhaoshuai:
recheck

kevin-zhaoshuai:
recheck

kevin-zhaoshuai:
@jtopjian Ping for review. Fix this acceptance problem

Issue #903, opened 2018-04-06
Cinder Backup


egrh3:
It looks like `gophercloud` doesn't support `cinder-backup`? Is it on your roadmap?

jtopjian:
@egrh3 Do you mean the backup API actions?

https://developer.openstack.org/api-ref/block-storage/v3/index.html#backups-backups

I don't think anyone is actively working on it at the moment. You're more than welcome to work on it if you'd like. Or else I can mark this issue as "Help Wanted" for anyone else to.

egrh3:
Yes. Sorry, I could have made my initial question more explicit by referencing what I was asking for.

I may not have the time available to assist with this. I don't in the short-term ( next couple of weeks ). If you can tag this is "help wanted", that'd be great. I'll know more when I'm done with my current list of work / tasks.

jtopjian:
Sure, sounds good. There's no obligation at all.

jtopjian:
Implemented in #960 and #1552

PR #960, opened 2018-04-23
[wip] cinder backups

egrh3:
For #903 

Supporting OpenStack Code:
https://github.com/openstack/cinder/tree/master/cinder/backup
create: https://github.com/openstack/cinder/blob/master/cinder/backup/api.py#L199
list: https://github.com/openstack/cinder/blob/master/cinder/backup/api.py#L112

This is an initial request for pull. I've managed to list existing backups and create one via the new v2 and v3 clients / services.

Of note: the HTTP 400 I was receiving ( mentioned in commit message for 7e01987 ) was because cinder-backup returns a 400 when attempting to make a backup of an "in-use" volume without "force: true". The error message returned direct from Cinder in the response is: `"{"badRequest": {"message": "Invalid volume: Backing up an in-use volume must use the force flag.", "code": 400}}"`

### DIFF
--- a/openstack/blockstorage/v2/backups/doc.go
+++ b/openstack/blockstorage/v2/backups/doc.go
@@ -1,0 +1,5 @@
+// Package snapshots provides information and interaction with snapshots in the
+// OpenStack Block Storage service. A snapshot is a point in time copy of the
+// data contained in an external storage volume, and can be controlled
+// programmatically.
+package snapshots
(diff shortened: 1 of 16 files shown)


jtopjian:
@egrh3 Thank you for taking the time to work on this. This is marked as `wip` but I wanted to add a few comments early on.

Over all this looks good. The structure and code are well-formed. Nice work! :smiley: I haven't done an exhaustive review, yet, though.

We recently made a [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial) for new contributors. If you haven't read it, it has some helpful pointers:

That this feature will need to be broken up into multiple PRs: one PR per "method" (Get, List, Create, etc). You're more than welcome to keep everything in a single PR for now so you can leverage the automated testing, but when it's time for review, new PRs will need to be made.

Additionally, there should be acceptance tests bundled with each PR. Let me know if you need any help with that area.

As well, let me know if you have any questions in general. Again, thank you for working on this :)

PR #961, opened 2018-04-23
Messaging Message Delete

problemv:
For #836 

code in this PR:

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/messages.py#L281..L336


openstack/messaging/v2/messages/requests.go
```diff
@@ -103,3 +103,38 @@ func Create(client *gophercloud.ServiceClient, queueName string, opts CreateOpts
...
+}
+
+// ToMessagesDeleteQuery formats a DeleteMessageOpts structure into a query string.
+func (opts DeleteMessageOpts) ToMessagesDeleteQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
+// DeleteMessages deletes multiple messages based off of ID or number of messages.
+func DeleteMessages(client *gophercloud.ServiceClient, queueName string, opts DeleteMessageOptsBuilder) (r DeleteResult) {
```
jtopjian: This could also be named `Delete` since "messages" is the focus of this package.

But if there will be future functions in this package that would make `Delete` ambiguous and confusing, then `DeleteMessages` is fine.

openstack/messaging/v2/messages/results.go
```diff
@@ -20,6 +20,12 @@ type MessagePage struct {
 	pagination.LinkedPageBase
 }
 
+// DeleteResult is the result from a Delete operation. Call its ExtractErr
+// method to determine if the call succeeded or failed.
+type DeleteResult struct {
+	gophercloud.ErrResult
```
jtopjian: It looks like pop "pops" messages off of the queue and returns them?

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/messages.py#L320

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/messages.py#L331-L336

If so, there are two ways to approach this:

1. Change this to `gophercloud.Result` so the delete action acts just like the others which return messages:

```go
poppedMessages, err := messages.Delete(client, ...).Extract()
```

2. Keep `DeleteResult` as `gophercloud.ErrResult`, remove `Pop` from the `DeleteOpts`, and create a dedicated `messages.Pop` function. This function would have a `PopOpts` struct with the single field "IDs". The result would be `gophercloud.Result` and work as described above.

Thoughts?

openstack/messaging/v2/messages/requests.go
problemv: Yeah, thats why I had the distinction. Later there will be a "DeleteMessage" function.

openstack/messaging/v2/messages/results.go
problemv: @jtopjian Yeah you are right about this. I need to stop thinking the documentation is truth when it comes to Zaqar -.- .

"This operation does not accept a request body and does not return a response body."

But judging by the code it does actually return a response body. I believe implementing a new function called PopMessages is the best option. Do you want that in a separate PR or is this one fine to include that?

problemv:
@jtopjian I included them together for now. Let me know if you want me to split Pop and Delete up.

problemv:
@jtopjian I opened a bug report with Zaqar and submitted a PR to fix the docs thanks for rooting this out with me :).

Issue #963, opened 2018-04-25
Octavia loadbalancer implementation incorrectly uses "tenant_id" rather than "project_id"


waynr:
For reference in the Octavia source code see:
https://github.com/openstack/octavia/blob/54a4cf00cf304b108013ca4487c138c13f30983d/octavia/api/v2/types/load_balancer.py#L45

For reference in the Octavia docs see:
https://developer.openstack.org/api-ref/load-balancer/v2/index.html#show-load-balancer-details

I can also verify from experience that https://github.com/gophercloud/gophercloud/blob/e25975f29734719dc6071b3683ca40dbb1937cc1/openstack/loadbalancer/v2/loadbalancers/results.go#L21 always ends up being an empty string and changing `tenant_id` to `project_id` fixes it in my vendored gophercloud source.

jtopjian:
@waynr Thanks! We noticed this while we were adding the L7Policy stuff. I think @lingxiankong was going to follow-up with a patch. @lingxiankong would you still be able to work on this?

lingxiankong:
@jtopjian yeah, sure, it's still on my TODO list, sorry for some delay as I am working with openlab guys to fix some issue I met with the octavia-ingress-controller. 

@waynr please let me know if you are going to fix by yourself, so we won't duplicate effort :-)

jtopjian:
> please let me know if you are going to fix by yourself, so we won't duplicate effort :-)

+ 1

No worries if you don't have time, the above quote is exactly what I was more concerned about. If no one has time, I don't mind doing it, either. :)

lingxiankong:
@jtopjian let's wait for @waynr's answer, I can definitely do that if he is unavailable.

waynr:
I'll let someone else take this one, there are a couple other project/tenant id related issues i want to file coming out of the same work that led me to discover this one.

lingxiankong:
Then I'll take this one.

jtopjian:
Fixed in #964

PR #964, opened 2018-04-27
Use 'project_id' for octavia resources

lingxiankong:
For #963

According to Octavia API document
https://developer.openstack.org/api-ref/load-balancer/v2/index.html
'tenant_id' is no longer supported.


lingxiankong:
@jtopjian this patch is ready for review

Issue #970, opened 2018-05-03
Container v1: Move testing files to testing directory


jtopjian:
I didn't catch in #954 that some testing files were not in the `testing` directory. This is a side-effect of the orchestration testing files not being contained in the `testing` directory. However, since it's possible to easily clean this up, it should.

jtopjian:
Fixed in #971

PR #971, opened 2018-05-03
Container v1: Move testing files to testing directory

jtopjian:
For #970 

/cc @kevin-zhaoshuai


kevin-zhaoshuai:
@jtopjian  +1 Thanks for tweaking code.

Issue #972, opened 2018-05-03
Identity V3: Policies CRUD


zhangzujian:
- [x] List (#973)
- [x] Create (#981)
- [x] Get (#1000)
- [x] Update (#1001)
- [x] Delete (#999)

API reference:
[Identity API v3 - Policies](https://developer.openstack.org/api-ref/identity/v3/#policies)

Python source code:

List:
https://github.com/openstack/keystone/blob/master/keystone/policy/controllers.py#L51
https://github.com/openstack/keystone/blob/master/keystone/policy/core.py#L57
https://github.com/openstack/python-keystoneclient/blob/master/keystoneclient/v3/policies.py#L82

Create:
https://github.com/openstack/keystone/blob/master/keystone/policy/controllers.py#L41
https://github.com/openstack/keystone/blob/master/keystone/policy/core.py#L42
https://github.com/openstack/python-keystoneclient/blob/master/keystoneclient/v3/policies.py#L52

Get:
https://github.com/openstack/keystone/blob/master/keystone/policy/controllers.py#L59
https://github.com/openstack/keystone/blob/master/keystone/policy/core.py#L47
https://github.com/openstack/python-keystoneclient/blob/master/keystoneclient/v3/policies.py#L69

Update:
https://github.com/openstack/keystone/blob/master/keystone/policy/controllers.py#L65
https://github.com/openstack/keystone/blob/master/keystone/policy/core.py#L50
https://github.com/openstack/python-keystoneclient/blob/master/keystoneclient/v3/policies.py#L94

Delete:
https://github.com/openstack/keystone/blob/master/keystone/policy/controllers.py#L74
https://github.com/openstack/keystone/blob/master/keystone/policy/core.py#L64
https://github.com/openstack/python-keystoneclient/blob/master/keystoneclient/v3/policies.py#L114

jtopjian:
@zhangzujian If you plan on implementing all of Policies, you can modify this issue to house all PRs so you don't have to open a new issue for each related PR.

See https://github.com/gophercloud/gophercloud/issues/539 for a possible template to use.

jtopjian:
@zhangzujian Thank you for implementing this :)

:tada:

PR #973, opened 2018-05-03
Identity V3: List Policies

zhangzujian:
For #972 

API reference:
[Identity API v3 - List Policies](https://developer.openstack.org/api-ref/identity/v3/#list-policies)

Python source code:
https://github.com/openstack/keystone/blob/master/keystone/policy/routers.py#L21
https://github.com/openstack/keystone/blob/master/keystone/policy/controllers.py#L51
https://github.com/openstack/keystone/blob/master/keystone/policy/core.py#L57
https://github.com/openstack/python-keystoneclient/blob/master/keystoneclient/v3/policies.py#L82

Acceptance testing will be added after other policy APIs are supported.

### DIFF
--- a/acceptance/openstack/identity/v3/policies_test.go
+++ b/acceptance/openstack/identity/v3/policies_test.go
@@ -1,0 +1,29 @@
+// +build acceptance
+
+package v3
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/policies"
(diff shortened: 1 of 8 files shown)


zhangzujian:
Ready for review.

jtopjian:
@zhangzujian This looks perfect - nice work :)

Would it be possible to include an acceptance test?

jtopjian:
err - sorry, I just read this note:

> Acceptance testing will be added after other policy APIs are supported.

It's OK if no policies are returned. The test will simply verify the List call works without error. Later when it's possible to create policies, we can redo the List test.

jtopjian:
@zhangzujian Thank you :)

I'll wait until the OpenLab tests are done. If it fails, that's OK -- it's a separate issue. I've tested this myself and it works.

PR #979, opened 2018-05-03
Messaging Message GetMessages

problemv:
For #836

code in this PR:

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/messages.py#L254

### DIFF
--- a/openstack/messaging/v2/messages/requests.go
+++ b/openstack/messaging/v2/messages/requests.go
@@ -174,3 +174,37 @@ func PopMessages(client *gophercloud.ServiceClient, queueName string, opts PopMe
 	r.Err = err
 	return
 }
+
+// GetMessagesOptsBuilder allows extensions to add additional parameters to the
+// GetMessages request.
+type GetMessagesOptsBuilder interface {
+	ToGetMessagesListQuery() (string, error)
+}
+
+// GetMessagesOpts params to be used with GetMessages.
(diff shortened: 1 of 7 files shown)


problemv:
@jtopjian Sorry about the small hiatus. This is ready for review. :)

Quick note: There is a follow up function just for get which is to get a single message hence this function is called GetMessages.

openstack/messaging/v2/messages/requests.go
```diff
@@ -174,3 +174,37 @@ func PopMessages(client *gophercloud.ServiceClient, queueName string, opts PopMe
...
+
+// GetMessagesOptsBuilder allows extensions to add additional parameters to the
+// GetMessages request.
+type GetMessagesOptsBuilder interface {
+	ToGetMessagesListQuery() (string, error)
+}
+
+// GetMessagesOpts params to be used with GetMessages.
+type GetMessagesOpts struct {
+	Ids []string `q:"ids,omitempty"`
```
jtopjian: Let's use IDs here to match the `Delete` function.

problemv:
@jtopjian No problem my bad! fixed :)

jtopjian:
LGTM - thanks! 

The OpenLab failure is ignorable -- it's a separate issue being worked on.

Issue #978, opened 2018-05-03
Question: Should We Support Additional Filters for LIST APIs in Identity V3 ?


zhangzujian:
In Identity V3, **LIST** APIs support filters like **name_contains=STR**. If the filter is added into query parameters, the server will return a list of resources of which the names contain **STR**.

This feature is really useful, I think. Should we support it ?

Python source code:
https://github.com/openstack/keystone/blob/master/keystone/common/controller.py#L421

jtopjian:
This does seem useful. Unfortunately it looks like it doesn't work on Extra properties, though, which is disappointing. Otherwise you could do a query such as:

```
http://localhost:5000/v3/projects?email__contains=foo
```

So it seems filters only work on certain fields.

Given that, though, I think this is something we could/should implement. Perhaps adding the following to `ListOpts`:

```go
Filter map[string]string `q:"-"`
```

You would then iterate over the map and append the key/values to the URL during the `ToXListQuery` method.

The reason I think we should use `map[string]string` is because there are far too many combinations of filters to create a structured type with.

Thoughts?

zhangzujian:
@jtopjian Agreed.

jtopjian:
@zhangzujian I wasn't sure if everything was finished here.

Thank you very much for implementing this!

PR #980, opened 2018-05-04
Identity V3: Add Filters into Struct users.ListOpts

zhangzujian:
For #978

### DIFF
--- a/acceptance/openstack/identity/v3/users_test.go
+++ b/acceptance/openstack/identity/v3/users_test.go
@@ -43,7 +43,7 @@ func TestUsersList(t *testing.T) {
 	th.AssertEquals(t, found, true)
 
 	listOpts.Filters = map[string]string{
-		"name_contains": "dmi",
+		"name__contains": "dmi",
 	}
 
 	allPages, err = users.List(client, listOpts).AllPages()
@@ -65,7 +65,7 @@ func TestUsersList(t *testing.T) {
 	th.AssertEquals(t, found, true)
 
 	listOpts.Filters = map[string]string{
-		"name_contains": "foo",
+		"name__contains": "foo",
 	}
 
 	allPages, err = users.List(client, listOpts).AllPages()
--- a/acceptance/openstack/identity/v3/users_test.go
+++ b/acceptance/openstack/identity/v3/users_test.go
@@ -41,6 +41,50 @@ func TestUsersList(t *testing.T) {
 	}
 
 	th.AssertEquals(t, found, true)
+
+	listOpts.Filters = map[string]string{
+		"name_contains": "dmi",
+	}
+
+	allPages, err = users.List(client, listOpts).AllPages()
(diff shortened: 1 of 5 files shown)


acceptance/openstack/identity/v3/users_test.go
```diff
@@ -41,6 +41,50 @@ func TestUsersList(t *testing.T) {
 	}
 
 	th.AssertEquals(t, found, true)
+
+	listOpts.Filters = map[string]string{
+		"name_contains": "dmi",
```
jtopjian: I think it's two underscores:

```
name__contains
```

params.go
```diff
@@ -377,6 +377,25 @@ func BuildQueryString(opts interface{}) (*url.URL, error) {
...
+	u, err := BuildQueryString(opts)
+	if err != nil {
+		return nil, err
+	}
+
+	params := u.Query()
+	for k, v := range filters {
+		params.Add(k, v)
+	}
+	return &url.URL{RawQuery: params.Encode()}, nil
```
jtopjian: I realize this function will reduce the amount of code in the identity package, but since it's not used outside of identity, it's better to modify _each_ `ToXListQuery` function.

This is already done in a handful of other Gophercloud packages. 

See here for a simple example: https://github.com/gophercloud/gophercloud/blob/master/openstack/clustering/v1/webhooks/requests.go#L22-L32

zhangzujian:
Ready for review.

zhangzujian:
recheck

openstack/identity/v3/users/requests.go
```diff
@@ -49,11 +50,22 @@ type ListOpts struct {
...
 	UniqueID string `q:"unique_id"`
+
+	// Filters filters the response by custom filters such as
+	// 'name__contains=foo'
+	Filters map[string]string `q:"-"`
 }
 
 // ToUserListQuery formats a ListOpts into a query string.
 func (opts ListOpts) ToUserListQuery() (string, error) {
 	q, err := gophercloud.BuildQueryString(opts)
```
jtopjian: let's do an error check here.

openstack/identity/v3/users/requests.go
```diff
@@ -49,11 +50,22 @@ type ListOpts struct {
...
+
+	params := q.Query()
+	for k, v := range opts.Filters {
+		params.Add(k, v)
```
jtopjian: At a minimum, the filter must have two underscores, so maybe we should do a string check to make sure each key has two underscores? Otherwise `Filter` could be used to pass any arbitrary key into the query. Thoughts?

openstack/identity/v3/users/requests.go
zhangzujian: You are right. Error check is necessary.

zhangzujian:
@jtopjian Please review.

jtopjian:
@zhangzujian LGTM!

PR #976, opened 2018-05-03
Move CI job definition to gophercloud repo

yongfengdu:
Add the jobs definition to .zuul.yaml, and the ansible playbooks.

For #974

PR #981, opened 2018-05-04
Identity V3: Create Policy

zhangzujian:
For #972

### DIFF
--- a/openstack/identity/v3/policies/requests.go
+++ b/openstack/identity/v3/policies/requests.go
@@ -24,7 +24,7 @@ func (opts ListOpts) ToPolicyListQuery() (string, error) {
 	return q.String(), err
 }
 
-// List enumerates the roles to which the current token has access.
+// List enumerates the policies to which the current token has access.
 func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pager {
 	url := listURL(client)
 	if opts != nil {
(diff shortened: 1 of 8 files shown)


zhangzujian:
recheck

zhangzujian:
recheck

zhangzujian:
Ready for review.

openstack/identity/v3/policies/requests.go
```diff
@@ -39,3 +39,52 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
...
+	ToPolicyCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts provides options used to create a policy.
+type CreateOpts struct {
+	// Type is the MIME media type of the serialized policy blob.
+	Type string `json:"type" required:"true"`
+
+	// Blob is the policy rule as a serialized blob.
+	Blob string `json:"blob" required:"true"`
```
jtopjian: Orchestration and Container also have support for fields which take JSON or YAML structures and we use `[]byte` as the type. For consistency, let's change this to `[]byte`.

https://github.com/gophercloud/gophercloud/blob/master/openstack/container/v1/capsules/requests.go#L40-L52

I don't see a huge need to have a separate `Template` struct/field/type here. Mainly because I'm not sure all of the MIME types that the policy API supports and so adding validation and such might be difficult. But it's up to you if you want to go that route.

zhangzujian: The Keystone service does not check the validation of **blob** and **type**. Arbitrary string is accepted. The only limitation is the length of **type**.

https://github.com/openstack/keystone/blob/master/keystone/policy/schema.py#L14

So I think there is no need to add a **Template** type. Instead, we can check the length of **type**.

I agree to change type of **Blob** from string to []byte for consistency.

zhangzujian:
Ready for review.

zhangzujian:
@jtopjian 

I noticed the OpenLab checks (including ones triggered by **recheck**) didn't check the PR at all. Nor the HTML report or the output archives includes anything about the PR.

To verify the problem, I create a new PR #992 which contains some tests that won't pass. For now, the unit test running on OpenLab has passed. The acceptance test is still running and I think it will pass, too.

I also noticed PR #976 will be merged recently. It seems the PR will not fix this problem.

jtopjian:
@zhangzujian The tests run in #992 won't be run because only certain packages under `acceptance` are configured to run.

However, _all_ `identity/v3` tests should be running.

@yongfengdu @animationzl Do you folks know why `TestPoliciesCreate` isn't being run? The strange part is that `TestPoliciesList` (in the same file) is run.

jtopjian:
@yongfengdu @animationzl per review of #989, it looks like even modified tests aren't being run -- not just new tests. We need to have the tests from the PR branch run.

yongfengdu:
@jtopjian I've take a look at this job, maybe the go get ./... overwrites the PR with master? Will discuss with @animationzl for this offline and fix it.

jtopjian:
@yongfengdu Sounds good - thank you :)

It looks like the same is happening in Terraform. For example:

https://github.com/terraform-providers/terraform-provider-openstack/pull/314

That PR has two tests with the substring `TestAccNetworkingV2SubnetRoute` that should have been tested.

 I want to say this is a new problem because I think new tests used to run... but maybe I'm wrong?

yongfengdu:
recheck

zhangzujian:
recheck

liu-sheng:
recheck

liu-sheng:
recheck

zhangzujian:
@jtopjian ping for review

PR #987, opened 2018-05-04
Update subnets package to support removing all Host Routes from a sub…

### LINKED ISSUE #984: Unable to remove all Subnet HostRoutes
As observed whilst implementing https://github.com/terraform-providers/terraform-provider-openstack/pull/314, currently it is not possible to remove all `HostRoutes` from a subnet via Gophercloud. 

This is due to: https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/subnets/requests.go#L179, which prevents sending an empty `HostRoutes` value. 

The corresponding logic in the [routers](https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/layer3/routers/requests.go#L103) package does allow providing an empty `Routes` value. 

PR incoming to resolve.

fatmcgav:
…net. Add test coverage.

Fixes #984

### DIFF
--- a/openstack/networking/v2/subnets/requests.go
+++ b/openstack/networking/v2/subnets/requests.go
@@ -176,7 +176,7 @@ type UpdateOpts struct {
 	DNSNameservers []string `json:"dns_nameservers,omitempty"`
 
 	// HostRoutes are any static host routes to be set via DHCP.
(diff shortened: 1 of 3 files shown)


fatmcgav:
Right, think this is ready for review @jtopjian :)

PR #998, opened 2018-05-10
Add BlockStorage Client accepting micro versions

j-griffith:
This just adds a NewBlockStorage client setup that accepts and sets
microversions.

This doesn't necessarily solve the issue of micro version support, or expose anything new in the OpenStack API's.  What it does do however is let those that are interested and have a need to utilize a micro version of the Cinder API that isn't formally implemented in Gophercloud to roll their own solution.

An example usage can be found here:  
https://gist.github.com/j-griffith/0dee9a3c871a3e6988d11be530bbd9bb

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -342,6 +342,17 @@ func NewBlockStorageV3(client *gophercloud.ProviderClient, eo gophercloud.Endpoi
 	return initClientOpts(client, eo, "volumev3")
 }
 
+// NewBlockStorageMicroVersion creates a ServiceClient that may be used to access the v3 block storage service with microversions.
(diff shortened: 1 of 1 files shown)


jtopjian:
@j-griffith I apologize if I'm missing something, but isn't this the same as:

```go
client, err := gophercloud.NewBlockStorageV3(client, eo)
client.Microversion = mv
```

j-griffith:
Oh my goodness!  For some reason I completely confused the serviceclient and the providerclient!!!  I'm real sorry, but you're absolutely correct, everything needed is already in place.  Sorry again for wasting your time on that.

jtopjian:
There's absolutely nothing to apologize for. It took me a good 20 minutes to follow the code between provider client, service client, `initReqOpts`, and `initClientOpts`. And even after that, I genuinely considered I might still be missing something :)

j-griffith:
Baahhh!!  I owe ya for the wasted time!

Issue #1005, opened 2018-05-27
Auth from env: OS_TOKEN support


ozerovandrei:
Hello.

Right now Gophercloud doesn't support `OS_TOKEN` or `OS_AUTH_TOKEN` environment variable for `AuthOptionsFromEnv()` function.
Some of the projects that uses Gophercloud implemented this feature by itself: [terraform-provider-openstack#6081](https://github.com/hashicorp/terraform/pull/6081)

Is this a recommended way to deal with lack of `OS_TOKEN` environment variable support?

If not, can we implement it? It seems really easy to do, and I linked a PR for this issue.

Documentation (wiki) reference: https://wiki.openstack.org/wiki/OpenStackClient/Authentication
Code references are provided in the linked PR.

jtopjian:
At the moment I'm not keen to add further support to `AuthOptionsFromEnv`. I think it serves a basic purpose for use in testing and development, but I would prefer users to use `gophercloud/utils/clientconfig` for anything beyond testing.

ozerovandrei:
@jtopjian oh I see.
So, right now the most recommended way for external packages to build `AuthOptions` from yaml or env variables is to use [clientconfig/AuthOptions](https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/requests.go#L119)?

jtopjian:
@ozerovandrei correct :)

The key here is "external packages" - I'd like for applications which consume Gophercloud to start using `clientconfig` for authentication since that package will support all authentication options users/devs expect to see.

Granted, unless someone has browsed some existing issues here, they would not be aware of `clientconfig`. At some point I hope to get that documented soon.

ozerovandrei:
@jtopjian thank you for the clarification.

Yeah, some tips in documentation would be helpful.
Closing this issue then.

PR #1006, opened 2018-05-27
Auth from env: add support for OS_TOKEN variable

ozerovandrei:
This PR adds support for the `OS_TOKEN` and `OS_AUTH_TOKEN` environment
variables.
If user provides one of those variables, he doesn't need to use username and
password authentication.

For #1005 

code in this PR:

First, `python-openstackclient` specifies token option in the `TokenEndpoint` and uses `keystoneauth1.loading.Opt`:
https://github.com/openstack/python-openstackclient/blob/stable/queens/openstackclient/api/auth_plugin.py#L56

Second, this option transforms to `OS_TOKEN` in the `argparse_default()` method of the `keystoneauth1.loading.Opt` class:
https://github.com/openstack/keystoneauth/blob/stable/queens/keystoneauth1/loading/opts.py#L144

### DIFF
--- a/openstack/auth_env.go
+++ b/openstack/auth_env.go
@@ -38,6 +38,7 @@ func AuthOptionsFromEnv() (gophercloud.AuthOptions, error) {
 	tenantName := os.Getenv("OS_TENANT_NAME")
 	domainID := os.Getenv("OS_DOMAIN_ID")
 	domainName := os.Getenv("OS_DOMAIN_NAME")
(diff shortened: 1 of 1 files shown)


ozerovandrei:
Closing because it's already implemented in [gophercloud/utils](https://github.com/gophercloud/utils/blob/master/openstack/clientconfig/requests.go#L119) repository.

External packages should use it to build `AuthOptions`.

Issue #1009, opened 2018-05-29
Add an external network filter option


kayrus:
I have a use case that requires to filter out external networks from network requests. It could be done by adding the following option:

* [router:external](https://github.com/openstack/neutron-lib/blob/f42bcbcb3fdcfa9cb3d0ffe1bba810a317ba3c9b/neutron_lib/api/definitions/external_net.py#L21)

[Reference](https://specs.openstack.org/openstack/neutron-specs/specs/api/external_networks__external-net_.html)

I found [this code](https://github.com/gophercloud/gophercloud/blob/c53578fc5176a5508a400609e00cd669d412f45c/openstack/networking/v2/extensions/external/requests.go), but it looks like it is applicable only when you modify the network.

jtopjian:
Thanks for digging into this.  :smile: 

> I found this code, but it looks like it is applicable only when you modify the network.

Good find. This is the right file we need to modify.

Neutron's heavy use of extensions makes working with Gophercloud _very_ verbose. Both Neutron and Gophercloud have legitimate design decisions, but the two together can make even the smallest change require a good amount of work.

This isn't something I'd expect someone new to Gophercloud to be able to easily figure out, so I created #1010 which implements this.

#1010 also includes an amended doc example and an acceptance test on how to use this.

It's possible to create some code to cover this verbosity up and make it more easier to work with. I have this somewhere on my todo list.

Please let me know if you have any questions. :)

jtopjian:
Implemented in #1010

PR #1010, opened 2018-05-30
Networking v2: Add ability to list external networks

jtopjian:
This commit adds the ability to list external networks. It does this by
way of creating an extention to the networks package's ListOpts struct.

For #1009

### DIFF
--- a/acceptance/openstack/networking/v2/networks_test.go
+++ b/acceptance/openstack/networking/v2/networks_test.go
@@ -7,8 +7,10 @@ import (
 
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/external"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/portsecurity"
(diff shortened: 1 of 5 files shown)


jtopjian:
@yongfengdu @liu-sheng Sorry to bug you guys. Any idea why I can't see the test output in the `job-output.txt.gz` file any more?

jtopjian:
@yongfengdu @liu-sheng @animationzl I noticed that the stdout is in the JSON-formatted logfile -- it's just missing from the text formatted file. I can get by with reading the JSON output for now, but it would be nice to have the text-formatted output back :)

liu-sheng:
@jtopjian Hi Joe, there seems is a zuul bug, we have modified in CI environment, not sure it is the reason for this issue, let me recheck to see. sorry for reply lately.

liu-sheng:
recheck

liu-sheng:
Oops! the above result of acceptance `job-output.txt.gz` include many `Waiting on logger` errors, that because a security group configuration mistake, have been fixed. we can check other PR https://github.com/gophercloud/gophercloud/pull/1013 to confirm.

jtopjian:
@liu-sheng sounds good! thanks so much for looking into this -- I didn't think it was a late reply at all 😄

Issue #210, opened 2016-12-02
Can keypairs.Create().Extract().PrivateKey be a DER encoded key?


sb10:
I'm doing:

    kp, err = keypairs.Create(p.computeClient, keypairs.CreateOpts{Name: resources.ResourceName}).Extract()
    key, err := ssh.ParsePrivateKey([]byte(kp.PrivateKey))

This used to work fine, but now, after an upgrade in our OpenStack installation, I get:

> asn1: structure error: superfluous leading zeros in length

From reading https://github.com/mitchellh/packer/issues/2526 and https://github.com/golang/go/issues/14145, it seems OpenStack produces BER encoded keys, but Go's built-in parser only handles DER.

How are people currently coping with this, and could Gophercloud internally convert BER to DER to make life easier on end-users?

rickard-von-essen:
After a lengthy investigation the only viable solution (to not have to write a Go DER decoder) I found was to shell out and try to run OpenSSL to convert the key and get a BER encoded version.

The only inconvenience is on Windows (of course) where _openssl_ is not installed by default otherwise it's kind of transparent.

IMHO OpenStack should revert to only produce DER (which they did before _Liberty_).

For my workaround see https://github.com/mitchellh/packer/pull/3135

sb10:
Hmmm. My solution, and my recommendation for Gophercloud, is that when the user of Gophercloud supples keypairs.CreateOpts{...} without PublicKey specified, under the hood Gophercloud creates its own key and injects it:

                    // create a new keypair; we can't just let Openstack create one for
			// us because in latest versions it does not return a DER encoded
			// key, which is what GO built-in library supports.
			privateKey, errk := rsa.GenerateKey(rand.Reader, 1024)
			if errk != nil {
				err = errk
				return
			}
			privateKeyPEM := &pem.Block{Type: "RSA PRIVATE KEY", Bytes: x509.MarshalPKCS1PrivateKey(privateKey)}
			privateKeyPEMBytes := pem.EncodeToMemory(privateKeyPEM)
			pub, errk := ssh.NewPublicKey(&privateKey.PublicKey)
			if errk != nil {
				err = errk
				return err
			}
			publicKeyStr := ssh.MarshalAuthorizedKey(pub)
                        
                    // inject
                    userKeyparisCreateOpts.PublicKey = string(publicKeyStr)

                    // so that we end up doing the equivalent of:
[...]

PR #1013, opened 2018-06-04
Acc Tests: Parse the key returned by Nova

jtopjian:
For #210

### DIFF
--- a/acceptance/openstack/compute/v2/keypairs_test.go
+++ b/acceptance/openstack/compute/v2/keypairs_test.go
@@ -5,6 +5,8 @@ package v2
 import (
 	"testing"
 
+	"golang.org/x/crypto/ssh"
+
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/keypairs"
@@ -21,6 +23,12 @@ func TestKeypairsCreateDelete(t *testing.T) {
 	keyPair, err := CreateKeyPair(t, client)
 	th.AssertNoErr(t, err)
 	defer DeleteKeyPair(t, client, keyPair)
+
+	// There was a series of OpenStack releases, between Liberty and Ocata,
+	// where the returned SSH key was not parsable by Go.
+	// This checks if the issue is happening again.
+	_, err = ssh.ParsePrivateKey([]byte(keyPair.PrivateKey))
+	th.AssertNoErr(t, err)
 
 	tools.PrintResource(t, keyPair)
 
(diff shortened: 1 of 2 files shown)


jtopjian:
@liu-sheng job output looks good! Thank you!!

jtopjian:
recheck mitaka

jtopjian:
recheck/mitaka

jtopjian:
recheck stable/mitaka

jtopjian:
recheck stable/mitaka

jtopjian:
recheck stable/mitaka

liu-sheng:
recheck stable/mitaka

liu-sheng:
recheck stable/mitaka

jtopjian:
@liu-sheng I apologize for the late reply :frowning_face: 

Thank you for looking into this. I'd like to (eventually) get the failing tests ignored when testing against Mitaka (or any other older release). Is there an environment variable of some sort that I can use to skip certain tests? Similar to what's done here: https://github.com/gophercloud/gophercloud/blob/master/acceptance/clients/conditions.go

kiwik:
recheck stable/mitaka

liu-sheng:
@jtopjian hah, it is never late at all :smile: 

That seems is possible, but need to modify the job definition, we can add a var in the job definition, just like the `os_branch: stable/mitaka` in [here](https://github.com/gophercloud/gophercloud/blob/master/.zuul.yaml#L53), e.g. 'skip_tests: xxx' the variable can be passed to the [playbook](https://github.com/gophercloud/gophercloud/blob/master/.zuul/playbooks/gophercloud-acceptance-test/run.yaml) and need to be handled in the [script](https://github.com/gophercloud/gophercloud/blob/master/script/acceptancetest)

jtopjian:
@liu-sheng Rather than specifying individual tests in the job definition, is it possible to set `os_branch` in the environment? That way the acceptance tests can self-determine what release/branch is being run and skip dynamically. 

For example, in [`conditions.go`](https://github.com/gophercloud/gophercloud/blob/master/acceptance/clients/conditions.go), I could have:

```go
func SkipMitaka(t *testing.T) {
	if os.Getenv("OS_BRANCH") == "stable/mitaka" {
		t.Skip("this is not supported in mitaka")
	}
}

func SkipNewton(t *testing.T) {
	if os.Getenv("OS_BRANCH") == "stable/newton" {
		t.Skip("this is not supported in newton")
	}
}

```

And then for the breaking tests:

```go
func TestSomeNewFeature(t *testing.T) {
	clients.SkipMitaka(t)
	clients.SkipNewton(t)

        ...
}
```

liu-sheng:
@jtopjian that sounds easy to implement, will have a try :)

kiwik:
recheck stable/mitaka

kiwik:
recheck stable/mitaka

kiwik:
The latest failing is caused by client init failing for Container API, that is not supported in Mitaka release yet, so we can ignore it, that do not block this PR merging. And OpenLab is trying to make test case know which OpenStack release will be used in testing, still be in WIP, see https://github.com/theopenlab/openlab/issues/67 . If it is ready, we can decide which test cases will be executed in which OpenStack release, like @jtopjian say https://github.com/gophercloud/gophercloud/pull/1013#issuecomment-395633308

jtopjian:
recheck stable/mitaka

jtopjian:
This is working as intended. The test passes in stable and fails in Mitaka (which is expected).

PR #1014, opened 2018-06-04
DB v1: add support for fault information in response body

jtopjian:
For #224 
Amending #225


jtopjian:
AFAICT this is correct. I'm going to merge this.

Issue #1015, opened 2018-06-04
Creation of existing container leads to abandoned CLOSE_WAIT socket


awskii:
Hi.

When I send request to Create (put) container, which is already exists, i've got abandoned socket in close_wait state, which means, you don't send SYN to peer.

Proof:
```$ lsof -a -p 11520
11520 awskii    4u  a_inode   0,13        0     11625 [eventpoll]
11520 awskii    5u     IPv4 157447      0t0       TCP localhost:http-alt (LISTEN)
11520 awskii    6u     IPv6 159056      0t0       TCP *:iprop (LISTEN)
```
[create container with existing name via `containers.Create` method]
```
11520 awskii    4u  a_inode   0,13        0     11625 [eventpoll]
11520 awskii    5u     IPv4 157447      0t0       TCP localhost:http-alt (LISTEN)
11520 awskii    6u     IPv6 159056      0t0       TCP *:iprop (LISTEN)
11520 awskii    7u     IPv6 160355      0t0       TCP localhost:iprop->localhost:54010 (ESTABLISHED)
11520 awskii    8u     IPv4 157521      0t0       TCP W:42054->92.53.68.203:https (CLOSE_WAIT)
```
[yet another existing container creation]
```
11520 awskii    4u  a_inode   0,13        0     11625 [eventpoll]
11520 awskii    5u     IPv4 157447      0t0       TCP localhost:http-alt (LISTEN)
11520 awskii    6u     IPv6 159056      0t0       TCP *:iprop (LISTEN)
11520 awskii    7u     IPv6 160355      0t0       TCP localhost:iprop->localhost:54010 (ESTABLISHED)
11520 awskii    8u     IPv4 157521      0t0       TCP W:42054->92.53.68.203:https (CLOSE_WAIT)
11520 awskii    9u     IPv4 159367      0t0       TCP W:42108->92.53.68.203:https (CLOSE_WAIT)
```

Sockets are abandoned, and not closing for days.

Using `resp.Body.Close()` in `containers.Create` method solving this issue.

awskii:
In objects.Create you are using Close function. so, i think, it's a bug that you don't use it with containers creation.
This issue was there for months if no more and it was really hard to detect such careless socket usage.

jtopjian:
@awskii Thanks for reporting this.

Can you provide some further details about how to reproduce the problem? I'm unable to reproduce this.

I'm using the following test to trigger `containers.Create`:

```go
[...]

PR #1016, opened 2018-06-08
Object Storage v1: Close Body on Container Create

jtopjian:
For #1015


liu-sheng:
recheck stable/newton

liu-sheng:
recheck stable/ocata

liu-sheng:
recheck stable/pike

PR #1023, opened 2018-06-12
WIP: Expose the OS_BRANCH environment variable for acc testing

liu-sheng:
This change expose the OS_BRANCH environment variable for running acceptance tests, that allow specific tests can be skipped according to specific OpenStack branch.

For https://github.com/gophercloud/gophercloud/pull/1013#issuecomment-395633308


liu-sheng:
need to find a common approach to make things like this more generic....

PR #1007, opened 2018-05-28
Fix the CI report issue

yongfengdu:
CI report SUCCESS when some test cases fails.

For theopenlab/openlab-zuul-jobs#219

PR #1035, opened 2018-06-13
Acc Tests: Enable all compute tests

jtopjian:
For #818


jtopjian:
Looks like quotasets tests are failing due to no longer being supported. Merging this will have to wait until it's possible to detect which branch/release is being used.

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
@yongfengdu Just to keep you updated, I modified the `script/acceptancetest` file from the changes made in #1007. I found that with #1007, if an acceptance test failed, _all_ tests would halt, even in different OpenStack projects. This would prevent someone from seeing if their change worked if any other test failed.

One thing I can't figure out is how to get the xunit test results to show on a failure. This seems to still work for Terraform, but no longer for Gophercloud.

PR #1041, opened 2018-06-13
Messaging Claim Create

problemv:
For #836 

code in this PR:

https://github.com/openstack/zaqar/blob/8e0dd9763aba67bb7094cf92312daeba8d087866/zaqar/transport/wsgi/v2_0/claims.py#L60

### DIFF
--- a/openstack/messaging/v2/claims/requests.go
+++ b/openstack/messaging/v2/claims/requests.go
@@ -1,0 +1,66 @@
+package claims
+
+import (
+	"net/http"
+
+	"github.com/gophercloud/gophercloud"
+)
+
+// CreateOptsBuilder Builder.
+type CreateOptsBuilder interface {
+	ToClaimCreateRequest() (map[string]interface{}, string, error)
+}
+
+// CreateOpts params to be used with Create.
+type CreateOpts struct {
+	// Sets the TTL for the claim. When the claim expires un-deleted messages will be able to be claimed again.
+	TTL int `json:"ttl,omitempty"`
+
+	// Sets the Grace period for the claimed messages. The server extends the lifetime of claimed messages
+	// to be at least as long as the lifetime of the claim itself, plus the specified grace period.
+	Grace int `json:"grace,omitempty"`
+
+	// Set the limit of messages returned by create.
+	Limit int `q:"limit" json:"-"`
+}
+
+// ToClaimCreateRequest assembles a body and URL for a Create request based on
+// the contents of a CreateOpts.
+func (opts CreateOpts) ToClaimCreateRequest() (map[string]interface{}, string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	if err != nil {
+		return nil, q.String(), err
+	}
+
+	b, err := gophercloud.BuildRequestBody(opts, "")
+	if err != nil {
+		return b, "", err
+	}
+	return b, q.String(), err
+}
+
+// Create creates a Claim that claims messages on a specified queue.
+func Create(client *gophercloud.ServiceClient, queueName string, opts CreateOptsBuilder) (r CreateResult) {
+	b, q, err := opts.ToClaimCreateRequest()
+	if err != nil {
+		r.Err = err
+		return
+	}
+
+	url := createURL(client, queueName)
+	if q != "" {
+		url += q
+	}
+
+	var resp *http.Response
+	resp, r.Err = client.Post(url, b, nil, &gophercloud.RequestOpts{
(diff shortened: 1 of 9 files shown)


openstack/messaging/v2/claims/requests.go
```diff
@@ -0,0 +1,76 @@
...
+	q, err := gophercloud.BuildQueryString(opts)
+	if err != nil {
+		return nil, q.String(), err
+	}
+
+	b, err := gophercloud.BuildRequestBody(opts, "")
+	if err != nil {
+		return b, "", err
+	}
+	if _, ok := b["Limit"]; ok {
```
problemv: @jtopjian I don't know if this is the best way to accomplish what I am trying to do.

jtopjian: Try the following struct field:

```
Limit int `q:"limit" json:"-"`
```

A few things about this:

1. `omitempty` is a core Go/JSON tag. `q` is an in-house Gophercloud tag. This catches everyone off-guard :)
2. Gophercloud's `q` support will automatically `omitempty` if the value is the zero-value: https://github.com/gophercloud/gophercloud/blob/master/params.go#L332
3. The special `json:"-"` tag _should_ drop a field from being included in the body: https://github.com/gophercloud/gophercloud/blob/master/params.go#L119-L121

If this doesn't work, then what you're doing here is the next best option :)

problemv:

openstack/messaging/v2/claims/requests.go
```diff
@@ -0,0 +1,73 @@
...
+
+// CreateOptsBuilder Builder.
+type CreateOptsBuilder interface {
+	ToClaimCreateRequest() (map[string]interface{}, string, error)
+}
+
+// CreateOpts params to be used with Create.
+type CreateOpts struct {
+	// The name of the queue to claim messages from.
+	QueueName string `json:"queue_name" required:"true"`
```
jtopjian: The original time we did this was a very unique situation: an attribute of the resource was part of the URL and it felt really weird having it in the function signature.

An equivalent here would be if `Grace` was part of the URL rather in the request body.

Since the Queue Name isn't part of the `Claim` itself, it's a relational attribute, it probably should go in the `Create` function signature:

```
claim, err := claims.Create(client, queueName, opts).Extract()
```

Given `messages.Create` has `queueName` in the URL, too, we should follow that for consistency: https://github.com/gophercloud/gophercloud/blob/master/openstack/messaging/v2/messages/requests.go#L38

_technically_ `Claims` could even be bundled in the `queues` package. This happens occasionally - for example LBaaS has Pools and Members bundled because they share the same URL path. That's always felt a little weird to me, but if it's something you'd prefer to do, go for it.

openstack/messaging/v2/claims/requests.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Sets the TTL for the claim. When the claim expires un-deleted messages will be able to be claimed again.
+	TTL int `json:"ttl,omitempty"`
```
jtopjian: To make sure: 0 is not a valid TTL? If it is, this should be `*int`

openstack/messaging/v2/claims/requests.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// Sets the Grace period for the claimed messages. The server extends the lifetime of claimed messages
+	// to be at least as long as the lifetime of the claim itself, plus the specified grace period.
+	Grace int `json:"grace,omitempty"`
```
jtopjian: To make sure: 0 is not a valid Grace? If it is, this should be `*int`

openstack/messaging/v2/claims/requests.go
problemv: Personally I like having it separate from queues. Mainly because doing it the other way is a little confusing.

openstack/messaging/v2/claims/requests.go
problemv: "The value of grace value must be between 60 and 43200 seconds (12 hours)."

So 0 is not a valid value.

openstack/messaging/v2/claims/requests.go
problemv: "The ttl value must be between 60 and 43200 seconds (12 hours). "

openstack/messaging/v2/claims/requests.go
jtopjian: :+1: 

Thanks for confirming :)

Issue #1033, opened 2018-06-13
Container v1: Capsule Acceptance Test Failure


jtopjian:
Recently an error has been happening with the `TestCapsule` acceptance test:

```
convenience.go:35: Failure in capsules_test.go, line 84: unexpected error "parsing time "2018-06-13 05:25:24" as "2006-01-02 15:04:05-07:00": cannot parse "" as "-07:00"
```

@kevin-zhaoshuai Do you know if this is a recent change in Zun or is this a temporary error in devstack?

jtopjian:
Look like this is the commit which is causing problems:

https://github.com/openstack/zun/commit/647e44392a26f02283e38de84eb1415b71be0077

We'll need to now support both time formats since this has all been done within a single API version.

kevin-zhaoshuai:
@jtopjian Just find this issue. Will modify the code to support both time formats now.

jtopjian:
@kevin-zhaoshuai Actually I'm working on it right now :)

kevin-zhaoshuai:
@jtopjian That will be great! Thanks a lot

jtopjian:
Fixed in #1042

PR #1042, opened 2018-06-14
Container v1: Support new time format

jtopjian:
For #1033


jtopjian:
@kevin-zhaoshuai This should fix the problem. I'm going to go ahead and merge it, but let me know if you have any questions.

PR #1045, opened 2018-06-16
Fix the issue of unmarshal string into Go struct field MaxOverSubscriptionRatio 

wisererik:
The MaxOverSubscriptionRatio type should be string because max_over_subscription_ratio returned by cinder is a string, not float.

Fix:  #1046


xing-yang:
@dims @j-griffith Can you please take a look of this?  Thanks.

jtopjian:
> looks like we need to accomodate `auto`

This is indeed the use-case that defines this change. Otherwise it would have been debatable to do a string-to-float conversion.

PR #1052, opened 2018-06-19
Clustering v1: Updates and Additional Features

jtopjian:
For #823


jtopjian:
@JackKuei This PR bundles all current Senlin PRs and includes a bunch of refactoring and reorganizing.

First and foremost, all previously committed work and currently open PRs were (as far as I can tell) correct. I compared everything with the Senlin API service code and all looked good. So as far as core functionality goes, all was good :smile: 

The vast majority of the updates in this batch are style changes: 

* Inline documentation
* Acceptance test layout
* Unit test layout

What you had before was fine (or else I wouldn't have previously approved them), but I took the opportunity to change things around to match the style of the other Gophercloud packages.

I think the most notable changes to be aware of are the changes to the time unmarshalling. I simplified how it works and both unit and acceptance tests pass for me, but let me know if something no longer works on your side.

One note about the acceptance tests is that I made it compatible with Gophercloud's standard set of environment variables for running tests. This will allow Senlin to eventually be tied into the OpenLab CI system.

Unless you have any objections, I will merge this batch since all unit and acceptance tests are passing.

JackKuei:
@jtopjian Sound good! Can go ahead and merge the code. I will run against our environment with this PR in the near future and open bugs against it if needed. The only gotchas I ran into before was with different time formats within senlin itself.

Issue #1054, opened 2018-06-20
AllPages is making one extra call to OpenStack API


garbage-collector:
Hi !

While starting to use the gophercloud library, I noticed that using the AllPages() function (https://github.com/gophercloud/gophercloud/blob/master/pagination/pager.go#L125) from the pager ends with performing one not needed call to the OpenStack.

For listing flavors, for example, it performs two calls to OpenStack API.

The first call is done to get the "testPage" (https://github.com/gophercloud/gophercloud/blob/master/pagination/pager.go#L132).
The second call is done to get results by calling the EachPage function.

So if you have N pages of results, it looks like the code will perform N+1 calls to the OpenStack API.

This looks like an issue as all GET calls will end with a higher response time because of the double calls.

Is this a normal behavior ?

One way to fix it is to consider the "testPage" as a "firstPage" but I prefer to discuss about the issue before creating a PR.

Thanks

jtopjian:
> Is this a normal behavior ?

Maybe? :)

I can confirm this behavior exists. It's possible it was an oversight or doing a test/probe of the first page is required to bootstrap the `AllPages` process. 

If you're able to fix the issue, by all means go for it. The CI environment will run enough `AllPages` calls to determine if anything breaks (hopefully).

garbage-collector:
Pull request created https://github.com/gophercloud/gophercloud/pull/1056 :)

jtopjian:
Implemented in #1056

PR #1056, opened 2018-06-21
AllPages() - Do not fetch twice the first page

garbage-collector:
For #1054

This is the quickest way I found to fix the issue. Other ways I found generate a bigger rewriting.

The goal is to change the "testPage" as a "firstPage". Once it has been already fetched, there is no need to fetch it again.

It avoids N+1 calls on each usage of AllPages()

### DIFF
--- a/pagination/pager.go
+++ b/pagination/pager.go
@@ -41,6 +41,8 @@ type Pager struct {
 
 	createPage func(r PageResult) Page
 
+	firstPage Page
+
 	Err error
 
 	// Headers supplies additional HTTP headers to populate on each paged request.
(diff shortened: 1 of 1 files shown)


jtopjian:
@garbage-collector Nice! :)

@jrperritt Thoughts on this?

Dexus:
Looks good. Do what you want and definitely save a request! If I look at some of our tests, that's 30k requests a day less!

garbage-collector:
On our side, we are thinking to move our codebase from some homemade code to gophercloud, and this patch will save around 8.000.000 calls per day on our OpenStack infrastructure, so thanks for approving, we can continue to rewrite stuff on our codebase :)

jrperritt:
Seems fine to me

jtopjian:
@garbage-collector Thank you for digging into this and submitting a fix :)

Quick side note about saving calls: if you want to minimize the amount of communication traffic, I recommend using `EachPage` instead of `AllPages`. Most of the documentation will use `AllPages` because it's more succinct, but for greater control and the least amount of calls, `EachPage` is the way to go.

jtopjian:
Looks like go tip is not liking this patch for some reason... it might be a temporary issue or some new feature in Go within the past few days:

https://travis-ci.org/gophercloud/gophercloud/jobs/398304237

Dexus:
The GO tests never run there...
So it looks like a temp failure

Von meinem iPhone gesendet

jtopjian:
> The GO tests never run there...

The test ran before and passed.

> So it looks like a temp failure

It's possible or it's possible it's a new change in Go that will be permanent and we'll need to figure out what's different

Go 1.10 tests are still passing, so it's something specific in tip.

jtopjian:
I don't think the Travis error is due to this merge. I think it's coincidental that Travis began failing after this was merged.

While the error in question is due to a miscount of the actual number of pages, I think it's due to the actual test fixture itself. I'm going to open a PR to confirm this.

PR #1057, opened 2018-06-22
Senlin: ProfileTypes Get/List

JackKuei:

code in this PR:

[[profiletypes:get/list]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/profile_types.py#L42)

### DIFF
--- a/openstack/clustering/v1/profiletypes/results.go
+++ b/openstack/clustering/v1/profiletypes/results.go
@@ -15,10 +15,13 @@ type GetResult struct {
 	commonResult
 }
 
+type Schema map[string]interface{}
+type SupportStatus map[string]interface{}
+
 type ProfileType struct {
-	Name          string                 `json:"name"`
-	Schema        map[string]interface{} `json:"schema"`
-	SupportStatus map[string]interface{} `json:"support_status"`
+	Name          string                     `json:"name"`
+	Schema        map[string]Schema          `json:"schema"`
+	SupportStatus map[string][]SupportStatus `json:"support_status"`
 }
 
 func (r commonResult) Extract() (*ProfileType, error) {
--- a/openstack/clustering/v1/profiletypes/results.go
+++ b/openstack/clustering/v1/profiletypes/results.go
@@ -1,0 +1,50 @@
+package profiletypes
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// commonResult is the response of a base result.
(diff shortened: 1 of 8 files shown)


openstack/clustering/v1/profiletypes/results.go
```diff
@@ -0,0 +1,50 @@
...
+
+// GetResult is the response of a Get operations.
+type GetResult struct {
+	commonResult
+}
+
+type ProfileType struct {
+	Name          string                 `json:"name"`
+	Schema        map[string]interface{} `json:"schema"`
+	SupportStatus map[string]interface{} `json:"support_status"`
```
jtopjian: Is it possible to create a more strict type for both `Schema` and `SupportStatus`?

Could the _field_ `Schema` be of type `map[string]Schema` and the _type_ `Schema` be `map[string]interface{}`?

For `SupportStatus`, could the field be `map[string][]SupportStatus` and type `SupportStatus` be `map[string]interface{}` or `map[string]string`?

I'm not implying that we should create _really_ strict types that define fields such as `description`, `since`, etc (unless you feel that's manageable and possible), but something a little more tailored for the response.

openstack/clustering/v1/profiletypes/testing/fixtures.go
```diff
@@ -0,0 +1,244 @@
...
+				"required": false,
+				"type": "Map",
+				"updatable": false
+			}
+		}
+	}
+}
+`
+
+const GetResponse15 = `
```
jtopjian: I might be overlooking something obvious but I don't see how this microversion response is different from the standard response:

https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/profile_types.py#L35-L38

openstack/clustering/v1/profiletypes/results.go
JackKuei: Good call on changing Schema and SupportStatus as it will make the code much cleaner. 
Don't think we should create really strict field definitions as I think will be hard to manage.

JackKuei:
@jtopjian Ready for review

openstack/clustering/v1/profiletypes/testing/fixtures.go
jtopjian: Perfect. I must have been comparing the wrong fixtures. Thank you :)

Issue #1058, opened 2018-06-22
Key Manger v1 Support


jtopjian:
Supersedes #331 

#### Secrets

Docs:
* https://developer.openstack.org/api-guide/key-manager/secrets.html
* https://docs.openstack.org/barbican/latest/api/reference/secrets.html

Model:
* https://github.com/openstack/barbican/blob/fda4948efbf4aa67e68ca41dc20e927d918c593c/barbican/model/models.py#L266
* https://github.com/openstack/barbican/blob/fda4948efbf4aa67e68ca41dc20e927d918c593c/barbican/plugin/interface/secret_store.py#L264
* https://github.com/openstack/barbican/blob/1a9dcc12612d8fd5c2c811b12ff8aa12cfd3db3f/barbican/common/hrefs.py#L67-L69

List:
* https://github.com/openstack/barbican/blob/c75924017fa74fb74044eb908922787b34e19bdf/barbican/model/repositories.py#L732
* https://github.com/openstack/barbican/blob/c75924017fa74fb74044eb908922787b34e19bdf/barbican/api/controllers/secrets.py#L347

Get:
* https://github.com/openstack/barbican/blob/c75924017fa74fb74044eb908922787b34e19bdf/barbican/api/controllers/secrets.py#L118

Get Payload:
* https://github.com/openstack/barbican/blob/c75924017fa74fb74044eb908922787b34e19bdf/barbican/api/controllers/secrets.py#L152

Create:
* https://github.com/openstack/barbican/blob/c75924017fa74fb74044eb908922787b34e19bdf/barbican/api/controllers/secrets.py#L435-L445

Update:
* https://github.com/openstack/barbican/blob/c75924017fa74fb74044eb908922787b34e19bdf/barbican/api/controllers/secrets.py#L219

Delete:
* https://github.com/openstack/barbican/blob/c75924017fa74fb74044eb908922787b34e19bdf/barbican/api/controllers/secrets.py#L256

---

#### Secret Metadata

Docs:
* https://developer.openstack.org/api-guide/key-manager/secret_metadata.html
* https://docs.openstack.org/barbican/latest/api/reference/secret_metadata.html

Model:
* https://github.com/openstack/barbican/blob/master/barbican/model/models.py#L398
* https://github.com/openstack/barbican/blob/c75924017fa74fb74044eb908922787b34e19bdf/barbican/model/repositories.py#L880

List Metadata:
[...]

PR #1059, opened 2018-06-22
Key Manager v1 Support

jtopjian:
For #1058

### DIFF
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -542,6 +542,27 @@ func NewContainerV1Client() (*gophercloud.ServiceClient, error) {
 	})
 }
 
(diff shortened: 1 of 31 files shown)


jtopjian:
This is something I've been meaning to get to for a year now and had some to do it in bulk. I'm quite confident the code is good. There might be a few inline doc typos, maybe a question about a name here and there, but that's about it.

Unfortunately Coveralls doesn't take acceptance tests into consideration as every feature is extensively tested, including creating all possible types of secrets and containers. 

I'm going to break protocol and do a bulk merge of this since due to the time constraints of core reviewers (including myself), getting around to reviewing individual PRs would take months. I'm hesitant to merge this into a feature branch since we ran into some problems doing that with Zun. Given this is all self-contained in the `keymanager` subpackage, it won't affect other packages.

Issue #1003, opened 2018-05-20
Magnum support


eugenetaranov:
Hello,

is magnum api considered stable now? I'd like to cleanup/update where appropriate magnum related code and get it merged.

ozerovandrei:
+1 to that question.

Gophercloud doesn't have support for the `clusters`, `cluster templates` and `quotas`.
And `bays` and `baymodels` are deprecated names now: [ocata release notes](https://docs.openstack.org/releasenotes/magnum-ui/ocata.html#prelude).

Also Magnum uses `Container infrastructure management` term for its service name instead of the `Container orchestration`.

jtopjian:
No doubt a good amount of updates are needed in the magnum branch. I think the best way forward is for someone to audit it to see what's still relevant and begin copying it to the master branch in small pieces. Usual Gophercloud rules apply.

I can certainly add this to my list of items, but that list is really long right now :smile:

eugenetaranov:
thanks @jtopjian for confirmation, will work on this next weeks

JackKuei:
@jtopjian @eugenetaranov Will have some bandwidth to contribute to this project as well. See how we can work together on this.

mrhillsman:
@jtopjian @eugenetaranov @JackKuei @ozerovandrei if any of you could use a development server to work on OpenLab would be glad to assist. Please just fill out "Get Started" form at https://openlabtesting.org

openstacker:
@eugenetaranov @jtopjian I'm active contributor for OpenStack Magnum now and I'm happy to work with you guys together. The long term goal is supporting Magnum in Terraform ;) thanks.

jtopjian:
@mrhillsman 

> if any of you could use a development server to work on OpenLab would be glad to assist.

We're looking to deploy Magnum at `$work` so I should have a magnum environment available soon. If someone gets started before me (which is the most likely possibility), that'll at least help with verifying PRs.

If we could have a CI job for Magnum, that would really help!

@openstacker 

> I'm active contributor for OpenStack Magnum now and I'm happy to work with you guys together.

Thanks! Once work gets started, I might ping you if any questions about the API come up.

> The long term goal is supporting Magnum in Terraform

That would be a welcomed addition, too. Adding support to Gophercloud is the first step and then I can assist with the Terraform side, too.

[...]

PR #1065, opened 2018-06-26
[WIP] Magnum: ClusterTemplates: Create & Get

openstacker:
For #1003 

code in this PR:

CREATE API:
https://developer.openstack.org/api-ref/container-infrastructure-management/#create-new-cluster-template

CREATE code:
https://github.com/openstack/magnum/blob/stable/queens/magnum/api/controllers/v1/cluster_template.py#L372

GET API:
https://developer.openstack.org/api-ref/container-infrastructure-management/#create-new-cluster-template

GET code:
https://github.com/openstack/magnum/blob/stable/queens/magnum/api/controllers/v1/cluster_template.py#L336


jtopjian:
@openstacker Looks like there's some issue with the code, which is why Travis is failing.

openstacker:
@jtopjian  Thanks for the review, Joe. At this moment, I'm not sure if I should put more effort on this given @eugenetaranov and @JackKuei are keen to upstream their code. So I will just hold it bit to see if there are patches from them until next Monday, so that I can help review. Otherwise, I will continue this one. Cheers.

jtopjian:
@openstacker Sounds good!

Issue #1062, opened 2018-06-25
Migrate Magnum Branch Onto Master


JackKuei:
Migrate existing outdated magnum code onto master branch of gophercloud.
Refer to https://github.com/gophercloud/gophercloud/issues/1003

[https://developer.openstack.org/api-ref/container-infrastructure-management/](url)
---
- [ ] Manage Bay: Create
- [ ] Manage Bay: List
- [ ] Manage Bay: Get
- [ ] Manage Bay: Update
- [ ] Manage Bay: Delete
---
- [ ] Manage Baymodels: Create
- [ ] Manage Baymodels: List
- [ ] Manage Baymodels: Get
- [ ] Manage Baymodels: Update
- [ ] Manage Baymodels: Delete
---
- [ ] Manage Cluster: Create
- [ ] Manage Cluster: List
- [ ] Manage Cluster: Get
- [ ] Manage Cluster: Update
- [ ] Manage Cluster: Delete
---
- [ ] Manage Cluster Templates: Create
- [ ] Manage Cluster Templates: List
- [ ] Manage Cluster Templates: Get
- [ ] Manage Cluster Templates: Update
- [ ] Manage Cluster Templates: Delete
---
- [ ] Manage Certificates: Create
- [ ] Manage Certificates: Get
- [ ] Manage Certificates: Update
---
- [ ] Magnum Service:  List
---
- [ ] Magnum Service:  Stats
---
- [ ] Magnum Quota: Create
- [ ] Magnum Quota: List
- [ ] Magnum Quota: Get
- [ ] Magnum Quota: Update
- [ ] Magnum Quota: Delete
---

PR #1067, opened 2018-06-26
Magnum: Clustertemplates Create

JackKuei:

code in this PR:

[[clustertemplates:create]](https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/cluster_template.py#L372)

### DIFF
--- a/acceptance/openstack/containerinfra/v1/containerinfra.go
+++ b/acceptance/openstack/containerinfra/v1/containerinfra.go
@@ -1,0 +1,92 @@
+package v1
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/containerinfra/v1/clustertemplates"
+	th "github.com/gophercloud/gophercloud/testhelper"
+)
+
+// CreateClusterTemplate will create a random cluster tempalte. An error will be returned if the
+// cluster-template could not be created.
+func CreateClusterTemplate(t *testing.T, client *gophercloud.ServiceClient) (*clustertemplates.ClusterTemplate, error) {
+	choices, err := clients.AcceptanceTestChoicesFromEnv()
+	if err != nil {
+		return nil, err
+	}
+
+	name := tools.RandomString("TESTACC-", 8)
+	t.Logf("Attempting to create cluster template: %s", name)
+
+	createOpts := clustertemplates.CreateOpts{
+		Name:                name,
+		Labels:              map[string]string{},
+		FixedSubnet:         "",
+		MasterFlavorID:      "",
+		NoProxy:             "10.0.0.0/8,172.0.0.0/8,192.0.0.0/8,localhost",
+		HTTPSProxy:          "http://10.164.177.169:8080",
+		TLSDisabled:         false,
+		KeyPairID:           "kp",
+		Public:              false,
+		HTTPProxy:           "http://10.164.177.169:8080",
+		DockerVolumeSize:    3,
+		ServerType:          "vm",
+		ExternalNetworkID:   choices.ExternalNetworkID,
+		ImageID:             choices.ImageID,
+		VolumeDriver:        "cinder",
+		RegistryEnabled:     false,
+		DockerStorageDriver: "devicemapper",
+		NetworkDriver:       "flannel",
+		FixedNetwork:        "",
+		COE:                 "kubernetes",
+		FlavorID:            "m1.small",
+		MasterLBEnabled:     true,
+		DNSNameServer:       "8.8.8.8",
+	}
+
+	res := clustertemplates.Create(client, createOpts)
+	if res.Err != nil {
+		return nil, res.Err
+	}
+
+	requestID := res.Header.Get("X-OpenStack-Request-Id")
+	th.AssertEquals(t, true, requestID != "")
+
+	t.Logf("Cluster Template %s request ID: %s", name, requestID)
+
+	clusterTemplate, err := res.Extract()
+	if err != nil {
+		return nil, err
+	}
+
+	t.Logf("Successfully created cluster template: %s", clusterTemplate.Name)
+
+	tools.PrintResource(t, clusterTemplate)
+	tools.PrintResource(t, clusterTemplate.CreatedAt)
+
+	th.AssertEquals(t, name, clusterTemplate.Name)
+	th.AssertEquals(t, choices.ExternalNetworkID, clusterTemplate.ExternalNetworkID)
+	th.AssertEquals(t, choices.ImageID, clusterTemplate.ImageID)
+
+	return clusterTemplate, nil
+}
+
+/*
+// DeleteClusterTemplate will delete a given cluster-template. A fatal error will occur if the
+// cluster-template could not be deleted. This works best as a deferred function.
+func DeleteClusterTemplate(t *testing.T, client *gophercloud.ServiceClient, id string) {
+	t.Logf("Attempting to delete cluster-template: %s", id)
+
+	err := clustertemplates.Delete(client, id).ExtractErr()
+	if err != nil {
+		t.Fatalf("Error deleting cluster-template %s: %s:", id, err)
+	}
+
+	t.Logf("Successfully deleted cluster-template: %s", id)
+
+	return
+}
+*/
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -414,3 +414,9 @@ func NewKeyManagerV1(client *gophercloud.ProviderClient, eo gophercloud.Endpoint
 	sc.ResourceBase = sc.Endpoint + "v1/"
 	return sc, err
 }
+
+// NewContainerInfraV1 creates a ServiceClient that may be used with the v1 container infra management
+// package.
+func NewContainerInfraV1(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
+	return initClientOpts(client, eo, "container-infra")
+}
--- a/openstack/containerinfra/v1/clustertemplates/doc.go
+++ b/openstack/containerinfra/v1/clustertemplates/doc.go
@@ -1,0 +1,89 @@
+// Package clustertemplates contains functionality for working with Magnum Cluster Templates
+// resources.
+/*
+Package clustertemplates provides information and interaction with the cluster-templates through
+the OpenStack Container Infra service.
+
+Example to Create Cluster Template
+
+	createOpts := clustertemplates.CreateOpts{
+		Name:                "test-cluster-template",
+		Labels:              map[string]string{},
+		FixedSubnet:         nil,
+		MasterFlavorID:      nil,
+		NoProxy:             "10.0.0.0/8,172.0.0.0/8,192.0.0.0/8,localhost",
+		HTTPSProxy:          "http://10.164.177.169:8080",
+		TLSDisabled:         false,
+		KeyPairID:           "kp",
+		Public:              false,
+		HTTPProxy:           "http://10.164.177.169:8080",
+		DockerVolumeSize:    3,
+		ServerType:          "vm",
+		ExternalNetworkID:   "public",
+		ImageID:             "fedora-atomic-latest",
+		VolumeDriver:        "cinder",
+		RegistryEnabled:     false,
+		DockerStorageDriver: "devicemapper",
+		NetworkDriver:       "flannel",
+		FixedNetwork:        nil,
+		COE:                 "kubernetes",
+		FlavorID:            "m1.small",
+		MasterLBEnabled:     true,
+		DNSNameServer:       "8.8.8.8",
+	}
+
+	clustertemplate, err := clustertemplates.Create(serviceClient, createOpts).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+Example to List Clusters Templates
+
+	listOpts := clustertemplates.ListOpts{
+		Limit: 20,
+	}
+
+	allPages, err := clustertemplates.List(serviceClient, listOpts).AllPages()
+	if err != nil {
+		panic(err)
+	}
+
+	allClusterTemplates, err := clusters.ExtractClusterTemplates(allPages)
+	if err != nil {
+		panic(err)
+	}
+
+	for _, clusterTemplate := range allClusterTemplates {
+		fmt.Printf("%+v\n", clusterTemplate)
+	}
+
+Example to Update Cluster Template
+
+	updateOpts := []clustertemplates.UpdateOptsBuilder{
+		clustertemplates.UpdateOpts{
+			Path:  "/master_lb_enabled",
+			Value: true,
+			Op:    "replace",
+		},
+		clustertemplates.UpdateOpts{
+			Path:  "/registry_enabled",
+			Value: true,
+			Op:    "replace",
+		},
+	}
+
+	clustertemplate, err := clustertemplates.Update(serviceClient, updateOpts).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+Example to Delete Cluster Template
+
+	clusterTemplateID := "dc6d336e3fc4c0a951b5698cd1236ee"
+	err := clustertemplates.Delete(serviceClient, clusterTemplateID).ExtractErr()
+	if err != nil {
+		panic(err)
+	}
+
+*/
+package clustertemplates
--- a/openstack/containerinfra/v1/clustertemplates/requests.go
+++ b/openstack/containerinfra/v1/clustertemplates/requests.go
@@ -1,0 +1,63 @@
+package clustertemplates
+
+import (
+	"net/http"
+
+	"github.com/gophercloud/gophercloud"
+)
+
+// CreateOptsBuilder Builder.
+type CreateOptsBuilder interface {
+	ToClusterCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts params
+type CreateOpts struct {
+	Labels              map[string]string `json:"labels,omitempty"`
+	FixedSubnet         string            `json:"fixed_subnet,omitempty"`
+	MasterFlavorID      string            `json:"master_flavor_id,omitempty"`
+	NoProxy             string            `json:"no_proxy,omitempty"`
+	HTTPSProxy          string            `json:"https_proxy,omitempty"`
+	HTTPProxy           string            `json:"http_proxy,omitempty"`
+	TLSDisabled         bool              `json:"tls_disabled,omitempty"`
+	KeyPairID           string            `json:"keypair_id,omitempty"`
+	Public              bool              `json:"public,omitempty"`
+	DockerVolumeSize    int               `json:"docker_volume_size"`
+	ServerType          string            `json:"server_type,omitempty"`
+	ExternalNetworkID   string            `json:"external_network_id,omitempty"`
+	ImageID             string            `json:"image_id" required:"true"`
+	VolumeDriver        string            `json:"volume_driver,omitempty"`
+	RegistryEnabled     bool              `json:"registry_enabled,omitempty"`
+	DockerStorageDriver string            `json:"docker_storage_driver,omitempty"`
+	Name                string            `json:"name,omitempty"`
+	NetworkDriver       string            `json:"network_driver,omitempty"`
+	FixedNetwork        string            `json:"fixed_network,omitempty"`
+	COE                 string            `json:"coe" required:"true"`
+	FlavorID            string            `json:"flavor_id,omitempty"`
+	MasterLBEnabled     bool              `json:"master_lb_enabled,omitempty"`
+	DNSNameServer       string            `json:"dns_nameserver,omitempty"`
+}
+
+// ToClusterCreateMap constructs a request body from CreateOpts.
+func (opts CreateOpts) ToClusterCreateMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "")
+}
+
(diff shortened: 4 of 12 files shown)


acceptance/openstack/containerinfra/v1/containerinfra.go
```diff
@@ -0,0 +1,92 @@
...
+
+	createOpts := clustertemplates.CreateOpts{
+		Name:                name,
+		Labels:              map[string]string{},
+		FixedSubnet:         "",
+		MasterFlavorID:      "",
+		NoProxy:             "10.0.0.0/8,172.0.0.0/8,192.0.0.0/8,localhost",
+		HTTPSProxy:          "http://10.164.177.169:8080",
+		TLSDisabled:         false,
+		KeyPairID:           "kp",
```
jtopjian: Will this fail if a keypair of "kp" doesn't exist? You can import `github.com/gophercloud/gophercloud/acceptance/openstack/compute/v2` and use https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/compute/v2/compute.go#L275 to generate a keypair:

```go
import (
  compute "github.com/gophercloud/gophercloud/acceptance/openstack/compute/v2"
)

func CreateClusterTemplate(...) (...) {
   keypair, err := compute.CreateKeyPair(t, computeClient)
}
```

acceptance/openstack/containerinfra/v1/containerinfra.go
```diff
@@ -0,0 +1,92 @@
...
+		VolumeDriver:        "cinder",
+		RegistryEnabled:     false,
+		DockerStorageDriver: "devicemapper",
+		NetworkDriver:       "flannel",
+		FixedNetwork:        "",
+		COE:                 "kubernetes",
+		FlavorID:            "m1.small",
+		MasterLBEnabled:     true,
+		DNSNameServer:       "8.8.8.8",
+	}
```
jtopjian: Are all of these options necessary to create a basic template? The install docs only show a few options needed: https://docs.openstack.org/magnum/latest/install/launch-instance.html#provision-a-docker-swarm-cluster-and-create-a-container

openstack/client.go
```diff
@@ -414,3 +414,9 @@ func NewKeyManagerV1(client *gophercloud.ProviderClient, eo gophercloud.Endpoint
 	sc.ResourceBase = sc.Endpoint + "v1/"
 	return sc, err
 }
+
+// NewContainerInfraV1 creates a ServiceClient that may be used with the v1 container infra management
+// package.
+func NewContainerInfraV1(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
+	return initClientOpts(client, eo, "container-infra")
```
jtopjian: In the `urls.go` file you are trying to determine if `/v1` is included in the URL. There isn't precedence for that which would make this a very rare case. I recommend adding `v1` here like how it's done in the other clients in this file.

openstack/containerinfra/v1/clustertemplates/doc.go
```diff
@@ -0,0 +1,89 @@
...
+
+Example to Delete Cluster Template
+
+	clusterTemplateID := "dc6d336e3fc4c0a951b5698cd1236ee"
+	err := clustertemplates.Delete(serviceClient, clusterTemplateID).ExtractErr()
+	if err != nil {
+		panic(err)
+	}
+
+*/
```
jtopjian: I recommend removing the other examples here. It could confuse users until the other methods are added.

openstack/containerinfra/v1/clustertemplates/requests.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/gophercloud/gophercloud"
+)
+
+// CreateOptsBuilder Builder.
+type CreateOptsBuilder interface {
+	ToClusterCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts params
+type CreateOpts struct {
```
jtopjian: Where in the Magnum API code are these fields defined?

openstack/containerinfra/v1/clustertemplates/requests.go
```diff
@@ -0,0 +1,63 @@
...
+// Create requests the creation of a new cluster.
+func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResult) {
+	b, err := opts.ToClusterCreateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	var result *http.Response
+	result, r.Err = client.Post(createURL(client), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 201, 202},
```
jtopjian: Using the `openstack --debug ...` command, you should be able to see what single response code is valid. I recommend starting with 1 code unless you can point to a reference where all 3 codes are valid.

openstack/containerinfra/v1/clustertemplates/results.go
```diff
@@ -0,0 +1,89 @@
...
+
+// Extract is a function that accepts a result and extracts a cluster-template resource.
+func (r commonResult) Extract() (*ClusterTemplate, error) {
+	var s *ClusterTemplate
+	err := r.ExtractInto(&s)
+	return s, err
+}
+
+// Represents a template for a Cluster Template
+type ClusterTemplate struct {
```
jtopjian: Where in the Magnum API code are these attributes defined?

openstack/containerinfra/v1/clustertemplates/results.go
```diff
@@ -0,0 +1,89 @@
...
+	CreatedAt           time.Time          `json:"-"`
+	NetworkDriver       string             `json:"network_driver"`
+	FixedNetwork        string             `json:"fixed_network"`
+	COE                 string             `json:"coe"`
+	FlavorID            string             `json:"flavor_id"`
+	MasterLBEnabled     bool               `json:"master_lb_enabled"`
+	DNSNameServer       string             `json:"dns_nameserver"`
+}
+
+func (r *ClusterTemplate) UnmarshalJSON(b []byte) error {
```
jtopjian: For Senlin, you mentioned that it's possible for time values to be returned as both `""` and `null`. That's why we need a special `UnmarshalJSON` method.

Are you sure Magnum does the same thing? If Magnum returns empty times as `null` (which would be the normal behavior), then we can remove this entirely. `time.RFC3339` is the standard time format which Go will correctly unmarshal without any help.

If you are unsure, I recommend removing this.

openstack/containerinfra/v1/clustertemplates/testing/fixtures.go
```diff
@@ -0,0 +1,197 @@
...
+	"testing"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/containerinfra/v1/clustertemplates"
+	th "github.com/gophercloud/gophercloud/testhelper"
+	fake "github.com/gophercloud/gophercloud/testhelper/client"
+)
+
+const ClusterTemplateResponse = `
```
jtopjian: These fixtures look to be pulled from the old `magnum` branch and are outdated (note how the link includes `baymodels`). Please generate fresh fixtures from a working Magnum environment.

acceptance/openstack/containerinfra/v1/containerinfra.go
JackKuei: Yes...this one will probably be defined from env var.

acceptance/openstack/containerinfra/v1/containerinfra.go
JackKuei: Will probably need to discuss what is actually needed. From the code only see couple mandatory fields, but when ran with just those mandatory fields, it doen't work.

openstack/containerinfra/v1/clustertemplates/results.go
JackKuei: Magnum is using ISO-8601, so shouldn't return both "" and null. Can test some more on it but time.RFC3339 should unmarshal properly

openstack/client.go
JackKuei: Magnum includes version in its endpoint, so adding it will make sc.ResourceBase have double "v1". But testing doesn't include "v1" for its endpoint. Will look at it more to see what needs to be done.

acceptance/openstack/containerinfra/v1/containerinfra.go
jtopjian: That's really curious. This is what is being POST'd when running the command shown in the above doc link:

```json
{
    "coe": "swarm",
    "dns_nameserver": "8.8.8.8",
    "docker_storage_driver": "devicemapper",
    "docker_volume_size": null,
    "external_network_id": "public",
    "fixed_network": null,
    "fixed_subnet": null,
    "flavor_id": "m1.small",
    "http_proxy": null,
    "https_proxy": null,
    "image_id": "fedora-atomic-latest",
    "keypair_id": null,
    "labels": {},
    "master_flavor_id": "m1.small",
    "master_lb_enabled": false,
    "name": "swarm-cluster-template",
    "network_driver": null,
    "no_proxy": null,
    "public": false,
    "registry_enabled": false,
    "server_type": "vm",
    "tls_disabled": false,
    "volume_driver": null
}
```

The notable omissions are the IPs and proxy-related fields. 

What I would be curious about is if Magnum wants `null` to be explicitly passed or if the fields can be omitted. If the former, then that is going to be a unique behavior of Magnum (all OpenStack services have their quirks) and we'll need to account for that accordingly (ie: no `omitempty`).

acceptance/openstack/containerinfra/v1/containerinfra.go
jtopjian: The other option is to not specify a keypair at all. Since I don't think any test is going to actually SSH to the created cluster, it should be safe to omit.

JackKuei:
@jtopjian Thanks for the review as these WIP are a good starting point to discuss with @eugenetaranov  and @openstacker with the differences in codes that we have and see how we can reconcile the differences.

openstack/client.go
jtopjian: ah - I understand.

In this case, I recommend setting `/v1` in the ServiceClient and removing it from `urls.go`. The `testhelper` functions and testclient will work with this configuration.

openstack/containerinfra/v1/clustertemplates/requests.go
openstacker: Basically here https://github.com/openstack/magnum/blob/stable/queens/magnum/api/controllers/v1/cluster_template.py#L36 but some of them are generated by Magnum not user input.

openstack/containerinfra/v1/clustertemplates/requests.go
```diff
@@ -13,29 +13,33 @@ type CreateOptsBuilder interface {
...
-	Public              bool              `json:"public,omitempty"`
-	DockerVolumeSize    int               `json:"docker_volume_size"`
-	ServerType          string            `json:"server_type,omitempty"`
 	ExternalNetworkID   string            `json:"external_network_id,omitempty"`
-	ImageID             string            `json:"image_id" required:"true"`
+	FixedNetwork        string            `json:"fixed_network,omitempty"`
+	FixedSubnet         string            `json:"fixed_subnet,omitempty"`
+	NetworkDriver       string            `json:"network_driver,omitempty"`
+	APIServerPort       int               `json:"apiserver_port,omitempty"`
+	DockerVolumeSize    int               `json:"docker_volume_size,omitempty"`
```
openstacker: It's required.

openstack/containerinfra/v1/clustertemplates/requests.go
```diff
@@ -13,29 +13,33 @@ type CreateOptsBuilder interface {
...
+	ClusterDistro       string            `json:"cluster_distro,omitempty"`
```
openstacker: Unfortunately, I think Magnum's API ref doc is not good and which is causing your misunderstanding here. Attributes like **ClusterDistro** is not required from the user side, it's extracted by Magnum from the given image.

openstacker: I would suggest just follow the openstackclient to figure out the parameters from user side instead of just using all the attributes at here https://github.com/openstack/magnum/blob/stable/queens/magnum/api/controllers/v1/cluster_template.py#L44 Cheers.

openstack/containerinfra/v1/clustertemplates/requests.go
jtopjian: According to https://docs.openstack.org/magnum/latest/install/launch-instance.html, `DockerVolumeSize` isn't required. And in the JSON request I pasted yesterday, it's omitted.

jtopjian: Unless you mean "required" as in "it must be included in the request body, even if it's null" ?

openstack/containerinfra/v1/clustertemplates/requests.go
JackKuei: For the fields, the interesting thing was, I was able to generate clustertemplates successfully both with and without the above four fields. Are you seeing something differently where these fields are required? Maybe we are running on different versions? I'll go ahead and remove the ClusterDistro. Yes the fields in the doc are incomplete and makes it confusing.

JackKuei: BTW, feel free to make modifications

openstack/containerinfra/v1/clustertemplates/doc.go
JackKuei: Done

openstack/containerinfra/v1/clustertemplates/requests.go
JackKuei: These fields will work with omitempty.

openstacker: That's a change we may need to make in openstack client, for now, user can create cluster template without the docker_volume_size setting, but it will be failed when user try to create cluster with that template. I will double check that and get back.

jtopjian: ahh understood.

So while it logically makes sense to require `docker_volume_size` because 99% of the time it's needed in order to create a cluster, that kind of validation is too high above the API for us to be worried about.

Gophercloud is a very thin, low communication layer between OpenStack and the Go language. If the API service will happily accept a cluster with no `docker_volume_size`, then we should allow that, even though the user will probably be shooting themselves in the foot. But at the same time, maybe there is such a possible template where `docker_volume_size` is not required.

jtopjian: Another way of thinking about this is: always assume Gophercloud will _not_ be end-user facing. Gophercloud is meant for developers to use to create OpenStack compatible applications. It is in those applications where the enforcement of settings such as `docker_volume_size` will be required.

openstacker: That totally makes sense, cheers.

JackKuei:
@jtopjian @openstacker @eugenetaranov Did you have a chance to take this PR for a spin? What would be some good values for the createOpts? We are running k8s so might not have the minimalist fields.

acceptance/openstack/containerinfra/v1/containerinfra.go
```diff
@@ -0,0 +1,94 @@
...
+		NoProxy:             "10.0.0.0/8,172.0.0.0/8,192.0.0.0/8,localhost",
+		HTTPSProxy:          "http://10.164.177.169:8080",
+		TLSDisabled:         &boolFalse,
+		KeyPairID:           "kp",
+		Public:              &boolFalse,
+		HTTPProxy:           "http://10.164.177.169:8080",
+		DockerVolumeSize:    3,
+		ServerType:          "vm",
+		ExternalNetworkID:   choices.ExternalNetworkID,
+		ImageID:             choices.ImageID,
```
jtopjian: Usually the `ImageID` is set to a "cirros" image. I don't think the standard Magnum templates will allow this -- they're looking for a Fedora Atomic image.

I think we should add a new Acceptance Test variable of `MagnumImageID` (`OS_MAGNUM_IMAGE_ID`) here: 

https://github.com/gophercloud/gophercloud/blob/master/acceptance/clients/clients.go#L18
https://github.com/gophercloud/gophercloud/blob/master/acceptance/clients/clients.go#L50
https://github.com/gophercloud/gophercloud/blob/master/acceptance/clients/clients.go#L100

But don't return an error if it's missing -- we'll have to update existing test infrastructure to account for this before we can make it a required setting.

openstack/containerinfra/v1/clustertemplates/requests.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/gophercloud/gophercloud"
+)
+
+// CreateOptsBuilder Builder.
+type CreateOptsBuilder interface {
+	ToClusterCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts params
+type CreateOpts struct {
```
jtopjian: Because of the number of fields, let's sort these in alphabetical order so they're easier to manage.

openstack/containerinfra/v1/clustertemplates/results.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Extract is a function that accepts a result and extracts a cluster-template resource.
+func (r commonResult) Extract() (*ClusterTemplate, error) {
+	var s *ClusterTemplate
+	err := r.ExtractInto(&s)
+	return s, err
+}
+
+// Represents a template for a Cluster Template
+type ClusterTemplate struct {
```
jtopjian: Because of the number of fields, let's sort these in alphabetical order.

openstack/containerinfra/v1/clustertemplates/requests.go
JackKuei: I was following the order of the doc. Will change to alphabetical.

openstack/containerinfra/v1/clustertemplates/results.go
JackKuei: I was following the order of the doc. Will change to alphabetical.

acceptance/openstack/containerinfra/v1/containerinfra.go
JackKuei: Will add the new env var OS_MAGNUM_IMAGE_ID

openstack/containerinfra/v1/clustertemplates/results.go
jtopjian: Yeah, in general that would be fine. But given the amount of options here, I think deviating and going with alphabetical is a more manageable layout here.

JackKuei:
@jtopjian @openstacker @eugenetaranov This is ready for PR, please look over the CreateOpts for the acceptance test  as I've trim the arg for docker swarm. https://github.com/gophercloud/gophercloud/pull/1067/files#diff-4967fbaa317c5486f775b78730ffc9aaR25
BTW, couldn't get my devstack to spin up docker swarm as it kept failing on creating swarm master. However tested on k8s was successful

jtopjian:
@JackKuei I'll take a look later today

> BTW, couldn't get my devstack to spin up docker swarm as it kept failing on creating swarm master. However tested on k8s was successful

I'll try this out, too. It's possible it's a devstack issue. I last tested this on my work's cloud and got Swarm to work (we have some custom Swarm templates in production)

jtopjian:
@JackKuei This works for me.

I think the reason Swarm is failing for you is because you're not enabling floating IPs. There was recently a patch merged to allow Swarm clusters without floating IPs: 

https://review.openstack.org/#/c/571200

If you're using Devstack, I would imagine that patch would be included, but who knows. I have a set of Magnum drivers that has that patch manually implemented, but it also does some other customizations for our cloud:

https://github.com/cybera/magnum-templates

I verified that a template created with this code can be used to create a fully working Swarm cluster. 

In general, this PR looks good to me. It's a large PR due to the amount of available fields. The current field configuration looks correct to me, though it's hard to obtain a confident list of _inputs_ from the Magnum code. 

It's also difficult to get a perfect acceptance test that'll be useful in devstack given devstack is an all-in-one environment and Magnum requires so many resources to test end-to-end. We can tweak the acceptance tests as we work through all of the PRs.

I'm happy with this and can merge it if you'd like. Really nice work on it.

Oh, and I have no idea why Coveralls is reporting such a massive decrease in coverage. I'm not too concerned about it right now.

acceptance/openstack/containerinfra/v1/containerinfra.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil, err
+	}
+
+	name := tools.RandomString("TESTACC-", 8)
+	t.Logf("Attempting to create cluster template: %s", name)
+
+	boolFalse := false
+	createOpts := clustertemplates.CreateOpts{
+		Name:                name,
+		MasterFlavorID:      "m1.small",
```
openstacker: Why don't just reuse the choices.FlavorID?

openstacker:
@JackKuei Thanks for the great work. Overall it looks good for me. And I agree with @jtopjian , we can improve any small issue later when we do the /clusters and /certificates API support.

acceptance/openstack/containerinfra/v1/containerinfra.go
jtopjian: That's probably a good idea. This can be updated in a future PR.

JackKuei:
@jtopjian @openstacker Yes, MasterFlavorID should use choices, was an oversight on me. Will fixed it during other clustertemplates APIs.

Thanks for the tip, I'll try enabling floating IPs and hopefully can get my devstack to create cluster.

Not really sure why the coverage dropped so much, before the "Added OS_MAGNUM_IMAGE_ID" commit, it was passing. Also when locally ran, got a coverage of 92% for clustertemplates.

JackKuei:
@jtopjian Since this involves so many commits, should I do a squash?

PR #1068, opened 2018-06-26
Magnum: Manage Cluster Templates Delete

JackKuei:

code in this PR:

[[clustertemplates:delete]](https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/cluster_template.py#L461)

### DIFF
--- a/openstack/containerinfra/v1/clustertemplates/results.go
+++ b/openstack/containerinfra/v1/clustertemplates/results.go
@@ -1,6 +1,7 @@
 package clustertemplates
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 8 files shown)


JackKuei:
@jtopjian This one is ready for review

openstack/containerinfra/v1/clustertemplates/results.go
```diff
@@ -58,3 +65,17 @@ type ClusterTemplate struct {
 	UserID              string             `json:"user_id"`
 	VolumeDriver        string             `json:"volume_driver"`
 }
+
+func (r DeleteResult) Extract() (string, error) {
```
jtopjian: This is a unique way of handling retrieval of the Request ID. Is this really needed or can this be handled with the standard implementation of:

```go
res := clustertemplates.Delete(client, id)
if res.Err != nil {
  panic(res.Err)
}

requestID := res.Header["X-Openstack-Request-Id"][0]
```

JackKuei: Agree, should keep it consistent. Will fix.

PR #1071, opened 2018-06-26
Magnum: Manage Cluster Templates Update

JackKuei:

code in this PR:

[[clustertemplates:update]](https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/cluster_template.py#L414)

### DIFF
--- a/acceptance/openstack/containerinfra/v1/clustertemplates_test.go
+++ b/acceptance/openstack/containerinfra/v1/clustertemplates_test.go
@@ -58,10 +58,6 @@ func TestClusterTemplatesCRUD(t *testing.T) {
 			Path:  "/labels/test",
 			Value: "test",
 		},
-		clustertemplates.UpdateOpts{
-			Op:   clustertemplates.AddOp,
-			Path: "/labels/test1",
-		},
 	}
 
 	updateClusterTemplate, err := clustertemplates.Update(client, clusterTemplate.UUID, updateOpts).Extract()
--- a/acceptance/openstack/containerinfra/v1/clustertemplates_test.go
+++ b/acceptance/openstack/containerinfra/v1/clustertemplates_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
 	"github.com/gophercloud/gophercloud/openstack/containerinfra/v1/clustertemplates"
 	th "github.com/gophercloud/gophercloud/testhelper"
 )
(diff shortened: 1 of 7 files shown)


JackKuei:
@jtopjian This one is ready for review

openstack/containerinfra/v1/clustertemplates/doc.go
```diff
@@ -66,5 +66,25 @@ Example to List Clusters Templates
 		fmt.Printf("%+v\n", clusterTemplate)
 	}
 
+Example to Update Cluster Template
+
+	updateOpts := []clustertemplates.UpdateOptsBuilder{
+		clustertemplates.UpdateOpts{
+			Path:  "/master_lb_enabled",
+			Value: true,
```
jtopjian: doc nit: true should be "true"

openstack/containerinfra/v1/clustertemplates/doc.go
```diff
@@ -66,5 +66,25 @@ Example to List Clusters Templates
...
+			Op:    "replace",
+		},
+		clustertemplates.UpdateOpts{
+			Path:  "/registry_enabled",
+			Value: true,
```
jtopjian: doc nit: true should be "true"

openstack/containerinfra/v1/clustertemplates/requests.go
```diff
@@ -123,3 +124,51 @@ func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
 	}
 	return
 }
+
+type UpdateOpts struct {
+	Path  string `json:"path" required:"true"`
+	Value string `json:"value,omitempty"`
+	Op    string `json:"op" required:"true"`
```
jtopjian: Let's create a type for `Op`. It looks like Zaqar Queues implements a similar API pattern, so we can steal from what we've done in Gophercloud: https://github.com/gophercloud/gophercloud/blob/master/openstack/messaging/v2/queues/requests.go#L134-L147

openstack/containerinfra/v1/clustertemplates/requests.go
```diff
@@ -123,3 +124,51 @@ func Get(client *gophercloud.ServiceClient, id string) (r GetResult) {
...
+type UpdateOptsBuilder interface {
+	ToClusterTemplateUpdateMap() (map[string]interface{}, error)
+}
+
+// ToClusterUpdateMap assembles a request body based on the contents of
+// UpdateOpts.
+func (opts UpdateOpts) ToClusterTemplateUpdateMap() (map[string]interface{}, error) {
+	if opts.Op != "remove" && opts.Value == "" {
+		return nil, fmt.Errorf("Value field must be provied for Op=[%s]. Only when Op=remove can have empty Value field", opts.Op)
+	}
```
jtopjian: Will the API return an error if an illegal combination is given? If so, let's just let the API error be returned to the user rather than masking it with our own error.

openstack/containerinfra/v1/clustertemplates/doc.go
JackKuei: Good catch!

openstack/containerinfra/v1/clustertemplates/requests.go
JackKuei: Sounds good

openstack/containerinfra/v1/clustertemplates/requests.go
JackKuei: Will fix and let back-end handle validation rather than our own masking.

JackKuei:
@jtopjian This one is ready

acceptance/openstack/containerinfra/v1/clustertemplates_test.go
```diff
@@ -40,32 +41,34 @@ func TestClusterTemplatesCRUD(t *testing.T) {
...
+			Value: "false",
+		},
+		clustertemplates.UpdateOpts{
+			Op:    clustertemplates.AddOp,
+			Path:  "/labels/test",
+			Value: "test",
+		},
+		clustertemplates.UpdateOpts{
+			Op:   clustertemplates.AddOp,
+			Path: "/labels/test1",
```
jtopjian: The addition of this is now causing the test to fail.

If you wanted to test for this error, it might be best to do it in a separate `Update` call so you can test for success in one call and test for an error in another call.

JackKuei: This got in by err. Fixed by removing this block.

jtopjian:
LGTM - thank you!

Issue #1076, opened 2018-06-28
Manila: missing revoke access functionality


gman0:
gophercloud's Manila interface is missing "revoke access" functionality, which is required by the upcoming changes in Manila external-provisioner in [kubernetes/cloud-provider-openstack](https://github.com/kubernetes/cloud-provider-openstack). Will follow-up with a PR.

jtopjian:
Implemented in #1077

PR #1077, opened 2018-06-28
Add revoke access to Manila

gman0:
For #1076

code in this PR:

- revoke access request https://github.com/openstack/manila/blob/master/api-ref/source/samples/share-actions-revoke-access-request.json
- removing an access rule from DB https://github.com/openstack/manila/blob/master/manila/share/access.py#L224

### DIFF
--- a/openstack/sharedfilesystems/v2/shares/testing/fixtures.go
+++ b/openstack/sharedfilesystems/v2/shares/testing/fixtures.go
@@ -198,6 +198,25 @@ func MockGrantAccessResponse(t *testing.T) {
 	})
 }
 
+var revokeAccessRequest = `{
+	"deny_access": {
+		"access_id": "a2f226a5-cee8-430b-8a03-78a59bd84ee8"
+	}
+}`
+
+func MockRevokeAccessResponse(t *testing.T) {
+	th.Mux.HandleFunc(shareEndpoint+"/"+shareID+"/action", func(w http.ResponseWriter, r *http.Request) {
+		th.TestMethod(t, r, "POST")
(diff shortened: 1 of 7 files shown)


gman0:
cc @jtopjian

openstack/sharedfilesystems/v2/shares/testing/fixtures.go
```diff
@@ -198,6 +198,25 @@ func MockGrantAccessResponse(t *testing.T) {
...
+func MockRevokeAccessResponse(t *testing.T) {
+	th.Mux.HandleFunc(shareEndpoint+"/"+shareID+"/action", func(w http.ResponseWriter, r *http.Request) {
+		th.TestMethod(t, r, "POST")
+		th.TestHeader(t, r, "X-Auth-Token", fake.TokenID)
+		th.TestHeader(t, r, "Content-Type", "application/json")
+		th.TestHeader(t, r, "Accept", "application/json")
+		th.TestJSONRequest(t, r, revokeAccessRequest)
+		w.Header().Add("Content-Type", "application/json")
+		w.WriteHeader(http.StatusOK)
+		fmt.Fprintf(w, grantAccessResponse)
```
jtopjian: I don't think this API call actually returns anything, correct? If so, then let's remove this line. Maybe change `StatusOK` to `StatusAccepted` to emulate the 202 response, too. I'm OK with leaving 200 as a valid `OkCode` in addition to the 202.

gman0: Ah yes, absolutely!

gman0:
Thanks!

PR #1083, opened 2018-06-30
Acc Tests: Enable ImageService v2 testing in OpenLab

jtopjian:
For #818


jtopjian:
@jrperritt ping for review of the second commit here.

jtopjian:
recheck

jtopjian:
Failure is a Zun failure. It's possible it's broke in master right now. I'll recheck it tomorrow.

Issue #1086, opened 2018-07-04
Bug in ExtractIntoSlicePtr


jtopjian:
I've found a bug in the `ExtractIntoSlicePtr`, though I'm not sure of the cause just yet.

Given a response such as:

```json
{                                                           
    "networks": [                                           
        {                                                   
            "status": "ACTIVE",                             
            "subnets": [                                    
                "54d6f61d-db07-451c-9ab3-b9609b6b6f0b"      
            ],                                              
            "name": "public",                               
            "admin_state_up": true,                         
            "tenant_id": "4fd44f30292945e481c7b8a0c8908869",
            "shared": true,                                 
            "id": "d32019d3-bc6e-4319-9c1d-6722fc136a22",   
            "provider:segmentation_id": 9876543210,         
            "provider:physical_network": null,              
            "provider:network_type": "local",               
            "router:external": true,                        
            "port_security_enabled": true                   
        },                                                  
        {                                                   
            "status": "ACTIVE",                             
            "subnets": [                                    
                "08eae331-0402-425a-923c-34f7cfe39c1b"      
            ],                                              
            "name": "private",                              
            "admin_state_up": true,                         
            "tenant_id": "26a7980765d0414dbc1fc1f88cdb7e6e",
            "shared": false,                                
            "id": "db193ab3-96e3-4cb3-8fc5-05f4296d0324",   
            "provider:segmentation_id": 1234567890,         
            "provider:physical_network": null,              
            "provider:network_type": "local",               
            "router:external": false,                       
            "port_security_enabled": false                  
        }                                                   
    ]                                                       
}
```                                                          

Both networks will have the subnet of `08eae331-0402-425a-923c-34f7cfe39c1b`.

jtopjian:
Fixed in #1092

PR #1092, opened 2018-07-05
Core: Fixing issue with extractIntoPtr and Slices

jtopjian:
For #1086


jtopjian:
recheck

jtopjian:
The failing test is the prior failing capsule test which I believe has been fixed.

I'm going to mere this.

Issue #1094, opened 2018-07-05
Manila: missing share listing


gman0:
gophercloud's Manila interface is missing share listing functionality, will follow-up with a PR.

jtopjian:
Implemented in #1095

PR #1095, opened 2018-07-05
Manila: added ListDetail for detailed Share listing 

gman0:
For #1094

code in this PR:

- example response: https://github.com/openstack/manila/blob/master/api-ref/source/samples/shares-list-detailed-response.json
- `get_all` function: https://github.com/openstack/manila/blob/master/manila/share/api.py#L1469

### DIFF
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -1,6 +1,9 @@
 package shares
 
-import "github.com/gophercloud/gophercloud"
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
 
 // CreateOptsBuilder allows extensions to add additional parameters to the
 // Create request.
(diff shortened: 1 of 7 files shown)


openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -66,6 +69,81 @@ func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r Create
 	return
 }
 
+// ListOpts holds options for listing Shares. It is passed to the
+// shares.List function.
+type ListOpts struct {
```
jtopjian: As best as I can tell, this looks correct. I've found some fields also defined in:

* https://github.com/openstack/manila/blob/e4acaa5bf14c1e188a21aab5f15d9613ebb0aa6e/manila/api/v1/shares.py#L123
* https://github.com/openstack/manila/blob/e4acaa5bf14c1e188a21aab5f15d9613ebb0aa6e/manila/api/v1/shares.py#L187

The v1 code is called from v2 by way of inheritance: https://github.com/openstack/manila/blob/e4acaa5bf14c1e188a21aab5f15d9613ebb0aa6e/manila/api/v2/shares.py#L421-L450

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -101,6 +104,84 @@ type CreateResult struct {
...
+	}
+
+	q := currentURL.Query()
+	q.Set("offset", mark)
+	currentURL.RawQuery = q.Encode()
+	return currentURL.String(), nil
+}
+
+// LastMarker returns the last offset in a ListResult.
+func (r SharePage) LastMarker() (string, error) {
```
jtopjian: I realize this was copied from `sharenetworks` but I wanted to flag it for a sanity check. 

My interpretation of this is that Gophercloud will request an `offset` of `9223372036854775807` in order to deliberately make an erroneous call which will return no shares. Is this really the only way to determine this? Wouldn't this cause a lot of unnecessary GETs over time?

openstack/sharedfilesystems/v2/shares/requests.go
gman0: I've added `metadata`, `extra_specs` and the missing fields from your second link. Thanks!

openstack/sharedfilesystems/v2/shares/results.go
gman0: Should be fixed now: `LastMarker` now returns an "invalid marker" (-1) which is recognized by `NextPageURL` which then returns an empty string - this stops the page iteration. Thanks!

gman0:
@jtopjian PTAL

openstack/sharedfilesystems/v2/shares/requests.go
jtopjian: Nice - thank you.

openstack/sharedfilesystems/v2/shares/results.go
jtopjian: Thank you for this as well!

gman0:
Thanks for the review and merge!

PR #1097, opened 2018-07-05
Convert type of 'command' from string to list

hongbin:
The field 'command' of container resource is changed from string
to list [1]. This change should not break API consumer if they
pin a specific API version. However, Gophercloud doesn't support
API microversion yet, so the CI is complaining. This patch attempts
to fix it.

[1] https://review.openstack.org/#/c/575709/

For #[PUT ISSUE NUMBER HERE]

code in this PR:

[PUT URLS HERE]

### DIFF
--- a/openstack/container/v1/capsules/results.go
+++ b/openstack/container/v1/capsules/results.go
@@ -196,7 +196,7 @@ type Container struct {
 	ImageDriver string `json:"image_driver"`
 
 	// Command for the container
-	Command string `json:"command"`
+	Command []string `json:"command"`
 
 	// Capsule ID for the container
 	CapsuleID int `json:"capsule_id"`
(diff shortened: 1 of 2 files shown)


hongbin:
recheck

hongbin:
@jtopjian ,

I am not sure if the Travis CI failure is relevant or irrelevant to this PR. Could you advice?

jtopjian:
@hongbin The error in travis is:

```# github.com/gophercloud/gophercloud/openstack/container/v1/capsules/testing
openstack/container/v1/capsules/testing/fixtures.go:449:15: cannot use "testcmd" (type string) as type []string in field value```

It looks like you'll need to make sure previous uses of `Command` are now `[]string` and not `string`. For example:

https://github.com/gophercloud/gophercloud/blob/master/openstack/container/v1/capsules/testing/fixtures.go#L449

hongbin:
recheck

hongbin:
recheck

jtopjian:
@hongbin no problem at all :)

Issue #1100, opened 2018-07-05
ETag calculation could use some documentation (/warning)


dropje86:
As already filed here #805 there is a fairly major issue when uploading large files in the default library behavior. The Create method's CreateOpts.Content expects an io.Reader which causes some issues down the line calculating the md5 hash as there's a need to temporarily copy the content into memory.

It took a while to figure out why my app was consuming massive amounts of memory during the upload of large files.

Could it be explicitly documented that in large file use cases it's better to calculate the hash outside of the lib and provide an ETag or explicitly require ETag calculation instead of doing it for us, or (maybe even better), switch to an io.ReadSeeker entirely?

Hopefully this will save someone some time in the future..

PR #1102, opened 2018-07-06
Use ReadSeeker to calculate ETag if possible

dropje86:
Addresses issue #1100

  If a readseeker is passed as CreateOpts.Content we do not have to copy
  the entire content into a temporary buffer, we can just reset the
  offset instead.

### DIFF
--- a/openstack/objectstorage/v1/objects/requests.go
+++ b/openstack/objectstorage/v1/objects/requests.go
@@ -190,13 +190,26 @@ func (opts CreateOpts) ToObjectCreateParams() (io.Reader, map[string]string, str
 	}
 
 	hash := md5.New()
+
+	if readSeeker, isReadSeeker := opts.Content.(io.ReadSeeker); isReadSeeker {
+		if _, err := io.Copy(hash, readSeeker); err != nil {
+			return nil, nil, "", err
+		}
+		readSeeker.Seek(0, io.SeekStart)
+
+		h["ETag"] = fmt.Sprintf("%x", hash.Sum(nil))
+
+		return readSeeker, h, q.String(), nil
+	}
+
+	// If ETag isn't set by the caller and the Content isn't a ReadSeeker we need to
+	// copy the content into a buffer to be able to calculate the md5 sum.
 	buf := bytes.NewBuffer([]byte{})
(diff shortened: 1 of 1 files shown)


dropje86:
The purpose is to use a ReadSeeker if that's passed as Content to prevent having to copy the content into a temporary buffer in the first place. This change is backwards compatible but way more efficient if you're passing an os.File or anything else that implements Read() and Seek().

So in short, it prevents temporarily storing content into a buffer to calculate the md5 sum while retaining backwards compatibility. It's better than just documentation as the library with this patch will choose the most efficient way of calculating the hash.

jtopjian:
@dropje86 Cool. That makes total sense to me and a really nice improvement.

But on the topic of documentation, we should explain the behavior (both new and old) as a comment on line 146. I realize the other struct fields are missing comments, but let's ensure at least one is commented from this point forward.

In addition, could you provide a unit test to confirm this works without crashing?

dropje86:
@jtopjian Sure, I'll make sure to update with comments (I did add 1 for the old behavior). I just looked into the current tests, this one is covering the "new" behavior already:
```go
func TestCreateObject(t *testing.T) {
	th.SetupHTTP()
	defer th.TeardownHTTP()

	content := "Did gyre and gimble in the wabe"

	HandleCreateTextObjectSuccessfully(t, content)

	options := &objects.CreateOpts{ContentType: "text/plain", Content: strings.NewReader(content)}
	res := objects.Create(fake.ServiceClient(), "testContainer", "testObject", options)
	th.AssertNoErr(t, res.Err)
}
```
objects.CreatOpts.Content is set to `strings.Reader` which implements `Seek()`. I can however explicitly add a test with something that doesn't implement it, do you want me to do that?

jtopjian:
>  I can however explicitly add a test with something that doesn't implement it, do you want me to do that?

Yes, exactly. Something that explicitly tests the new feature and would trigger the new code.

dropje86:
@jtopian I’ve made it a bit nicer by just always putting the content into a readseeker if wasn’t already.

I’ve added 2 tests which cover both paths of passing an object with and without `Seek()`

Issue #1096, opened 2018-07-05
Manila: missing get-share-by-name


gman0:
gophercloud's Manila interface is missing a get-share-by-name functionality. Since there's no such Manila API call, the Manila CLI client circumvents this by listing all shares and searching for the ones with matching name. Given this fact, is gophercloud the correct place to implement such thing as well, or are all the programs requiring such functionality implement get-by-name themselves? I believe the latter option would just lead to unnecessary code duplicity across projects. If it's ok to have it in gophercloud lib, I'm more than happy to deliver it ASAP.

jtopjian:
There's precedence for these kinds of functions in Gophercloud:

https://github.com/gophercloud/gophercloud/search?utf8=%E2%9C%93&q=IDFromName&type=

I've been meaning to move these to https://github.com/gophercloud/utils as I think that's a better home for them. However, rather than have _one_ function in utils while the rest are still here, go ahead and submit a PR for something similar here. In the future I'll move it to utils.

gman0:
Cool, will do!

gman0:
@jtopjian With share's `ListDetail` now merged, is it ok for it to be used in `IDFromName` or should I rather supply the "brief" `List` method? TBH at first i was expecting the get-by-name function would return the whole `Share` and not just its ID, that's why I went for `ListDetail`.

gman0:
I'm ok with either.

jtopjian:
I see no harm in using `ListDetail` as the basis of the `IDFromName` function, but I'm OK with either, too.

PR #1103, opened 2018-07-06
Manila: IDFromName for shares

gman0:
For #1096 

Uses `ListDetail` with the supplied share name in the `ListOpts` for filtering.


jtopjian:
@gman0 This looks good to me. I'm assuming it's ready for review/merge? Just want to confirm.

gman0:
@jtopjian Yes

I'll keep in mind to state that in the future! :D

gman0:
Thanks!

Issue #1098, opened 2018-07-05
Zun: the acceptance test should wait for the capsule to be 'Running' after create


hongbin:
Here is the capsule acceptance test: https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/container/v1/capsules_test.go . The test created a capsule but doesn't wait for the capsule to become 'Running'. It is better to do that.

As a reference, here is how we tested it in tempest: https://review.openstack.org/#/c/546007/

Before starting a PR, please read our [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial).

jtopjian:
@hongbin This is pretty easy to implement. You can see an example of how the acceptance tests wait for Compute Instances to become active here:

https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/compute/v2/compute.go#L937-L957

https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/compute/v2/compute.go#L470-L472

The `WaitFor` function can go in a file called `acceptance/openstack/container/v1/capsules.go` which will allow it to be re-used in other acceptance tests in the future.

hongbin:
@jtopjian , ACK. Thanks for the hints.

PR #1106, opened 2018-07-07
Zun: assert capsule become 'Running' after create

hongbin:
In the Zun acceptance test, add a step to wait for the capsule
to become 'Running' after it is created.

Fixes: #1098

For #[PUT ISSUE NUMBER HERE]

code in this PR:

[PUT URLS HERE]

### DIFF
--- a/acceptance/openstack/container/v1/capsules.go
+++ b/acceptance/openstack/container/v1/capsules.go
@@ -1,0 +1,31 @@
+package v1
+
+import (
+	"fmt"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/container/v1/capsules"
+)
+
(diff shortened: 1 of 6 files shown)


hongbin:
@jtopjian ,

The CI complains the missing of coverage of the new WaitForCapsuleStatus method, but I somehow couldn't find an example of testing the WaitFor* method in other modules. Any advice?

jtopjian:
@hongbin Most of the time coverage failures are acceptable - all depends on the situation.

In this case, the `WaitForCapsuleStatus` function should actually go in a new file called `acceptance/openstack/container/v1/capsules.go`. The files under `acceptance` do not count toward total coverage, so that _should_ resolve the issue. But if it doesn't, no big deal.

hongbin:
@jtopjian cool. Thanks again for your advice

hongbin:
@kevin-zhaoshuai FYI

hongbin:
@jtopjian Yes, I do like this merged

Issue #1107, opened 2018-07-07
Zun: remove internal fields from the Capsule/Container struct


hongbin:
The fields below are for internal usage and should not be exposed via API. We have plan (https://review.openstack.org/#/c/575610/) to remove those fields from the server side. Gophercloud needs to remove those fields as well.

```
	// Container ID for the container
	ContainerID string `json:"container_id"`

	// Websocket url for the container
	WebsocketUrl string `json:"websocket_url"`

	// Websocket token for the container
	WebsocketToken string `json:"websocket_token"`

	// Capsule ID for the container
	CapsuleID int `json:"capsule_id"`

	// Meta for the container
	Meta map[string]string `json:"meta"`

	// ID for the container
	ID int `json:"id"`

	// ID for the capsule
	ID int `json:"id"`
```

PR #1108, opened 2018-07-07
Zun: remove internal fields from Capsule/Container

hongbin:
* Remove the following fields from Capsule: ID
* Remove the following fields from Container: ContainerID, WebsocketUrl,
  WebsocketToken, CapsuleID, Meta, ID.

All those fields are for internal usage and should not be exposed via API.
There is a plan (https://review.openstack.org/#/c/575610/) to remove those
fields from the server side. This is the corresponding change in Gophercloud.

Fixed: #1107

### DIFF
--- a/openstack/container/v1/capsules/results.go
+++ b/openstack/container/v1/capsules/results.go
@@ -44,9 +44,6 @@ type Capsule struct {
 	// UUID for the capsule
 	UUID string `json:"uuid"`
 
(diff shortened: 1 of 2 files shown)


hongbin:
@kevin-zhaoshuai FYI

hongbin:
recheck

hongbin:
@jtopjian Yes, I do like this merged

PR #1112, opened 2018-07-11
Senlin: Clusters CompleteLifecycle

JackKuei:

code in this PR:

[[clusters:completelifecycle]](https://github.com/openstack/senlin/blob/master/senlin/api/openstack/v1/clusters.py#L232)

### DIFF
--- a/openstack/clustering/v1/clusters/doc.go
+++ b/openstack/clustering/v1/clusters/doc.go
@@ -212,5 +212,16 @@ Example to Check a Cluster
 	if err != nil {
 		panic(err)
 	}
+
+Example to Complete Life Cycle
+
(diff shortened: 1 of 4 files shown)


JackKuei:
@jtopjian This PR is ready for review.

dkt26111:
This API was missing from the official Senlin API docs.  I have created a patch set to add:
https://review.openstack.org/#/c/581907/

You can review the output of the doc build here:
http://logs.openstack.org/07/581907/1/check/build-openstack-api-ref/6c97e95/html/#complete-lifecycle-on-a-cluster-action

JackKuei:
@jtopjian let's keep the convention following the python source code then. Will change it to CompleteLifecycle

JackKuei:
@jtopjian Will do more prod testing for this API. Will open bug if there is problem.

Issue #1116, opened 2018-07-13
Bug: Senlin Node's should use "project" field rather than "project_id"


JackKuei:
Environment: DevStack, stable/queen branch
gophercloud version: HEAD of master, 45c2d035713fa44f366921035a44610ac5d45beb

Observed behavior:
When invoking nodes.List(...) the project UUID is returning empty due to the response should use "project" rather than "project_id". 

Expected behaviour:
The project field should be properly populated

jtopjian:
Fixed in #1117

PR #1117, opened 2018-07-13
Fixed Senlin Node's should use "project" field rather than "project_i…

JackKuei:
…d" #1116

For #[1116 [Bug: Senlin Node's should use "project" field rather than "project_id"]](https://github.com/gophercloud/gophercloud/issues/1116)

code in this PR:

[[nodes:get]](https://github.com/openstack/senlin/blob/master/senlin/objects/node.py#L45)


jtopjian:
@JackKuei Since this is a change to how the code interprets the Senlin API service, we need the link to where `project` is defined.

jtopjian:
Oh, duh. I missed the link.

Issue #1122, opened 2018-07-13
Senlin webhook acceptance test is incomplete


dkt26111:
The implementation of Senlin webhook acceptance test is incomplete:
https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/clustering/v1/webhooktrigger_test.go#L27

The acceptance test needs to create profile, cluster and receiver and then call webhook trigger with version option.

dkt26111:
The fix has been merged

PR #1123, opened 2018-07-13
Clean up Senlin webhook implementation and test

dkt26111:
- Clean up webhook test
- Remove unnecessary import for crypto openpgp errors

For #1122 

code in this PR:

https://github.com/openstack/senlin/blob/master/senlin/api/openstack/v1/webhooks.py


jtopjian:
@dkt26111 Travis is reporting a small gofmt fix needed. The order of the imports should be:

```go
import (
  stdlib
  packages

  github.com/foo/bar
)
```

dkt26111:
@jtopjian Done.

Issue #1101, opened 2018-07-05
Support for Orchestration service's "stack update" PATCH request verb.


scott-chaney:
[Openstack Orchestration API](https://developer.openstack.org/api-ref/orchestration/v1/index.html#stacks)

The Openstack Heat orchestration API distinguishes between PATCH and PUT verbs for the stack update call to afford usage of the existing stack template or enforcement of a specified template being passed in to the call.  This is noted under the "existing" request parameter in the API documentation.

Since the PUT verb requires the template object be specified, it would be useful to have support for the PATCH verb to simply update template's parameters when a template is not available locally.

Note: The Openstack API documentation seems slightly misleading.  The template object is required when using PUT regardless of the --existing request parameter (functionality verified in Mitaka and Queens).

Related Openstack code:
[stack's API update_patch() definition](https://github.com/openstack/heat/blob/master/heat/api/openstack/v1/stacks.py#L494)

also see https://bugs.launchpad.net/python-heatclient/+bug/1224828

scott-chaney:
I am working on this btw.

I wasn't able to find another gophercloud feature that differentiated between the PUT and PATCH for updates.  I implemented the logic in the currnet [Update()](https://github.com/gophercloud/gophercloud/blob/master/openstack/orchestration/v1/stacks/requests.go#L333) function using the "exsiting" flag in the UpdateOpts, similar to the Openstack CLI client pattern.  I thought about breaking it out into an UpdatePatch() function, but that seems more contra to GohperCloud's update pattern.

@jtopjian would you or any of the project maintainers have thoughts on the public method to expose this?

jtopjian:
To make sure I'm understanding this correctly: `UpdatePatch` takes the same options as `Update` but the template and template files are optional?

> I thought about breaking it out into an UpdatePatch() function, but that seems more contra to GopherCloud's update pattern.

If my above assumption is true, I think this is the best implementation. Sure, it creates a second "update" call, but if you omit the name from the design, it's really just any other API call. `UpdatePatch` matches what's in the Heat API code, so that's probably a good name to keep.

The alternative of overloading the existing `Update` function and conditionally calling PUT or PATCH would be the type of design that isn't normal for Gophercloud.

Does that help? Or am I misunderstanding this?

scott-chaney:
Your understanding is correct.  I'll go ahead and create a separate function and get rid of the need for the "existing" flag.

Thanks for the input!

scott-chaney:
[...]

PR #1131, opened 2018-07-16
Implements heat/orchestration stack update PATCH verb.

scott-chaney:
For #[[1101](https://github.com/gophercloud/gophercloud/issues/1101)]

Related Openstack code:
https://github.com/openstack/heat/blob/master/heat/api/openstack/v1/stacks.py#L494
https://github.com/openstack/heat/blob/master/heat/api/openstack/v1/__init__.py#L218

Some notes on the implementation:
Both PUT and PATCH update verbs are acceptable, but the PATCH verb allows template parameter updates without needing to specify the template to the API call, where as PUT does require the template.  PATCH does accept a template, but does not delete the environment overlay as PUT does.  See `openstack/orchestration/v1/stacks/doc.go` for more information on PUT vs PATCH functionality.

I ran this implementation by my colleagues and gophercloud contributors @problemv, @JackKuei  and @dkt26111 and we had some uncertainty on the most gophercloud appropriate manner to implement the PUT vs PATCH, specifically on both routes usage of` ToStackUpdateMaps()`.

We explored the following options:
- Current method of having the Update functions pass in an "existing" flag to the `ToStackUpdateMap() `to reason if a template is required
- Adding a private member to the `UpdateOpts` struct, "existing", that is set by the Update functions and then checked via new interface method in the `ToStackUpdateMap()` function
- Using completely divergent paths for `Update()` and `UpdatePatch()` functions (new structs, new interfaces, new functions)
- Moving the logic from the `ToStackUpdateMap()` function into the different update functions

Since this feature of distinguishing between PUT and PATCH is somewhat unique to stack updates in GopherCloud,  we are hoping to get some feedback on which implementation fits best with in the language and project paradigms.

### DIFF
--- a/openstack/orchestration/v1/stacks/doc.go
+++ b/openstack/orchestration/v1/stacks/doc.go
@@ -5,4 +5,121 @@
 // application framework or component specified (in the template). A stack is a
 // running instance of a template. The result of creating a stack is a deployment
 // of the application framework or component.
+
+/* Update Stack examples using the different update methods.
+Summary of behavior between the different methods :
+
+Function | Test Case | Result
+
+Update()	| Template AND Parameters WITH Conflict | Parameter takes priority, parameters are set in raw_template.environment overlay
+Update()	| Template ONLY | Template updates, raw_template.environment overlay is removed
+Update()	| Parameters ONLY | No update, template is required
+
+UpdatePatch() 	| Template AND Parameters WITH Conflict | Parameter takes priority, parameters are set in raw_template.environment overlay
+UpdatePatch() 	| Template ONLY | Template updates, but raw_template.environment overlay is not removed, existing parameter values will remain
+UpdatePatch() 	| Parameters ONLY | Parameters (raw_template.environment) is updated, excluded values are unchanged
+
+The PUT Update() function will remove parameters from the raw_template.environment overlay
+if they are excluded from the operation, whereas PATCH Update() will never be destructive to the
+raw_template.environment overlay.  It is not possible to expose the raw_template values with a
+patch update once they have been added to the environment overlay with the PATCH verb, but
+newly added values that do not have a corresponding key in the overlay will display the
+raw_template value.
+*/
+
+// example stack update using the Update() (PUT) method
+
+/*
+t := make(map[string]interface{})
+f, err := ioutil.ReadFile("template.yaml")
+if err != nil {
+	panic(err)
+}
+err = yaml.Unmarshal(f, t)
+if err != nil {
+	panic(err)
+}
+
+template := stacks.Template{}
+template.TE = stacks.TE{
+	Bin: f,
+}
+
+var params = make(map[string]interface{})
+params["number_of_nodes"] = 2
+
+stackName := "my_stack"
+stackId := "d68cc349-ccc5-4b44-a17d-07f068c01e5a"
+
+stackOpts := &stacks.UpdateOpts{
+	Parameters: params,
+	TemplateOpts: &template,
+}
+
+// client is an instance of *gophercloud.ServiceClient
+res := stacks.Update(client, stackName, stackId, stackOpts)
+if res.Err != nil {
+	panic(res.Err)
+}
+
+//example template.yaml of a Heat::ResourceGroup with 3 nodes
+*/
+
+//example stack update using the UpdatePatch()  (PATCH) method
+
+/*
+var params = make(map[string]interface{})
+params["number_of_nodes"] = 2
+
+stackName := "my_stack"
+stackId := "d68cc349-ccc5-4b44-a17d-07f068c01e5a"
+
+stackOpts := &stacks.UpdateOpts{
+	Parameters: params,
+}
+
+res := stacks.UpdatePatch(client, stackName, stackId, stackOpts)
+if res.Err != nil {
+	panic(res.Err)
+}
+
+
+//example template.yaml of a Heat::ResourceGroup with 3 nodes
+*/
+
+/*
+heat_template_version: 2016-04-08
+
+parameters:
+	number_of_nodes:
+		type: number
+		default: 3
+		description: the number of nodes
+	node_flavor:
+		type: string
+		default: m1.small
+		description: node flavor
+	node_image:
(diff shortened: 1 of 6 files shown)


scott-chaney:
recheck

scott-chaney:
The failure noted on the checks doesn't give much info, and from what I can tell the recheck didn't trigger the check again. Hoping someone who has access to the logs can take a look and let me know if it was a problem on the code side or a hiccup in the CI job. Thanks!

jtopjian:
@scott-chaney It looks like the failure is from Travis and I've gone ahead and restarted the job. The `recheck` command is only for OpenLab tests, confusingly.

jtopjian:
@scott-chaney Thank you (and the rest of the team) for looking into this. 

Given the minor difference between the `Update` and `UpdatePatch` functions, I think we can proceed with reusing `UpdateOpts` for both.

The modification to the interface method's signature (`existing`) works, but I'm a little hesitant to do that just yet.

What if we did something like this: https://github.com/jtopjian/gophercloud/commit/3dd3cc2b7888b47f1da21d1dd25df88cf458783b

The idea is that we extract the common code of building the map to a private function in the package. We then have two interfaces, both of which `UpdateOpts` can satisfy. One of the interface methods guards against a nil `TemplateOpts`. 

I quickly put this together so some comments might need changed, maybe a better name for the private function, etc. Or this might not be a good approach at all.

Thoughts?

scott-chaney:
Hey @jtopjian thanks for the feedback! This looks like a cleaner implementation.  I didn't like having to change the function signature and pass a random bool.  I'll integrate these changes and test them out.  Thanks!

openstack/orchestration/v1/stacks/doc.go
```diff
@@ -5,4 +5,121 @@
 // application framework or component specified (in the template). A stack is a
 // running instance of a template. The result of creating a stack is a deployment
 // of the application framework or component.
+
```
jtopjian: More documentation is always better than no documentation, so we can definitely keep everything here. However, let's strip out all `//` and `/* */` marks from this document and just have one `/*` at the beginning and one `*/` at the end of this file.

openstack/orchestration/v1/stacks/doc.go
```diff
@@ -5,4 +5,121 @@
...
+The PUT Update() function will remove parameters from the raw_template.environment overlay
+if they are excluded from the operation, whereas PATCH Update() will never be destructive to the
+raw_template.environment overlay.  It is not possible to expose the raw_template values with a
+patch update once they have been added to the environment overlay with the PATCH verb, but
+newly added values that do not have a corresponding key in the overlay will display the
+raw_template value.
+*/
+
+// example stack update using the Update() (PUT) method
```
jtopjian: Example to Update a Stack using the Update (PUT) Method

Lines like this should have a "title" format using upper-case when appropriate.

openstack/orchestration/v1/stacks/doc.go
```diff
@@ -5,4 +5,121 @@
...
+
+stackName := "my_stack"
+stackId := "d68cc349-ccc5-4b44-a17d-07f068c01e5a"
+
+stackOpts := &stacks.UpdateOpts{
+	Parameters: params,
+	TemplateOpts: &template,
+}
+
+// client is an instance of *gophercloud.ServiceClient
```
jtopjian: Instead of this line, just call client `orchestrationClient` or something like that.

openstack/orchestration/v1/stacks/doc.go
```diff
@@ -5,4 +5,121 @@
...
+res := stacks.Update(client, stackName, stackId, stackOpts)
+if res.Err != nil {
+	panic(res.Err)
+}
+
+//example template.yaml of a Heat::ResourceGroup with 3 nodes
+*/
+
+//example stack update using the UpdatePatch()  (PATCH) method
```
jtopjian: Example to Update a Stack using the UpdatePatch (PATCH) Method

openstack/orchestration/v1/stacks/requests.go
```diff
@@ -265,42 +265,65 @@ type UpdateOptsBuilder interface {
...
 		return nil, err
 	}
-	opts.TemplateOpts.fixFileRefs()
-	b["template"] = string(opts.TemplateOpts.Bin)
 
 	files := make(map[string]string)
-	for k, v := range opts.TemplateOpts.Files {
-		files[k] = v
+
+	// Template is only required on PUT update
```
jtopjian: We can nix this comment now.

openstack/orchestration/v1/stacks/requests.go
```diff
@@ -340,6 +363,18 @@ func Update(c *gophercloud.ServiceClient, stackName, stackID string, opts Update
 	return
 }
 
+// Update accepts an UpdateOpts struct and updates an existing stack using the
+//  http PATCH verb with the values provided. opts.TemplateOpts is required.
```
jtopjian: opts.TemplateOpts is _not_ required?

openstack/orchestration/v1/stacks/requests.go
```diff
@@ -265,42 +265,65 @@ type UpdateOptsBuilder interface {
 	ToStackUpdateMap() (map[string]interface{}, error)
 }
 
+type UpdatePatchOptsBuilder interface {
```
jtopjian: Let's add a doc comment here:

```
// UpdatePatchOptsBuilder is the interface options structs have to satisfy in order
// to be used in the UpdatePatch operation in this package.
```

scott-chaney:
@jtopjian Thanks for the feedback! I hope that I addressed all of the points that you mentioned.

> I don't use Gophercloud+Heat, so I haven't tried this first-hand -- have you?
Yes, the info in the documentation notes were gathered from functionally running the changes in GopherCloud in one of our Openstack environments for both `Update `and `UpdatePatch` and then inspecting the stack and the Heat database directly.  I also compared the behavior to the direct Heat API calls.

Let me know if there are any other changes you think would improve this.  Thanks again for the help.

PR #1137, opened 2018-07-18
Magnum certificates create

dkt26111:
For #1062 

code in this PR:

https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/certificate.py#L152

### DIFF
--- a/openstack/containerinfra/v1/certificates/doc.go
+++ b/openstack/containerinfra/v1/certificates/doc.go
@@ -11,5 +11,17 @@ Example to get certificates
 		panic(err)
 	}
 
+Example to create certificates
+
+	opts := certificates.CreateOpts{
+		BayUUID:	"d564b18a-2890-4152-be3d-e05d784ff727",
+		Csr:		"-----BEGIN CERTIFICATE REQUEST-----\nMIIEfzCCAmcCAQAwFDESMBAGA1UEAxMJWW91ciBOYW1lMIICIjANBgkqhkiG9w0B\n-----END CERTIFICATE REQUEST-----\n",
+	}
+
+	response, err := certificates.Create(sc, opts).Extract()
+	if err != nil {
+		panic(err)
+	}
+
 */
 package certificates
--- a/openstack/containerinfra/v1/certificates/doc.go
+++ b/openstack/containerinfra/v1/certificates/doc.go
@@ -15,7 +15,7 @@ Example to create certificates
 
 	opts := certificates.CreateOpts{
 		BayUUID:	"d564b18a-2890-4152-be3d-e05d784ff727",
-		Csr:		"-----BEGIN CERTIFICATE REQUEST-----\nMIIEfzCCAmcCAQAwFDESMBAGA1UEAxMJWW91ciBOYW1lMIICIjANBgkqhkiG9w0B\n-----END CERTIFICATE REQUEST-----\n",
+		CSR:		"-----BEGIN CERTIFICATE REQUEST-----\nMIIEfzCCAmcCAQAwFDESMBAGA1UEAxMJWW91ciBOYW1lMIICIjANBgkqhkiG9w0B\n-----END CERTIFICATE REQUEST-----\n",
 	}
 
 	response, err := certificates.Create(sc, opts).Extract()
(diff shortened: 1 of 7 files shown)


jtopjian:
@dkt26111 It's another `gofmt` issue:

```diff
 import (
-	"github.com/gophercloud/gophercloud"
 	"net/http"
+
+	"github.com/gophercloud/gophercloud"
 )
```

If possible, it usually helps to have your editor run `gofmt` upon save.

dkt26111:
Sorry about the format errors.  It was a goimports error and I didn't run that.  But yea, I setup my Intellij to run both gofmt and goimports on save so that should help avoid these types of problems.  Thanks for the tip.

dkt26111:
I have been thinking of having one struct as well.  From the code that you pointed out. it makes sense to have a single struct for both APIs so I'll make that change.

openstack/containerinfra/v1/certificates/doc.go
```diff
@@ -11,5 +11,17 @@ Example to get certificates
 		panic(err)
 	}
 
+Example to create certificates
+
+	opts := certificates.CreateOpts{
+		BayUUID:	"d564b18a-2890-4152-be3d-e05d784ff727",
+		Csr:		"-----BEGIN CERTIFICATE REQUEST-----\nMIIEfzCCAmcCAQAwFDESMBAGA1UEAxMJWW91ciBOYW1lMIICIjANBgkqhkiG9w0B\n-----END CERTIFICATE REQUEST-----\n",
```
jtopjian: CSR

openstack/containerinfra/v1/certificates/requests.go
```diff
@@ -1,9 +1,28 @@
...
 
+// CreateOptsBuilder allows extensions to add additional parameters
+// to the Create request.
+type CreateOptsBuilder interface {
+	ToCertificateCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts represents options used to create a certificate.
+type CreateOpts struct {
+	BayUUID string `json:"bay_uuid" required:"true"`
```
jtopjian: I tried an experiment by adding `ClusterUUID` as a field and, sure enough, it still worked. I think that's due to this line of code: 

https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/certificate.py#L89

Given that, we can have the following in `CreateOpts`:

```go
ClusterUUID string `json:"cluster_uuid,omitempty" xor:"BayUUID"`
BayUUID     string `json:"bay_uuid,omitempty" xor:"ClusterUUID"`
```

`xor` is an internal Gophercloud tag that will require either one or the other.

openstack/containerinfra/v1/certificates/testing/fixtures.go
```diff
@@ -28,10 +28,37 @@ const CertificateResponse = `
 	]
 }`
 
+const CreateCertificateResponse = `
```
jtopjian: Here's the response I'm seeing:

```json
{
  "bay_uuid": "7341c88d-027d-46fe-ac2e-5a0cc0ee8a73",
  "cluster_uuid": "7341c88d-027d-46fe-ac2e-5a0cc0ee8a73",
  "csr": "FOO",
  "links": [
    {
      "href": "http://example.com:9511/v1/certificates/7341c88d-027d-46fe-ac2e-5a0cc0ee8a73",
      "rel": "self"
    },
    {
      "href": "http://example.com:9511/certificates/7341c88d-027d-46fe-ac2e-5a0cc0ee8a73",
      "rel": "bookmark"
    }
  ],
  "pem": "BAR"
}
```

`cluster_uuid` is returned, but I might be testing with a newer version of Magnum.

One thing that might be helpful is to set `OS_DEBUG=1` when running the acceptance tests to see the API traffic:

```shell
$ OS_DEBUG=1 go test -v -tags "fixtures acceptance" -run TestCertificatesCRUD github.com/gophercloud/gophercloud/acceptance/openstack/containerinfra/v1 -count=1
```

Usually I write the acceptance test before the unit test so I can use the acceptance test to generate the fixtures.

jtopjian: > Support both bay_uuid and cluster_uuid in gophercloud?

This is the safest option and is probably the best way to go for the time being. At least then we have initial support for both and can drop the old `bay`-style in the future.

openstack/containerinfra/v1/certificates/doc.go
dkt26111: Done

Issue #1143, opened 2018-07-24
Feature: Add /image/v2/tasks requests


serboox:
Please add this requests in upstream:
1.  Create task -> `
curl -X POST http://openstack/image/v2/tasks -H "X-Auth-Token: ####" -d {
    "type": "import",
    "input": {
        "import_from": "http://app-catalog.openstack.example.org/groovy-image",
        "import_from_format": "qcow2",
        "image_properties": {
            "disk_format": "vhd",
            "container_format": "ovf"
        }
    }
}
`
2. List tasks -> `curl -X GET http://openstack/image/v2/tasks -H "X-Auth-Token: ####"`
3. Show task details -> `curl -X GET http://openstack/image/v2/tasks/{task_id} -H "X-Auth-Token: ####"`

Documentation [link](https://developer.openstack.org/api-ref/image/v2/)

ozerovandrei:
To clarify some things.

Imageservice tasks could be deprecated after couple of releases in favor of the [Import API](https://developer.openstack.org/api-ref/image/v2/#interoperable-image-import).
But that API is implemented in the Queens release only (Image API v2.6).

There is a deprecation message in the code: [tasks.py#L47](https://github.com/openstack/glance/blob/stable/queens/glance/api/v2/tasks.py#L47).

Additional documentation:
[Tasks spec](https://docs.openstack.org/glance/latest/admin/tasks.html)
[Import-refactor spec](https://specs.openstack.org/openstack/glance-specs/specs/mitaka/approved/image-import/image-import-refactor.html)

What I propose is to add the Imageservice Import API in addition to tasks API.
I can create an additional issue or I can just add some PRs into that issue.

ozerovandrei:
All PRs are ready for review.

PR #1147, opened 2018-07-25
Imageservice V2: implement tasks Get method

ozerovandrei:
Implement Get method for the Imageservice tasks.
Add unit test with documentation example.

For #1143

code in this PR:

API
https://github.com/openstack/glance/blob/stable/queens/glance/api/v2/tasks.py#L121
https://github.com/openstack/glance/blob/stable/queens/glance/api/v2/tasks.py#L309

DB
https://github.com/openstack/glance/blob/stable/queens/glance/db/sqlalchemy/api.py#L1559

### DIFF
--- a/openstack/imageservice/v2/tasks/results.go
+++ b/openstack/imageservice/v2/tasks/results.go
@@ -1,0 +1,117 @@
+package tasks
+
+import (
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// TaskStatus represents valid task status.
+type TaskStatus string
+
+const (
+	// TaskStatusPending represents status of the pending task.
+	TaskStatusPending TaskStatus = "pending"
+
+	// TaskStatusProcessing represents status of the processing task.
+	TaskStatusProcessing TaskStatus = "processing"
+
+	// TaskStatusSuccess represents status of the success task.
+	TaskStatusSuccess TaskStatus = "success"
+
+	// TaskStatusFailure represents status of the failure task.
+	TaskStatusFailure TaskStatus = "failure"
+)
+
+// Task represents a single task of the OpenStack Image service.
+type Task struct {
+	// ID is a unique identifier of the task.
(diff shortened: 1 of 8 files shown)


openstack/imageservice/v2/tasks/results.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// GetResult represents the result of a Get operation. Call its Extract
+// method to interpret it as a Task.
+type GetResult struct {
+	commonResult
+}
+
+// TaskStatus represents valid task status.
+type TaskStatus string
```
jtopjian: Let's move this + the consts to `requests.go`

openstack/imageservice/v2/tasks/results.go
```diff
@@ -0,0 +1,123 @@
...
+// Task represents a single task of the OpenStack Image service.
+type Task struct {
+	// ID is a unique identifier of the task.
+	ID string `json:"id"`
+
+	// Type represents the type of the task.
+	Type string `json:"type"`
+
+	// Status represents current status of the task.
+	Status TaskStatus `json:"status"`
```
jtopjian: This should be a `string`

ozerovandrei:
@jtopjian 
Oh well I've gotten those custom types all wrong then. I've been thinking that they're here to allow only well-known values (`IPv4` and `IPv6` for the `IPVersion` for example).

As for documentation. I think that we need at least add inline docs for the fields that can have pre-defined well known values.
In that case those people who are checking `godoc` or reading those comments will know that they can use additional well-known values for some fields.

I've updated all PRs and they're ready.

jtopjian:
> I've been thinking that they're here to allow only well-known values (IPv4 and IPv6 for the IPVersion for example)

Right - I used to think the same, but then I noticed that some fields are _not_ like this. Upon further investigation and discussion, those fields were accidentally glanced over and need to be cleaned up at some point. I still accidentally let one slip by every now and then. I'll try to comb through Gophercloud and get them all fixed at once. 

> As for documentation. I think that we need at least add inline docs for the fields that can have pre-defined well known values.

Good point.

> I've updated all PRs and they're ready.

Sounds good. I'll review when I have a moment. Thank you!

Issue #1151, opened 2018-07-26
Hotfix: ToServerRebootMap


serboox:
Link to method: [link](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/requests.go)
Now:
// ToServerRebootMap builds a body for the reboot request.
`func (opts *RebootOpts) ToServerRebootMap() (map[string]interface{}, error) {
	return gophercloud.BuildRequestBody(opts, "reboot")
}`
We get an error:
`cannot use rebootOpts (type servers.RebootOpts) as type servers.RebootOptsBuilder in argument to servers.Reboot:
  servers.RebootOpts does not implement servers.RebootOptsBuilder (ToServerRebootMap method has pointer receiver)`

Should be:
// ToServerRebootMap builds a body for the reboot request.
`func (opts RebootOpts) ToServerRebootMap() (map[string]interface{}, error) {
	return gophercloud.BuildRequestBody(opts, "reboot")
}`

jtopjian:
Fixed in #1152

PR #1152, opened 2018-07-26
Fix ToServerRebootMap() method

serboox:
For #1151 
Documentation: [link](https://godoc.org/github.com/gophercloud/gophercloud/openstack/compute/v2/servers)

Example to Reboot a Server
``` go
rebootOpts := servers.RebootOpts{
	Type: servers.SoftReboot,
}

serverID := "d9072956-1560-487c-97f2-18bdf65ec749"

err := servers.Reboot(computeClient, serverID, rebootOpts).ExtractErr()
if err != nil {
	panic(err)
}
```

### DIFF
--- a/acceptance/openstack/compute/v2/servers_test.go
+++ b/acceptance/openstack/compute/v2/servers_test.go
@@ -271,7 +271,7 @@ func TestServersActionReboot(t *testing.T) {
 	th.AssertNoErr(t, err)
 	defer DeleteServer(t, client, server)
 
-	rebootOpts := &servers.RebootOpts{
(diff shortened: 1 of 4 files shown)


jtopjian:
@serboox Thanks for fixing this. None of our tests were showing the error reported in #1151, but for consistency, I agree with the fix.

Just a side note, the tests that were updated in https://github.com/gophercloud/gophercloud/pull/1152/commits/ff9c3032a5a36dab414a2d0f7f4e55855717fcf1 didn't have to be made. I'm OK with committing them since it's still legal code.

Thanks again for working on this.

Issue #1144, opened 2018-07-24
Feature: Add /network/v2.0/network-ip-availability requests


serboox:
Please add this requests in upstream:
1. List Network IP Availability -> `curl -X GET http://openstack/network/v2.0/network-ip-availability -H "X-Auth-Token: ####"`
2. Show Network IP Availability -> `curl -X GET http://openstack/network/v2.0/network-ip-availabilities/{network_id} -H "X-Auth-Token: ####"`

Documentation [link](https://developer.openstack.org/api-ref/network/v2/)

jtopjian:
Implemented in #1154

PR #1153, opened 2018-07-26
Networking V2: Implement IP availabilities List

ozerovandrei:
Add the new Networking service extensions "networkipavailabilities" package with the List method.
Add unit and acceptance tests with documentation.

For #1144 

code in this PR:

API definition:
https://github.com/openstack/neutron-lib/blob/stable/queens/neutron_lib/api/definitions/network_ip_availability.py

Filtetring options:
https://github.com/openstack/neutron/blob/master/neutron/extensions/network_ip_availability.py#L47
https://github.com/openstack/neutron/blob/stable/pike/neutron/db/network_ip_availability_db.py#L28

DB structs populations (network and subnet availability):
https://github.com/openstack/neutron/blob/stable/pike/neutron/db/network_ip_availability_db.py#L153
https://github.com/openstack/neutron/blob/stable/pike/neutron/db/network_ip_availability_db.py#L170

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/networkipavailabilities/networkipavailabilities_test.go
+++ b/acceptance/openstack/networking/v2/extensions/networkipavailabilities/networkipavailabilities_test.go
@@ -1,0 +1,32 @@
+// +build acceptance networking networkipavailabilities
+
+package networkipavailabilities
+
+import (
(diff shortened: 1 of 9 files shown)


jtopjian:
Implemented in #1154

Issue #1158, opened 2018-07-27
Compute V2: remote consoles support


ozerovandrei:
The following OpenStack Compute service actions were deprecated:
 - `os-getRDPConsole`
 - `os-getSerialConsole`
 - `os-getSPICEConsole`
 - `os-getVNCConsole`

Instead, users must use a unified Server Remote Consoles API.

Documentation:
[API](https://developer.openstack.org/api-ref/compute/#server-remote-consoles)
[Spec](https://specs.openstack.org/openstack/nova-specs/specs/liberty/implemented/consolidate-console-api.html)

Right now there is only single `Create` request for that API and I want to add it in the Gophercloud.

PR #1159, opened 2018-07-27
Compute V2: implement remoteconsoles Create method

ozerovandrei:
Add the new "remoteconsoles" Compute extension package with needed
structures and Create method.
Add unit and acceptance test with documentation.

For #1158

code in this PR:

API create request handler:
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/remote_consoles.py#L160

API definitions:
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/schemas/remote_consoles.py#L91
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/schemas/remote_consoles.py#L115

Allowed protocols and types:
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/schemas/remote_consoles.py#L123
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/schemas/remote_consoles.py#L127

### DIFF
--- a/openstack/compute/v2/extensions/remoteconsoles/doc.go
+++ b/openstack/compute/v2/extensions/remoteconsoles/doc.go
@@ -1,14 +1,20 @@
 /*
 Package remoteconsoles provides the ability to create server remote consoles
 through the Compute API.
+You need to specify at least "2.6" microversion for the ComputeClient to use
+that API.
 
 Example of Creating a new RemoteConsole
+
+  computeClient, err := openstack.NewComputeV2(providerClient, endpointOptions)
+  computeClient.Microversion = "2.6"
 
   createOpts := remoteconsoles.CreateOpts{
     Protocol: string(remoteconsoles.RemoteConsoleVNCProtocol),
     Type:     string(remoteconsoles.RemoteConsoleNoVNCType),
 	}
-	serverID := "b16ba811-199d-4ffd-8839-ba96c1185a67"
+  serverID := "b16ba811-199d-4ffd-8839-ba96c1185a67"
+
   remtoteConsole, err := remoteconsoles.Create(computeClient, serverID, createOpts).Extract()
   if err != nil {
     panic(err)
--- a/openstack/compute/v2/extensions/remoteconsoles/doc.go
+++ b/openstack/compute/v2/extensions/remoteconsoles/doc.go
@@ -1,0 +1,19 @@
+/*
+Package remoteconsoles provides the ability to create server remote consoles
+through the Compute API.
+
+Example of Creating a new RemoteConsole
+
+  createOpts := remoteconsoles.CreateOpts{
+    Protocol: string(remoteconsoles.RemoteConsoleVNCProtocol),
+    Type:     string(remoteconsoles.RemoteConsoleNoVNCType),
+	}
+	serverID := "b16ba811-199d-4ffd-8839-ba96c1185a67"
+  remtoteConsole, err := remoteconsoles.Create(computeClient, serverID, createOpts).Extract()
+  if err != nil {
+    panic(err)
+  }
+
+  fmt.Printf("%+v\n", task)
+*/
+package remoteconsoles
--- a/openstack/compute/v2/extensions/remoteconsoles/doc.go
+++ b/openstack/compute/v2/extensions/remoteconsoles/doc.go
@@ -10,9 +10,9 @@ Example of Creating a new RemoteConsole
   computeClient.Microversion = "2.6"
 
   createOpts := remoteconsoles.CreateOpts{
-    Protocol: string(remoteconsoles.RemoteConsoleVNCProtocol),
-    Type:     string(remoteconsoles.RemoteConsoleNoVNCType),
-	}
+    Protocol: remoteconsoles.ConsoleProtocolVNC,
+    Type:     remoteconsoles.ConsoleTypeNoVNC,
+  }
   serverID := "b16ba811-199d-4ffd-8839-ba96c1185a67"
 
   remtoteConsole, err := remoteconsoles.Create(computeClient, serverID, createOpts).Extract()
@@ -20,6 +20,6 @@ Example of Creating a new RemoteConsole
     panic(err)
   }
 
-  fmt.Printf("%+v\n", task)
+  fmt.Printf("Console URL: %s\n", remtoteConsole.URL)
 */
 package remoteconsoles
--- a/openstack/compute/v2/extensions/remoteconsoles/requests.go
+++ b/openstack/compute/v2/extensions/remoteconsoles/requests.go
@@ -4,49 +4,49 @@ import (
 	"github.com/gophercloud/gophercloud"
 )
 
-// RemoteConsoleProtocol represents valid remote console protocol.
+// ConsoleProtocol represents valid remote console protocol.
 // It can be used to create a remote console with one of the pre-defined protocol.
-type RemoteConsoleProtocol string
+type ConsoleProtocol string
 
 const (
-	// RemoteConsoleVNCProtocol represents the VNC console protocol.
-	RemoteConsoleVNCProtocol RemoteConsoleProtocol = "vnc"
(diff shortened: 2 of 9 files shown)


ozerovandrei:
Well, this won't work until we pass the `X-OpenStack-Nova-API-Version` header with at least `2.6` version value.

openstack/compute/v2/extensions/remoteconsoles/requests.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/gophercloud/gophercloud"
+)
+
+// RemoteConsoleProtocol represents valid remote console protocol.
+// It can be used to create a remote console with one of the pre-defined protocol.
+type RemoteConsoleProtocol string
+
+const (
+	// RemoteConsoleVNCProtocol represents the VNC console protocol.
+	RemoteConsoleVNCProtocol RemoteConsoleProtocol = "vnc"
```
jtopjian: How about `ConsoleProtocolVNC`. Same pattern for the rest.

openstack/compute/v2/extensions/remoteconsoles/requests.go
```diff
@@ -0,0 +1,83 @@
...
+	RemoteConsoleMKSProtocol RemoteConsoleProtocol = "mks"
+)
+
+// RemoteConsoleType represents valid remote console type.
+// It can be used to create a remote console with one of the pre-defined type.
+type RemoteConsoleType string
+
+const (
+	// RemoteConsoleNoVNCType represents the VNC console type.
+	RemoteConsoleNoVNCType RemoteConsoleType = "novnc"
```
jtopjian: Similar here: `ConsoleTypeNoVNC`. Same pattern for the rest, where the unique part is at the end of the name.

openstack/compute/v2/extensions/remoteconsoles/requests.go
```diff
@@ -0,0 +1,83 @@
...
+
+// CreateOptsBuilder allows to add additional parameters to the Create request.
+type CreateOptsBuilder interface {
+	ToRemoteConsoleCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts specifies parameters to the Create request.
+type CreateOpts struct {
+	// Protocol specifies the protocol of a new remote console.
+	Protocol string `json:"protocol" required:"true"`
```
jtopjian: Should this be:

```go
Protocol RemoteConsoleProtocol `json:"protocol" reqauired:"true"`
```

openstack/compute/v2/extensions/remoteconsoles/requests.go
```diff
@@ -0,0 +1,83 @@
...
+
+	// Type specifies the type of a new remote console.
+	Type string `json:"type" required:"true"`
```
jtopjian: Should this be:

```go
Type RemoteConsoleType `json:"type" required:"true"`
```

openstack/compute/v2/extensions/remoteconsoles/requests.go
```diff
@@ -0,0 +1,83 @@
+package remoteconsoles
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// RemoteConsoleProtocol represents valid remote console protocol.
+// It can be used to create a remote console with one of the pre-defined protocol.
+type RemoteConsoleProtocol string
```
jtopjian: Since the package name is `remoteconsoles`, I think we can shorten this to:

```go
type ConsoleProtocol string
```

And the name will still have a clear description of the type.

openstack/compute/v2/extensions/remoteconsoles/requests.go
```diff
@@ -0,0 +1,83 @@
...
+	// RemoteConsoleSerialProtocol represents the serial console protocol.
+	RemoteConsoleSerialProtocol RemoteConsoleProtocol = "serial"
+
+	// RemoteConsoleMKSProtocol represents the MKS console protocol.
+	RemoteConsoleMKSProtocol RemoteConsoleProtocol = "mks"
+)
+
+// RemoteConsoleType represents valid remote console type.
+// It can be used to create a remote console with one of the pre-defined type.
+type RemoteConsoleType string
```
jtopjian: Similar here, we can probably shorten this to:

```go
type ConsoleType string
```

But feel free to keep the original name if you feel it's better.

openstack/compute/v2/extensions/remoteconsoles/doc.go
```diff
@@ -0,0 +1,25 @@
...
+
+Example of Creating a new RemoteConsole
+
+  computeClient, err := openstack.NewComputeV2(providerClient, endpointOptions)
+  computeClient.Microversion = "2.6"
+
+  createOpts := remoteconsoles.CreateOpts{
+    Protocol: string(remoteconsoles.RemoteConsoleVNCProtocol),
+    Type:     string(remoteconsoles.RemoteConsoleNoVNCType),
+	}
```
jtopjian: nit: indentation needs fixed here.

openstack/compute/v2/extensions/remoteconsoles/requests.go
ozerovandrei: It looks like the same issue as we discussed in [PR#1147](https://github.com/gophercloud/gophercloud/pull/1147#issuecomment-407868998).

Do you think that we need to strict users to pre-defined values in case of this change? I don't actually get how it's different from the #1147 😅

ozerovandrei:
> Let's wait for a decision with the other microversion stuff. If it ends up taking too long, I might go ahead and merge this. Thoughts?

I kinda need this API implementation 😅. Also I don't want to create deprecated stuff that can work without microversions (`os-getRDPConsole`, `os-getSerialConsole`, `os-getSPICEConsole`, `os-getVNCConsole`) or to implement my own temporary solution in the project that I work with.

Maybe you can merge this as an exception? This package differs from PRs #1163 or #1115 because microversions won't change some parts of the responses or options, they completely turn it on or off.

openstack/compute/v2/extensions/remoteconsoles/requests.go
ozerovandrei: Seems better, yeah.

openstack/compute/v2/extensions/remoteconsoles/requests.go
ozerovandrei: Seems better, yeah.

openstack/compute/v2/extensions/remoteconsoles/requests.go
jtopjian: No worries - since it's not actually written anywhere, it's hard to grasp :)

The rule is: custom types for "input", standard types for "output". So any type of `CreatOpts`, `UpdateOpts`, etc should have types which can provide better/safer input. Result structures should only contain "standard" (`string`, `int`, `[]string`, etc fields). If a struct contains a struct, then that can be a defined struct.

ozerovandrei: Alright, seems legit.

ozerovandrei:
recheck

ozerovandrei:
recheck

ozerovandrei:
recheck

ozerovandrei:
@jtopjian what do you think about current state of this PR after microversion implementation in Gophercloud?

Tricky part is: we can't disable/enable some fields in request opts, response results. All of the Nova API remoteconsoles controller won't be available without setting the Compute microversion.

ozerovandrei:
recheck

jtopjian:
> Tricky part is: we can't disable/enable some fields in request opts, response results. All of the Nova API remoteconsoles controller won't be available without setting the Compute microversion.

If I understand correctly, this should be fine. Well, kind of fine. You will just have to set the microversion on the client when calling making any request to `remoteconsoles`.

ozerovandrei:
recheck

Issue #1160, opened 2018-07-27
Compute V2: support for extended server information


ozerovandrei:
Currently OpenStack Compute service provides additional `OS-*` fields in the server struct that needs to be unmarshalled.
All of the extended fields are implemented in the separated extensions inside the Nova API. So it seems that it is better to implement them as separate packages in the Gophercloud.

For now Gophercloud implements only `extendedstatus` for the `OS-EXT-STS` fields (#659).
We need to also add support for the following info:

- [x] `OS-EXT-STS` Extended Status Extension.
- [x] `OS-DCF` Disk Management Extension.
- [x] `OS-EXT-AZ` Extended Availability Zone support. 
- [x] `OS-EXT-SRV-ATTR` partial Extended Server Attributes support. (#1163)
- [x] `OS-EXT-SRV-ATTR` full (depends on #1115 microversioning) Extended Server Attributes support (#1501)
- [x] `OS-SRV-USG` Usage (launched_at and terminated_at) support. (#1161)

ozerovandrei:
It seems that `OS-DCF` extension was merged into main server struct:
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/views/servers.py#L150

So what's better way to implement that field?
Do I need to separate it in the different extension package or it is completely fine to add it into the `Server` struct?

jtopjian:
@ozerovandrei Thank you for spotting the missing features :)

The disk config extension is already defined here:

https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions/diskconfig

If both the request and response are part of the base API, we could add it to the base server `CreateOpts` and `Server` struct.

Availability Zones are already defined here: https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/availabilityzones/results.go

It looks like `OS-EXT-AZ` is indeed missing from Gophercloud. I can review this shortly.

For `OS-EXT-SRV-ATTR`: since it requires a microversion, I'm hesitant to merge it right now. There should be a solution once #1115 is merged.

ozerovandrei:
@jtopjian oh, I missed those two packages for some reason.

Well, I personally wanted to propose a separate package for the `OS-DCF` so the main `compute/v2/servers` package won't become a huge mess (at least for now).
But it seems that Nova extensions are in slow progress of being merged into single module inside the Nova API server. So I guess i some future we will actually need to merge all of those stuff into one package.
[...]

PR #1161, opened 2018-07-27
Compute V2: implement server OS-SRV-USG extension

ozerovandrei:
Add the new "serverusage" package with basic structure and UnmarshalJSON
method.
Add unit test and update TestServersWithExtensionsCreateDestroy aceptance
test.
Add documentation example.

For #1160 

code in this PR:

Main class:
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/server_usage.py#L22

### DIFF
--- a/acceptance/openstack/compute/v2/servers_test.go
+++ b/acceptance/openstack/compute/v2/servers_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/extendedstatus"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/lockunlock"
 	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/pauseunpause"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/serverusage"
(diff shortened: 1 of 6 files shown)


acceptance/openstack/compute/v2/servers_test.go
```diff
@@ -105,6 +107,7 @@ func TestServersWithExtensionsCreateDestroy(t *testing.T) {
 	th.AssertEquals(t, int(extendedServer.PowerState), extendedstatus.RUNNING)
 	th.AssertEquals(t, extendedServer.TaskState, "")
 	th.AssertEquals(t, extendedServer.VmState, "active")
+	th.AssertEquals(t, extendedServer.LaunchedAt, extendedServer.Updated)
```
jtopjian: The first time I ran this, it worked, but on the second attempt, these two values were slightly different:

```
"OS-SRV-USG:launched_at": "2018-07-28T02:17:57.000000"
"updated": "2018-07-28T02:18:02Z"
```

How about we check to see if `LaunchedAt` does not equal a zero-value `time.Time`?

ozerovandrei:
recheck

ozerovandrei:
recheck

PR #1163, opened 2018-07-27
Compute V2: add server OS-EXT-SRV-ATTR extension

ozerovandrei:
Add the new "extendedserverattributes" package with basic structure.
Add unit test and update TestServersWithExtensionsCreateDestroy aceptance test.
Add documentation example.

For #1160

code in this PR:

Main class:
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/extended_server_attributes.py#L24


ozerovandrei:
@jtopjian yeah, this is ready.

I've also updated issue #1160 to not forget adding all other fields in this extension.

PR #1165, opened 2018-07-30
Imageservice V2: add imageimport Get request

ozerovandrei:
Add the new "imageimport" package and implement Get request to retrieve Import API information.
Add unit and acceptance tests with documentation.

For #1143 

code in this PR:

Main class:
https://github.com/openstack/glance/blob/stable/queens/glance/api/v2/discovery.py#L26

Import methods are populated from configuration file:
https://github.com/openstack/glance/blob/stable/queens/glance/api/v2/discovery.py#L38
https://github.com/openstack/glance/blob/stable/queens/glance/common/config.py#L729

### DIFF
--- a/openstack/imageservice/v2/imageimport/requests.go
+++ b/openstack/imageservice/v2/imageimport/requests.go
@@ -1,0 +1,20 @@
+package imageimport
+
+import "github.com/gophercloud/gophercloud"
+
+// ImportMethod represents valid Import API method.
+type ImportMethod string
+
+const (
+	// GlanceDirectMethod represents glance-direct Import API method.
+	GlanceDirectMethod ImportMethod = "glance-direct"
+
+	// WebDownloadMethod represents web-download Import API method.
+	WebDownloadMethod ImportMethod = "web-download"
+)
+
+// Get retrieves Import API information data.
+func Get(c *gophercloud.ServiceClient) (r GetResult) {
+	_, r.Err = c.Get(infoURL(c), &r.Body, nil)
+	return
+}
--- a/openstack/imageservice/v2/imageimport/results.go
+++ b/openstack/imageservice/v2/imageimport/results.go
@@ -1,0 +1,32 @@
+package imageimport
+
+import "github.com/gophercloud/gophercloud"
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// GetResult represents the result of a get operation. Call its Extract method
+// to interpret it as ImportInfo.
+type GetResult struct {
+	commonResult
+}
+
+// ImportInfo represents information data for the Import API.
+type ImportInfo struct {
+	ImportMethods ImportMethods `json:"import-methods"`
+}
+
+// ImportMethods contains information about available Import API methods.
+type ImportMethods struct {
+	Description string   `json:"description"`
+	Type        string   `json:"type"`
+	Value       []string `json:"value"`
+}
+
+// Extract is a function that accepts a result and extracts ImportInfo.
+func (r commonResult) Extract() (*ImportInfo, error) {
+	var s *ImportInfo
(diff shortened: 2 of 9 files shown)


openstack/imageservice/v2/imageimport/results.go
```diff
@@ -0,0 +1,32 @@
...
+
+// GetResult represents the result of a get operation. Call its Extract method
+// to interpret it as ImportInfo.
+type GetResult struct {
+	commonResult
+}
+
+// ImportInfo represents information data for the Import API.
+type ImportInfo struct {
+	ImportMethods ImportMethods `json:"import-methods"`
```
jtopjian: Fields in `results.go` (which can be thought of as "outputs") get standard types. In this case `string`.

openstack/imageservice/v2/imageimport/requests.go
```diff
@@ -0,0 +1,20 @@
...
+const (
+	// GlanceDirectMethod represents glance-direct Import API method.
+	GlanceDirectMethod ImportMethod = "glance-direct"
+
+	// WebDownloadMethod represents web-download Import API method.
+	WebDownloadMethod ImportMethod = "web-download"
+)
+
+// Get retrieves Import API information data.
+func Get(c *gophercloud.ServiceClient) (r GetResult) {
```
jtopjian: Since this API call isn't returning anything about an import task, should we rename this to `GetImportMethods`? It might make it more clear to users.

openstack/imageservice/v2/imageimport/results.go
jtopjian: You are absolutely correct. I'm sorry!

openstack/imageservice/v2/imageimport/results.go
```diff
@@ -0,0 +1,32 @@
...
+	Type        string   `json:"type"`
+	Value       []string `json:"value"`
+}
+
+// Extract is a function that accepts a result and extracts ImportInfo.
+func (r commonResult) Extract() (*ImportInfo, error) {
+	var s *ImportInfo
+	err := r.ExtractInto(&s)
+	return s, err
+}
```
jtopjian: @ozerovandrei If `ImportInfo` is only being used as a "shell" for `ImportMethods`, then we should be able to do this:

```go
func (r commonResult) Extract() (*ImportMethods, error) {
  var s struct {
    ImportMethods ImportMethods `json:"import-methods"`
  }
  err := r.ExtractInto(&s)
  return s.ImportMethods, err
}
```

This removes the need for the `ImportInfo` struct.

ozerovandrei: I disagree because in the [API reference](https://developer.openstack.org/api-ref/image/v2/#import-methods-and-values-discovery) it says:
`In the Image API v2.6, this discovery call contains only the import-methods field.`

And if you research the [Importing API spec](https://specs.openstack.org/openstack/glance-specs/specs/mitaka/approved/image-import/image-import-refactor.html), you will find the much bigger schema for the `GET v2/info/import` value discovery (it's really huge).

So i think `ImportInfo` response will grow bigger in the future so it's better to have current two structures in that case.

jtopjian: OK sounds good.

openstack/imageservice/v2/imageimport/requests.go
jtopjian: Given this might return more than just import methods in the future, `GetImportMethods` is probably not a good name.

I guess since this is the only `GET` method under `imageimports` at this time, we can keep this as `Get`. That feels a little strange, but I guess that's how it is :)

ozerovandrei:
@jtopjian I think this is ready.

PR #1166, opened 2018-07-30
Imageservice V2: add Stage request to imagedata

ozerovandrei:
Add Stage request for the imagedata package with tests and documentation.
Provide acceptance test with new helper functions to work with local image files.

For #1143

code in this PR:

API route:
https://github.com/openstack/glance/blob/stable/queens/glance/api/v2/router.py#L471

Main method:
https://github.com/openstack/glance/blob/stable/queens/glance/api/v2/image_data.py#L270


ozerovandrei:
Implementation was added into the `imagedata` package because it's implemented in the `image_data.py` in the Glance API server.

ozerovandrei:
@jtopjian this is ready.

Issue #425, opened 2017-07-19
Compute v2: Make Rescue an Extension


jtopjian:
rescue is currently defined under the [`servers`](https://github.com/gophercloud/gophercloud/blob/177ebbf9d6d507338a9e9d7edeabd0cd26b9f9a9/openstack/compute/v2/servers/requests.go#L515-L546) package, but this should be moved to a dedicated extension since the `rescue` action is an extension:

https://github.com/openstack/nova/blob/stable/ocata/nova/api/openstack/compute/rescue.py

jtopjian:
I just noticed that Nova is in the process of removing the concept of "extensions". You can see the difference between the original link I gave vs what is in the master branch:

https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/rescue.py

The file layout looks to be the same, though. `rescue.py` is not being merged into `servers.py`.

I think it still makes sense to break rescue out of the `servers` package. If all possible actions to a server were housed under the `servers` package, it would become quite unwieldy. 

I propose moving all packages currently under `extensions` to the `compute/v2` directory. I thought about suggesting the creation of an `actions` or `serveractions` directory, but given some packages  in `extensions` are not actions (ex: keypairs, servergroups, etc), it would make organizing various compute features difficult and bike-sheddy.

@jrperritt thoughts?

jrperritt:
> thoughts?

I'd like to hear more about it. Can you elaborate on the desire to refactor this? Is it just to stay in sync with the OpenStack repo's structure? Are there any other services (or, in general, code) in Gophercloud that require a change like this? The reason I ask is because we'll have to weigh the benefit of that against the weight of this change breaking all the code using any of those extensions (if those users update their Gophercloud code to a commit with this change, of course).

jtopjian:
The Gerrit topic for this change is here: https://review.openstack.org/#/c/453571/

It looks like the move is 90% complete. I did a search through Nova and found a few extensions that haven't been changed yet ([ex](https://github.com/openstack/nova/blob/5a39fe58ffbd3dab78543052c972989cc645262b/nova/api/openstack/compute/flavor_rxtx.py).

> Is it just to stay in sync with the OpenStack repo's structure?

It's mostly that. 

[...]

PR #1168, opened 2018-07-31
Compute V2: migrate rescue to separate package

ozerovandrei:
Move the Rescue action of the Compute service to its own package to leverage common Gophercloud structuring.
Add more options, documentation and tests.

For #425

code in this PR:

Main method:
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/rescue.py#L43

New `rescue_image_ref ` option:
https://github.com/openstack/nova/blob/stable/queens/nova/api/openstack/compute/rescue.py#L47

### DIFF
--- a/openstack/compute/v2/extensions/rescueunrescue/results.go
+++ b/openstack/compute/v2/extensions/rescueunrescue/results.go
@@ -1,0 +1,18 @@
+package rescueunrescue
+
+import "github.com/gophercloud/gophercloud"
+
+// RescueResult is the response from a Rescue operation. Call its ExtractErr
+// method to determine if the call succeeded or failed.
+type RescueResult struct {
+	gophercloud.ErrResult
+}
+
+// Extract interprets any RescueResult as an AdminPass, if possible.
+func (r RescueResult) Extract() (string, error) {
+	var s struct {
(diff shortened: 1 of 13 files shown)


openstack/compute/v2/extensions/rescueunrescue/results.go
```diff
@@ -0,0 +1,18 @@
+package rescueunrescue
+
+import "github.com/gophercloud/gophercloud"
+
+// RescueResult is the response from a Rescue operation. Call its ExtractErr
+// method to determine if the call succeeded or failed.
+type RescueResult struct {
+	gophercloud.ErrResult
```
jtopjian: I realize this is how `RescueResult` is currently implemented, but I wonder if we can take the opportunity to make this more aligned with the rest of Gophercloud.

There are probably a few edge cases, but normally `ErrResult` is for responses that wouldn't contain a body - only an error.

What are your thoughts on making this `gophercloud.Result`?

ozerovandrei: Yes, it seems more appropriate, I'll change it.

jtopjian:
Implemented in #1169

Issue #1171, opened 2018-07-31
Senlin Bug: Policies Delete Should Use ExtractErr Instead Of Extract


JackKuei:
To keep it consistent with other gophercloud senlin APIs, rather than use Extract with DeleteHeader, caller should use ExtractErr, and obtain the header with something like res.Header["X-Openstack-Request-Id"][0]

[DeleteHeader source link](https://github.com/gophercloud/gophercloud/blob/master/openstack/clustering/v1/policies/results.go#L142 )
```golang
// DeleteHeader contains the delete information from a delete policy request.
type DeleteHeader struct {
	RequestID string `json:"X-OpenStack-Request-ID"`
}
```

[Extract source link](https://github.com/gophercloud/gophercloud/blob/master/openstack/clustering/v1/policies/results.go#L147)
```golang
func (r DeleteResult) Extract() (*DeleteHeader, error) {
	var s *DeleteHeader
	err := r.HeaderResult.ExtractInto(&s)
	return s, err
}
```

jtopjian:
Fixed in #1172

PR #1172, opened 2018-07-31
Senlin Bug: Policies Delete Should Use ExtractErr Instead Of Extract

JackKuei:

For #[1171 [Senlin Bug: Policies Delete Should Use ExtractErr Instead Of Extract]](https://github.com/gophercloud/gophercloud/issues/1171)

code in this PR:

[[policies::delete]](https://github.com/openstack/senlin/blob/master/senlin/api/openstack/v1/policies.py#L89)


JackKuei:
@jtopjian Think coverage percentage went down because deleted lines of codes.

jtopjian:
I think we can live with -0.01 ;)

In general, we're not too concerned about coverage calculations. Of course unit and acceptance tests should be provided where applicable. But the actual calculation had some bugs, so no big deal.

JackKuei:
@jtopjian Yeah, just want to let you know that did look at cause of the reason of the coverage percentage going down. This one is ready to be merged.

PR #1173, opened 2018-07-31
Magnum Certificates Update implementation

dkt26111:
For #1062 

code in this PR:

https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/certificate.py#L171


dkt26111:
@jtopjian This is ready for review when you get a chance.  Thanks.

PR #1183, opened 2018-08-14
Update README.md

sc68cal:
Fix link to issue tracker.

`/issues` was being transformed into https://github.com/gophercloud/gophercloud/blob/master/issues, which 404's.


jtopjian:
@sc68cal Thanks for noticing this and fixing it.

Just a heads up that per the Pull Request template, all PRs need to have a corresponding issue, no matter how small the PR is. I've gone ahead and opened #1184 as a reference for this PR.

Issue #1185, opened 2018-08-14
Magnum: ClusterTemplate APIServerPort should be int instead of string


ozerovandrei:
Hello.

I've just started implementing Magnum Cluster templates resource for the [terraform-provider-openstack#402](https://github.com/terraform-providers/terraform-provider-openstack/issues/402) and I got a JSON unmarshalling error regarding "apiserver_port" field.

It seems that there should be `int` type instead of `string`: [cluster_template.py#L82](https://github.com/openstack/magnum/blob/stable/rocky/magnum/api/controllers/v1/cluster_template.py#L82).

jtopjian:
Fixed in #1186

PR #1186, opened 2018-08-14
Magnum: fix Cluster templates APIServerPort type

ozerovandrei:
OpenStack Magnum Cluster template response uses int type instead of string.
This commit fixes ClusterTemplate field type and unit tests.

For #1185 

code in this PR:

https://github.com/openstack/magnum/blob/stable/rocky/magnum/api/controllers/v1/cluster_template.py#L82

### DIFF
--- a/openstack/containerinfra/v1/clustertemplates/testing/fixtures.go
+++ b/openstack/containerinfra/v1/clustertemplates/testing/fixtures.go
@@ -14,7 +14,7 @@ import (
 
 const ClusterTemplateResponse = `
 {
-	"apiserver_port": null,
+	"apiserver_port": 8081,
 	"cluster_distro": "fedora-atomic",
 	"coe": "kubernetes",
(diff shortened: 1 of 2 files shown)


openstack/containerinfra/v1/clustertemplates/testing/fixtures.go
```diff
@@ -232,7 +232,6 @@ var ExpectedClusterTemplate = clustertemplates.ClusterTemplate{
 }
 
 var ExpectedClusterTemplate_EmptyTime = clustertemplates.ClusterTemplate{
-	APIServerPort:       "",
```
JackKuei: Would it be better to have it as APIServerPort: 0?

jtopjian: six one way, half a dozen the other, I think.

Having it set to 0 would be more explicit, but both ways are valid and work.

Issue #1187, opened 2018-08-17
Image Service v2: Support Updating Properties


jtopjian:
The `imageservice/v2/images` package currently lacks support for updating image properties:

https://github.com/openstack/glance/blob/b421f18a62f0b6cdd17761a0429724580ce29608/glance/api/v2/images.py#L273-L328

jtopjian:
Implemented in #1188

PR #1188, opened 2018-08-17
Image Service v2: Support Updating Properties

jtopjian:
This commit adds support to update image properties. Technically it can
update any image field with a string value, so this duplicates the
functionality of some existing Update calls. However, there's no other
way to implement this since the OpenStack Image Service treats all
properties as first-class fields.

For #1187


jtopjian:
The acceptance test failure is unrelated to this PR. The acceptance test included in this PR passes.

The failed acceptance test might be a temporary error or caused by a recent change in Keystone. I will have to look at this separately.

I'm going to merge this PR, though. I've been trying to clean up a bunch of image issues and this PR should be the last piece.

Issue #1182, opened 2018-08-14
Infinite loop on reauth process


garbage-collector:
I was performing some tests on gophercloud when I encountered an issue.

Let's imagine that compute always give a 401 answer but keystone is always answering 201 on token creation.

This situation ends with an infinite loop on reauthentication process.

You can easily reproduce the issue with this code (I'm using gock to fake the OpenStack API):

```golang
package main

import (
	"log"

	"github.com/gophercloud/gophercloud"
	"github.com/gophercloud/gophercloud/openstack"
	"github.com/gophercloud/gophercloud/openstack/compute/v2/servers"
	"github.com/gophercloud/gophercloud/openstack/identity/v3/tokens/testing"
	"github.com/h2non/gock"
)

func main() {
	gock.New("http://127.0.0.1:8774/v2.1/a99e9b4e620e4db09a2dfb6e42a01e66").
		Get("/servers/detail").
		MatchHeader("X-Auth-Token", "TOKEN").
		Persist().
		Reply(401)

	gock.New("http://127.0.0.1:5000/v3/").
		Post("/auth/token").
		MatchType("json").
		Persist().
		Reply(201).
		SetHeader("X-Subject-Token", "TOKEN").
		JSON(testing.TokenOutput)

	opts := gophercloud.AuthOptions{
		IdentityEndpoint: "http://127.0.0.1:5000/v3/",
		Username:         "{username}",
		Password:         "{password}",
		DomainID:         "{domain-id}",
		AllowReauth:      true,
	}

	provider, err := openstack.AuthenticatedClient(opts)
	if err != nil {
		log.Fatal(err)
	}

	client, err := openstack.NewComputeV2(provider, gophercloud.EndpointOpts{
		Region: "RegionOne",
	})
	if err != nil {
		log.Fatal(err)
	}

	pages, err := servers.List(client, nil).AllPages()
	if err == nil {
		log.Fatal("authentication required expected")
	}
	if pages != nil {
		log.Fatal("pages expected to be nil")
	}
}
```

This code will have an infinite loop, without any logs or print, so don't forget to stop the program.

After digging the code, the only behaviour handled by the reauthentication are those two:
* Compute returns HTTP 401
* Keystone returns HTTP 201
* Compute returns HTTP 200

or

* Compute returns HTTP 401
* Keystone returns HTTP 401

This scenario does not seems to be handled:
* Compute returns HTTP 401
* Keystone returns HTTP 201
* Compute returns HTTP 401

[...]

PR #1192, opened 2018-08-20
Set reauth func to nil before making a new request

garbage-collector:
For #1182 

This patch avoids infinite loop on reauthentication process in case of component is always answering 401.

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -295,7 +295,11 @@ func (client *ProviderClient) Request(method, url string, options *RequestOpts) 
 						seeker.Seek(0, 0)
 					}
 				}
+				// make a new call to request with a nil reauth func in order to avoid infinite loop
+				reauthFunc := client.ReauthFunc
+				client.ReauthFunc = nil
 				resp, err = client.Request(method, url, options)
+				client.ReauthFunc = reauthFunc
 				if err != nil {
 					switch err.(type) {
 					case *ErrUnexpectedResponseCode:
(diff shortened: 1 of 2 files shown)


jtopjian:
@garbage-collector Thank you for working on this. You can ignore the test failure for now - I need to fix that in a separate PR.

Forgive my ignorance, but I want to make sure I understand the effects of this implementation (I'm cautious about changing the Reauth stuff in case it creates unintended side effects):

You're setting `client.ReauthFunc` to nil before making the post-reauth request in order to stem the `ReauthFunc` from being triggered again, effectively making the reauthentication system a flat system that has no possibility of calling itself, right?

garbage-collector:
@jtopjian yes, this patch aims to avoid the ReauthFunc being triggered again and again on the same call, while keeping the provider client clean for another call to another API (this is why the ReauthFunc is set again after the call).
By setting the ReauthFunc to nil, we ensure that the code will not go through the Reauth process a second time.

jtopjian:
This makes sense to me.

Would it be possible to include a unit test (sans-Gock, though that's only due to the external dependency)? This will help prevent any regressions in the future.

garbage-collector:
For sure. I'll add some unit tests asap :)

jtopjian:
@garbage-collector Thank you! 😄 

@jrperritt Any thoughts/concerns on this PR?

garbage-collector:
Tests have been added !

Issue #1199, opened 2018-08-24
Workflow V2: Adding Mistral APIs


garbage-collector:
It looks like gophercloud is missing workflow v2 implementation: https://docs.openstack.org/mistral/latest/

I'd be happy to propose a Pull Request to handle Mistral APIs, what do you think about this?

I've already prepared the [branche here](https://github.com/garbage-collector/gophercloud/tree/workflow-api), waiting your "LGTM" to propose the Pull Request.

Thanks !

jtopjian:
Very cool. 

If you haven't yet, please read the [contributor's tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial) which should explain the right process for this (let me know if it's unclear). Basically, please do not open one large PR of Mistral - it'll have to be broken out into several PRs.

When someone already has a large body of work done, I admit that this can be a pain, but it has proven worthwhile every time.

The current Magnum issue (https://github.com/gophercloud/gophercloud/issues/1062) might be a good template to go off of.

Short of that, this sounds good to me and we really appreciate the contribution. :)

garbage-collector:
Ok, I'll have a look on this next week, I'll split my actual work into 3 PRs (I've worked on the workflows, executions and crontriggers APIs).

Acceptance tests are working on a DevStack with Mistral installed, so last work is to split into different PRs, not such a pain :)

Thanks !

jrperritt:
It will need more than 3 PRs. See [here](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-05-pull-requests.md#adding-an-entire-openstack-project).

garbage-collector:
Here it is!

I really tried to follow all the contributing process, feel free to tell if I'm doing it wrong.

[Mistral API Doc](https://docs.openstack.org/mistral/latest/api/index.html)
[Mistral source code](https://github.com/openstack/mistral)

---

[Workflows](https://docs.openstack.org/mistral/latest/api/v2.html#workflows)

- [x] Manage Workflow: Create
- [x] Manage Workflow: Delete
- [x] Manage Workflow: Get
- [x] Manage Workflow: List
- [ ] Manage Workflow: Update

---

[Executions](https://docs.openstack.org/mistral/latest/api/v2.html#executions)

- [x] Manage Execution: Create
- [x] Manage Execution: Delete
- [x] Manage Execution: Get
- [x] Manage Execution: List
- [ ] Manage Execution: Update

---

[...]

PR #1201, opened 2018-08-27
Add Workflowv2 clients

garbage-collector:
For #1199 

https://github.com/openstack/mistral/blob/master/devstack/plugin.sh#L41


jtopjian:
Added in #1202

PR #1202, opened 2018-08-27
Workflowv2 workflows create

garbage-collector:

For #1199 

https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/workflow.py#L141

### DIFF
--- a/openstack/workflow/v2/workflows/requests.go
+++ b/openstack/workflow/v2/workflows/requests.go
@@ -1,0 +1,56 @@
+package workflows
+
+import (
+	"io"
+
+	"github.com/gophercloud/gophercloud"
+)
+
+// CreateOptsBuilder allows extension to add additional parameters to the Create request.
+type CreateOptsBuilder interface {
+	ToWorkflowCreateQuery() (io.Reader, string, error)
+}
+
+// CreateOpts specifies parameters used to create a cron trigger.
+type CreateOpts struct {
+	// Scope is the scope of the workflow.
+	// Allowed values are "private" and "public".
+	Scope string `q:"scope"`
+
+	// Namespace will define the namespace of the workflow.
+	Namespace string `q:"namespace"`
+
+	// Definition is the workflow definition written in Mistral Workflow Language v2.
+	Definition io.Reader
+}
+
+// ToWorkflowCreateQuery constructs a request query string from CreateOpts.
+func (opts CreateOpts) ToWorkflowCreateQuery() (io.Reader, string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return opts.Definition, q.String(), err
+}
+
+// Create requests the creation of a new execution.
+func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResult) {
+	url := createURL(client)
+	var b io.Reader
+	if opts != nil {
+		tmpB, query, err := opts.ToWorkflowCreateQuery()
+		if err != nil {
+			r.Err = err
+			return
+		}
+		url += query
+		b = tmpB
+	}
+
+	_, r.Err = client.Post(url, nil, &r.Body, &gophercloud.RequestOpts{
+		RawBody: b,
+		MoreHeaders: map[string]string{
+			"Content-Type": "text/plain",
+			"Accept":       "", // Drop default JSON Accept header
+		},
+	})
+
+	return
+}
--- a/openstack/workflow/v2/workflows/requests.go
+++ b/openstack/workflow/v2/workflows/requests.go
@@ -8,7 +8,7 @@ import (
 
 // CreateOptsBuilder allows extension to add additional parameters to the Create request.
 type CreateOptsBuilder interface {
-	ToWorkflowCreateQuery() (io.Reader, string, error)
+	ToWorkflowCreateParams() (io.Reader, string, error)
 }
 
 // CreateOpts specifies parameters used to create a cron trigger.
@@ -24,8 +24,8 @@ type CreateOpts struct {
 	Definition io.Reader
 }
 
-// ToWorkflowCreateQuery constructs a request query string from CreateOpts.
-func (opts CreateOpts) ToWorkflowCreateQuery() (io.Reader, string, error) {
+// ToWorkflowCreateParams constructs a request query string from CreateOpts.
+func (opts CreateOpts) ToWorkflowCreateParams() (io.Reader, string, error) {
 	q, err := gophercloud.BuildQueryString(opts)
 	return opts.Definition, q.String(), err
 }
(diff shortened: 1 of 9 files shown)


garbage-collector:
Ready to review!

openstack/workflow/v2/workflows/requests.go
```diff
@@ -0,0 +1,56 @@
...
+	var b io.Reader
+	if opts != nil {
+		tmpB, query, err := opts.ToWorkflowCreateQuery()
+		if err != nil {
+			r.Err = err
+			return
+		}
+		url += query
+		b = tmpB
+	}
```
jtopjian: Can this instead be:

```go
if opts != nil {
  query, err := opts.ToWorkflowCreatQuery()
  if err != nil {
    r.Err = err
    return
  }
  url += query
  b = opts.Definition
}
```

Or is there a reason to have `opts.Definition` come from `ToWorkflowCreateQuery()`?

openstack/workflow/v2/workflows/results.go
```diff
@@ -0,0 +1,56 @@
...
+	gophercloud.Result
+}
+
+// CreateResult is the response of a Post operations. Call its Extract method to interpret it as a Workflow.
+type CreateResult struct {
+	commonResult
+}
+
+// Extract helps to get a Workflow struct from a Get function.
+func (r commonResult) Extract() (*Workflow, error) {
```
jtopjian: If I'm understanding correctly, "create" will return a list (https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/workflow.py#L169-L173), even if there's only one workflow, but GET will return a single workflow?

If so, then maybe have a `GetResult` and a `CreateResult` rather than one inheriting from the other - might add better separation.

Also, slight nit, but this shouldn't be in this PR. No doubt this was accidental, though, so it's OK to keep it. 😄

openstack/workflow/v2/workflows/results.go
```diff
@@ -0,0 +1,56 @@
...
+func (r CreateResult) Extract() ([]Workflow, error) {
+	var s struct {
+		Workflows []Workflow `json:"workflows"`
+	}
+	err := r.ExtractInto(&s)
+	return s.Workflows, err
+}
+
+// Workflow represents a workflow execution on OpenStack mistral API.
+type Workflow struct {
```
jtopjian: It appears these fields are defined here:

https://github.com/openstack/mistral/blob/632e4bda67ca1e6b579607f264f258890b0c1d8e/mistral/api/controllers/v2/resources.py#L88-L101

(for future PRs, we'll need references like this to validate the returned fields and their types)

It looks like this is missing a few fields such as tags, created_at, and updated_at?

openstack/workflow/v2/workflows/requests.go
garbage-collector: I based this from what I saw in objectstorage package, because workflows have the same way of working regarding the create process: the POST request takes a definition file as body (text/plain), and other parameters are sent through URL query.

This is how it's handled on objectstorage:
https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/objects/requests.go#L171 

I just kept the same signature. I'll probably rename the function "ToWorkflowCreateQuery" to "ToWorkflowCreateParams" to be less ambiguous.

We can't make your proposal as we have a CreateOptsBuilder which is an interface, so no access to opts.Definition.

openstack/workflow/v2/workflows/results.go
garbage-collector: Added missing fields

garbage-collector:
Thanks for the review !

All comments have been answered and fixed, and I rebased all workflowsv2-workflows-* branches with the required fixes.

But it looks like I've made some mess with my rebases as we lost some coverage on files that I didn't touch at all.

Can you help me to understand what happened ? Not really used to git workflows like this (dependencies between branches and stuff like this), so I'd love to have some help to understand what happened and what I missed.

openstack/workflow/v2/workflows/requests.go
jtopjian: > I based this from what I saw in objectstorage package, because workflows have the same way of working regarding the create process: the POST request takes a definition file as body (text/plain), and other parameters are sent through URL query.

Right, and re-using existing patterns is definitely the way to go. At least as a start.

What I was pointing out is that you aren't actually doing anything with `Definition` inside `ToWorkflowCreateParams`, though. For example, in the `objects` package, the `Content` is used to create the etag hash.

My question was more _will_ you ever be doing anything with `Definition`? 

However,

> We can't make your proposal as we have a CreateOptsBuilder which is an interface, so no access to opts.Definition.

Oh, right - duh. I totally overlooked that. Busy day :)

garbage-collector:
BTW @jtopjian :
I've ran the various tests and they were successfull. I can provide you an access to a devstack working with Mistral if you want, just give me your pubkey and I can spawn a devstack to perform the acceptance tests. Not an issue to build one :)

openstack/workflow/v2/workflows/requests.go
jtopjian: > What do you think about it?

Nah, that'll be a large break from the current patterns. Let's leave it as-is.

jtopjian:
> But it looks like I've made some mess with my rebases as we lost some coverage on files that I didn't touch at all.

Don't worry about the test coverage results. It can be quite buggy at times. 

As well, you don't need to worry about the OpenLab tests, either, since Mistral isn't set up. Any OpenLab failures are probably transient. 

This is why running the acceptance tests offline is important for this series of PRs.

> I can provide you an access to a devstack working with Mistral if you want

Thank you for the offer! I should be able to get Mistral set up within the next week or so - just pressed for time at the moment :)

openstack/workflow/v2/workflows/results.go
```diff
@@ -53,4 +42,13 @@ type Workflow struct {
 	// Scope is the scope of the workflow.
 	// Values can be "private" or "public".
 	Scope string `json:"scope"`
+
+	// Tags is a list of tags associated to the workflow.
+	Tags []string `json:"tags"`
+
+	// CreatedAt is the creation date of the workflow.
+	CreatedAt gophercloud.JSONRFC3339ZNoTNoZ `json:"created_at"`
```
jtopjian: We'll want this to be of type `time.Time` and you'll need to implement a custom `UnmarshalJSON` func. See here for an example: https://github.com/gophercloud/gophercloud/blob/master/docs/FAQ.md#overriding-default-unmarshaljson-method

Or just grep/search for `UnmarshalJSON` in this repository as this happens frequently.

openstack/workflow/v2/workflows/results.go
```diff
@@ -53,4 +42,13 @@ type Workflow struct {
...
+
+	// UpdatedAt is the last update date of the workflow.
+	UpdatedAt *gophercloud.JSONRFC3339ZNoTNoZ `json:"updated_at"`
```
jtopjian: Same here

garbage-collector:
Fixed Created and Updated field to time.Time.

I rebased all the workflows PR with this.

Thanks :)

PR #1205, opened 2018-08-27
Workflowv2 workflows list

garbage-collector:
For #1199 

https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/workflow.py#L201


garbage-collector:
Rebased and ready to review

PR #1209, opened 2018-08-27
Workflowv2 executions list

garbage-collector:
For #1199 

https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/execution.py#L321

### DIFF
--- a/openstack/workflow/v2/executions/requests.go
+++ b/openstack/workflow/v2/executions/requests.go
@@ -1,6 +1,15 @@
 package executions
 
-import "github.com/gophercloud/gophercloud"
+import (
+	"encoding/json"
+	"fmt"
+	"net/url"
+	"reflect"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
 
 // CreateOptsBuilder allows extension to add additional parameters to the Create request.
 type CreateOptsBuilder interface {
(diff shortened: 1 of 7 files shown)


garbage-collector:
Ready to review!

garbage-collector:
Need rebase, waiting workflowv2-workflows-* branches to be accepted before rebasing it.

garbage-collector:
Ready to review

openstack/workflow/v2/executions/requests.go
```diff
@@ -64,3 +67,55 @@ func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
...
 	return
 }
+
+// ListOptsBuilder allows extension to add additional parameters to the List request.
+type ListOptsBuilder interface {
+	ToExecutionListQuery() (string, error)
+}
+
+// ListOpts filters the result returned by the List() function.
+type ListOpts struct {
```
jtopjian: Same as the other PR. It looks like there are some missing fields as well as possibly renaming some fields:

https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/execution.py#L321-L327

jtopjian:
@garbage-collector It looks like all of the fields are included, but I think the API implementation supports more than just specifying values.

Looking at this section of code:

https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/execution.py#L376-L396

It appears there are two kinds of list parameters: "standard" and "filters".

Standard appears to be: marker, limit, sort_keys, sort_dirs, all_projects, marker, limit, sort_keys, sort_dirs, filters, all_projects.

Filters appear to be everything else.

A filtered parameter can be in the form of either:

```
value

filter_type:value
```

as described here: https://github.com/openstack/mistral/blob/211e3964111816ef52588f914d4b7821e05ab625/mistral/utils/filter_utils.py

There are two ways to go about this:

1. Just change all filter params to a `string` type and have the user build the filter themselves. This includes `created_at` and `deleted_at` where the user would have to convert the time to a string themselves.

2. Go all out and create a `ListFilter` struct that looks kind of like:

```go
type ListFilter struct {
  FilterType FilterType
  Value string
}
```

Where `FilterType` is a set of string consts like `eq`, `neq`, etc.

You could get even more detailed and make different `ListFilters` depending on the type. For example, the `CreatedAt` and `UpdatedAt` fields would have a `ListDateFilter` that took a `time.Time` for their `Value`.

There's an example of this here: https://github.com/gophercloud/gophercloud/blob/master/openstack/imageservice/v2/images/types.go#L85-L104

in short: you can get really detailed with this. It also shows just how much can go undocumented in the API docs :)

I'm OK with going with Option 1. If you or someone wanted to put the time into Option 2, by all means go for it. The Image filter I linked to above was added quite a long time after the original implementation, so there's somewhat precedence for being lazy with this kind of situation.

As well, for weird filters like `all_projects` that _should_ be a bool but take an integer (`=1`), just use a type of `int` (or `*int` if specifying 0 is important). We've done that in a few other areas.

garbage-collector:
I just updated the Pull Request with the requested changes.

I implemented the second solution, with special types for filtering.

I definitely need to take more time on reading OpenStack code, it's helpful in this case of situation.

You will probably notice in the PR that some parameters from [here](https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/execution.py#L376-L396) are not implemented as eligible for filtering.
Despite the fact that they are defined in this part of code as "filterable", the [call definition](https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/execution.py#L315-L327) expects special types for some fields and checks the inputs.

For example, ```workflow_id``` expects to be a valid ```uuid```.
I tried it on my devstack, and if you try to filter with values like ```workflow_id=eq:8531789a-92d1-4bd9-9d56-8a1bd28b2a7b```, it ends with an error: ```Invalid input for field/attribute workflow_id. Value: 'eq:8531789a-92d1-4bd9-9d56-8a1bd28b2a7b'. unable to convert to uuid. Error: Expected a uuid but received eq:8531789a-92d1-4bd9-9d56-8a1bd28b2a7b.```.

As it's not working with the current workflow API definition, I choose do not allow filtering on those fields.

Thanks for the tips, I hope this time is the good time !

garbage-collector:
I tested it, and json fields are checking for strict equality :(

openstack/workflow/v2/executions/requests.go
```diff
@@ -64,3 +73,165 @@ func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
...
+func (l ListDateFilter) String() string {
+	return fmt.Sprintf("%s%s", l.Filter, l.Value.Format(gophercloud.RFC3339ZNoTNoZ))
+}
+
+// FilterType represents a valid filter to use for filtering executions.
+type FilterType string
+
+const (
+	// FilterEq checks equality.
+	FilterEq = "eq"
```
jtopjian: Let's use capital letters where applicable, like how it's done here: https://github.com/gophercloud/gophercloud/blob/master/openstack/imageservice/v2/images/types.go#L90-L95

So, capitalize everything except `Has` since that's an actual word. Change `NotEq` to `NEQ`, etc.

openstack/workflow/v2/executions/requests.go
```diff
@@ -64,3 +73,165 @@ func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
...
+
+// ListFilter allows to filter string parameters with different filters.
+// Empty value for Filter checks for equality.
+type ListFilter struct {
+	Filter FilterType
+	Value  string
+}
+
+func (l ListFilter) String() string {
+	return fmt.Sprintf("%s%s", l.Filter, l.Value)
```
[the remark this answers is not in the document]
garbage-collector: Fixed. The ```String()``` function for ```FilterType``` was a helper for handling zero value of filter, but I moved the code elsewhere.

garbage-collector:
Ready to review (again) ! :)

I also fixed the tests.

Issue #1214, opened 2018-08-27
Implement Keystone Application Credentials support


zioproto:
This is a feature request.

Keystone Application Credentials seems to be not yet supported by gophercloud:

https://docs.openstack.org/keystone/queens/user/application_credentials.html

Because these credentials are meant to be used with scripts and deployments tools, it would be very nice to have them supported in gophercloud.

Thank you

zioproto:
Related github issues blocked by this missing feature:

- https://github.com/appscode/guard/issues/183
- https://github.com/kubernetes/cloud-provider-openstack/issues/232

jtopjian:
Hi @zioproto !

> Related github issues blocked by this missing feature:

Just to give you a realistic expectation, the most likely way for this to get added is for someone to submit it. I've added a "help wanted" label to more easily identify it.

zioproto:
@jtopjian does it make sense to copy the folder [tokens](https://github.com/gophercloud/gophercloud/tree/master/openstack/identity/v3/tokens) into a new folder called `application_credentials` to use the same code structure ? I would like to try to implement the basic functionality of authenticating with the application credential instead of the token. It is possible to add later the functionalities like create/delete of application credentials. If you can point me to the right direction on how to start I can try to implement this. Thank you

jtopjian:
> does it make sense to copy the folder tokens into a new folder called application_credentials to use the same code structure ?

Off the top of my head, I'm not sure. I can review the application credential API later on today and let you know if that makes sense. However, if you're just looking for a template to use, I recommend reading the [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial) since it includes a template as well as a lot of other useful information on how to quickly get code merged.

Let me know if you need any help or have any questions. I'm at a conference this week, but I'll do my best to get back quickly 😄

zioproto:
@jtopjian I have a question, but take your time to answer because I am leaving the office and I will be offline the next 16 hours ;) 

My 1st goal is to issue a user token, but instead of username and password I want to use a keystone application credential.

As explained here:
https://specs.openstack.org/openstack/keystone-specs/specs/keystone/backlog/application-credentials.html

The URL will be `POST /v3/auth/tokens`

and the example payload looks like:

```
    "auth": {
        "identity": {
            "methods": [
                "application_credential"
            ],
            "application_credential": {
                "name": "backup",
                "project": {
[...]

PR #1224, opened 2018-09-03
Add support to authenticate with application credential

zioproto:
For #1214 

code in this PR:

- https://github.com/openstack/keystoneauth/blob/stable/rocky/keystoneauth1/identity/v3/application_credential.py#L48-L67
 - https://github.com/openstack/keystone/blob/stable/rocky/keystone/conf/constants.py#L20
 - https://github.com/openstack/python-keystoneclient/blob/stable/rocky/keystoneclient/v3/application_credentials.py#L42

### DIFF
--- a/auth_options.go
+++ b/auth_options.go
@@ -84,6 +84,12 @@ type AuthOptions struct {
 
 	// Scope determines the scoping of the authentication request.
 	Scope *AuthScope `json:"-"`
+
+	// Authentication through Application Credentials requires supplying name, project and secret
+	// For project we can use TenantID
+	ApplicationCredentialID     string `json:"-"`
+	ApplicationCredentialName   string `json:"-"`
+	ApplicationCredentialSecret string `json:"-"`
 }
 
 // AuthScope allows a created token to be limited to a specific domain or project.
@@ -142,7 +148,7 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
 	type userReq struct {
 		ID       *string    `json:"id,omitempty"`
 		Name     *string    `json:"name,omitempty"`
-		Password string     `json:"password"`
+		Password string     `json:"password,omitempty"`
 		Domain   *domainReq `json:"domain,omitempty"`
 	}
 
@@ -154,10 +160,18 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
 		ID string `json:"id"`
 	}
 
+	type applicationCredentialReq struct {
+		ID     *string  `json:"id,omitempty"`
+		Name   *string  `json:"name,omitempty"`
+		User   *userReq `json:"user,omitempty"`
+		Secret *string  `json:"secret,omitempty"`
+	}
+
 	type identityReq struct {
-		Methods  []string     `json:"methods"`
-		Password *passwordReq `json:"password,omitempty"`
-		Token    *tokenReq    `json:"token,omitempty"`
+		Methods               []string                  `json:"methods"`
+		Password              *passwordReq              `json:"password,omitempty"`
+		Token                 *tokenReq                 `json:"token,omitempty"`
+		ApplicationCredential *applicationCredentialReq `json:"application_credential,omitempty"`
 	}
 
 	type authReq struct {
@@ -171,6 +185,7 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
 	// Populate the request structure based on the provided arguments. Create and return an error
 	// if insufficient or incompatible information is present.
 	var req request
+	var userRequest userReq
 
 	if opts.Password == "" {
 		if opts.TokenID != "" {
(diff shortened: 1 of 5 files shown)


jtopjian:
@zioproto Travis is flagging a `go fmt` error. If you can, run [`go fmt`](https://blog.golang.org/go-fmt-your-code) on the changed files - that should fix things up.

zioproto:
@rdl87  I pushed a new version of the patch that works with any possible combination of ID or NAME env vars. Can you review it ? Are you working on the tests ?

RdL87:
@zioproto i'm working on this just now. So i'm going to recompile the client with the new library you've just pushed. I will ping you as soon as the keystone become available.

zioproto:
@jtopjian I believe we are ready for review. I added two tests. I tried to to give correct code pointers to keystone codebase. thank you

auth_options.go
```diff
@@ -194,8 +209,35 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
 			req.Auth.Identity.Token = &tokenReq{
 				ID: opts.TokenID,
 			}
+
+		} else if opts.ApplicationCredentialID != "" {
+			// Configure the request for ApplicationCredentialID authentication.
+			// https://github.com/openstack/keystone-specs/blob/master/specs/keystone/queens/application-credentials.rst#using-an-application-credential-to-obtain-a-token
```
jtopjian: Let's replace the above URL with:

https://github.com/openstack/keystoneauth/blob/ebe781a3ea0386d6ff088a84e8dde26e538b856d/keystoneauth1/identity/v3/application_credential.py#L48-L67

auth_options.go
```diff
@@ -194,8 +209,35 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
...
+			// There are three kinds of possible application_credential requests
+			// 1. application_credential id + secret.
+			// 2. application_credential name + user_id + secret
+			// 3. application_credential name + user_name + user_domain_id/user_domain_name
```
jtopjian: How about rewording this to:

1. application_credential id + secret
2. application_credential name + secret + user_id
3. application_credential name + secret + username + domain_id / domain_name

Secondly, since an app credential secret is always required (I _think_ it's required, right? https://github.com/openstack/keystoneauth/blob/c2ae9e298e2acdc4e653bce18735294cf4df2454/keystoneauth1/loading/_plugins/identity/v3.py#L270-L272), right below this line, let's make sure one is set:

```go
if opts.ApplicationCredentialSecret == "" {
  return nil, ErrAppCredMissingSecret{}
}
```

You can add the new error to the bottom of this file: https://github.com/gophercloud/gophercloud/blob/master/errors.go#L453

auth_options.go
```diff
@@ -194,8 +209,35 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
...
+			req.Auth.Identity.Methods = []string{"application_credential"}
+			req.Auth.Identity.ApplicationCredential = &applicationCredentialReq{
+				ID:     &opts.ApplicationCredentialID,
+				Secret: &opts.ApplicationCredentialSecret,
+			}
+		} else if opts.ApplicationCredentialName != "" {
```
jtopjian: Again, right below this line, let's make sure a secret was provided - same way as above.

In addition, let's also make sure that only one of `DomainName` or `DomainID` were provided, the same way it's done in other places in this file:

```go
if opts.DomainID == "" && opts.DomainName == "" {
  return nil, ErrDomainIDOrDomainName{}
}
```

auth_options.go
```diff
@@ -194,8 +209,35 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
...
+			req.Auth.Identity.Methods = []string{"application_credential"}
+			if opts.DomainID != "" {
+				userRequest = userReq{Name: &opts.Username, Domain: &domainReq{ID: &opts.DomainID}}
```
jtopjian: nit: use multiple lines:

```go
userRequest = userReq{
  Name: &opts.Username,
  ...
}
```

Same for the other occurrences.

auth_options.go
```diff
@@ -194,8 +209,35 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
...
+			} else if opts.DomainName != "" {
+				userRequest = userReq{Name: &opts.Username, Domain: &domainReq{Name: &opts.DomainName}}
+			} else {
+				userRequest = userReq{Name: &opts.Username}
```
jtopjian: I don't think this variation is possible?

zioproto: Probably you are right. @RdL87 I will drop this

zioproto: Yes you are right this cannot work without domain. If we had implemented using UsedID that it would have make sense to create a request without DomainID/Name, but it is not our case.

zioproto:
Thanks for the review. I pushed changes that should address all the points of the review

zioproto:
@jtopjian I am not sure what the problem with the coveralls test not passing

jtopjian:
@zioproto You can ignore Coveralls. It acts weird sometimes.

Issue #1141, opened 2018-07-24
proposal: add ProviderClient.AuthResult() method


majewsky:
ProviderClient (and hence, ServiceClient) is responsible for obtaining a Keystone token (i.e. sending the `POST /v3/auth/tokens` request to Keystone), but it only allows very limited access to the response to that request. The token itself can be retrieved via `ProviderClient.Token()`, and the service catalog is indirectly accessible via `ProviderClient.EndpointLocator`, but that's it.

There are other usecases where you need to look at the token:

1. The response contains the list of roles that you have, which may be interesting to selectively enable/disable functionality in the client.

2. My own usecase (which prompted this proposal) is that I'm building a service that needs to manipulate role assignments for its own service user. It therefore needs to know its own user ID. It gets its configuration from the OS_* env vars (via `clientconfig.AuthenticatedClient()`), so it could look at those vars, but that's not guaranteed to work.

    - Option 1: `OS_USER_ID` is supplied, and that's what we need.
    - Option 2: `OS_USERNAME` is supplied, but the service user may lack permission to list users, so it cannot resolve its own user name into an ID.
    - Option 3: `OS_TOKEN` is supplied. No way to transform this into a user directly.

    However, the auth response already directly contains the user ID in question, so no guessing is needed.

### Proposal

**NOTE: Cannot be implemented like this, [see below for amendments](https://github.com/gophercloud/gophercloud/issues/1141#issuecomment-415388044).**

Add the following API in the toplevel package:

```
func (p *ProviderClient) AuthResult() AuthResult { ... }

type AuthResult struct {
  V2Result *Result //response from POST /v2/tokens
  V3Result *Result //response from POST /v3/auth/tokens
}
```

The structure of the AuthResult type encodes the information which Identity API version was used to authenticate: Either all fields are nil (failed authentication), or exactly one field is non-nil (depending on which Identity API version was used). It is therefore not possible to e.g. inadvertently decode a V2 response with a V3 result type. The type is designed to allow simple future expansion when a new Identity API version is released.

Add the following API in package `openstack/identity/v3/tokens` (and analogously for v2):

```
func FromAuthResult(p *ProviderClient) *CreateResult {
  result := p.AuthResult().V3Result
  if result == nil {
    return nil
  }
  return &CreateResult{ commonResult { *result } }
}
```

This indirection is needed because `type AuthResult` cannot directly reference `tokens.CreateResult` as the type of its `V3Result` field. (That would create an import cycle.)

[...]

PR #1226, opened 2018-09-06
add gophercloud.AuthResult

majewsky:
For #1141.

I have not tested this yet since I first want to have some more eyes on the API design. I ended up doing it in a way that differs from all my previous proposals in #1141.

Pros:
- This does not require any new public API in the `openstack` package, and only small changes to packages `openstack/identity/{v2,v3}/tokens`.
- Using `ProviderClient.GetAuthResult` feels natural. There is no excessive casting between different result types.

Cons:
- Keeping the `ProviderClient.authResult` field private requires some additional API surface between `v2auth/v3auth` and ProviderClient, in the form of `CopyTokenFrom`.
- `IsAnAuthResult` is slightly awkward.

The new `SetTokenAndAuthResult` method is strictly required to ensure that the mutex gets used. (It could have been `SetAuthResult` instead, but I figured that since token and AuthResult are always set at the same time, we may as well save ourselves the double locking/unlocking of the mutex that keeping `SetAuthResult` separate from `SetToken` would entail.)

### DIFF
--- a/auth_result.go
+++ b/auth_result.go
@@ -9,10 +9,7 @@ The following types satisfy this interface:
     github.com/gophercloud/gophercloud/openstack/identity/v2/tokens.CreateResult
     github.com/gophercloud/gophercloud/openstack/identity/v3/tokens.CreateResult
 
-Unfortunately, those types do not share any common methods (or rather, none
-with identical type signatures), so this interface contains a bogus method
-implemented by both of them to ensure that other types cannot be used with this
-interface. Usage of this type might look like this:
+    Usage example:
 
   import (
     "github.com/gophercloud/gophercloud"
@@ -44,7 +41,12 @@ interface. Usage of this type might look like this:
       panic(fmt.Sprintf("got unexpected AuthResult type %t", r))
     }
   }
+
+  Both implementing types share a lot of methods by name, like ExtractUser() in
+  this example. But those methods cannot be part of the AuthResult interface
+  because the return types are different (in this case, type tokens2.User vs.
+  type tokens3.User).
 */
 type AuthResult interface {
-	IsAnAuthResult()
+	ExtractTokenID() (string, error)
 }
--- a/auth_result.go
+++ b/auth_result.go
@@ -6,46 +6,46 @@ client's Keystone token. It is returned from ProviderClient.GetAuthResult().
 
 The following types satisfy this interface:
 
-    github.com/gophercloud/gophercloud/openstack/identity/v2/tokens.CreateResult
-    github.com/gophercloud/gophercloud/openstack/identity/v3/tokens.CreateResult
+	github.com/gophercloud/gophercloud/openstack/identity/v2/tokens.CreateResult
+	github.com/gophercloud/gophercloud/openstack/identity/v3/tokens.CreateResult
 
-    Usage example:
+Usage example:
 
-  import (
-    "github.com/gophercloud/gophercloud"
-    tokens2 "github.com/gophercloud/gophercloud/openstack/identity/v2/tokens"
-    tokens3 "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
-  )
+	import (
+		"github.com/gophercloud/gophercloud"
+		tokens2 "github.com/gophercloud/gophercloud/openstack/identity/v2/tokens"
+		tokens3 "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
+	)
 
-  func GetAuthenticatedUserID(providerClient *gophercloud.ProviderClient) (string, error) {
-    r := providerClient.AuthResult()
-    if r != nil {
-      //ProviderClient did not use openstack.Authenticate(), e.g. because token
-      //was set manually with ProviderClient.SetToken()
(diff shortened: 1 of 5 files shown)


jtopjian:
@majewsky Thanks for submitting this. Just a quick update that I haven't had time to review it, but I should in the next few days.

jtopjian:
@majewsky I read through this a few times and think it's pretty good. It does a good job balancing the differences between the auth result types and presenting it in a unified way. In addition, unit tests pass, so there doesn't seem to be any up-front regression issues. Nice work.

Perhaps one issue might be for anyone who has implemented a custom tokens package. They'll need to update their package to adhere to the interface.

> Keeping the ProviderClient.authResult field private requires some additional API surface between v2auth/v3auth and ProviderClient, in the form of CopyTokenFrom.

I think this is OK. 

> IsAnAuthResult is slightly awkward.

It is, but I think the benefit outweighs the awkwardness here. I'd be inclined to name the interface `AuthResultProvider` since it might better convey a characteristic.

Another idea is that instead of the bogus method, what about a simple method that returned the Token ID as a string? The functionality is a little redundant since it can be obtained from the client, but at least the method is doing _something_:

```go
func (r CreateResult) GetTokenID(client *gophercloud.ProviderClient) (string, error) {
  authResult := client.GetAuthResult()
  if authResult == nil {
    return "", nil
  }

  if res, ok := authResult.(*CreateResult); ok {
    token, err := res.ExtractToken()
    if err != nil {
      return "", err
    }

    return token.ID, nil
  }

  return "", nil
}
```

It can also serve as a documented example of how to use the `GetAuthResult` method.

Thoughts?

majewsky:
Sorry for taking so long to get back to you. Work has had other priorities and this fell under the table.

> I'd be inclined to name the interface AuthResultProvider since it might better convey a characteristic.

I don't know. The implementing types do not "provide" results, they "are" results.

> Another idea is that instead of the bogus method, what about a simple method that returned the Token ID as a string?

In the most recent commit, the `AuthResult` interface now contains the method `ExtractTokenID() (string, error)` which does exactly that. I also changed `SetTokenAndAuthResult` with two args into `SetTokenFromAuthResult` with only one arg since the token ID can now be obtained from the AuthResult.

majewsky:
I rebased to resolve the conflicts between this and #1408.

@jtopjian Bump.

@jrperritt This PR might also be interesting to you.

auth_result.go
```diff
@@ -0,0 +1,52 @@
...
+client's Keystone token. It is returned from ProviderClient.GetAuthResult().
+
+The following types satisfy this interface:
+
+    github.com/gophercloud/gophercloud/openstack/identity/v2/tokens.CreateResult
+    github.com/gophercloud/gophercloud/openstack/identity/v3/tokens.CreateResult
+
+    Usage example:
+
+  import (
```
jtopjian: Looks like there's an indenting mismatch between the code block and the text?

auth_result.go
```diff
@@ -0,0 +1,52 @@
...
+    "github.com/gophercloud/gophercloud"
+    tokens2 "github.com/gophercloud/gophercloud/openstack/identity/v2/tokens"
+    tokens3 "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
+  )
+
+  func GetAuthenticatedUserID(providerClient *gophercloud.ProviderClient) (string, error) {
+    r := providerClient.AuthResult()
```
jtopjian: `GetAuthResult()`?

auth_result.go
```diff
@@ -0,0 +1,52 @@
...
+    if r != nil {
```
jtopjian: `if r == nil`?

provider_client.go
```diff
@@ -115,6 +117,20 @@ func (client *ProviderClient) UseTokenLock() {
 	client.reauthmut = new(reauthlock)
 }
 
+//GetAuthResult returns the result from the request that was used to obtain a provider
```
jtopjian: Comments should be:

```
// Foo
```

Instead of

```
//Foo
```

provider_client.go
```diff
@@ -126,13 +142,53 @@ func (client *ProviderClient) Token() string {
...
 		defer client.mut.Unlock()
 	}
 	client.TokenID = t
+	client.authResult = nil
+}
+
+//SetTokenFromAuthResult safely sets the value of the auth token in the ProviderClient
+//and also records the AuthResult that was returned from the token creation
+//request. Applications may use this method in a custom ReauthFunc.
+func (client *ProviderClient) SetTokenFromAuthResult(r AuthResult) error {
```
jtopjian: `SetTokenFromAuthResult` is also setting `client.authResult`. The longest part of me re-reviewing this was figuring out how `authResult` was being set. It would have been more immediately clear to me, and so most likely others, if the name was `SetTokenAndAuthResult`. Or have two separate methods: one to set the token and one to set the auth result.

majewsky: I wouldn't want two separate methods, so I renamed to `SetTokenAndAuthResult`.

majewsky:
@jtopjian Should all be resolved now.

Issue #1233, opened 2018-09-12
BuildRequestBody failing with time.Time parameter


garbage-collector:
This code is failing:

```golang
func TestBuildRequestBody(t *testing.T) {
	type Params struct {
		ID   string    `json:"id"`
		Date time.Time `json:"date"`
	}

	params := &Params{
		ID:   "Foo",
		Date: time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC),
	}

	body, err := BuildRequestBody(params, "params")
	th.AssertNoErr(t, err)

	th.AssertJSONEquals(t, `{ "params": { "id": "Foo", "date": "1970-01-01T00:00:00Z" } }`, body)
}
```

```
Failure in params_test.go, line 22: unexpected error "json: cannot unmarshal string into Go value of type map[string]interface {}"
```

The way ```BuildRequestBody``` is working tries to unmarshal a string into a map[string]interface{} because time.Time is considered as a struct I think. Looking for a fix, except if there is a workaround for this issue?

garbage-collector:
After some digging, it looks like when the parameter type is a struct with a custom JSON marshaller that returns a ```string``` instead of a ```map[string]interface{}```, then the BuildRequestBody will never work.

jtopjian:
Here's an example of how to handle `time.Time` within a request struct:

https://github.com/gophercloud/gophercloud/blob/master/openstack/keymanager/v1/orders/requests.go#L75

https://github.com/gophercloud/gophercloud/blob/master/openstack/keymanager/v1/orders/requests.go#L97-L110

garbage-collector:
I just looked the ```keymanager/v1/orders``` package, and this is a way to handle ```time.Time```, but it looks more to a workaround than a solution.

And it looks more to a way of handling OpenStack time format than handling the issue in the BuildRequestBody code.

For example, we will still continue to have issues with this kind of code:

```
type ComplexField struct {
	ID   string
	Name string
}

func (c ComplexField) MarshalJSON() ([]byte, error) {
	return json.Marshal(c.ID + "." + c.Name)
}

func TestBuildRequestBody(t *testing.T) {
	type Params struct {
		ID      string       `json:"id"`
		Date    time.Time    `json:"date"`
		Complex ComplexField `json:"complex"`
	}

	params := &Params{
		ID:   "Foo",
		Date: time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC),
		Complex: ComplexField{
[...]

PR #1234, opened 2018-09-12
Fix BuildRequestBody for struct with specific json marshal

garbage-collector:
For #1233

We avoid to call this part of code

```
	b, err := json.Marshal(opts)
	if err != nil {
		return nil, err
	}

	//fmt.Printf("string(b): %s\n", string(b))

	err = json.Unmarshal(b, &optsMap)
	if err != nil {
		return nil, err
	}
```

For structs that implements a custom JSON Marshaller that does not ends with a ```map[string]interface{}```.

It's the case for ```time.Time```, where JSON Marshaller returns a string ```"1970-01-01T00:00:00Z"```.

I just dissociate the struct checks (required and not zero stuff) with the creation of the map[string]interface{}.

### DIFF
--- a/params_test.go
+++ b/params_test.go
@@ -1,0 +1,25 @@
+package gophercloud
+
+import (
+	"testing"
+	"time"
+
+	th "github.com/gophercloud/gophercloud/testhelper"
+)
+
+func TestBuildRequestBody(t *testing.T) {
+	type Params struct {
+		ID   string    `json:"id"`
+		Date time.Time `json:"date"`
+	}
+
+	params := &Params{
+		ID:   "Foo",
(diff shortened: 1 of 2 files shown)


params_test.go
```diff
@@ -0,0 +1,25 @@
...
+	params := &Params{
+		ID:   "Foo",
+		Date: time.Date(1970, time.January, 1, 0, 0, 0, 0, time.UTC),
+	}
+
+	body, err := BuildRequestBody(params, "params")
+	th.AssertNoErr(t, err)
+
+	th.AssertJSONEquals(t, `{"params": {"id": "Foo", "date": "1970-01-01T00:00:00Z"}}`, body)
+}
```
jtopjian: This should actually go in https://github.com/gophercloud/gophercloud/blob/master/testing/params_test.go

PR #402, opened 2017-06-21
Add tenants V2 update support

mozhuli:
For #397

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:

delete_tenant general implementation
https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/core.py#L305

delete_tenant sql driver implementation
https://github.com/openstack/keystone/blob/5c8dcd2f2f1a7645f93f39c3f5784920e2099998/keystone/identity/backends/sql.py#L340

PR #1235, opened 2018-09-14
Magnum: Manage Quotas Create

JackKuei:

code in this PR:

[[quotas:create]](https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/quota.py#L166)

### DIFF
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -38,9 +38,6 @@ type AcceptanceTestChoices struct {
 	// NetworkName is the name of a network to launch the instance on.
 	NetworkName string
 
-	// ProjectName contains the name of a valid project.
-	ProjectName string
-
 	// ExternalNetworkID is the network ID of the external network.
 	ExternalNetworkID string
 
@@ -63,7 +60,6 @@ func AcceptanceTestChoicesFromEnv() (*AcceptanceTestChoices, error) {
 	magnumImageID := os.Getenv("OS_MAGNUM_IMAGE_ID")
 	magnumKeypair := os.Getenv("OS_MAGNUM_KEYPAIR")
 	networkName := os.Getenv("OS_NETWORK_NAME")
-	projectName := os.Getenv("OS_PROJECT_NAME")
 	floatingIPPoolName := os.Getenv("OS_POOL_NAME")
 	externalNetworkID := os.Getenv("OS_EXTGW_ID")
 	shareNetworkID := os.Getenv("OS_SHARE_NETWORK_ID")
@@ -117,7 +113,6 @@ func AcceptanceTestChoicesFromEnv() (*AcceptanceTestChoices, error) {
 		MagnumImageID:      magnumImageID,
 		MagnumKeypair:      magnumKeypair,
 		NetworkName:        networkName,
-		ProjectName:        projectName,
 		ExternalNetworkID:  externalNetworkID,
 		ShareNetworkID:     shareNetworkID,
 		DBDatastoreType:    dbDatastoreType,
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -35,9 +35,6 @@ type AcceptanceTestChoices struct {
 	// MagnumImageID contains the ID of a valid magnum image.
 	MagnumImageID string
 
-	// MagnumProjectID contains the ID of a valid magnum image.
-	MagnumProjectID string
-
 	// NetworkName is the name of a network to launch the instance on.
 	NetworkName string
 
@@ -65,7 +62,6 @@ func AcceptanceTestChoicesFromEnv() (*AcceptanceTestChoices, error) {
 	flavorIDResize := os.Getenv("OS_FLAVOR_ID_RESIZE")
 	magnumImageID := os.Getenv("OS_MAGNUM_IMAGE_ID")
 	magnumKeypair := os.Getenv("OS_MAGNUM_KEYPAIR")
-	magnumProjectID := os.Getenv("OS_MAGNUM_PROJECT_ID")
 	networkName := os.Getenv("OS_NETWORK_NAME")
 	projectName := os.Getenv("OS_PROJECT_NAME")
 	floatingIPPoolName := os.Getenv("OS_POOL_NAME")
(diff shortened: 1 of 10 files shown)


jtopjian:
recheck

JackKuei:
@jtopjian This one is ready to be reviewed

openstack/containerinfra/v1/quotas/results.go
```diff
@@ -0,0 +1,57 @@
...
+	return s, err
+}
+
+type Quotas struct {
+	Resource  string    `json:"resource"`
+	CreatedAt time.Time `json:"created_at"`
+	UpdatedAt time.Time `json:"updated_at"`
+	HardLimit int       `json:"hard_limit"`
+	ProjectID string    `json:"project_id"`
+	ID        string    `json:"-"`
```
jtopjian: The API code defines this as an int: https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/quota.py#L39

But I'm OK with using a string. IDs of finite types usually don't last long and are eventually replaced with UUIDs, so using a string here might future-proof this package.

openstack/containerinfra/v1/quotas/results.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*r = Quotas(s.tmp)
+
+	switch t := s.ID.(type) {
+	case int, int8, int16, int32, int64, uint, uint8, uint16, uint32, uint64, float32, float64:
```
jtopjian: Go should always unmarshal a number to `float64`: https://golang.org/pkg/encoding/json/#Unmarshal

So I think `float64` is the only type that needs to be checked here.

openstack/containerinfra/v1/quotas/results.go
JackKuei: Yes, originally I coded as int, but ran into situation where json was returning string, thus the special handling.

openstacker:
Sorry, I just found this list, https://github.com/terraform-providers/terraform-provider-openstack/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+magnum+

So for cluster and cluster template, we just need another patch to support the cluster datasource, right? Thanks.

ozerovandrei:
@openstacker hello.

Yeah, you right, there is only cluster datasource left. You can also check or discuss terraform-provider-openstack Magnum progress in the following issue: [#402](https://github.com/terraform-providers/terraform-provider-openstack/issues/402).

JackKuei:
@openstacker Yes @ozerovandrei  made great progress on the terraform provider side. I should be able to continue on magnum quotas shortly.

acceptance/openstack/containerinfra/v1/containerinfra.go
```diff
@@ -177,3 +179,87 @@ func WaitForCluster(client *gophercloud.ServiceClient, clusterID string, status
 		return false, nil
 	})
 }
+
+// CreateProject will create a project with a random name.
+// It takes an optional createOpts parameter since creating a project
+// has so many options. An error will be returned if the project was
+// unable to be created.
+func CreateProject(t *testing.T, client *gophercloud.ServiceClient, c *projects.CreateOpts) (*projects.Project, error) {
```
jtopjian: Instead of recreating this function, you can import it from the acceptance tests. For example:

https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/networking/v2/extensions/layer3/routers_test.go#L9

https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/networking/v2/extensions/layer3/routers_test.go#L20

So here it would be:

```go
import (
  identity "github.com/gophercloud/gophercloud/acceptance/openstack/identity/v3"
)

project, err := identity.CreateProject(t, client, nil)
```

JackKuei:
@jtopjian Regarding creating a temp project, currently our environment locks down the creations of projects. Could the logic be that if project name is specified use the project name, otherwise create a temp project?

jtopjian:
@JackKuei I hate to veto this one, but I'd prefer not. It adds a good amount of logic that only exists for one organization.

Looking at this from the point of view of true acceptance testing and CI, these tests should be run in a disposable environment with admin privileges and no restrictions. 

Granted, creating disposable environments can be time consuming and using an already existing environment is a convenience, but tailoring the tests to those convenient existing environments isn't the best long-term design - if that makes sense.

There are two options I can see here: 

1. Create a devstack environment with Magnum. I fully understand that launching a Magnum cluster within devstack is an absolute pain, but since this is only dealing with quotas, it should be quite manageable.

2. I can write the acceptance test portion of this if you'd like.

JackKuei:
@jtopjian Totally understand from the point of view of during dev and launching a disposable devstack environment to perform testing. But thought expanding the feature can also help users to run tests on existing environment, especially our users. But if you feel strongly about it, maybe we can keep this extended feature in house.

jtopjian:
Another option is to use the values set in `OS_TENANT_NAME` or `OS_TENANT_ID`. If the Magnum API requires an ID but a name is set, then you should be able to use the Identity API to do that kind of resolution - this is assuming that managing Magnum quotas is an admin-only API.

If your cloud has done some customization to allow non-admin users to manage a magnum quota, then we're really getting into customized territory and I would then highly recommend running these tests in a true vanilla OpenStack environment.

But just a warning that if/when we integrate Magnum into OpenLab / CI, these tests might be refactored to match to how other tests are written.

JackKuei:
@jtopjian Let's go with assuming using the vanilla OpenStack environment for now then. Will create another ticket if find a better solution.
Understood about refactoring these code since seems like magnum is still in flux

acceptance/clients/clients.go
```diff
@@ -113,6 +117,7 @@ func AcceptanceTestChoicesFromEnv() (*AcceptanceTestChoices, error) {
 		MagnumImageID:      magnumImageID,
 		MagnumKeypair:      magnumKeypair,
 		NetworkName:        networkName,
+		ProjectName:        projectName,
```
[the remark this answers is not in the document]
JackKuei: Sorry to put this in the back burner for a while. This is one is ready to be merged.

jtopjian:
@JackKuei No problem at all!

Issue #1249, opened 2018-09-27
Bug: Senlin profile create is missing "version" field for spec when sending to api backend


JackKuei:
Running the clustering profile acceptance test failed because the version field is missing in the request json results in 500 error being returned by the api backend.

`{
  "profile": {
    "name": "TESTACC-InhooUHk",
    "spec": {
      "properties": {
        "flavor": "t2.small",
        "image": "centos7.3-latest",
        "name": "TESTACC-InhooUHk",
        "networks": [
          {
            "network": "sandbox-internal-net"
          }
        ],
        "security_groups": ""
      },
      "type": "os.nova.server"
    }
  }
}`

The 'version' key is missing from the provided spec map.

Reference:  #[823 [Add support for Senlin clustering]](https://github.com/gophercloud/gophercloud/issues/823)

jtopjian:
Fixed in #1250

PR #1250, opened 2018-09-28
Fixed where creating profile and policies will send spec "version" in…

JackKuei:

For #[1249 [Bug: Senlin profile create is missing "version" field for spec when sending to api backend]](https://github.com/gophercloud/gophercloud/issues/1249)
Ref #[823 [Add support for Senlin clustering]](https://github.com/gophercloud/gophercloud/issues/823)

### DIFF
--- a/openstack/clustering/v1/policies/results.go
+++ b/openstack/clustering/v1/policies/results.go
@@ -94,6 +94,19 @@ func (r *Spec) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+func (r Spec) MarshalJSON() ([]byte, error) {
+	spec := struct {
+		Type       string                 `json:"type"`
+		Version    string                 `json:"version"`
+		Properties map[string]interface{} `json:"properties"`
+	}{
+		Type:       r.Type,
+		Version:    r.Version,
+		Properties: r.Properties,
+	}
+	return json.Marshal(spec)
+}
+
 // policyResult is the resposne of a base Policy result.
 type policyResult struct {
 	gophercloud.Result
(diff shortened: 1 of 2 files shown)


jtopjian:
The reason Travis is giving an error is because you can't specify a JSON tag when the `UnmarshalJSON` method implements that same field/tag. 

Of course, the reason you're trying to add it is to have "version" included in the JSON request. So there are two possible options here:

1. Reverify that that Senlin is known to return multiple different formats for "version" in Profiles and Policies. If the past few releases of Senlin are always returning the same format and there is no need to convert from string to float and such, then remove the `UnmarshalJSON` method.

2. Create a corresponding `MarshalJSON` method which is the opposite of `UnmarshalJSON`: it customizes the way in which the request body is built so you can inject the version tag.

JackKuei:
@jtopjian Seems like jumping through lots of hoops just to bypass the go vet error. when just adding a "version" tag would work for both marshalling and unmarshalling. Tried to create a corresponding MarshalJSON(), but it is not triggering it.

JackKuei:
@jtopjian here is the code that failed to trigger during marshalling
```
// MarshalJSON marshals the spec properties
func (r *Spec) MarshalJSON() ([]byte, error) {
	spec := struct {
		Type       string                 `json:"type"`
		Version    string                 `json:"version"`
		Properties map[string]interface{} `json:"properties"`
	}{
		Type:       r.Type,
		Version:    r.Version,
		Properties: r.Properties,
	}
	return json.Marshal(spec)
}
```

JackKuei:
nvm found the problem....Was using pointer receiver.

jtopjian:
> Seems like jumping through lots of hoops just to bypass the go vet error

_true_, but another way of thinking about it is jumping through a lot of hoops for an API that doesn't provide a consistent input/output :)

JackKuei:
that comment was before getting the MarshalJSON() to work. Using MarshalJSON is alot less painful and cleaner

jtopjian:
Ah, OK :)

Glad you got it sorted out. Let me know if this is ready for review/merge.

PR #1256, opened 2018-10-02
Fix - add X-Auth-Token header on tokens.Validate

cygmris:
Under stable/rocky of keystone
`tokens.Validate` with only `X-Subject-Token` won't get correct response but with also `X-Auth-Token`.

```
HEAD /v3/auth/tokens HTTP/1.1
Host: 127.0.0.1:5000
User-Agent: gophercloud/1.0.0
Connection: close
Accept: application/json
X-Subject-Token: gAAAAABbsJ2xgMz0O5ec33rym04sW1o-StKSOTOu7-jh03mevreKuYxwECuJ7l2CLCbE18HkRBjqIBtZz_pb981QuL3Sm30VtR7yPvcY6OYCQdoIxZxnKV0KrmzaQUke--PXPgqWxULQKDjha4CeFjDeY3keA9lh0A

HTTP/1.1 401 Unauthorized
Date: Sun, 30 Sep 2018 09:56:01 GMT
Server: Apache/2.4.34 (Ubuntu)
WWW-Authenticate: Keystone uri="http://127.0.0.1:5000"
Content-Length: 114
Vary: X-Auth-Token
X-Distribution: Ubuntu
x-openstack-request-id: req-f04eabfb-94f3-4740-958a-dbfa330f5234
Connection: close
```

### DIFF
--- a/openstack/identity/v3/tokens/requests.go
+++ b/openstack/identity/v3/tokens/requests.go
@@ -99,6 +99,7 @@ func (opts *AuthOptions) CanReauth() bool {
 func subjectTokenHeaders(c *gophercloud.ServiceClient, subjectToken string) map[string]string {
 	return map[string]string{
 		"X-Subject-Token": subjectToken,
+		"X-Auth-Token":    subjectToken,
 	}
 }
 


jtopjian:
@cygmris Thanks for submitting this. Could you take a look at our [contributor guide](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial) and make sure this patch follows the right process? Please let me know if you have any questions.

jtopjian:
After re-reviewing this PR, I don't think this is needed at all.

The `X-Auth-Token` header should be added automatically by the HTTP client, and Gophercloud is doing this correctly:

```
2020/01/13 04:14:52 [DEBUG] OpenStack Request URL: HEAD https://keystone-yyc.cloud.cybera.ca:5000/v3/auth/tokens
2020/01/13 04:14:52 [DEBUG] OpenStack request Headers:
Accept: application/json
User-Agent: gophercloud/2.0.0
X-Auth-Token: ***
X-Subject-Token: ***
2020/01/13 04:14:53 [DEBUG] OpenStack Response Code: 200
```

Additionally, adding the token that is being validated as the `X-Auth-Token` will cause the authentication to be done with the `X-Subject-Token`. If the token is not valid, the request will return an unauthentication error rather than an invalid token response. And authenticating as a supplied token might also be a security issue.

Issue #1257, opened 2018-10-04
Networking v2: Trunk extension CRUD


celebdor:
The OpenStack networking API includes, in its second version, a very useful tool for using a single device to multiplex traffic of different ports. This Issue is about adding support for the trunk operations described in:

    https://developer.openstack.org/api-ref/network/v2/#trunk-networking

* [x]  List 
* [X]  Get
* [x]  Create
* [x]  Delete
* [x]  Update
* [x]  AddSubports
* [x]  RemoveSubports
* [x]  GetSubports

celebdor:
All the API merged!

PR #1258, opened 2018-10-04
Networking trunk extension support Create/Delete

celebdor:
This patch adds the Networking extension trunk support for Create and Delete operations.

The reason for having both create and delete together is because otherwise the acceptance tests can't remove any resources that are added to the trunk (ports, subnet and network).

For #1257

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/trunks/trunks.go
+++ b/acceptance/openstack/networking/v2/extensions/trunks/trunks.go
@@ -1,0 +1,36 @@
+package trunks
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
+)
+
+func CreateTrunk(t *testing.T, client *gophercloud.ServiceClient, parentPortID string, subportIDs ...string) (trunk *trunks.Trunk, err error) {
+	trunkName := tools.RandomString("TESTACC-", 8)
+	iTrue := true
+	opts := trunks.CreateOpts{
+		Name:         trunkName,
+		Description:  "Trunk created by gophercloud",
+		AdminStateUp: &iTrue,
+		PortID:       parentPortID,
+	}
+
+	opts.Subports = make([]trunks.Subport, len(subportIDs))
+	for id, subportID := range subportIDs {
+		opts.Subports[id] = trunks.Subport{
+			SegmentationID:   id + 1,
+			SegmentationType: "vlan",
+			PortID:           subportID,
+		}
+	}
+
+	t.Logf("Attempting to create trunk: %s", opts.Name)
+	trunk, err = trunks.Create(client, opts).Extract()
+	if err == nil {
+		t.Logf("Successfully created trunk")
+	}
+	return
+}
--- a/openstack/networking/v2/extensions/trunks/results.go
+++ b/openstack/networking/v2/extensions/trunks/results.go
@@ -1,0 +1,74 @@
+package trunks
+
+import (
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+)
+
+type Subport struct {
+	SegmentationID   int    `json:"segmentation_id"`
+	SegmentationType string `json:"segmentation_type"`
+	PortID           string `json:"port_id"`
+}
+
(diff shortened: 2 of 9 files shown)


jtopjian:
Thank you for working on this. 

> The reason for having both create and delete together is because otherwise the acceptance tests can't remove any resources that are added to the trunk (ports, subnet and network).

Having acceptance tests which can't be cleaned up is OK. Acceptance tests are run in a disposable environment (either automatically triggered through OpenLab or the contributor using a devstack environment), so cleaning up isn't really required (although it's nice). I recommend not deleting the network and subnet for now, either. That way should allow the test to pass.

Please have a look at our [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial) which will cover some other items as well. Notably, we'll need to see some references to the Neutron code per Step 3.

Thanks again and let me know if you have any questions. :)

celebdor:
Thanks for the review. I'll split it further then.

celebdor:
Neutron code:

https://github.com/openstack/neutron/blob/master/neutron/services/trunk/plugin.py#L199-L234
https://github.com/openstack/neutron/blob/master/neutron/db/standard_attr.py#L28-L64

mrhillsman:
recheck

openstack/networking/v2/extensions/trunks/results.go
```diff
@@ -0,0 +1,101 @@
...
+	gophercloud.Result
+}
+
+// CreateResult is the response from a Create operation. Call its Extract method
+// to interpret it as a Trunk.
+type CreateResult struct {
+	commonResult
+}
+
+type Trunk struct {
```
jtopjian: Just for posterity / reference:

https://github.com/openstack/neutron/blob/master/neutron/objects/trunk.py#L91-L99

https://github.com/openstack/neutron/blob/master/neutron/objects/trunk.py#L131

https://github.com/openstack/neutron/blob/master/neutron/objects/trunk.py#L142

acceptance/openstack/networking/v2/extensions/trunks/trunks.go
```diff
@@ -0,0 +1,34 @@
...
+	"testing"
+
+	"github.com/gophercloud/gophercloud"
+	trunks "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
+)
+
+func CreateTrunk(t *testing.T, client *gophercloud.ServiceClient, parentPortID string, subportIDs ...string) (trunk *trunks.Trunk, err error) {
+	iTrue := true
+	opts := trunks.CreateOpts{
+		Name:         "gophertrunk",
```
jtopjian: nit: randomized names are usually used

acceptance/openstack/networking/v2/extensions/trunks/trunks.go
```diff
@@ -0,0 +1,34 @@
+package trunks
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud"
+	trunks "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/trunks"
```
jtopjian: nit: no need for an alias here.

openstack/networking/v2/extensions/trunks/results.go
```diff
@@ -0,0 +1,101 @@
...
+
+	// Display description.
+	Description string `json:"description"`
+
+	// A list of tags associated with the trunk
+	Tags []string `json:"tags,omitempty"`
+}
+
+// UnmarshalJSON to override default
+func (r *Trunk) UnmarshalJSON(b []byte) error {
```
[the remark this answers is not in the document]
celebdor: Cool! I didn't know that. It is practically my first golang code ever.

acceptance/openstack/networking/v2/extensions/trunks/trunks.go
celebdor: Alright, changing it!

acceptance/openstack/networking/v2/extensions/trunks/trunks.go
celebdor: Thanks!

jtopjian:
@celebdor I think it's a safe assumption that this is ready to be merged, so I'm going to go ahead. In the future, it's helpful to post when something is ready to be re-reviewed so I don't make an incorrect assumption.

PR #1259, opened 2018-10-04
Neutron Standard Attributes Tag Extension - ReplaceAll support

hardys:
This initial patch adds only the ability to Replace all tags on
a given resource, as described in the API docs:

https://developer.openstack.org/api-ref/network/v2/#replace-all-tags

If this approach is acceptable subsequent PRs can add the other
operations supported by this API.

For #1260

code in this PR:

Note this isn't quite a normal CRUD API since it operates on existing resources:

PUT that adds either a single tag (append):
https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L115

... or replaces all of them with a list:
https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L127

Delete one tag:
https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L140

... or delete them all:
https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L152

When these are implemented it would be good to look at List filtering by tag (which Neutron also supports) but I guess that will be a separate issue - getting the ability to set/unset tags is a prerequisite.

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/attributestags_test.go
+++ b/acceptance/openstack/networking/v2/extensions/attributestags_test.go
@@ -1,0 +1,32 @@
+// +build acceptance networking tags
+
+package extensions
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	networking "github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
+	th "github.com/gophercloud/gophercloud/testhelper"
+)
+
+func TestTags(t *testing.T) {
+	client, err := clients.NewNetworkV2Client()
+	th.AssertNoErr(t, err)
+
+	// Create Network
+	network, err := networking.CreateNetwork(t, client)
+	th.AssertNoErr(t, err)
+	defer networking.DeleteNetwork(t, client, network.ID)
+
+	tagReplaceAllOpts := attributestags.ReplaceAllOpts{
+		Tags: []string{"abc", "123"},
+	}
+	tags, err := attributestags.ReplaceAll(client, "networks", network.ID, tagReplaceAllOpts).Extract()
+	th.AssertNoErr(t, err)
+	th.AssertDeepEquals(t, tags, []string{"abc", "123"})
+
+	// FIXME(shardy) - when the networks Get schema supports tags we should
+	// verify the tags are set in the Get response
+}
--- a/openstack/networking/v2/extensions/attributestags/doc.go
+++ b/openstack/networking/v2/extensions/attributestags/doc.go
@@ -1,0 +1,19 @@
+/*
+Package attributestags manages Tags on Resources created by the OpenStack Neutron Service.
+
+This enables tagging via a standard interface for resources types which support it.
+
+See https://developer.openstack.org/api-ref/network/v2/#standard-attributes-tag-extension for more information on the underlying API.
+
+Example to ReplaceAll Resource Tags
+
+    network, err := networks.Create(conn, createOpts).Extract()
+
+    tagReplaceAllOpts := attributestags.ReplaceAllOpts{
+        Tags:         []string{"abc", "123"},
+    }
+    attributestags.ReplaceAll(conn, "networks", network.ID, tagReplaceAllOpts)
+
+
+*/
+package attributestags
--- a/openstack/networking/v2/extensions/attributestags/requests.go
+++ b/openstack/networking/v2/extensions/attributestags/requests.go
@@ -1,0 +1,36 @@
+package attributestags
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// ReplaceAllOptsBuilder allows extensions to add additional parameters to
+// the ReplaceAll request.
+type ReplaceAllOptsBuilder interface {
+	ToAttributeTagsReplaceAllMap() (map[string]interface{}, error)
+}
+
+// ReplaceAllOpts provides options used to create Tags on a Resource
+type ReplaceAllOpts struct {
+	Tags []string `json:"tags" required:"true"`
+}
+
+// ToAttributeTagsReplaceAllMap formats a ReplaceAllOpts into the body of the
+// replace request
+func (opts ReplaceAllOpts) ToAttributeTagsReplaceAllMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "")
+}
+
+// ReplaceAll updates all tags on a resource, replacing any existing tags
+func ReplaceAll(client *gophercloud.ServiceClient, resourceType string, resourceID string, opts ReplaceAllOptsBuilder) (r ReplaceAllResult) {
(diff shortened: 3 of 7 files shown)


hardys:
Note this only shows the PUT replace-all API for now, happy to push additional PRs for the other actions (and tests for this one of course, it works for me locally though) - wanted to check if the general approach is OK before proceeding though, thanks for any feedback!

hardys:
@jtopjian Thanks for the prompt and detailed feedback! 

It all sounds reasonable to me - Thanks for the hint re the package name, fairly new to Go atm (coming from Python/OpenStack) so still figuring out the conventions - I did consider just "tags", but I wanted to avoid confusion with the older tags extension, I guess attributetags or stdattributetags would be fine.

I'll go ahead and make the changes requested, add some tests and add a checklist to the issue to track each action I hope to implement.

hardys:
Ok I pushed a commit with the code review fixes, this works for me locally but I'll add tests via follow-up commits, please let me know if I missed anything

hardys:
Ok added a basic unit test for this initial PR which impments the ReplaceAll item in the issue checklist.

Should I squash these now, or at least rename the top commit to reflect the fact it only implements the ReplaceAll interface to tags?

I'll also look at adding an acceptance test, then I think this should be ready to un-WIP, unless there's anything else I'm missing?

jtopjian:
> Should I squash these now

Sure, go ahead and squash. We ask to only amend commits once an official review has started.

Let me know once it's ready for another review. I should be able to look it over some time today.

hardys:
@jtopjian thanks for the reply, I squashed the latest commits to b48c338 - this should be ready for review now as it fully implements the ReplaceAll action and I think I fixed all the code review issues previously raised.

This has been locally tested against a recent OpenStack deployment (TripleO not devstack, but it's a stock upstream environment) and it works, but I'm not entirely clear on how to approach the acceptance test, e.g should I create a new test that creates one (or more?) resourceTypes then tags them, or add some tag testing to all the existing network resource tests that are expected to support tags?

The API for this is supposed to be fairly generic, so I'm not sure how exhaustive the acceptance test should be vs a simple smoke test for a single resource type.  For reference I've been testing it for a network resource ref http://paste.openstack.org/show/731597/.

openstack/networking/v2/extensions/attributestag/doc.go
```diff
@@ -0,0 +1,19 @@
...
+    network, err := networks.Create(conn, createOpts).Extract()
+
+    tagReplaceAllOpts := attributestag.ReplaceAllOpts{
+        Tags:         []string{"abc", "123"},
+    }
+    attributestag.ReplaceAll(conn, "networks", network.ID, tagReplaceAllOpts)
+
+
+*/
+package attributestag
```
jtopjian: I'm _really_ sorry, but this will bother me forever otherwise: can we call this `attributestags`?

openstack/networking/v2/extensions/attributestag/requests.go
```diff
@@ -0,0 +1,37 @@
...
+// ToAttributeTagsReplaceAllMap formats a ReplaceAllOpts into the body of the
+// replace request
+func (opts ReplaceAllOpts) ToAttributeTagsReplaceAllMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "")
+}
+
+// ReplaceAll updates all tags on a resource, replacing any existing tags
+func ReplaceAll(client *gophercloud.ServiceClient, resourceType string, resourceID string, opts ReplaceAllOptsBuilder) (r ReplaceAllResult) {
+	b, err := opts.ToAttributeTagsReplaceAllMap()
+	// The type/id are part of the path not the body
```
jtopjian: nit: this comment can be removed

openstack/networking/v2/extensions/attributestag/results.go
```diff
@@ -0,0 +1,26 @@
+package attributestag
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// "tagResult" denoting that "tags" are returned and can be used for both the ReplaceAll and Get actions
+
```
jtopjian: nit: let's remove the empty line. This comment was also just for the example I gave. You can either remove it or do something more formal like:

```
// tagResult represents a result which will contain a resource's tags.
```

openstack/networking/v2/extensions/attributestag/results.go
```diff
@@ -0,0 +1,26 @@
...
+type tagResult struct {
+	gophercloud.Result
+}
+
+// Extract interprets any tagResult as Tags.
```
jtopjian:
```
// Extract interprets any tagResult as a slice of strings
```

The use of `Tags` in this method is only used internally and isn't exposed to the user. The user only sees the slice.

hardys:
Note when implementing the acceptance test, I noticed one strange thing which is the API to create e.g networks doesn't support tags (you have PUT the tags after creation via the attributestag extension), but the GET for networks does support tags:

https://developer.openstack.org/api-ref/network/v2/#show-network-details

This with a change like the following we can easily get the tags with a single API call, so it'd be nice to enable this in a future PR (as well as querying for lists of resources via tags).

diff --git a/openstack/networking/v2/networks/results.go b/openstack/networking/v2/networks/results
index 62f4b3c..b7c52ae 100644
--- a/openstack/networking/v2/networks/results.go
+++ b/openstack/networking/v2/networks/results.go
@@ -76,6 +76,8 @@ type Network struct {
        // Availability zone hints groups network nodes that run services like DHCP, L3, FW, and ot
        // Used to make network resources highly available.
        AvailabilityZoneHints []string `json:"availability_zone_hints"`
+
+       Tags []string `json:"tags"`

hardys:
@jtopjian ready for another review pass when you have a moment ref 04bbc2d

Note I left one FIXME in the acceptance test as IMO we should really check the tags are really set on the network vs just checking the API response, but as mentioned above that needs a schema change to the networks API so I guess that would be best considered via a separate PR?

If you agree such an addition is reasonable I can add an item to the checklist, or raise a separate issue for those changes.

hardys:
https://github.com/gophercloud/gophercloud/pull/1271 pushed which adds the List support and allowed me to remove the FIXME in the acceptance test.

I did notice one unexpected thing - although the neutron docs say it's a list of tags, it's returned sorted and duplicates are not allowed, e.g it's a set - the acceptance test was adjusted in the second PR to account for that.

jtopjian:
> Note when implementing the acceptance test, I noticed one strange thing which is the API to create e.g networks doesn't support tags 

This isn't unusual. There isn't anyone overseeing the existing resources in OpenStack upstream and Gophercloud to ensure that feature parity exists. The same can happen when initially adding a resource (it's only partially added) but we try to prevent that.

Having this added in a separate PR would be appreciated.

> Note I left one FIXME in the acceptance test as IMO we should really check the tags are really set on the network vs just checking the API response, but as mentioned above that needs a schema change to the networks API so I guess that would be best considered via a separate PR?

Right.

It's OK to have partial acceptance tests while a set of API calls is in progress, so adding better validation later (once available) is fine.

PR #1262, opened 2018-10-06
Acc Tests: Enable networking trunks to be tested

jtopjian:
For #818


jtopjian:
Looks like the trunk extension isn't enabled.

PR #1268, opened 2018-10-06
Networking v2 Trunk support - Delete

celebdor:
This patch adds the Networking extension trunk support for the Delete
operation.

Signed-off-by: Antoni Segura Puimedon <celebdor@gmail.com>
For #1257

Neutron master code showing the delete operation:
https://github.com/openstack/neutron/blob/master/neutron/services/trunk/plugin.py#L261-L272


celebdor:
Thanks a lot!

PR #1270, opened 2018-10-08
Resolve #209 - Always be able to handle large integers in JSON parsing.

### LINKED ISSUE #209: quotasets.Get().Extract() fails with JSON parser error on large values

sb10:
For #209

The bug in #209 seems to have fixed itself since being reported. This PR alters some tests to ensure no reversion.

It also implements the (no longer required) fix discussed in #209, in case any other system has to deal large integers.

### DIFF
--- a/openstack/compute/v2/extensions/quotasets/testing/fixtures.go
+++ b/openstack/compute/v2/extensions/quotasets/testing/fixtures.go
@@ -22,7 +22,7 @@ const GetOutput = `
       "injected_file_content_bytes" : 10240,
       "injected_files" : 5,
       "metadata_items" : 128,
-      "ram" : 200000,
+      "ram" : 9216000,
       "key_pairs" : 10,
       "injected_file_path_bytes" : 255,
 	  "server_groups" : 2,
@@ -73,7 +73,7 @@ const GetDetailsOutput = `
       },
       "ram" : {
           "in_use": 0,
-          "limit": 200000,
+          "limit": 9216000,
           "reserved": 0
       },
       "key_pairs" : {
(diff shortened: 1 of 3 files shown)


sb10:
Can I please request a code review? I don't know how to see why the openlab check failed.

jtopjian:
recheck

jtopjian:
> The bug in #209 seems to have fixed itself since being reported. 

I wonder if it was a newer version of Go that fixed this problem internally.

> This PR alters some tests to ensure no reversion.

Good idea - thank you.

> It also implements the (no longer required) fix discussed in #209, in case any other system has to deal large integers.

I'd prefer not to make this modification. I don't like the idea of altering the provider client in anticipation of problems since that might end up causing unintended problems. :)

jtopjian:
OpenLab might be having issues right now.

sb10:
Ok, it's just the altered tests now.

Issue #1260, opened 2018-10-04
Add support for Neutron Standard Attributes Tag Extension


hardys:
I've been working with gophercloud and noticed there's no supporting for tagging neutron resources.

Over the last few releases it seems there's a fairly standard way of tagging multiple resource types, documented here:

https://developer.openstack.org/api-ref/network/v2/#standard-attributes-tag-extension

The OpenStack CLI supports this via e.g openstack network create foo --tag abc, but internally this is two API calls, one to create the network/port/subnet/router etc, then another to resource_type/resource_id/tags to set the resource tags.

This potentially makes it easier to add support for all resource types that support tagging, as we hopefully don't need to update every resource, but instead expose this tags API via gophercloud.

I'll push a PR demonstrating this so we can discuss, open to other ideas if there's a better approach.

Line references to the code which implements this in Neutron:

Note this isn't quite a normal CRUD API since it operates on existing resources:

PUT that adds either a single tag (append):
https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L115

... or replaces all of them with a list:
https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L127

Delete one tag:
https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L140

... or delete them all:
https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L152

When these are implemented it would be good to look at List filtering by tag (which Neutron also supports) but I guess that will be a separate issue - getting the ability to set/unset tags is a prerequisite.

Task List (one task per API action referenced in the API docs above):
- [x] Replace all tags
- [x] Remove all tags
- [x] Confirm a tag
- [x] Add a tag
- [x] List tags
- [x] Remove a tag

jtopjian:
This sounds good to me. If you'd like, you can use this issue to house all PRs which implement the actions. It's common to create a checklist. For example: https://github.com/gophercloud/gophercloud/issues/1199

jtopjian:
@hardys I think this suite is complete? If so, thank you so much - really nice work!

hardys:
@jtopjian yes this is now complete, thanks for all the reviews! :)

PR #1271, opened 2018-10-08
Neutron attributestags List support

hardys:
This patch adds the ability to List all tags on a given resource.
    
As described in the API docs:
    
https://developer.openstack.org/api-ref/network/v2/#obtain-tag-list
    
Issue: #1260

Code from neutron related to this PR:

https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L98

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/attributestags_test.go
+++ b/acceptance/openstack/networking/v2/extensions/attributestags_test.go
@@ -21,12 +21,16 @@ func TestTags(t *testing.T) {
 	defer networking.DeleteNetwork(t, client, network.ID)
 
 	tagReplaceAllOpts := attributestags.ReplaceAllOpts{
-		Tags: []string{"abc", "123"},
+		// Note Neutron returns tags sorted, and although the API
+		// docs say list of tags, it's a set e.g no duplicates
+		Tags: []string{"a", "b", "c"},
 	}
 	tags, err := attributestags.ReplaceAll(client, "networks", network.ID, tagReplaceAllOpts).Extract()
 	th.AssertNoErr(t, err)
-	th.AssertDeepEquals(t, tags, []string{"abc", "123"})
+	th.AssertDeepEquals(t, tags, []string{"a", "b", "c"})
 
-	// FIXME(shardy) - when the networks Get schema supports tags we should
-	// verify the tags are set in the Get response
+	// Verify the tags are set in the List response
+	tags, err = attributestags.List(client, "networks", network.ID).Extract()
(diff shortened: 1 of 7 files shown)


hardys:
@jtopjian thanks for merging the ReplaceAll patch, this is the next in the branch, adds List support and resolves the FIXME in the acceptance test

hardys:
I've been trying to figure out how to make PRs depend on each other, and AFAICS it doesn't really work?

Such a workflow is mentioned in the contributor docs but it seems each dependent PR contains a diff of all patches, and that appears to be a long-standing github issue.

Hints on how to handle that appreciated - I could obviously push all the commits to a single PR/branch but your contributor docs explicitly advise against that.  Thanks!

hardys:
@jtopjian FYI I pushed https://github.com/hardys/gophercloud/commits/neutron_std_attr_tags_wip which contains all the patches to close the issue - I'll rebase/push the PRs one at a time unless there's a better workflow I can use to enable easy review of the branch (sorry bit unfamiliar coming from gerrit reviews where this is pretty easy)

After that branch merges I'd like to add Get & list/query support for all the resources that support it, but I'll raise a separate issue for that with details.

jtopjian:
> Such a workflow is mentioned in the contributor docs but it seems each dependent PR contains a diff of all patches, and that appears to be a long-standing github issue.

That's correct. Github doesn't have a good way of handling this as other systems do. I'm not familiar with Gerritt / OpenStack workflow, but from what others have said, it's quite different.

When you submit a PR that is dependent on another PR, Github will display it as if it was its own dedicated PR, including all commits of the parent commits/PR.

If I were to merge a parent PR as-is, then even though Github will still show those commits as pending in the child PR, I can merge it and Github will only apply the unmerged commits.

If you submitted 2 PRs, a child and parent, and requested a review on both, and if both were good to merge, I could skip directly to the child PR and Github would show the parent as also having been merged.

If you submitted 2 PRs and needed to make changes on the parent, then you'll have to rebase and force-push the child manually. 

So it _kind of_ supports dependent PRs, but not intuitively nor provides a lot of help managing them.

jtopjian:
> Hints on how to handle that appreciated - I could obviously push all the commits to a single PR/branch but your contributor docs explicitly advise against that. Thanks!

This is correct, too.

It can be frustrating to split feature implementations up into multiple PRs but it has proved very valuable in the long-run. Even if someone is on a good streak with submitting perfect PRs, there will be one PR that requires a good amount of discussion and would have caused all subsequent PRs to have been rewritten in a major way.

Issue #1242, opened 2018-09-20
Network v2: networkipavailabilities int out of bounds


jtopjian:
While implementing #1240, OpenLab reported the following error:

```
2018-09-19 03:27:45.522524 | ubuntu-xenial |    convenience.go:35: Failure in networkipavailabilities_test.go, line 22: unexpected error "json: cannot unmarshal number 18446744073709552000 into Go struct field SubnetIPAvailability.total_ips of type int"
```

We'll need to change the type of `total_ips` to something that can accommodate that large of a number.

jtopjian:
Fixed in #1273

PR #1273, opened 2018-10-09
Network v2: Fix Upper Limit of IPs

jtopjian:
The networkipavailabilities extension requires an increase to the
upper limit for both Total IPs and Used IPs. This is to
accommodate a large amount of IPv6 addresses.

This commit changes the type of these fields from int to big.Int.

For #1242 

With this in place, the acceptance tests now passes:

```
=== RUN   TestNetworkIPAvailabilityList
--- PASS: TestNetworkIPAvailabilityList (0.13s)
        tools.go:72: {
                  "network_id": "a8976812-bbaf-46f0-b534-d672e77b1cf8",
                  "network_name": "private",
                  "project_id": "29b363f3aca74d608757a51c23c36fad",
                  "tenant_id": "29b363f3aca74d608757a51c23c36fad",
                  "subnet_ip_availability": [
                    {
                      "subnet_id": "203ffe15-a1b5-424f-96d8-6dbb0e33d819",
                      "subnet_name": "private-subnet",
                      "cidr": "10.0.0.0/26",
                      "ip_version": 4,
                      "total_ips": 61,
                      "used_ips": 2
                    },
                    {
                      "subnet_id": "510a4389-1a3a-42ae-b564-1a5e0e8a2715",
                      "subnet_name": "ipv6-private-subnet",
                      "cidr": "fdbc:bf53:567e::/64",
                      "ip_version": 6,
                      "total_ips": 18446744073709552000,
                      "used_ips": 2
                    }
                  ],
                  "total_ips": {},
                  "used_ips": {}
                }
```

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/networkipavailabilities/networkipavailabilities_test.go
+++ b/acceptance/openstack/networking/v2/extensions/networkipavailabilities/networkipavailabilities_test.go
@@ -22,6 +22,10 @@ func TestNetworkIPAvailabilityList(t *testing.T) {
 	th.AssertNoErr(t, err)
 
 	for _, availability := range allAvailabilities {
-		tools.PrintResource(t, availability)
+		for _, subnet := range availability.SubnetIPAvailabilities {
+			tools.PrintResource(t, subnet)
+			tools.PrintResource(t, subnet.TotalIPs)
+			tools.PrintResource(t, subnet.UsedIPs)
+		}
 	}
 }
--- a/openstack/networking/v2/extensions/networkipavailabilities/results.go
+++ b/openstack/networking/v2/extensions/networkipavailabilities/results.go
@@ -1,6 +1,8 @@
 package networkipavailabilities
 
 import (
+	"math/big"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
@@ -43,10 +45,10 @@ type NetworkIPAvailability struct {
 	SubnetIPAvailabilities []SubnetIPAvailability `json:"subnet_ip_availability"`
 
 	// TotalIPs represents a number of IP addresses in the network.
-	TotalIPs int `json:"total_ips"`
+	TotalIPs big.Int `json:"total_ips"`
 
 	// UsedIPs represents a number of used IP addresses in the network.
-	UsedIPs int `json:"used_ips"`
+	UsedIPs big.Int `json:"used_ips"`
 }
 
 // SubnetIPAvailability represents availability details for a single subnet.
@@ -64,10 +66,10 @@ type SubnetIPAvailability struct {
 	IPVersion int `json:"ip_version"`
 
 	// TotalIPs represents a number of IP addresses in the subnet.
-	TotalIPs int `json:"total_ips"`
+	TotalIPs big.Int `json:"total_ips"`
 
 	// UsedIPs represents a number of used IP addresses in the subnet.
-	UsedIPs int `json:"used_ips"`
+	UsedIPs big.Int `json:"used_ips"`
 }
 
 // NetworkIPAvailabilityPage stores a single page of NetworkIPAvailabilities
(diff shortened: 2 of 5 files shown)


jtopjian:
@jrperritt Do you know if there's a better way to do this?

`math/big` has native support for JSON un/marshalling so it _seems_ appropriate. The downside is that all interaction with the fields will need to be done through `math/big`'s methods.

jrperritt:
Hmm, where did that number (`18446744073709552000`) come from? That looks like (2^64)-1 with truncated precision. I'd expect it to be `18446744073709551615`, in which case I think a `uint64` would work.

jtopjian:
`18446744073709551616` should be the total amount of IPs in an IPv6 `/64` subnet, but for some reason Neutron is returning `18446744073709552000`.

Googling `18446744073709552000` turns up lots of results with varying causes.

Googling `18446744073709552000 openstack` turns up:

https://github.com/fog/fog-openstack/commit/1ed17e2c63124f1564f50499e7d306984d172b67#diff-78d550100f3ae421db70105b98c48c86

So at least it's not just in my environments.

Whether the number is being malformed from MySQL or Python, who knows, but I don't think we should focus on that number specifically because a `/63` of IPv6 addresses is `36893488147419103232` in length. A `/32` is `79228162514264337593543950336`.

So there's definitely a need for _really_ large numbers here. :)

jrperritt:
But it's the number that's important: I think that''s the highest a 64-bit computer can address with a single word. Regardless, if the numbers can get bigger than that, I'd consider 2 options:
1. set the field as a `string` and let users deal with it,
2. have 2 fields, one for `uint64` and one for `*big.Int`, only one of which is ever populated

jtopjian:
> But it's the number that's important: I think that''s the highest a 64-bit computer can address with a single word.

Yep - totally understand.

> set the field as a string and let users deal with it,

I'm going to opt for this. On one hand it feels like a lazy way out, but on the other hand, working with strings is easier to most than with `math/big`.

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
@kiwik @liu-sheng @yongfengdu @mrhillsman Hi all, we're getting consistent failed test runs across all of our PRs.

mrhillsman:
@jtopjian i have to run to drop off kid at school and then will review. at first glance at least the failure in this PR is due to AMQP; leaving details in case anyone gets to it before me:

2018-10-11 03:24:53.714 | Oct 11 03:24:46 ubuntu zun-compute[30137]: 2018-10-11 03:24:46.060 30137 ERROR oslo_service.service MessageDeliveryFailure: Unable to connect to AMQP server on 127.0.0.1:5672 after inf tries: (0, 0): (403) ACCESS_REFUSED - Login was refused using authentication mechanism AMQPLAIN. For details see the broker logfile.

=ERROR REPORT==== 11-Oct-2018::03:24:32 ===
Error on AMQP connection <0.2881.0> (192.168.200.90:54256 -> 192.168.200.90:5672 - barbican-keystone-listener:28259:3b51cd95-e585-49a2-ab9c-bef60a7e45e9, user: 'stackrabbit', state: opening):
access to vhost 'None' refused for user 'stackrabbit'

=ERROR REPORT==== 11-Oct-2018::03:24:46 ===
Error on AMQP connection <0.2888.0> (127.0.0.1:34364 -> 127.0.0.1:5672, state: starting):
AMQPLAIN login refused: user 'guest' - invalid credentials

jtopjian:
@mrhillsman Thanks! Definitely not an emergency - whenever someone is able to get to it 😄

jtopjian:
recheck

jtopjian:
@mrhillsman looks like tests are back on track - thank you :)

jtopjian:
@jrperritt This is what I came up with for the "string" type. I still had to use `big.Int` as an intermediary. Thoughts?

jrperritt:
That looks right

jtopjian:
Cool - thank you! :smile:

PR #1276, opened 2018-10-10
Neutron attributestags DeleteAll support

hardys:
As described in the API docs:

https://developer.openstack.org/api-ref/network/v2/#remove-all-tags

Issue: #1260

Code from neutron related to this PR:

https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L152


hardys:
recheck

hardys:
recheck

Issue #1279, opened 2018-10-10
BuildRequestBody ignores required fields in slice of structs fields


celebdor:
BuildRequestBody correctly checks the children structs of the opts parameter. However, if one of the struct fields, instead of being a struct, is a slice of structs, the fields of all the structs in the slice will not be checked for field requirements.

jtopjian:
Fixed / Implemented in #1280

PR #1280, opened 2018-10-10
Networking trunks - Verify subport missing fields

celebdor:
builquery had a bug that made it overlook the required fields in slices
of structs. This patch addresses it and adds a unit test for trunk
creation.

For #1279

### DIFF
--- a/openstack/networking/v2/extensions/trunks/results.go
+++ b/openstack/networking/v2/extensions/trunks/results.go
@@ -8,9 +8,9 @@ import (
 )
 
 type Subport struct {
-	SegmentationID   int    `json:"segmentation_id"`
-	SegmentationType string `json:"segmentation_type"`
-	PortID           string `json:"port_id"`
+	SegmentationID   int    `json:"segmentation_id" required:"true"`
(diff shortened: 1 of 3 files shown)


celebdor:
Seems it broke other unit tests. I'll throw the debugger at it.

celebdor:
This should have fixed it

jtopjian:
@celebdor The modifications to `BuildRequestBody` look OK to me, but we'll want that in a dedicated PR with some unit tests added to https://github.com/gophercloud/gophercloud/blob/master/testing/params_test.go.

jtopjian:
Actually, this PR is OK, but let's change the commit message to focus on the changes to `BuildRequestBody` and have the modifications to `openstack/networking/v2/extensions/trunks/results.go` be in a separate commit of this PR - if that makes sense?

celebdor:
Done!

celebdor:
Sure, why not?

celebdor:
I think this makes it ready!

celebdor:
Thank you for the reviews and the improvement that came from them

PR #1281, opened 2018-10-10
Handle filters on list parameters

garbage-collector:
For #1199

https://github.com/openstack/mistral/blob/master/mistral/api/controllers/v2/workflow.py#L201-L252

Please notice that this is a breaking change with the current API, but impacted code was merged a couple of weeks ago (see #1205).

This allows to filter workflows according to filter accepted by Mistral API.

### DIFF
--- a/acceptance/openstack/workflow/v2/workflow.go
+++ b/acceptance/openstack/workflow/v2/workflow.go
@@ -19,6 +19,9 @@ version: '2.0'
 %s:
   description: Simple workflow example
   type: direct
+  tags:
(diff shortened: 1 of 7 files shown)


jtopjian:
@garbage-collector Is this ready for review/merge?

> Please notice that this is a breaking change with the current API, but impacted code was merged a couple of weeks ago (see #1205).

This is totally fine. Thank you for noting it.

garbage-collector:
Oups! Ready to review!

PR #1283, opened 2018-10-11
Neutron std attr tags confirm

hardys:
As described in the API docs:

https://developer.openstack.org/api-ref/network/v2/#standard-attributes-tag-

https://developer.openstack.org/api-ref/network/v2/#confirm-a-tag

Issue: #1260

Code from neutron related to this PR: https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L103

### DIFF
--- a/openstack/networking/v2/extensions/attributestags/results.go
+++ b/openstack/networking/v2/extensions/attributestags/results.go
@@ -38,3 +38,20 @@ type DeleteResult struct {
 type AddResult struct {
 	gophercloud.ErrResult
 }
+
+// ConfirmResult is the result from an Confirm operation.
+type ConfirmResult struct {
+	gophercloud.Result
+}
+
+func (r ConfirmResult) Extract() (bool, error) {
+	exists := r.Err == nil
+
+	if r.Err != nil {
+		if _, ok := r.Err.(gophercloud.ErrDefault404); ok {
+			r.Err = nil
+		}
+	}
+
+	return exists, r.Err
+}
--- a/acceptance/openstack/networking/v2/extensions/attributestags_test.go
+++ b/acceptance/openstack/networking/v2/extensions/attributestags_test.go
@@ -52,6 +52,13 @@ func TestTags(t *testing.T) {
 	sort.Strings(tags)
 	th.AssertDeepEquals(t, []string{"b", "c", "d"}, tags)
 
+	// Confirm tags exist/don't exist
+	exists, err := attributestags.Confirm(client, "networks", network.ID, "d").Extract()
+	th.AssertNoErr(t, err)
+	th.AssertEquals(t, true, exists)
+	noexists, err := attributestags.Confirm(client, "networks", network.ID, "a").Extract()
+	th.AssertEquals(t, false, noexists)
+
 	// Delete all tags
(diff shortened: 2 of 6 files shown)


openstack/networking/v2/extensions/attributestags/results.go
```diff
@@ -38,3 +38,15 @@ type DeleteResult struct {
...
+// Extract interprets ConfirmResult to return bool
+func (r ConfirmResult) Extract() (bool, error) {
+	err := r.ExtractErr()
+	return err == nil, err
+}
+
+// ConfirmResult is the result from an Confirm operation.
+// Call its ExtractErr method to determine if the call succeeded or failed.
+type ConfirmResult struct {
+	gophercloud.ErrResult
```
jtopjian: `ErrResult` is only for results which don't return a value (eg there is no `Extract()` method. Let's change this to `gophercloud.Result`

jtopjian:
@hardys When making the change for this, it will be best to update master, rebase this branch with master, and then do a force push. The result will appear in Github as 1 commit (the latest commit).

hardys:
> @hardys When making the change for this, it will be best to update master, rebase this branch with master, and then do a force push. The result will appear in Github as 1 commit (the latest commit).

Thanks for the review, change made and rebased onto master

openstack/networking/v2/extensions/attributestags/results.go
```diff
@@ -38,3 +38,13 @@ type DeleteResult struct {
 type AddResult struct {
 	gophercloud.ErrResult
 }
+
+// Extract interprets ConfirmResult to return bool
+func (r ConfirmResult) Extract() (bool, error) {
+	return r.Err == nil, r.Err
```
jtopjian: I apologize, I might not have looked closely enough at this last time.

The logic to determine if the tag exists makes sense to me, but there's a possible issue where a true error is ignored because the user focused only on the `bool` portion. For example, what if a 401 or 500 was returned?

How about something like this:

```go
func (r ConfirmResult) Extract() (bool, error) {
  exists := r.Err == nil

  if r.Err != nil {
    if _, ok := r.Err.(gophercloud.ErrDefault404); ok {
      r.Err = nil
    } 
  } 

  return exists, r.Err
}
```

So we're considering 404 errors to be acceptable and setting them to `nil`. This way a user can now do:

```go
exists, err := attributestags.Confirm(client, "networks", network.ID, "a").Extract() 
if err != nil {
  panic(err)
}

if !exists {
  panic("Tag does not exist")
}
```

Thoughts?

hardys: Yes good point, I was mostly thinking of the 404 case but you're right we do need to handle other errors - I'll update it as you describe, thanks!

hardys:
@jtopjian thanks for the review, now updated as you described.

I guess there are still some other 404 cases we won't catch here (for example an invalid resourceType will return false but err will be nil, which isn't ideal).

I can't think of a good way to solve that atm without another API call (which would appear to negate the reason I assume this API exists, e.g to check a tag exists without retrieving the entire resource) - I guess that's really a limitation of the underlying API so I'm thinking this is probably sufficient for now?

jtopjian:
Right - there's some trade-offs with this implementation, but it's the best we can do for now.

Issue #1285, opened 2018-10-11
Expose Neutron tags via resource schema


hardys:
Following on from https://github.com/gophercloud/gophercloud/issues/1260, we now have the ability to set resource tags on all resources that support it but we don't expose the tags returned as part of the GET for each resource.

https://docs.openstack.org/neutron/pike/contributor/internals/tag.html

The above doc says these resources support standard attribute tags:

    floatingips
    networks
    policies
    ports
    routers
    security_groups
    subnetpools
    subnets
    trunks

Grepping the neutron code seems to validate that list:

```
[shardy@localhost neutron]$ grep -R -B1 "tag_support = True" ./* | grep -v tests
./db/models/l3.py-    collection_resource_map = {l3_apidef.ROUTERS: l3_apidef.ROUTER}
./db/models/l3.py:    tag_support = True
--
./db/models/l3.py-    collection_resource_map = {l3_apidef.FLOATINGIPS: l3_apidef.FLOATINGIP}
./db/models/l3.py:    tag_support = True
--
./db/models/securitygroup.py-    collection_resource_map = {sg.SECURITYGROUPS: 'security_group'}
./db/models/securitygroup.py:    tag_support = True
--
./db/models_v2.py-                               port_def.RESOURCE_NAME}
./db/models_v2.py:    tag_support = True
--
./db/models_v2.py-                               subnet_def.RESOURCE_NAME}
./db/models_v2.py:    tag_support = True
--
./db/models_v2.py-                               subnetpool_def.RESOURCE_NAME}
./db/models_v2.py:    tag_support = True
--
./db/models_v2.py-    collection_resource_map = {net_def.COLLECTION_NAME: net_def.RESOURCE_NAME}
./db/models_v2.py:    tag_support = True
--
./db/qos/models.py-    collection_resource_map = {'policies': 'policy'}
./db/qos/models.py:    tag_support = True
--
./services/trunk/models.py-    collection_resource_map = {'trunks': 'trunk'}
./services/trunk/models.py:    tag_support = True
```

It should be possible to simply add the tag to the schema for each resource struct e.g:

```
[shardy@localhost gophercloud]$ git diff
diff --git a/openstack/networking/v2/networks/results.go b/openstack/networking/v2/networks/results
index 62f4b3c..4d683d9 100644
--- a/openstack/networking/v2/networks/results.go
+++ b/openstack/networking/v2/networks/results.go
@@ -76,6 +76,9 @@ type Network struct {
[...]

PR #1277, opened 2018-10-10
Neutron std attr tags add

hardys:
As described in the API docs:
    
https://developer.openstack.org/api-ref/network/v2/#standard-attributes-tag-
    
https://developer.openstack.org/api-ref/network/v2/#add-a-tag
    
Issue: #1260

Code from neutron related to this PR: https://github.com/openstack/neutron/blob/master/neutron/extensions/tagging.py#L115

Note this PR depends on https://github.com/gophercloud/gophercloud/pull/1276
--

PR #1286, opened 2018-10-11
Add Tags to resources that support std-attributes-tags

hardys:
Since we added the ability to set resource tags via the attributestags
extensione, it is helpful to expose those tags as they are included
in the GET response for resources that support tagging ref:

https://docs.openstack.org/neutron/pike/contributor/internals/tag.html

Issue: #1285

Edit adding links to show the resources which have tag_support = True:

Router: https://github.com/openstack/neutron/blob/master/neutron/db/models/l3.py#L67
FloatingIP: https://github.com/openstack/neutron/blob/master/neutron/db/models/l3.py#L112
SecurityGroup: https://github.com/openstack/neutron/blob/master/neutron/db/models/securitygroup.py#L32
Port: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L115
Subnet: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L210
SubnetPool: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L250
Network: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L278
QoS/RBAC Policy: https://github.com/openstack/neutron/blob/master/neutron/db/qos/models.py#L36
Trunk: https://github.com/openstack/neutron/blob/master/neutron/services/trunk/models.py#L49

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/attributestags_test.go
+++ b/acceptance/openstack/networking/v2/extensions/attributestags_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	networking "github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2"
 	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/attributestags"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
 	th "github.com/gophercloud/gophercloud/testhelper"
 )
 
@@ -28,6 +29,13 @@ func TestTags(t *testing.T) {
 	tags, err := attributestags.ReplaceAll(client, "networks", network.ID, tagReplaceAllOpts).Extract()
 	th.AssertNoErr(t, err)
 	sort.Strings(tags) // Ensure ordering, older OpenStack versions aren't sorted...
+	th.AssertDeepEquals(t, []string{"a", "b", "c"}, tags)
+
+	// Verify the tags are also set in the object Get response
+	gnetwork, err := networks.Get(client, network.ID).Extract()
+	th.AssertNoErr(t, err)
+	tags = gnetwork.Tags
+	sort.Strings(tags)
 	th.AssertDeepEquals(t, []string{"a", "b", "c"}, tags)
 
 	// Add a tag
--- a/acceptance/openstack/networking/v2/extensions/attributestags_test.go
+++ b/acceptance/openstack/networking/v2/extensions/attributestags_test.go
@@ -3,6 +3,7 @@
 package extensions
 
 import (
+	"sort"
 	"testing"
 
 	"github.com/gophercloud/gophercloud/acceptance/clients"
@@ -21,13 +22,13 @@ func TestTags(t *testing.T) {
 	defer networking.DeleteNetwork(t, client, network.ID)
 
 	tagReplaceAllOpts := attributestags.ReplaceAllOpts{
-		// Note Neutron returns tags sorted, and although the API
-		// docs say list of tags, it's a set e.g no duplicates
+		// docs say list of tags, but it's a set e.g no duplicates
 		Tags: []string{"a", "b", "c"},
 	}
 	tags, err := attributestags.ReplaceAll(client, "networks", network.ID, tagReplaceAllOpts).Extract()
 	th.AssertNoErr(t, err)
-	th.AssertDeepEquals(t, tags, []string{"a", "b", "c"})
+	sort.Strings(tags) // Ensure ordering, older OpenStack versions aren't sorted...
+	th.AssertDeepEquals(t, []string{"a", "b", "c"}, tags)
 
 	// Add a tag
 	err = attributestags.Add(client, "networks", network.ID, "d").ExtractErr()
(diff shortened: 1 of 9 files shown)


hardys:
@jtopjian Hi if you have time to check this simple patch it'd be much appreciated - I'm working on adding support for tagging to terraform-cloud-provider-openstack and enabling the tag response via the Get calls makes the acceptance test for that much easier - should be generally useful too since it avoids making two calls to get the resource details and the tags via the attributestags list call.

Thanks!

jtopjian:
@hardys At first glance this looks good. We'll need to see the actual Neutron code which shows each of these supports a tag field, though.

hardys:
> @hardys At first glance this looks good. We'll need to see the actual Neutron code which shows each of these supports a tag field, though.

Thanks for the reply - yes I did spend some time on this already, but it's not super-obvious due to the way the API responses are constructed (inheritiance/mixins etc with data derived from the DB object model, which I grepped and pasted into the issue) - I'll take another look and update the PR description (disclaimer, I'm not a Neutron developer...)

hardys:
@jtopjian code references added, although I'm not super familiar with the translation layer between the plugin DB model and the API responses (there's no hard-coded API schema AFAICS like in some other services, and also in gophercloud/terraform etc).

Let me know if that's sufficient, I can add an acceptance test too if needed but I was a bit worried about burning CI cycles for little coverage (I guess I can perhaps just add tags to some existing tests though).

jtopjian:
Thanks! Looking for code is _not_ fun, but if I had at penny every time the code highlighted a notable difference than what the docs described... I'd have a lot of pennies.

This all looks good. How about integrating one resource into a Tag acceptance test? That should be good enough to confidently say that `tag_support` does indeed provide proper tag support for each resource.

hardys:
@jtopjian thanks for the feedback!  Do you mean just e.g add a Get to https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/networking/v2/extensions/attributestags_test.go that proves the Tags exist when getting the Network, or should I do something covering more/different resources?

jtopjian:
I think we're on the same page. Basically use the `attributetags` extension to add a tag to a network and then perform a `networking/v2/networks` Get exercise the newly added `Tags` field in the `networking/v2/networks` package to verify that tag exists.

hardys:
@jtopjian thanks - I updated the patch with an acceptance test for networks.Get - I know the CI has been having some issues so FWIW here's my local run passing:

```
[shardy@localhost gophercloud]$ go test -v -tags "fixtures acceptance" -run "TestTags" github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2/extensions -count=1
=== RUN   TestTags
--- PASS: TestTags (4.64s)
	networking.go:32: Attempting to create network: TESTACC-P1v8n5YU
	networking.go:39: Successfully created network.
	networking.go:399: Attempting to delete network: 176875e0-9057-4675-a37b-8a3ee9ffbeb9
	networking.go:406: Deleted network: 176875e0-9057-4675-a37b-8a3ee9ffbeb9
PASS
ok  	github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2/extensions	4.658s
```

jtopjian:
@hardys Yeah, the OpenLab team is aware of the issue. Thanks for the test output - I'll also quickly test it offline. Similar for your other PRs: if you have any that are ready for review/test, let me know and I'll have time to do it at some point today or tomorrow.

hardys:
@jtopjian thanks, I think all my other PRs are ready for review now, the only issue is the acceptance test changes may need a rebase if this patch lands first.

I branched each patch so they need to land in this order:

https://github.com/gophercloud/gophercloud/pull/1276
https://github.com/gophercloud/gophercloud/pull/1277
https://github.com/gophercloud/gophercloud/pull/1282
https://github.com/gophercloud/gophercloud/pull/1283

Those all work fine for me locally so if you have time I'd welcome any review feedback - thanks!

hardys:
Rebased to address merge conflict in acceptance/openstack/networking/v2/extensions/attributestags_test.go

acceptance/openstack/networking/v2/extensions/attributestags_test.go
```diff
@@ -48,4 +49,9 @@ func TestTags(t *testing.T) {
 	tags, err = attributestags.List(client, "networks", network.ID).Extract()
 	th.AssertNoErr(t, err)
 	th.AssertDeepEquals(t, tags, []string{})
+
+	// Verify the tags are also set in the object Get response
+	gnetwork, err := networks.Get(client, network.ID).Extract()
+	th.AssertNoErr(t, err)
+	th.AssertDeepEquals(t, gnetwork.Tags, []string{"a", "b", "c"})
```
jtopjian: This must be some kind of bug with `AssertDeepEquals` - I'm not sure why it's passing. 

If you change this to:

```go
t.Logf("Network tags: %#v", gnetwork.Tags)
th.AssertEquals(t, len(gnetwork.Tags), 3)
```

You'll see that no tags exist and the assert now fails. But if you move these lines to before the tags are deleted (line 46), all is good.

hardys: > You'll see that no tags exist and the assert now fails. But if you move these lines to before the tags are deleted (line 46), all is good.

Ouch sorry about that, rebase mistake - as you say it does pass which is strange, I'll see if I can figure out why.

hardys:
Thanks for spotting the rebase mistake, it's confusing that the test still passed, now fixed.

hardys:
Well that is strange, the DeepEquals now fails because the order of the Network object Tags response is different e.g gnetwork.Tags=[b a c] - I'll sort the results in the test since the order doesn't seem to be consistent over the attributetags and Network etc API

hardys:
Ok it seems the tag ordering changes depending on the OpenStack version - I don't have access to my normal trunk/lastest test environment today so I ran the tests against an Ocata cloud and the tests show the tags are returned in a different order for both the attributetags and Network Get API calls on the older version.

I'll sort the results in the test since that means the test will work on both versions and we probably shouldn't care about the tag ordering.

hardys:
Ok I pushed https://github.com/gophercloud/gophercloud/pull/1300 to fix the test issues, then I'll rebase this on that change to ensure the test additions are correct.

jtopjian:
@hardys both this and #1300 look good, so I'll just merge them both.

PR #1292, opened 2018-10-13
Acc Tests: Disabling Zun

jtopjian:


jtopjian:
Prior OpenLab errors showed zun not being able to connect to rabbit, so I thought zun might be the cause. Another PR just passed without this patch, so zun was not the problem.

Issue #1293, opened 2018-10-13
LBaaS v2 / Octavia: Support Member weight of 0


jtopjian:
Load Balancer members are able to have a weight of 0:

https://github.com/openstack/octavia/blob/c64a4574ad31eed1412f2677edc2ce2bc546a5e2/octavia/common/constants.py#L81

But we currently `omitempty`, which prevents a weight of 0 from being possible:

* https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/lbaas_v2/pools/requests.go#L277
* https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/lbaas_v2/pools/requests.go#L326
* https://github.com/gophercloud/gophercloud/blob/master/openstack/loadbalancer/v2/pools/requests.go#L268
* https://github.com/gophercloud/gophercloud/blob/master/openstack/loadbalancer/v2/pools/requests.go#L317

PR #1294, opened 2018-10-13
Lbaas weight 0

jtopjian:
For #1293


jtopjian:
@jrperritt before I merged this, I wanted to give you a heads up on it since I think your team works with LBaaS quite a bit?

jrperritt:
>  I wanted to give you a heads up on it since I think your team works with LBaaS quite a bit?

Thanks

Issue #1289, opened 2018-10-12
Support List query by tag for Neutron resources


hardys:
Since we can now set tags on Neutron resources it would be helpful to enable querying by tag, e.g like the following CLI actions:

```
[shardy@localhost ~]$ openstack --debug network list --all-tags wibble 2>&1 | grep -e "GET /v2.0/networks" -e terraform_test
[shardy@localhost ~]$ openstack --debug network list --tags wibble 2>&1 | grep -e "GET /v2.0/networks"
http://192.168.1.12:9696 "GET /v2.0/networks?tags=wibble HTTP/1.1" 200 721
[shardy@localhost ~]$ openstack --debug network list --any-tags wibble 2>&1 | grep -e "GET /v2.0/networks"
http://192.168.1.12:9696 "GET /v2.0/networks?tags-any=wibble HTTP/1.1" 200 721
[shardy@localhost ~]$ openstack --debug network list --not-tags wibble 2>&1 | grep -e "GET /v2.0/networks"
http://192.168.1.12:9696 "GET /v2.0/networks?not-tags=wibble HTTP/1.1" 200 33099
[shardy@localhost ~]$ openstack --debug network list --not-any-tags wibble 2>&1 | grep -e "GET /v2.0/networks"
http://192.168.1.12:9696 "GET /v2.0/networks?not-tags-any=wibble HTTP/1.1" 200 33099
```

The same resources added for Get support via https://github.com/gophercloud/gophercloud/issues/1285 should support this query interface (line number where the query is processed TODO, will include in the PR).

jtopjian:
Implemented in #1296

PR #1296, opened 2018-10-15
Enable list query by tag for Neutron resources

hardys:
For resources that support std-attribute-tags we can query by tag so
add the ability to do this, with an acceptance test showing the
interface working with Network resources.

Issue: #1289

These neutron code references show the resources which have tag_support = True:

Router: https://github.com/openstack/neutron/blob/master/neutron/db/models/l3.py#L67
FloatingIP: https://github.com/openstack/neutron/blob/master/neutron/db/models/l3.py#L112
SecurityGroup: https://github.com/openstack/neutron/blob/master/neutron/db/models/securitygroup.py#L32
Port: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L115
Subnet: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L210
SubnetPool: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L250
Network: https://github.com/openstack/neutron/blob/master/neutron/db/models_v2.py#L278
QoS/RBAC Policy: https://github.com/openstack/neutron/blob/master/neutron/db/qos/models.py#L36
Trunk: https://github.com/openstack/neutron/blob/master/neutron/services/trunk/models.py#L49

This code shows the API which enables list filtering based on tags:
https://github.com/openstack/neutron/blob/master/neutron/objects/tag.py#L37

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/attributestags_test.go
+++ b/acceptance/openstack/networking/v2/extensions/attributestags_test.go
@@ -3,40 +3,49 @@
 package extensions
 
 import (
+	"fmt"
 	"sort"
 	"testing"
 
+	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	networking "github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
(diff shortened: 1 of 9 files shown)


hardys:
Rebased to resolve merge conflicts with acceptance test

Issue #1299, opened 2018-10-16
Fix issues in attributestags_test.go


hardys:
Discussion on https://github.com/gophercloud/gophercloud/pull/1286 highlighted some issues with the recently added attributestags acceptance test:

* The order of the DeepEquals is wrong, which means if the actual result is an empty list no equivalence is tested
* The DeepEquals can't be used to test the result of DeleteAll, for the same reason as above
* On older OpenStack versions the tags aren't returned in a sorted order, so we need to sort the slices before testing them with DeepEquals

jtopjian:
Fixed in #1286 / #1300

PR #1300, opened 2018-10-16
Fix issues in attributestags acceptance test

hardys:
Discussion on #1286 highlighted some issues:

* The order of the DeepEquals is wrong, which means if the actual
  result is an empty list no equivalence is tested
* The DeepEquals can't be used to test the result of DeleteAll, for the same reason
* On older OpenStack versions the tags aren't returned in a sorted order

Closes Issue: #1299


kiwik:
recheck

kiwik:
recheck

Issue #1269, opened 2018-10-07
Blockstorage V2: panic: interface conversion : interface {} is nil, not []interface {}


waynr:
Before starting a PR, please read our [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial).

```
panic: interface conversion: interface {} is nil, not []interface {}                     
                                                                                         
goroutine 1 [running]:                                                                                                                                                                                                                                                        
github.com/rcbops/mk8s/vendor/github.com/gophercloud/gophercloud.Result.extractIntoPtr(0xc000317b30, 0xc00056c2a0, 0xc00056c360, 0x0, 0x0, 0x103aba0, 0xc0003ce320, 0x1233e24, 0x7, 0x0, ...)                                                                                    
        /go/src/github.com/rcbops/mk8s/vendor/github.com/gophercloud/gophercloud/results.go:93 +0xc18                                                                                                                                                                            
github.com/rcbops/mk8s/vendor/github.com/gophercloud/gophercloud.Result.ExtractIntoSlicePtr(0xc000317b30, 0xc00056c2a0, 0xc00056c360, 0x0, 0x0, 0x103aba0, 0xc0003ce320, 0x1233e24, 0x7, 0xc00014b418, ...)
        /go/src/github.com/rcbops/mk8s/vendor/github.com/gophercloud/gophercloud/results.go:200 +0x1ba                                                                                
github.com/rcbops/mk8s/vendor/github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes.ExtractVolumesInto(0x1390260, 0xc0005ac300, 0x103aba0, 0xc0003ce320, 0x1390260, 0xc0005ac0c0)                                                                                 
        /go/src/github.com/rcbops/mk8s/vendor/github.com/gophercloud/gophercloud/openstack/blockstorage/v2/volumes/results.go:146 +0xd6
```

I have tracked the issue down to this exact commit: https://github.com/gophercloud/gophercloud/commit/f1bae72a73954e0636d2de309b417bcfa4bd7de7

I have confirmed that this commit is what leads to the panic shown above in two ways. First off, the commit immediately preceeding it does not exhibit the behavior. Second, I have reverted it in the latest master branch and the problem goes away.

Here is the code block which results in the above stack trace:

```
type VolumeWithTenant struct {
	volumes.Volume
	volumetenants.VolumeTenantExt
}

[...]

PR #1305, opened 2018-10-19
Core: Fix Unchecked Type Assertion

jtopjian:
This commit fixes an unchecked type assertion when extracting results
into a slice pointer. The type assertion would previously fail when a
paged result was empty.

For #1269


jtopjian:
@jrperritt when you have a moment, can you review this?

PR #1310, opened 2018-10-25
Senlin: Clusters RemoveNodes

JackKuei:

code in this PR:

[[clusters:removenodes]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/clusters.py#L115)

### DIFF
--- a/openstack/clustering/v1/clusters/requests.go
+++ b/openstack/clustering/v1/clusters/requests.go
@@ -546,3 +546,25 @@ func AddNodes(client *gophercloud.ServiceClient, id string, opts AddNodesOpts) (
 	r.Header = result.Header
 	return
 }
+
+func (opts RemoveNodesOpts) ToClusterNodeMap(nodeAction string) (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, nodeAction)
+}
+
+type RemoveNodesOpts struct {
+	Nodes []string `json:"nodes" required:"true"`
+}
+
+func RemoveNodes(client *gophercloud.ServiceClient, clusterID string, opts RemoveNodesOpts) (r DeleteResult) {
+	b, err := opts.ToClusterNodeMap("del_nodes")
+	if err != nil {
+		r.Err = err
+		return
+	}
+	var result *http.Response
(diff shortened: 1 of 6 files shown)


JackKuei:
@jtopjian This one is ready to be reviewed. Thx

openstack/clustering/v1/clusters/requests.go
```diff
@@ -546,3 +546,25 @@ func AddNodes(client *gophercloud.ServiceClient, id string, opts AddNodesOpts) (
 	r.Header = result.Header
 	return
 }
+
+func (opts RemoveNodesOpts) ToClusterNodeMap(nodeAction string) (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, nodeAction)
+}
```
jtopjian: Comparing this with `AddNodeOpts`, I see what you're doing here.

If `AddNodeOpts` and `RemoveNodeOpts` take a `nodeAction` parameter, then what actually distinguishes them from one another?

Wouldn't these be the same:

```
removeOpts.ToClusterNodeMap("del_nodes")
addOpts.ToClusterNodeMap("del_nodes")
```

I can think of two options here:

1. Rename `ToClusterNodeMap` to `ToClusterAddNodeMap` and `ToClusterRemoveNodeMap` and hard-code the `add_nodes` and `del_node` string.

    The benefit of this is creating distinct add and remove interfaces which will never clash.

2. Rename `AddNodeOpts` to just `NodeOpts`, keep `ToClusterNodeMap` as-is. Get rid of `RemoveNodeOpts`.

    The benefit is that you are sharing a common struct between actions that take a list/array of nodes. If there are further actions which take a list/array of nodes, this might be the better way to go.

Thoughts?

JackKuei: @jtopjian good points. I prefer to use the second option with code sharing, but might ended up with option 1 because if these api diverges, then will have to use option 1 anyway, so might as well start with option 1.

jtopjian:
@JackKuei Just want to confirm that this is good to merge? I'm quite sure it is, but I never want to assume when there's been work done after a review request.

JackKuei:
@jtopjian Yes it is. Thanks for the confirmation.

PR #1313, opened 2018-10-30
Add DefaultPoolID update support

kayrus:
Openstack v2 API supports `default_pool_id` update
https://developer.openstack.org/api-ref/load-balancer/v2/index.html?expanded=update-a-listener-detail#update-a-listener

PR #1314, opened 2018-10-30
lbaas: l7policies: don't omit invert: bool=false

kayrus:
When the rule was initially created with `invert=true`, it is not possible to update it to `invert=false`, because json engine omits `false` value.

### DIFF
--- a/openstack/loadbalancer/v2/l7policies/requests.go
+++ b/openstack/loadbalancer/v2/l7policies/requests.go
@@ -306,7 +306,7 @@ type UpdateRuleOpts struct {
 
 	// When true the logic of the rule is inverted. For example, with invert true,
 	// equal to would become not equal to. Default is false.
-	Invert bool `json:"invert,omitempty"`
+	Invert *bool `json:"invert,omitempty"`
 }
 
 // ToRuleUpdateMap builds a request body from UpdateRuleOpts.
--- a/openstack/loadbalancer/v2/l7policies/testing/fixtures.go
+++ b/openstack/loadbalancer/v2/l7policies/testing/fixtures.go
@@ -76,7 +76,7 @@ var (
 		Value:        "/images*",
 		ProjectID:    "e3cd678b11784734bc366148aa37580e",
 		Key:          "",
-		Invert:       false,
+		Invert:       true,
 		AdminStateUp: true,
 	}
 	RuleHostName = l7policies.Rule{
@@ -239,7 +239,7 @@ const SingleRuleBody = `
 {
 	"rule": {
 		"compare_type": "REGEX",
-		"invert": false,
+		"invert": true,
 		"admin_state_up": true,
 		"value": "/images*",
 		"key": null,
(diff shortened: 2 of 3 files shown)


jtopjian:
@kayrus Thank you for submitting this, but the same comment I made still applies here: https://github.com/gophercloud/gophercloud/pull/1313#issuecomment-434353227

kayrus:
@jtopjian, I'll take care about your suggestions soon.

I've noticed that there are a lot of inconsistencies like this. E.g. for the `admin_state_up` values.

As far as I understood there could be only two cases:

* using the pointer `AdminStateUp *bool `json:"admin_state_up,omitempty"``
* using the pure bool `AdminStateUp bool `json:"admin_state_up"``

```sh
$ grep -r ' bool `json:' | grep omit
compute/v2/extensions/quotasets/requests.go:    Force bool `json:"force,omitempty"`
networking/v2/extensions/trunks/results.go:     AdminStateUp bool `json:"admin_state_up,omitempty"`
networking/v2/extensions/subnetpools/requests.go:       Shared bool `json:"shared,omitempty"`
networking/v2/extensions/subnetpools/requests.go:       IsDefault bool `json:"is_default,omitempty"`
sharedfilesystems/v2/shares/results.go: IsPublic bool `json:"is_public,omitempty"`
blockstorage/v3/volumes/requests.go:    Multiattach bool `json:"multiattach,omitempty"`
blockstorage/extensions/quotasets/requests.go:  Force bool `json:"force,omitempty"`
blockstorage/extensions/volumeactions/requests.go:      Force bool `json:"force,omitempty"`
loadbalancer/v2/l7policies/requests.go: Invert bool `json:"invert,omitempty"`
loadbalancer/v2/l7policies/requests.go: Invert bool `json:"invert,omitempty"`
```

Does it make sense to fix that as well?

kayrus:
same behavior with empty strings (e.g `Name` or `Description` fields)

jtopjian:
@kayrus It's really best to open an issue to discuss these changes rather than first submit a Pull Request for one change and then discuss further changes in the same Pull Request.

As for the validity of these changes, it's possible there are some parts of the code that need updated to use pointers and it's possible that the current code is correct and the various OpenStack API services are not consistent. We will need to see references to actual Python code in order to determine this.

_All of this (and more!) is covered in our [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial)_ 😄

kiwik:
recheck

jtopjian:
@kayrus If you remove `omitempty` from these fields, then they will _always_ send `false`, even if the field wasn't specified. This is why Travis is reporting an error.

Instead of setting `omitempty`, you will need to change the field from `bool` to `*bool`. If you grep the Gophercloud repo for `*bool`, you'll see a lot of examples to go off of.

kayrus:
@jtopjian I'm aware about the pointer. I didn't want to break the compatibility. That is why I created a #1332 issue to discuss changes.

jtopjian:
Sure - I'll jump over there :)

kayrus:
Closing. Fix will be completely reworked according to #1332.

jtopjian:
recheck

jtopjian:
@kayrus Nice - thank you for reviewing everything :)

Is this ready to review/merge?

kayrus:
@jtopjian yes.

Issue #1322, opened 2018-11-12
 Octavia: Load balancer pool support for PROXY


itsc0lin:
The create pool POST operation protocol attribute supports an additional argument: PROXY in the Rocky release of Octavia. This operation is currently not successful when using gophercloud due to the absence of the protocol definition in the listener and pools requests of the lbaas_v2 driver for Openstack.

itsc0lin:
Can see in this section that PROXY is among the supported protocols:

https://github.com/openstack/octavia/blob/a0112c4b378d02a4c3a0a86265859db54ce3de75/octavia/common/constants.py#L74-L76

jtopjian:
Implemented in #1323

PR #1323, opened 2018-11-12
Add PROXY protocol for listener/pool requests to LBaaS v2

itsc0lin:
For #1322

code in this PR:

https://github.com/openstack/octavia/blob/master/octavia/common/jinja/haproxy/templates/macros.j2#L180-L183

https://github.com/openstack/octavia/blob/a0112c4b378d02a4c3a0a86265859db54ce3de75/octavia/common/constants.py#L74-L76

### DIFF
--- a/openstack/loadbalancer/v2/listeners/requests.go
+++ b/openstack/loadbalancer/v2/listeners/requests.go
@@ -12,6 +12,7 @@ type Protocol string
 const (
 	ProtocolTCP   Protocol = "TCP"
 	ProtocolUDP   Protocol = "UDP"
+	ProtocolPROXY Protocol = "PROXY"
 	ProtocolHTTP  Protocol = "HTTP"
 	ProtocolHTTPS Protocol = "HTTPS"
 )
--- a/openstack/loadbalancer/v2/pools/requests.go
+++ b/openstack/loadbalancer/v2/pools/requests.go
@@ -67,6 +67,7 @@ const (
 
 	ProtocolTCP   Protocol = "TCP"
 	ProtocolUDP   Protocol = "UDP"
+	ProtocolPROXY Protocol = "PROXY"
 	ProtocolHTTP  Protocol = "HTTP"
 	ProtocolHTTPS Protocol = "HTTPS"
 )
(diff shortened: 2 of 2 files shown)


jtopjian:
@itsc0lin Thanks for this. This looks like it's ready for review/merge?

itsc0lin:
Hi @jtopjian wanted to ask before we merge if it would be alright to add this and the UDP support to the extensions subdir? Saw it was moved out of there and into this location after my last PR but am realizing that is going to lead to conflicts with sections like this one in cloud-provider-openstack: https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L29-L42 and am hoping not to have to refactor that work in order to refer to this new location for these resources.

jtopjian:
I'll be extra verbose just to make sure there's no confusion, since there's unfortunately always confusion between 

* https://github.com/gophercloud/gophercloud/tree/master/openstack/networking/v2/extensions/lbaas_v2
* https://github.com/gophercloud/gophercloud/tree/master/openstack/loadbalancer/v2

The first is for the deprecated Neutron LBaaS project which is an implementation of the LBaaS v2 API. The second is for the Octavia project which _also_ implements LBaaS v2 _plus_ more. 

So it's entirely possible to use the first to communicate with Octavia, but only up to a certain point. It appears `cloud-provider-openstack` is only using the second for cascading delete support (https://github.com/kubernetes/cloud-provider-openstack/blob/master/pkg/cloudprovider/providers/openstack/openstack_loadbalancer.go#L1528-L1533).

If you search the Neutron LBaaS v2 code and find that it also supports the `PROXY` protocol, you can add it to both locations.

If the `PROXY` protocol is only supported in Octavia, and someone wants to implement it in `cloud-provider-openstack`, then `cloud-provider-openstack` will have to start switching over to the Octavia packages. This will also limit compatibility to only Octavia deployments (rather than Neutron LBaaSv2).

zetaab:
@jtopjian that is true. However, it is much easier to start modifying those cloudprovider libs if this support is added to gophercloud. I have tested this and at least for me this PR works as should.

Could we please merge this?

jtopjian:
@zetaab Sure can.

zetaab:
thanks

Issue #1324, opened 2018-11-20
Reauth not occuring?


sb10:
I've been using AllowReauth with success and never thought about how and under what circumstances it works, but I hit a situation where I started getting "Authentication failed" and it never recovered.

I set up like:

```
opts, err := openstack.AuthOptionsFromEnv()
opts.AllowReauth = true
provider, err := openstack.AuthenticatedClient(opts)
computeClient, err = openstack.NewComputeV2(provider, gophercloud.EndpointOpts{
	Region: os.Getenv("OS_REGION_NAME"),
})
```

This then works fine when I use the computeClient to do things like create new servers over a period of days. Normally I don't get auth issues for any other usage either.

But on one occasion I got stuck here:

```
        pager := flavors.ListDetail(computeClient, flavors.ListOpts{})
	return pager.EachPage(func(page pagination.Page) (bool, error) {
		flavorList, err := flavors.ExtractFlavors(page)
		if err != nil {
			return false, err
		}

		for _, f := range flavorList {
			[...]
		}
		return true, nil
	})
```

This always returned err "Authentication failed" every time I called it. Since that fails it's possible I'm not attempting any other usage of computeClient.

Starting a new process with the exact same executable that calls the above code, with the same OS_* environment variables, from the same host, at the same time as that process continued to get "Authentication failed", showed no authentication issues.

In the above context, should Reauth be occurring? Or am I supposed to string match the err for "Authentication failed" and do a manual reauth? How?

mape90:
I have seen this same problem and it actually happens really often. I was able to identify the error and it seems to be because commit (3bb4676e61b8eda86b9d50ab2906708b78017115)

There the change isn't thread safe and in some cases other thread can fetch the nil function and then save it after its Request. After this no reauthentications will be made.

```
        reauthFunc := client.ReauthFunc
        client.ReauthFunc = nil
	resp, err = client.Request(method, url, options)
	client.ReauthFunc = reauthFunc
```

For solving this could it be possible to pass explicitly do we want to reauth or not? For example add parameter to `RequestOpts` to force reauth off. So that we could just copy the opts struct and set there the reauth off when we make the recursive `Request` call.

So we could do something like deep copy the options struct and set there the `optionsCopy.ReauthForceOff == true` and then change the reauth check from `client.ReauthFunc != nil ` to `( client.ReauthFunc != nil &&  options.ReauthForceOff == false )` and of cource add that `ReauthForceOff` to the `RequestOpts`

[...]

Issue #1244, opened 2018-09-21
Panic when Reauthenticate with multiple go routine


NBR41:
Hi,

it seems that there is a mistake introduced in the commit : 

```
commit 3bb4676e61b8eda86b9d50ab2906708b78017115
Author: Nathan Castelein <garbage-collector@users.noreply.github.com>
Date:   Fri Aug 24 04:02:31 2018 +0200

    Set reauth func to nil before making a new request (#1192)
    
    * Set reauth func to nil before making a new request
    
    * Add tests

diff --git a/provider_client.go b/provider_client.go
index 77368f0..95fa11a 100644
--- a/provider_client.go
+++ b/provider_client.go
@@ -295,7 +295,11 @@ func (client *ProviderClient) Request(method, url string, options *RequestOpts)
                                                seeker.Seek(0, 0)
                                        }
                                }
+                               // make a new call to request with a nil reauth func in order to avoid infinite loop
+                               reauthFunc := client.ReauthFunc
+                               client.ReauthFunc = nil
                                resp, err = client.Request(method, url, options)
+                               client.ReauthFunc = reauthFunc
                                if err != nil {
                                        switch err.(type) {
                                        case *ErrUnexpectedResponseCode:
```

I will try to explain the method to reproduce:
If two go routine need to Reauth with the same client when calling Request, the lock of client is too late in : 

```Go
func (client *ProviderClient) Reauthenticate(previousToken string) (err error) {
	if client.ReauthFunc == nil {
		return nil
	}

	if client.mut == nil {
		return client.ReauthFunc()
	}
	client.mut.Lock()
	defer client.mut.Unlock()

	client.reauthmut.Lock()
	client.reauthmut.reauthing = true
	client.reauthmut.Unlock()

	if previousToken == "" || client.TokenID == previousToken {
		err = client.ReauthFunc()
	}

	client.reauthmut.Lock()
	client.reauthmut.reauthing = false
	client.reauthmut.Unlock()
	return
}
``` 
Some times client.ReauthFunc is nil.
If two GoRoutine try to Reauth,
The first GoRoutine lock client, Reauth, release the Lock.
The second GoRoutine call Reauth, pass the test 
```if client.ReauthFunc == nil {```, 
and in the same time a second GoRoutine set client.ReauthFunc at nil in the method Request:
```
				reauthFunc := client.ReauthFunc
				client.ReauthFunc = nil
				resp, err = client.Request(method, url, options)
[...]

PR #1327, opened 2018-11-25
core: Revert setting reauth func to nil

jtopjian:
This commit reverts PR #1192. The goal of #1192 was to set the reauth
function to nil when performing reauthentication in order to prevent an
infinite loop. However, this was not done in a concurrent-safe way.
Having reviewed the use-case in more detail, I feel that scenarios in
which this will happen are more site-specific rather than an issue with
Gophercloud and a custom reauth function should be used.

Included in this PR is a simple unit test which verifies that a custom
reauthentication function can be used for these scenarios. However, it's
important to not use this example in production. Rather, users should
use the example provided in the FAQ.

For #1244 
For #1324

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -126,11 +126,11 @@ func (client *ProviderClient) SetToken(t string) {
 	client.TokenID = t
 }
 
-//Reauthenticate calls client.ReauthFunc in a thread-safe way. If this is
-//called because of a 401 response, the caller may pass the previous token. In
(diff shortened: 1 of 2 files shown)


jtopjian:
I'm going to go ahead and merge this PR since this fixes two issues.

If a fix such as #1192 is redetermined to be necessary, the unit test I've included fails when #1192 is re-applied, so it's a good way to determine if a new fix is valid or not.

PR #1329, opened 2018-11-29
Only set ResourceBase when we need to

clayg:
I'm having a bit of a time with go eco-system [1], I know PR could be more helpful, but it's all I was able to come up with tonight [2].

1. I got lost in the contributing guide.  Is godep still a thing?  it's dep now, maybe?  only golang 1.12 is all about the modules, right?
2. this is related to terraform-providers/terraform-provider-openstack#498

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -3,6 +3,7 @@ package openstack
 import (
 	"fmt"
 	"reflect"
+	"strings"
 
 	"github.com/gophercloud/gophercloud"
 	tokens2 "github.com/gophercloud/gophercloud/openstack/identity/v2/tokens"
@@ -321,7 +322,9 @@ func NewComputeV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpt
 // package.
 func NewNetworkV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
 	sc, err := initClientOpts(client, eo, "network")
-	sc.ResourceBase = sc.Endpoint + "v2.0/"
+	if !strings.HasSuffix(sc.Endpoint, "v2.0/") {
+		sc.ResourceBase = sc.Endpoint + "v2.0/"
+	}
 	return sc, err
 }
 


jtopjian:
@clayg Thanks for submitting this. Rather than start off with a PR, though, there should be a corresponding issue to discuss this change (as described here: https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-02-issues.md#step-2-create-an-issue). Tabling that for a moment:

Following the trail of links you gave, it looks as though you're trying to use the Terraform OpenStack provider with the Rackspace public cloud? If so, the Terraform provider is not compatible with Rackspace:

https://www.terraform.io/docs/providers/openstack/index.html#openstack-releases-and-versions

Gophercloud also doesn't guarantee compatibility.

I understand that _if_ we would merge this patch, you might get a little further with using the Terraform OpenStack provider and Rackspace, but there might be a good chance another error will crop up. But again, let's table Rackspace entirely.

Is there precedence for a "standard" OpenStack cloud to publish the network API endpoint with the `/v2.0` suffix? If so, that gives some weight to merging this. I understand that ideally we should be doing endpoint/version discovery within the client, but we're not at that point yet. I'm not totally against this patch, but given the only report we've had of Gophercloud's Network v2 client not working is with Rackspace, I'm not entirely for it. @jrperritt thoughts?

> I got lost in the contributing guide. Is godep still a thing? it's dep now, maybe? only golang 1.12 is all about the modules, right?

Which contributing guide? Gophercloud has its own here: https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial and Terraform OpenStack has some dev notes here: https://www.terraform.io/docs/providers/openstack/index.html#testing-and-development

As for vendor management, it's a bit of a mess right now and varies from project to project. Both Terraform core and the OpenStack provider use `govendor` but will be switching to Go modules soon. Gophercloud does not vendor at the moment.

If you have suggestions for either of the two guides I listed, please let me know - I'm always trying to minimize the barrier to entry for both of these projects.

clayg:
@jtopjian Hey!  Thanks for all the pointers - and your patience.  Does #1330 look like a good place to start?  Do you hang out on Freenode or Gitter or anything like that?  Maybe it'd be useful if we could high-bandwidth at some point - I think we're timezone compatible. :call_me_hand: :sunglasses:

jtopjian:
@clayg I've also opened #1331 as the "official" corresponding issue to this PR and tagged it in your original message here.

jrperritt:
> I'm inclined to merge it

Let's not. Like @jtopjian mentioned in his first comment, this library is for vanilla OpenStack. It probably can be made to work with Rackspace Public Cloud, but the onus is on the user for that. We don't want to start making (albeit small) changes to get Gophercloud working with all OpenStack-compatible clouds.

@clayg For this PR specifically, you could create the networking client and then modify the service client's `ResourceBase` field to suit your needs (e.g. using something like `strings.Replace(sc.ResourceBase, "/v2.0", "", 1)`).

jtopjian:
@jrperritt I agree, but in this case,

> For this PR specifically, you could create the networking client and then modify the service client's ResourceBase field to suit your needs (e.g. using something like strings.Replace(sc.ResourceBase, "/v2.0", "", 1)).

"you" is actually "me" since this fix is ultimately for Terraform, which is, selfishly, one reason I'm inclined to include it. 

From the notes I added to #1331, this isn't a Rackspace setting per-se. IMO it falls under one of those quick-and-easy fixes that make life easier for other people. Perhaps another option is to make some kind of abstracted function in `gophercloud/utils`?

jtopjian:
And with all of that said, adding this directly to Terraform directly is something I can certainly do, too.

jrperritt:
> From the notes I added to #1331, this isn't a Rackspace setting per-se.

I agree (`all OpenStack-compatible clouds`).

> IMO it falls under one of those quick-and-easy fixes that make life easier for other people.

True, but it only makes life easier for people on clouds with this specific setting. What if a cloud provider has that networking version set as `2.0` instead of `v2.0`? Anything deviating from OpenStack-proper can be handled in user code (or separate library).

jtopjian:
Fair point. This goes back to the automatic endpoint and version discovery being the best long-term solution and everything else being duct-tape.

I'm still keen on having the discussion in #1330 to at least have a written account of what all of the differences look like. 

@clayg For this specific fix, I don't mind adding this directly to Terraform [here](https://github.com/terraform-providers/terraform-provider-openstack/blob/master/openstack/config.go#L259-L264). But before I do that, would you be able to do some further testing to see how much further you get with using the Terraform OpenStack provider with Rackspace's networking service? If you're able to manage networks, subnets, and ports, IMO that's worthwhile of making the change, keeping the usual disclaimer of not officially supporting Rackspace.

clayg:
Hey thanks for all the input @jrperritt and @jtopjian - seems like the best places to make progress are on terraform-providers/terraform-provider-openstack#498 and #1330 - let's close this down!  :+1:

Issue #1304, opened 2018-10-18
Networking V2: VLAN transparency extension


ozerovandrei:
Gophercloud should support OpenStack Networking V2 `vlan-transparent` extension that adds VLAN transparency to deliver VLAN transparent trunk networks.

Links:
[API](https://developer.openstack.org/api-ref/network/v2/#vlan-transparency-extension)
[Spec](https://specs.openstack.org/openstack/neutron-specs/specs/kilo/nfv-vlan-trunks.html)
[Terraform-provider-openstack#430](https://github.com/terraform-providers/terraform-provider-openstack/issues/430)
[Code](https://github.com/openstack/neutron-lib/blob/56033ba643812a30577f6ab17648806c2ee494ba/neutron_lib/api/definitions/vlantransparent.py)

ozerovandrei:
@jtopjian what do you think is better way to implement this: in a separate `networking/v2/extensions/vlantransparent` package or just by adding a new field into the existing `Network` struct?

I guess that separate package is better so it will follow current Gophercloud extensions scheme. This package will work the same way as the `router:external` extension package (it will only add one field for now and provide customized `ToNetworkListQuery`, `ToNetworkCreateMap`, `ToNetworkUpdateMap` methods).

In any case we will need to disable acceptance tests for this extension because I'm not sure that OpenLab provides it. Devstack has the `DEFAULT/vlan_transparent` parameter set to `false` by default and there is no support for Openvswitch ML2 plugin yet: https://review.openstack.org/#/c/576687/

jtopjian:
@ozerovandrei Yes, let's implement this as an extension similar to `router:external`.

jtopjian:
@ozerovandrei I've also updated this Issue with a link to the code.

ozerovandrei:
@jtopjian thanks.

I added 3 PRs but they're failing because of gofmt issues in another packages
```
openstack/container/v1/capsules/template.go
openstack/orchestration/v1/stacks/utils.go
acceptance/openstack/networking/v2/extensions/trunks/trunks_test.go
```

jtopjian:
Must be a new version of `goimports` causing the changes. I'll get those cleaned up.

jtopjian:
@ozerovandrei everything should be working now. Travis tests should pass with no rebasing, but can you rebase the other PRs so we can get OpenLab to pass (even though they're not explicitly tested).

jtopjian:
Implemented in #1338 

Thank you! :tada:

PR #1337, opened 2018-12-04
NetworkingV2: add vlan-transparent Create method

ozerovandrei:
Add a method to create an OpenStack Networking V2 Network with the "vlan-transparent" attribute set.

For #1304

code in this PR:

[vlantransparent.py](https://github.com/openstack/neutron-lib/blob/stable/rocky/neutron_lib/api/definitions/vlantransparent.py)


jtopjian:
Merged as part of #1338

PR #1343, opened 2018-12-06
feature/shared file systems: update share

kayrus:
#129 - add share update action with tests

### DIFF
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -86,12 +86,16 @@ func AcceptanceTestChoicesFromEnv() (*AcceptanceTestChoices, error) {
 	if externalNetworkID == "" {
 		missing = append(missing, "OS_EXTGW_ID")
 	}
+
+	/* // Temporarily disabled, see https://github.com/gophercloud/gophercloud/issues/1345
 	if networkID == "" {
 		missing = append(missing, "OS_NETWORK_ID")
 	}
 	if subnetID == "" {
 		missing = append(missing, "OS_SUBNET_ID")
 	}
+	*/
+
 	if networkName == "" {
 		networkName = "private"
 	}
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -38,11 +38,14 @@ type AcceptanceTestChoices struct {
 	// NetworkName is the name of a network to launch the instance on.
 	NetworkName string
 
+	// NetworkID is the ID of a network to launch the instance on.
+	NetworkID string
+
+	// SubnetID is the ID of a subnet to launch the instance on.
+	SubnetID string
+
 	// ExternalNetworkID is the network ID of the external network.
 	ExternalNetworkID string
-
-	// ShareNetworkID is the Manila Share network ID
-	ShareNetworkID string
 
 	// DBDatastoreType is the datastore type for DB tests.
 	DBDatastoreType string
@@ -60,9 +63,10 @@ func AcceptanceTestChoicesFromEnv() (*AcceptanceTestChoices, error) {
 	magnumImageID := os.Getenv("OS_MAGNUM_IMAGE_ID")
 	magnumKeypair := os.Getenv("OS_MAGNUM_KEYPAIR")
 	networkName := os.Getenv("OS_NETWORK_NAME")
+	networkID := os.Getenv("OS_NETWORK_ID")
+	subnetID := os.Getenv("OS_SUBNET_ID")
 	floatingIPPoolName := os.Getenv("OS_POOL_NAME")
 	externalNetworkID := os.Getenv("OS_EXTGW_ID")
-	shareNetworkID := os.Getenv("OS_SHARE_NETWORK_ID")
 	dbDatastoreType := os.Getenv("OS_DB_DATASTORE_TYPE")
 	dbDatastoreVersion := os.Getenv("OS_DB_DATASTORE_VERSION")
 
@@ -82,11 +86,14 @@ func AcceptanceTestChoicesFromEnv() (*AcceptanceTestChoices, error) {
 	if externalNetworkID == "" {
 		missing = append(missing, "OS_EXTGW_ID")
 	}
+	if networkID == "" {
+		missing = append(missing, "OS_NETWORK_ID")
+	}
+	if subnetID == "" {
+		missing = append(missing, "OS_SUBNET_ID")
+	}
 	if networkName == "" {
 		networkName = "private"
-	}
-	if shareNetworkID == "" {
-		missing = append(missing, "OS_SHARE_NETWORK_ID")
 	}
 	notDistinct := ""
 	if flavorID == flavorIDResize {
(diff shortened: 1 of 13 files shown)


script/acceptancetest
```diff
@@ -107,6 +107,11 @@ if [[ $? != 0 ]]; then
   failed=1
 fi
 
+# Shared Filesystems v2
```
jtopjian: This is a good idea and I'm actually surprised any tests are passing. It looks like we'll have to review all of the tests and make sure they're using settings suitable for a standard devstack installation. For now, let's omit this.

jtopjian: Additionally, feel free to use issue #818 for any PRs that are only for acceptance tests.

kayrus: Seems so. I fixed the tests and now they run fine in my local manila environment, however they still don't run in openlabtesting.org env. I've added a stub in the shell script and created an issue: #1345

kayrus:
Ready to be merged

acceptance/clients/clients.go
```diff
@@ -82,12 +86,15 @@ func AcceptanceTestChoicesFromEnv() (*AcceptanceTestChoices, error) {
 	if externalNetworkID == "" {
 		missing = append(missing, "OS_EXTGW_ID")
 	}
+	if networkID == "" {
+		missing = append(missing, "OS_NETWORK_ID")
+	}
+	if subnetID == "" {
+		missing = append(missing, "OS_SUBNET_ID")
+	}
```
jtopjian: This might be a breaking change for OpenLab. The appropriate changes are being made to the `stackenv` script, but I want to make sure that script is being used both in Gophercloud and Terraform testing before merging this PR.

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -341,3 +341,41 @@ func Shrink(client *gophercloud.ServiceClient, id string, opts ShrinkOptsBuilder
...
+type UpdateOptsBuilder interface {
+	ToShareUpdateMap() (map[string]interface{}, error)
+}
+
+// UpdateOpts contain options for updating an existing Share. This object is passed
+// to the share.Update function. For more information about the parameters, see
+// the Share object.
+type UpdateOpts struct {
+	// Share name
+	Name *string `json:"display_name,omitempty"`
```
jtopjian: `CreateOpts` has both a `Name` and `DisplayName`. Are these actually aliases to the same field?

Either way, let's change this to `DisplayName` so it matches the JSON tag.

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -341,3 +341,41 @@ func Shrink(client *gophercloud.ServiceClient, id string, opts ShrinkOptsBuilder
...
+	// Share description
+	Description *string `json:"display_description,omitempty"`
```
jtopjian: Same here.

openstack/sharedfilesystems/v2/shares/requests.go
kayrus: No, they are not aliases. `create` and `get` functions do support aliases: [create](https://github.com/openstack/manila/blob/4f22d0b19e782800b06deba74e6ec15708e5db33/manila/api/v1/shares.py#L246..L255), [get](https://github.com/openstack/manila/blob/4f22d0b19e782800b06deba74e6ec15708e5db33/manila/api/v1/shares.py#L149..L156).

Let's keep common `Name` and `Description` fields names at least in gophercloud?I don't want to copy this confusion from openstack.

jtopjian: The concern I have is that the field name is different than the JSON tag:

```go
Name *string `json:"display_name,omitempty"`
```

I'd much prefer changing this to:

```go
DisplayName *string `json:"display_name,omitempty"`
```

I totally agree that Manila is handling `Name` and `DisplayName` in a weird and confusing way, but I'd prefer not to have Gophercloud fix this behavior. Gophercloud is a thin layer between an app and OpenStack - warts and all. If Gophercloud would fix weird behaviors in OpenStack, there would be a few thousand more lines of code throughout it :smile: 

The better solution would be to fix this in the app - for example, Terraform. In Terraform, you can either one of `Name` or `DisplayName` and hide the confusion within the app. If this becomes repetitive between different applications, then a central, shared utility function is the next best solution.

Does this make sense?

acceptance/openstack/sharedfilesystems/v2/shares.go
```diff
@@ -87,24 +87,17 @@ func GetAccessRightsSlice(t *testing.T, client *gophercloud.ServiceClient, share
...
 		t.Errorf("Unable to delete share %s: %v", share.ID, err)
 	}
 
-	<-done
+	err = waitForStatus(client, share.ID, "deleted", 600)
+	if err != nil {
+		t.Errorf("Failed to wait for 'deleted' status for %s share: %v", share.ID, err)
+	} else {
+		t.Logf("Deleted share: %s", share.ID)
+	}
```
jtopjian: Thank you for adding this commit.

I saw this when I did the review and debated asking to change it. On one hand the original solution was correct and I wanted to allow some artistic creativity, but on the other, the modified solution is the better way to go.

kayrus: I added go routine when tried to debug failed tests during deletion. Then, I realized that `waitForStatus` function doesn't correctly handle deleted resources. I fixed `waitForStatus`, but forgot to remove go routine.

openstack/sharedfilesystems/v2/shares/requests.go
kayrus: Ok, I'll keep display prefix.

acceptance/clients/clients.go
kayrus: How exactly this might be a breaking change for OpenLab?

jtopjian: Because this change is now requiring these two environment variables to be set (by adding them to the slice of `missing` variables). So the scripts which execute the acceptance tests need to have these variables set and configured.

kayrus: Ok, i'll just comment out these requirements.

jtopjian: You'll also have to make sure the tests for SFS/Manila aren't enabled for the time being. Once it's confirmed that these environment variables have been set in OpenLab, you can have SFS/Manila begin running through OpenLab.

If you want to hold off for another day or two, it shouldn't take long for the OpenLab team to get back to us.

kayrus: They are not enabled, see this comment:

```shell
# Temporarily disabled, see https://github.com/gophercloud/gophercloud/issues/1345
true || go test -v -timeout $timeout -tags "fixtures acceptance" ./acceptance/openstack/sharedfilesystems/v2/
```

kayrus: > If you want to hold off for another day or two, it shouldn't take long for the OpenLab team to get back to us.

We can enable unsupported tests in a new PR.

jtopjian: All sounds good - thank you!

kayrus:
@jtopjian could you please resolve the [requested change](https://github.com/gophercloud/gophercloud/pull/1343#pullrequestreview-182929349)? Github didn't detect that the code was commented.

jtopjian:
@kayrus Github doesn't automatically change review - has to be manually approved after a re-review. This looks good to me and I'll go ahead and merge it.

PR #1353, opened 2018-12-13
Senlin: Nodes Op

JackKuei:

code in this PR:

[[nodes:op api]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/api/openstack/v1/nodes.py#L197)
[[nodes:op engine]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/engine/node.py#L394)
[[nova op names]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/profiles/os/nova/server.py#L241)
[[heat op names]](https://github.com/openstack/senlin/blob/e5bc16b3fea2f10dead365f983c5a937f23d72b0/senlin/profiles/os/heat/stack.py#L88)

### DIFF
--- a/openstack/clustering/v1/nodes/doc.go
+++ b/openstack/clustering/v1/nodes/doc.go
@@ -79,7 +79,7 @@ Example to Perform an Operation on a Node
 		Operation: nodes.RebootOperation,
 		Params:    nodes.OperationParams{"type": "SOFT"},
 	}
-    actionID, err := nodes.Ops(serviceClient, nodeID, operationOpts).Extract()
+	actionID, err := nodes.Ops(serviceClient, nodeID, operationOpts).Extract()
 	if err != nil {
 		panic(err)
 	}
--- a/openstack/clustering/v1/nodes/doc.go
+++ b/openstack/clustering/v1/nodes/doc.go
@@ -70,5 +70,19 @@ Example to Get a Node
 	}
 
 	fmt.Printf("%+v\n", node)
+
+Example to Perform an Operation on a Node
+
+	serviceClient.Microversion = "1.4"
+	nodeID := "node123"
+	opsOpts := nodes.OpOpts{
+		Operation: nodes.RebootOperation,
+		Params:    nodes.OpParams{"type": "SOFT"},
+	}
+    actionID, err := nodes.Ops(serviceClient, nodeID, opsOpts).Extract()
+	if err != nil {
+		panic(err)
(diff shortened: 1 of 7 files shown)


JackKuei:
@jtopjian This one is ready to be reviewed. Thx.

JackKuei:
Yes, changing from OpOptsBuilder => OperationOptsBuilder would definitely make it clearer for the user. I was debating with that since all the other functions follows the pattern of OpOptsBuilder even though "OpOpts" definitely is very confusing. I'm OK with the wording changes.

jtopjian:
Sounds good.

Yes, I think this is definitely a case that warrants breaking standard styles :)

JackKuei:
As for the struct changes, it basically boils down to which one looks better.

Operation with params:
```
{ 
 Operation: nodes.RebootOperation, 
 Params:    nodes.OpParams{"type": "SOFT"}
},
```
OR
```
{
     nodes.RebootOperation: nodes.OpParams{
          "type": "SOFT",
     },
},
```

Operation without params:
```
{
   Operation: nodes.LockOperation
},
```
OR 
```
{
    nodes.LockOperation: nil,
}
```

jtopjian:
Let's keep the existing structure.

The alternative format is useful because it more closely mirrors the JSON body, but it feels a little awkward in Go. I prefer to have an explicit field for the Operation.

JackKuei:
Sounds good

openstack/clustering/v1/nodes/doc.go
```diff
@@ -70,5 +70,19 @@ Example to Get a Node
...
+	nodeID := "node123"
+	operationOpts := nodes.OperationOpts{
+		Operation: nodes.RebootOperation,
+		Params:    nodes.OperationParams{"type": "SOFT"},
+	}
+    actionID, err := nodes.Ops(serviceClient, nodeID, operationOpts).Extract()
+	if err != nil {
+		panic(err)
+	}
+
```
jtopjian: @JackKuei I'm so sorry - I didn't see this before. Looks like there's a formatting issue here.

Other than that, this LGTM and I can merge once fixed.

JackKuei: Good catch! Making a fix

Issue #1352, opened 2018-12-13
LBaaSv2 / Octavia: Add l7policy member for the listener structure


kayrus:
Mitaka neutron-lbaas extension doesn't return the listener_id for the l7policy, thus it is impossible to identify the loadbalancer status (pending, error, etc).

LB statuses API returns the whole tree of the dependent LB components. Therefore it is possible to detect the LB associated with the L7listener. Adding the the string below won't harm, since it is a result structure:

```diff
diff --git a/vendor/github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/listeners/results.go b/vendor/github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/listeners/results.go
index 81cda90..1ec6a5a 100644
--- a/vendor/github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/listeners/results.go
+++ b/vendor/github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/listeners/results.go
@@ -2,6 +2,7 @@ package listeners
 
 import (
        "github.com/gophercloud/gophercloud"
+       "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/l7policies"
        "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/lbaas_v2/pools"
        "github.com/gophercloud/gophercloud/pagination"
 )
@@ -55,6 +56,9 @@ type Listener struct {
        // Pools are the pools which are part of this listener.
        Pools []pools.Pool `json:"pools"`
 
+       // L7policies are the L7 policies which are part of this listener.
+       L7Policies []l7policies.L7Policy `json:"l7policies"`
+
        // The provisioning status of the listener.
        // This value is ACTIVE, PENDING_* or ERROR.
        ProvisioningStatus string `json:"provisioning_status"`
```

I honestly tried to find the corresponding openstack python code, which handles this tree behavior, but failed. I just know that it works in Mitaka. Test could be done via curl:

```
$ curl -H "Content-Type: application/json" -H "X-Auth-Token: ${OS_AUTH_TOKEN}" 'http://api:5000/v2.0/lbaas/loadbalancers/977ce39f-3326-489c-8594-023d8b6cc5cf/statuses' | jq -r
```

 Objections?

jtopjian:
[...]

PR #1354, opened 2018-12-13
LBaaSv2/Octavia: add "L7Policies" member into the Listerner struct

kayrus:
Resolves: #1352


kayrus:
@jtopjian can we merge it and I'll run tests in https://github.com/terraform-providers/terraform-provider-openstack/pull/520?

PR #1351, opened 2018-12-12
LBaaS_v2: L7policy redirect_pool_id and redirect_url should not be omitted during update

kayrus:
Resolves: #1350

PR #1356, opened 2018-12-14
LBaaS_v2: Listener's default_pool_id should not be omitted during update

### LINKED ISSUE #1355: LBaaS_v2: don't omit IDs during the update
Similar to #1350.

Listener field should be `null` if it is empty:
* default_pool_id [ref](https://github.com/openstack/neutron-lbaas/blob/mitaka-eol/neutron_lbaas/extensions/sharedpools.py#L55)

kayrus:
Resolves #1355
Missed this field.

### DIFF
--- a/acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
+++ b/acceptance/openstack/loadbalancer/v2/loadbalancers_test.go
@@ -213,6 +213,43 @@ func TestLoadbalancersCRUD(t *testing.T) {
 	defer DeleteL7Policy(t, lbClient, lb.ID, policy.ID)
 	defer DeleteL7Rule(t, lbClient, lb.ID, policy.ID, rule.ID)
 
+	// Update listener's default pool ID
+	updateListenerOpts = listeners.UpdateOpts{
+		DefaultPoolID: &pool.ID,
+	}
+	_, err = listeners.Update(lbClient, listener.ID, updateListenerOpts).Extract()
+	th.AssertNoErr(t, err)
+
+	if err := WaitForLoadBalancerState(lbClient, lb.ID, "ACTIVE", loadbalancerActiveTimeoutSeconds); err != nil {
+		t.Fatalf("Timed out waiting for loadbalancer to become active")
+	}
+
+	newListener, err = listeners.Get(lbClient, listener.ID).Extract()
+	th.AssertNoErr(t, err)
+
+	tools.PrintResource(t, newListener)
+
+	th.AssertEquals(t, newListener.DefaultPoolID, pool.ID)
+
+	// Remove listener's default pool ID
+	emptyPoolID := ""
+	updateListenerOpts = listeners.UpdateOpts{
+		DefaultPoolID: &emptyPoolID,
+	}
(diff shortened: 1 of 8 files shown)


jtopjian:
@kayrus After looking at this, I made a mistake with approving #1351.

The changes here and changes in #1351 will require the user to have to set the fields to their current value unless they want the value to be removed. This isn't what we want.

Instead, look at how `gateway_ip` is handled in Neutron subnets: https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/subnets/requests.go#L203-L205

So #1351 will need reverted/modified and the changes here will need reverted, too, and then a special case in the `UpdateMap` will need to be added. I apologize about this - it's my fault for not catching it earlier.

kayrus:
Maybe I got you wrong, but last time (sorry, cannot find the evidence, too many messages, give me some time, maybe I'll find it) you told me that gophercloud should not implement logic on input data, that is why I already prepared a commit for terraform, but not for gophercloud:

```diff
        if d.HasChange("default_pool_id") {
                defaultPoolID := d.Get("default_pool_id").(string)
-               updateOpts.DefaultPoolID = &defaultPoolID
+               if defaultPoolID != "" {
+                       updateOpts.DefaultPoolID = &defaultPoolID
+               }
        }
```

jtopjian:
Yes, in #1343, I was saying that we should not mask that Manila has `name` and `display_name`.

In this case, this is a limitation of Go. If someone were to pass an empty `UpdateOpts`, they would end up removing their `redirect_pool_id` et al, possibly unintentionally. That's Gophercloud and Go causing that - not Neutron / Octavia.

There are two ways to get around this:

1. The user _must_ specify `RedirectPoolID` et al regardless of if they are updating it or not. 
2. Have Gophercloud use an empty string as a way of saying "The user wants to pass `null`".

> UPD: In addition, unlike https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/subnets/requests.go#L203-L205, redirect_pool_id, redirect_url and key are not accepted by neutron API as an empty string, but only as null.

We're using an empty string to trigger a null value:

```go
// If `gateway_ip` is an empty string
if m := b["subnet"].(map[string]interface{}); m["gateway_ip"] == "" {
  // Then change it from an empty string to `nil` which will be sent as `null` in JSON form.
  m["gateway_ip"] = nil
}
```


Please let me know if this clarifies things.

kayrus:
Ok, I'll prepare an update. Do I need to create a new issue and a new PR for each component?

jtopjian:
Since this is my mistake for not catching it before, I'll go ahead and do a PR for #1351. This way you only need to re-do this PR.

kayrus:
This is not your mistake, I just got you wrong on [this comment](https://github.com/gophercloud/gophercloud/pull/1343#discussion_r240014764) and I explicitly didn't put additional logic. Nevertheless, thank you a lot for your effort helping me.

kayrus:
> Coverage decreased (-0.01%)

How come?

jtopjian:
Coveralls can act weird a lot of times - it's safe to ignore. :)

kayrus:
Fixed test. Forgot to explicitly set an empty string for the `DefaultPoolID`.

PR #1359, opened 2018-12-15
Manila: add messages support

### LINKED ISSUE #1358: Manila: add messages support
In order to identify the exact error message of failed manila share, [messages API](https://developer.openstack.org/api-ref/shared-file-system/#user-messages-since-api-2-37) could be used.

kayrus:
Resolves #1358

### DIFF
--- a/acceptance/openstack/sharedfilesystems/v2/messages/messages.go
+++ b/acceptance/openstack/sharedfilesystems/v2/messages/messages.go
@@ -1,0 +1,19 @@
+package messages
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/sharedfilesystems/v2/messages"
+)
+
+// DeleteMessage will delete a message. An error will occur if
+// the message was unable to be deleted.
+func DeleteMessage(t *testing.T, client *gophercloud.ServiceClient, message *messages.Message) {
+	err := messages.Delete(client, message.ID).ExtractErr()
+	if err != nil {
+		t.Fatalf("Failed to delete message %s: %v", message.ID, err)
+	}
+
+	t.Logf("Deleted message: %s", message.ID)
+}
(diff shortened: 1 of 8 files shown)


kayrus:
recheck

jtopjian:
If it's happening here, too, I'm leaning on an OpenLab problem. Since it's the weekend, it might not get resolved for another day or two.

kayrus:
recheck

kayrus:
Ready for review.

acceptance/openstack/sharedfilesystems/v2/messages/messages.go
```diff
@@ -0,0 +1,34 @@
...
+	err := messages.Delete(client, message.ID).ExtractErr()
+	if err != nil {
+		t.Fatalf("Failed to delete message %s: %v", message.ID, err)
+	}
+
+	t.Logf("Deleted message: %s", message.ID)
+}
+
+// PrintMessage will print a message and all of its attributes.
+func PrintMessage(t *testing.T, message *messages.Message) {
```
jtopjian: Can `tools.PrintResource` be used instead? If you really want to see the `CreatedAt` and `ExpiresAt` times, you should be able to do `tools.PrintResource(t, message.CreatedAt)`.

openstack/sharedfilesystems/v2/messages/results.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+const MinimumManilaMessagesMicroVersion = "2.37"
```
jtopjian: microversions are set during the request, so if we would be adding constants like this, it would go in `requests.go` instead. However, let's remove this - this opens the door to having Gophercloud maintain these kinds of contants for every microversion version and feature.

acceptance/openstack/sharedfilesystems/v2/messages/messages.go
kayrus: Done

PR #1369, opened 2018-12-20
Amend example code

npu21:
For #1372 

Amend example code in doc.go
client := openstack.NewComputeV2(provider, opts) should be
client, err := openstack.NewComputeV2(provider, opts)


npu21:
@jtopjian 
Sure,Thanks!

Issue #1375, opened 2018-12-25
NetworkingV2: Address scopes support


ozerovandrei:
Gophercloud should support the `v2.0/address-scopes` API.
This feature is needed for [terraform-provider-openstack#364](https://github.com/terraform-providers/terraform-provider-openstack/issues/364)

- [x] Address scopes List #1376
- [x] Address scopes Get #1377 
- [x] Address scopes Create #1381 
- [x] Address scopes Update
- [x] Address scopes Delete
- [x] Address scopes CRUD acceptance test

[API reference](https://developer.openstack.org/api-ref/network/v2/#address-scopes)

jtopjian:
@ozerovandrei Thank you! :tada:

PR #1376, opened 2018-12-25
NetworkingV2: implement address-scopes List call

ozerovandrei:
Add "address-scopes" package with List method.

For #1375 

code in this PR:

[API reference](https://developer.openstack.org/api-ref/network/v2/#address-scopes)
[Code reference(API)](https://github.com/openstack/neutron-lib/blob/stable/rocky/neutron_lib/api/definitions/address_scope.py)
[Code reference(get collection call)](https://github.com/openstack/neutron/blob/stable/rocky/neutron/extensions/address_scope.py#L31)
[Code reference(DB)](https://github.com/openstack/neutron/blob/stable/rocky/neutron/db/address_scope_db.py#L100)

### DIFF
--- a/openstack/networking/v2/extensions/layer3/addressscopes/urls.go
+++ b/openstack/networking/v2/extensions/layer3/addressscopes/urls.go
@@ -2,7 +2,7 @@ package addressscopes
 
 import "github.com/gophercloud/gophercloud"
 
-const resourcePath = "addressscopes"
+const resourcePath = "address-scopes"
 
 func rootURL(c *gophercloud.ServiceClient) string {
 	return c.ServiceURL(resourcePath)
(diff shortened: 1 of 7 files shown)


ozerovandrei:
This is ready.

openstack/networking/v2/extensions/layer3/addressscopes/urls.go
```diff
@@ -0,0 +1,13 @@
+package addressscopes
+
+import "github.com/gophercloud/gophercloud"
+
+const resourcePath = "addressscopes"
```
jtopjian: I think this should be `address-scopes`. The API docs have this and my test environment fails when using `addressscopes`.

jtopjian:
recheck

Issue #1361, opened 2018-12-17
Manila: support extended error messages returned for requests


kayrus:
Manila can [return](https://github.com/openstack/manila/blob/ef418b4e8db4c2037897986011f010a5e1d9780f/manila/api/openstack/wsgi.py#L1262..L1271) detailed error [messages](https://github.com/openstack/manila/blob/ef418b4e8db4c2037897986011f010a5e1d9780f/manila/api/openstack/wsgi.py#L1286..L1287) on failed requests, e.g.:

```json
{
  "itemNotFound": {
    "code": 404,
    "message": "ShareSnapshotNotFound: Snapshot 70bfbebc-d3ff-4528-8bbb-58422daa280b could not be found."
  }
}
```

It would be great if gophercloud can extract these messages and bypass as an error result.

In addition to `message` field, there is a `details` [field](https://github.com/openstack/manila/blob/ef418b4e8db4c2037897986011f010a5e1d9780f/manila/api/openstack/wsgi.py#L1337).

Is there a similar behavior somewhere in gophercloud code?

kayrus:
Some investigations:

This code [returns error](https://github.com/gophercloud/gophercloud/blob/94924357ebf6c7d448c70d65082ff7ca6f78ddc5/results.go#L45) right away, if the response code is not corresponding to expected ones.

In case of the error, here [only the error message](https://github.com/gophercloud/gophercloud/blob/94924357ebf6c7d448c70d65082ff7ca6f78ddc5/openstack/sharedfilesystems/v2/shares/results.go#L106) is returned.

If we create a new error structure and try to extract `err.(gophercloud.ErrDefault%%%).ErrUnexpectedResponseCode.Body` into this structure, we can print verbose error messages returned by the API.

kayrus:
The first attempt: https://github.com/kayrus/gophercloud/commit/9ded55b66bcb9446b619016e821b6200feed21f2

However I think it does make sense to put the error unmarshalling into the low level, e.g. into the `provider_client.go` -> `Request` func

jtopjian:
I think having access to a more detailed error struct is worthwhile, but I'm not sure this is something that we'd want to be handled automatically. Instead, it might be better to return the error as-is and provide the option of unmarshaling it into a struct.

Perhaps a package such as `openstack/sharedfilesystems/v2/errors` which would define the error struct and be used in the following way in an application:

```go
var detailedErr errors.ErrorDetails

[...]

PR #1378, opened 2019-01-03
Manila: add response error messages support

kayrus:
Resolves: #1361


kayrus:
@jtopjian, ready for review

PR #1381, opened 2019-01-04
NetworkingV2: add Address scopes Create method

ozerovandrei:
Implement Create method for the "address-scopes" extension with tests and documentation example.

For #1375 

code in this PR:

[API reference](https://developer.openstack.org/api-ref/network/v2/#address-scopes)
[Code reference(API)](https://github.com/openstack/neutron-lib/blob/stable/rocky/neutron_lib/api/definitions/address_scope.py)
[Code reference(DB)](https://github.com/openstack/neutron/blob/stable/rocky/neutron/db/address_scope_db.py#L71)


ozerovandrei:
All fixed and ready.

Issue #1380, opened 2019-01-04
Manila: add share snapshots support


kayrus:
manila api v2 uses [manila/api/v1/share_snapshots.py](https://github.com/openstack/manila/blob/e96e46d0d826f4b654773d081094b05e79f33c09/manila/api/v1/share_snapshots.py)

[additional info](https://developer.openstack.org/api-ref/shared-file-system/#list-share-snapshots)

PR #1382, opened 2019-01-06
Manila: add share snapshots support (list, get)

kayrus:
Part of the #1380 

* https://github.com/openstack/manila/blob/e96e46d0d826f4b654773d081094b05e79f33c09/manila/api/v1/share_snapshots.py#L88
* https://github.com/openstack/manila/blob/e96e46d0d826f4b654773d081094b05e79f33c09/manila/api/views/share_snapshots.py#L46

Acceptance tests don't make sense before create/delete methods are implemented.

@jtopjian not sure how exactly to name the package: `sharesnapshots` or `snapshots`

### DIFF
--- a/openstack/sharedfilesystems/v2/snapshots/requests.go
+++ b/openstack/sharedfilesystems/v2/snapshots/requests.go
@@ -1,0 +1,73 @@
+package snapshots
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// ListOpts holds options for listing Snapshots. It is passed to the
+// snapshots.List function.
+type ListOpts struct {
+	// (Admin only). Defines whether to list the requested resources for all projects.
+	AllTenants bool `q:"all_tenants"`
+	// The snapshot name.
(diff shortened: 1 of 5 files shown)


kayrus:
@jtopjian ready for review

kayrus:
@jtopjian any chance to take a look on this?

jtopjian:
> Acceptance tests don't make sense before create/delete methods are implemented.

That's fine. Acceptance tests before delete are usually fine since we anticipate this to run in a disposable environment and we don't care if the snapshot is deleted.

> not sure how exactly to name the package: sharesnapshots or snapshots

That's a good question. Are there any other kind of snapshots? If so, then use `sharesnapshots`. If not, then `snapshots` is fine. In hindsight, `sharenetworks` could have been `networks` but oh well.

kayrus:
According to Openstack internals, share networks have a `share_network` key in results. Share snapshots have `snapshots` key. Therefore I guess I'll rename this package. And the same will apply on the Terraform resource: `openstack_sharedfilesystem_snapshot_v2`.

kayrus:
recheck

PR #1338, opened 2018-12-04
NetworkingV2: add vlan-transparent Update method

ozerovandrei:
Add a method to update an OpenStack Networking V2 Network with the
    "vlan-transparent" attribute.

For #1304

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:

[vlantransparent.py](https://github.com/openstack/neutron-lib/blob/stable/rocky/neutron_lib/api/definitions/vlantransparent.py)

PR #1392, opened 2019-01-09
Manila: add share snapshot support (tests)

kayrus:
Part of the  #1380

### DIFF
--- a/acceptance/openstack/sharedfilesystems/v2/securityservices.go
+++ b/acceptance/openstack/sharedfilesystems/v2/securityservices.go
@@ -43,20 +43,3 @@ func DeleteSecurityService(t *testing.T, client *gophercloud.ServiceClient, secu
 
 	t.Logf("Deleted security service: %s", securityService.ID)
 }
-
-// PrintSecurityService will print a security service and all of its attributes.
-func PrintSecurityService(t *testing.T, securityService *securityservices.SecurityService) {
-	t.Logf("ID: %s", securityService.ID)
-	t.Logf("Project ID: %s", securityService.ProjectID)
-	t.Logf("Domain: %s", securityService.Domain)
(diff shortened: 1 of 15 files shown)


kayrus:
@jtopjian ready for review

kayrus:
recheck

kayrus:
recheck

jtopjian:
@kayrus This needs rebased. In addition, it's combining Create, Update, and Delete into a single PR. This needs to be three separate PRs.

Having Get and List are fine because Gophercloud's policy is 1 PR per "method" and both Get and List use `GET`. But POST, PUT/PATCH, and DELETE are three separate methods.

kayrus:
@jtopjian I'm not sure why do I need to split PRs, but ok.

jtopjian:
> I'm not sure why do I need to split PRs, but ok.

In general, a patch/PR should always be an atomic unit. `POST` can exist without `DELETE`. It can be tested, both with unit tests and acceptance tests, and not require any other methods. 

Additionally, it's part of the [rules](https://github.com/gophercloud/gophercloud/blob/master/docs/STYLEGUIDE.md) of this project, which I bend as far as I can go to make it easier for contributors.

If I merge PRs that blatantly disregard the project rules, I'll have my merge rights removed and then we're all screwed :)

kayrus:
@jtopjian I saw PRs, which introduced everything at once, like https://github.com/gophercloud/gophercloud/pull/1338. This added a confusion. Besides I didn't want to wait for each PR to be merged, then rebase and again and again.
But now I see that there were independent PRs and one combined PR like this.

jtopjian:
> Besides I didn't want to wait for each PR to be merged, then rebase and again and again.

Sometimes this is unavoidable, but I try to minimize this and fast-track as much as I can.

PR #1395, opened 2019-01-10
Manila: add share snapshots support (create)

kayrus:
Part of the #1380

### DIFF
--- a/openstack/sharedfilesystems/v2/snapshots/requests.go
+++ b/openstack/sharedfilesystems/v2/snapshots/requests.go
@@ -18,10 +18,6 @@ type CreateOptsBuilder interface {
 type CreateOpts struct {
 	// The UUID of the share from which to create a snapshot
 	ShareID string `json:"share_id" required:"true"`
-	// Indicates whether snapshot creation must be attempted when a share’s status is not
-	// available. Set to true to force snapshot creation when the share is busy performing
-	// other operations. Default is false.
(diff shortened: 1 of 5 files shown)


openstack/sharedfilesystems/v2/snapshots/requests.go
```diff
@@ -5,6 +5,56 @@ import (
...
+// passed to snapshots.Create(). For more information about these parameters,
+// please refer to the Snapshot object, or the shared file systems API v2
+// documentation
+type CreateOpts struct {
+	// The UUID of the share from which to create a snapshot
+	ShareID string `json:"share_id" required:"true"`
+	// Indicates whether snapshot creation must be attempted when a share’s status is not
+	// available. Set to true to force snapshot creation when the share is busy performing
+	// other operations. Default is false.
+	Force *bool `json:"force,omitempty"`
```
jtopjian: Where is this defined in the Manila API code? I see that it is defined in the API docs, but I am unable to see this defined in the service-side code you linked to.

Sometimes the API docs are out of date - maybe this is no longer a valid field?

kayrus: JFYI: https://bugs.launchpad.net/manila/+bug/1811336

jtopjian: Thank you for looking into this. I'll be able to do another review later on today.

Issue #1407, opened 2019-01-15
Multiple Reauthenticate requests when token expires


zimnx:
When ProviderClient is used concurrently and token expiration happens, all goroutines will try to reauthorize token. Since it may be time consuming operation, and reauth is synchronous accross all goroutines, it may block aplication.

In our scale environment we had over 4000 goroutines waiting on Reauthenticate mutex:
```
3200 @ 0xa84edc 0xa84fce 0xa96534 0xa9624d 0xacbbce 0xaccbcd 0xd1f7d1 0xd200d6 0xd23ab0 0xd2310c 0xec20c9 0x1105330 0x1104777 0x110829b 0x110c98d 0xab2112 0xb12bc6 0xb121a4 0x10f6e96 0x10f8e32 0x10f8d76 0x110d242 0xab207b 0xb12bc6 0xb121a4 0x10f6e96 0x10f8e32 0x10f8d76 0x10fd54b 0xab1ffb 0xb12bc6 0xb121a4
#	0xa9624c	sync.runtime_SemacquireMutex+0x3c										
#	0xacbbcd	sync.(*Mutex).Lock+0xed												
#	0xaccbcc	sync.(*RWMutex).Lock+0x2c											
#	0xd1f7d0	github.com/cloudwan/gohan/vendor/github.com/gophercloud/gophercloud.(*ProviderClient).Reauthenticate+0x60	
#	0xd200d5	github.com/cloudwan/gohan/vendor/github.com/gophercloud/gophercloud.(*ProviderClient).Request+0x775		
#	0xd23aaf	github.com/cloudwan/gohan/vendor/github.com/gophercloud/gophercloud.(*ServiceClient).Request+0x1cf		
#	0xd2310b	github.com/cloudwan/gohan/vendor/github.com/gophercloud/gophercloud.(*ServiceClient).Get+0x10b			
#	0xec20c8	github.com/cloudwan/gohan/vendor/github.com/gophercloud/gophercloud/openstack/identity/v3/tokens.Get+0x1a8	
#	0x110532f	github.com/cloudwan/gohan/cloud.(*keystoneV3Client).VerifyToken+0x5f						
#	0x1104776	github.com/cloudwan/gohan/cloud.(*KeystoneIdentity).VerifyToken+0x46						
#	0x110829a	github.com/cloudwan/gohan/server/middleware.(*CachedIdentityService).VerifyToken+0x8a				
[...]

PR #1408, opened 2019-01-15
Block requests requiring authorization until reauth process finishes

zimnx:
For #1407 

This approach blocks AuthenticatedHeaders if Reauthorize is running, and unblocks it when it finishes. 
This might be expected behavior since user expects to receive proper token.

Reauthorize notifies all blocked listeners when reauth request finishes. It also support multiple goroutines trying to reauthorize concurrently, which results in single request. In case of an error it's propagated to all Reauthorize callers. 

Without this PR, Auth Service might get DDoSed on token expiration. See #1407 for details.


provider_client.go
```diff
@@ -79,19 +79,20 @@ type ProviderClient struct {
...
+// authenticated service requests. Blocks if Reauthenticate is in progress.
 func (client *ProviderClient) AuthenticatedHeaders() (m map[string]string) {
 	if client.reauthmut != nil {
-		client.reauthmut.RLock()
-		if client.reauthmut.reauthing {
-			client.reauthmut.RUnlock()
-			return
+		client.reauthmut.Lock()
+		for client.reauthmut.reauthing {
+			client.reauthmut.done.Wait()
```
piotr-dulikowski: Automatic reauthentication will deadlock here. Reauthenticate() will set the reauthing flag, then (indirectly) call the Request() method to renew the token. Request() will call AuthenticatedHeaders() and will block here indefinitely.

zimnx: By indirect you mean that ReauthFunc is calling Request? Can you point me to place where it's called? I somehow missed that.

piotr-dulikowski: The call chain is as follows:
- ReauthFunc is assigned at openstack/client.go:224
- v3auth (called at openstack/client.go:225)
- Create (called at openstack/client.go:191)
- ServiceClient.Post (called at openstack/identity/v3/tokens/requests.go:120)
- ServiceClient.Request (called at service_client.go:85)
- ProviderClient.Request (called at service_client.go:149)

zimnx: Thanks, I'll revert change in AuthenticatedHeaders. That will prevent deadlock

zimnx: Resolved by your offline suggestion. Check recent commit.

jtopjian:
Reauthentication is my weakest spot because I don't use it for anything. I've asked @jrperritt to read it over when he has some time.

@majewsky might be another person who can give some input on this.

> Test took 11 hours, all requests finished within 15ms.

Thank you so much for the amount of attention you've given this issue.

majewsky:
> @majewsky might be another person who can give some input on this.

From what I can see in the comments and in the code, @zimnx has given the reauth topic much more thought than I ever have. I don't see anything objectionable in the code, but I think I'll abstain since I don't have too much time on my hands to dive into this more thoroughly.

jtopjian:
@majewsky No problem at all - thank you :)

I'll give @jrperritt a few more days to look at this. If he doesn't have the time (which is OK), I'll go ahead and merge this.

PR #1419, opened 2019-01-23
Auth: update application credential requirements logic

### LINKED ISSUE #1365: Auth: application credential auth doesn't require username
If you use openstack [application credential](https://docs.openstack.org/keystone/queens/user/application_credentials.html) auth method, username should be ignored. So far gophercloud requires it to be defined:

```
Missing one of the following environment variables [OS_USERNAME, OS_USERID]
```

However if you export `OS_USERNAME=any` along with `OS_APPLICATION_CREDENTIAL_*`, application will work fine.

[Ref1](https://github.com/gophercloud/gophercloud/blob/bdd8b1ecd793ffb03d43314c66565be3d6a7a582/openstack/auth_env.go#L41..L43)
[Ref2](https://github.com/gophercloud/gophercloud/blob/bdd8b1ecd793ffb03d43314c66565be3d6a7a582/openstack/auth_env.go#L62..L67)

kayrus:
Resolves #1365

Env requirements:

```
OS_APPLICATION_CREDENTIAL_SECRET:
  OS_APPLICATION_CREDENTIAL_ID: done
  OS_APPLICATION_CREDENTIAL_NAME:
    OS_USERID: done
    OS_USERNAME:
      OS_DOMAIN_NAME: done
      OS_DOMAIN_ID: done
```

### DIFF
--- a/auth_options.go
+++ b/auth_options.go
@@ -185,7 +185,6 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
 	// Populate the request structure based on the provided arguments. Create and return an error
 	// if insufficient or incompatible information is present.
 	var req request
-	var userRequest userReq
 
 	if opts.Password == "" {
 		if opts.TokenID != "" {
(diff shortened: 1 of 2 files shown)


auth_options.go
```diff
@@ -182,10 +182,45 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
 		Auth authReq `json:"auth"`
 	}
 
+	appCredReq := func(opts *AuthOptions) (*userReq, error) {
```
jtopjian: Let's put all of these conditionals inline with the rest of the logic below. Admittedly it's a long string of if/else statements and the whole thing could probably be refactored, but let's stick with consistency.

kayrus: Done

kayrus:
lbaas CRUD is being naughty again.

auth_options.go
```diff
@@ -229,25 +228,48 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
...
-			// make sure that only one of DomainName or DomainID were provided
-			if opts.DomainID == "" && opts.DomainName == "" {
-				return nil, ErrDomainIDOrDomainName{}
-			}
-			req.Auth.Identity.Methods = []string{"application_credential"}
-			if opts.DomainID != "" {
-				userRequest = userReq{
-					Name:   &opts.Username,
-					Domain: &domainReq{ID: &opts.DomainID},
+			userRequest, err := func(opts *AuthOptions) (*userReq, error) {
```
jtopjian: I apologize for being unclear - I meant to get rid of the function entirely. Unless I'm mistaken, the function isn't serving a purpose of being reusable, correct? You could add the contents of the function to be inline with the rest of the series of conditionals and it would be the same? Please let me know if I'm wrong about this.

kayrus: Function is used to simplify the logic. If I remove it, the logic will be full of if then else.

kayrus: Done

kayrus:
Again, TestLoadbalancersCRUD timeout.

Issue #1420, opened 2019-01-23
Bug: Messaging Claims Create Causes Crash 


JackKuei:
Got panic trace due to runtime error. Possibly triggered with authentication failure

Jan 17 07:08:34 xxxx[3961]: panic: runtime error: invalid memory address or nil pointer dereference
Jan 17 07:08:34 xxxx[3961]: [signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0xb56cb8]
Jan 17 07:08:34 xxxx[3961]: goroutine 99 [running]:
Jan 17 07:08:34 xxxx[3961]: xxxx/vendor/github.com/gophercloud/gophercloud/openstack/messaging/v2/claims.Create(0xc0001a3f90, 0xc000450e70, 0x25, 0xef0880, 0
Jan 17 07:08:34 xxxx[3961]: xxxx/vendor/github.com/gophercloud/gophercloud/openstack/messaging/v2/claims/requests.go:60 +0x1d8

jtopjian:
Fixed in #1421

PR #1421, opened 2019-01-23
Fix bug where messaging claims-create causes panic crash because HTTP…

JackKuei:
… POST encounters error and has no response object.

Reference:  #[1420 [Bug: Messaging Claims Create Causes Crash]](https://github.com/gophercloud/gophercloud/issues/1420)

Code: https://github.com/gophercloud/gophercloud/blob/e102249148e2e09eb5efab07b7c3ea74c161e935/openstack/messaging/v2/claims/requests.go#L55-L62


JackKuei:
@jtopjian This is ready to be reviewed.

Issue #1404, opened 2019-01-14
Octavia: tags support for load balancers


lingxiankong:
- [x] Include `tags` for getting load balancer
- [x] Include `tags` for listing load balancers
- [x] Specify `tags` for creating load balancer
- [x] Specify `tags` for updating load balancer
- [x] Query load balancer by tags

lingxiankong:
Waiting for this patch https://review.openstack.org/632013 in octavia.

jtopjian:
@lingxiankong Thank you for your work on this! I'm going to close this, but let me know if you'd like it re-opened for further work.

🎉

PR #1424, opened 2019-01-29
Octavia: filter load balancers by tags

lingxiankong:
For #1404

code in this PR:

- API doc: <https://developer.openstack.org/api-ref/load-balancer/v2/index.html#filtering-and-column-selection>
- Code: <https://github.com/openstack/octavia/blob/6f96401e9c9c0fbac879dc174f129cfcaafb5f09/octavia/api/common/pagination.py#L183>


lingxiankong:
@jtopjian Hi, this PR is ready for review

PR #1427, opened 2019-01-30
[WIP]: Fix counter concurrency in reauth tests

kayrus:
An attempt to fix:

```
$ ./script/coverage
--- FAIL: TestReauthEndLoop (0.01s)
	convenience.go:35: Failure in provider_client_test.go, line 204: expected 8 but got 6
FAIL
```

I might be wrong, but it looks like reauth tests from the #1408 were not adjusted in a proper way to support concurrency

P.S. expected and actual function parameters in [AssertEquals](https://github.com/gophercloud/gophercloud/blob/cc9c99918988100699c8ec6f0d7a73dfa5b71fda/testhelper/convenience.go#L215) are mixed up almost in every test...

### DIFF
--- a/testing/provider_client_test.go
+++ b/testing/provider_client_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jtopjian:
This was semi-fixed in #1435. The test isn't accurate and could probably be removed entirely - it was meant to test for a possible regression with a prior PR, so if it passes at all, it's fine. Thank you for looking into this, though.

Issue #1434, opened 2019-02-02
Core: Minor reauth concurrency issues


jtopjian:
#1408 introduced some minor concurrency issues that are occasionally showing up in the unit tests. Upon investigation, I spotted another few areas where concurrency can be improved.

PR #1435, opened 2019-02-02
Core: Reauth Concurrency fixes

jtopjian:
For #1434 

/cc @kayrus @zimnx @piotr-dulikowski

### DIFF
--- a/testing/provider_client_test.go
+++ b/testing/provider_client_test.go
@@ -61,14 +61,14 @@ func TestConcurrentReauth(t *testing.T) {
 	p.UseTokenLock()
 	p.SetToken(prereauthTok)
 	p.ReauthFunc = func() error {
-		p.IsThrowaway = true
+		p.SetThrowaway(true)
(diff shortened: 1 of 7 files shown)


testing/provider_client_test.go
```diff
@@ -201,8 +205,8 @@ func TestReauthEndLoop(t *testing.T) {
 	wg.Wait()
 	th.AssertEquals(t, info.reauthAttempts, 6)
 	th.AssertEquals(t, info.maxReauthReached, true)
-	th.AssertEquals(t, errAfter, 6)
-	th.AssertEquals(t, errUnable, 14)
+	th.AssertEquals(t, errAfter > 5, true)
+	th.AssertEquals(t, errUnable < 15, true)
```
jtopjian: I don't think it's possible to accurately get an exact number here. I looked at the other concurrency tests and there's always a small range of valid values, so I think checking for a range of values is good enough.

PR #1441, opened 2019-02-07
ContainerInfraV1: add ListDetail method

ozerovandrei:
Add ListDetail method with documentation and tests.
This method uses different URL than List method.

For #1442

code in this PR:

https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/cluster.py#L347


ozerovandrei:
recheck

ozerovandrei:
recheck

PR #1443, opened 2019-02-07
Identity V3: Add application credentials package

### LINKED ISSUE #1436: Identity: add an applicationcredentials users extension package
Openstack API ref: https://developer.openstack.org/api-ref/identity/v3/index.html?expanded=create-application-credential-detail#application-credentials

API methods to add:
* GET ([show](https://github.com/openstack/keystone/blob/fad6bbabcf062486570e46ca428ce08dfea82e78/keystone/api/users.py#L582..L591), [list](https://github.com/openstack/keystone/blob/fad6bbabcf062486570e46ca428ce08dfea82e78/keystone/api/users.py#L523..L534))
* POST ([create](https://github.com/openstack/keystone/blob/fad6bbabcf062486570e46ca428ce08dfea82e78/keystone/api/users.py#L536..L575))
* DELETE ([delete](https://github.com/openstack/keystone/blob/fad6bbabcf062486570e46ca428ce08dfea82e78/keystone/api/users.py#L593..L604))

Desired package path: `gophercloud/openstack/identity/v3/extensions/applicationcredentials`

struct draft:

```go
type ApplicationCredential struct {
        ID           string `json:"id"`
        Name         string `json:"name"`
        Description  string `json:"description"`
        Unrestricted bool   `json:"unrestricted"`
        // Secret can be computed and returned by an API only once during the POST request
        Secret    string       `json:"secret"`
        ProjectID string       `json:"project_id"`
        // only role ID or/and role name could be sent in the POST request
        Roles     []roles.Role `json:"roles"`
        ExpiresAt string       `json:"expires_at"`
}
```

kayrus:
Resolves #1436


kayrus:
I moved the package directly into the `gophercloud/openstack/identity/v3` path.

kayrus:
@jtopjian kindly look on this pr

kayrus:
Note: I have to verify whether it is possible to manage an application credential for any user being admin.

jtopjian:
@kayrus Thanks! At first glance, this looks good. I will have some time in the next day or two to look it over in more detail.

openstack/identity/v3/applicationcredentials/requests.go
```diff
@@ -0,0 +1,95 @@
...
+// CreateOptsBuilder allows extensions to add additional parameters to
+// the Create request.
+type CreateOptsBuilder interface {
+	ToApplicationCredentialCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts provides options used to create an application credential.
+type CreateOpts struct {
+	// The name of the application credential.
+	Name string `json:"name,omitempty"`
```
jtopjian: Looks like this is required? https://github.com/openstack/keystone/blob/ef87a202127f0ac84b857716e72595ce75058378/keystone/application_credential/schema.py#L48

Omitting the name in `CreateOpts` yields an error, too.

openstack/identity/v3/applicationcredentials/requests.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// CreateOptsBuilder allows extensions to add additional parameters to
+// the Create request.
+type CreateOptsBuilder interface {
+	ToApplicationCredentialCreateMap() (map[string]interface{}, error)
+}
+
+// CreateOpts provides options used to create an application credential.
+type CreateOpts struct {
```
jtopjian: It looks like this might support arbitrary extra fields: https://github.com/openstack/keystone/blob/ef87a202127f0ac84b857716e72595ce75058378/keystone/application_credential/schema.py#L49

This is common with the Keystone/Identity service. Basically it allows you to add _any_ additional field to the request and it will be stored in an `extra` db column.

See here for an example of how to implement this:

* https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/users/requests.go#L126
* https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/users/requests.go#L142

If this feature works for application credentials, a similar implementation can be done here. However, that's optional, especially for an initial commit.

openstack/identity/v3/applicationcredentials/results.go
```diff
@@ -0,0 +1,127 @@
...
+	"encoding/json"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+type Role struct {
+	// DomainID is the domain ID the role belongs to.
+	DomainID string `json:"domain_id,omitempty"`
```
jtopjian: It doesn't look like `domain_id` is specified in the schema: https://github.com/openstack/keystone/blob/ef87a202127f0ac84b857716e72595ce75058378/keystone/application_credential/schema.py#L22-L25

But I'm seeing it in the API output, so it's a legitimate field.

Nothing to do here, just a note for posterity.

openstack/identity/v3/applicationcredentials/requests.go
kayrus: Looks like it doesn't work, at least in rocky:

```json
2019/02/14 09:12:45 [DEBUG] OpenStack Request Body: {
  "application_credential": {
    "description": "Swift technical user",
    "expires_at": "2019-03-09T12:58:50Z",
    "foo": "bar",
    "name": "swift",
    "roles": [
      {
        "name": "swiftoperator"
      }
    ],
    "unrestricted": false
  }
}
2019/02/14 09:12:47 [DEBUG] OpenStack Response Code: 400
2019/02/14 09:12:47 [DEBUG] OpenStack Response Headers:
Content-Length: 131
Date: Thu, 14 Feb 2019 08:12:47 GMT
Server: nginx/1.15.3
Strict-Transport-Security: max-age=15724800; includeSubDomains
Vary: X-Auth-Token
X-Openstack-Request-Id: req-48f7bcdf-8f71-4003-af22-38667c601990
2019/02/14 09:12:47 [DEBUG] OpenStack Response Body: {
  "error": {
    "code": 400,
    "message": "u'foo' is an invalid keyword argument for ApplicationCredentialModel",
    "title": "Bad Request"
  }
}
```

openstack/identity/v3/applicationcredentials/requests.go
kayrus: I've never seen, that the `required:true` tag worked within go. It is more likely cosmetic change.
Done.

kayrus:
recheck

openstack/identity/v3/applicationcredentials/requests.go
jtopjian: It's a Gophercloud-specific tag: https://github.com/gophercloud/gophercloud/blob/master/params.go#L64

openstack/identity/v3/applicationcredentials/requests.go
jtopjian: Thank you for looking into it.

Issue #1412, opened 2019-01-16
Adding parallel to acceptance test


JackKuei:
Some acceptance tests can take a while to run, for the tests that can be run in parallel, should add t.Parallel() to the acceptance test. The test user can then decide whether or not to run test in parallel as well as number of parallel running threads.

jtopjian:
I've thought about this, too. I think the issue is that the tests which take a long time to run would run very poorly parallelized (possibly causing resource exhaustion in the test infrastructure). And the tests that run quickly wouldn't shorten the testing time by much if they run in parallel.

If your comment is more about the time it takes for OpenLab to run, a good 30 minutes is spent just setting up the environment. When you look at it that way, the tests themselves are running pretty quickly.

However, if you want to experiment with `t.Parallel`, by all means go for it :)

JackKuei:
It was kind of self serving, but I was hoping to PR the code with t.Parallel() in it, so don't have to perform any customization. So for OpenLab can run go test without parallel while on my local dev, I can run parallel tests.

jtopjian:
Sure, give it a try and we'll see how it goes.

PR #1445, opened 2019-02-08
Senlin: Test Parallel

JackKuei:

For #[1412 [Add Parallel To Acceptance Test]](https://github.com/gophercloud/gophercloud/issues/1412)

code in this PR:

N/A

### DIFF
--- a/acceptance/openstack/clustering/v1/actions_test.go
+++ b/acceptance/openstack/clustering/v1/actions_test.go
@@ -12,6 +12,7 @@ import (
 )
 
 func TestActionsList(t *testing.T) {
+	t.Parallel()
 	client, err := clients.NewClusteringV1Client()
 	th.AssertNoErr(t, err)
 
--- a/acceptance/openstack/clustering/v1/clusters_test.go
+++ b/acceptance/openstack/clustering/v1/clusters_test.go
@@ -15,6 +15,7 @@ import (
 )
 
 func TestClustersCRUD(t *testing.T) {
+	t.Parallel()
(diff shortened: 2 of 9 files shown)


ozerovandrei (dismissed):
LGTM

jtopjian:
I've been debating this one for a few days. I'd prefer to see how OpenLab handles this before it's merged. It's possible that so much parallel activity crushes the node running the tests.

JackKuei:
I'm not familiar with how OpenLab execute the tests. But running "go test" will not invoke running tests in parallel, only when using the parallel flag will it take effect such as "go test -parallel XX"

jtopjian:
It appears `t.Parallel()` and `-parallel XX` are two different ways to enable parallel execution and are independent of each other.

Once a test has `t.Parallel()`, all parallel tests will run in parallel without any additional test flags - I just confirmed this.

JackKuei:
I think this is a more complete explanation for parallel tests marked with t.Parallel(). ```Parallel tests are run in a number of goroutines. This number is defaulted to GOMAXPROCS, but can be set by passing the -parallel flag to go test.```

jtopjian:
Right, so the concern is that if/when Senlin is tested in OpenLab (which is on my list to do), would this PR cause the tests to crush the environment?

JackKuei:
Yeah, that's something you need to test on your side when you get a chance. Also a side note, those test cases might not execute in order when ran in parallel.

EmilienM:
This PR has been staled for a very long time, I'll close it. Feel free to re-open!

PR #1448, opened 2019-02-12
Enable go modules

radeksimko:
This is in preparation to switching [Terraform OpenStack provider](https://github.com/terraform-providers/terraform-provider-openstack) to go modules (see https://github.com/terraform-providers/terraform-provider-openstack/issues/656).

Enabling go modules throughout the dependency tree of the provider will make the switch and future upgrades easier.

I ran all tests via `go test ./...` in Go `1.11.5`, but I don't have access to any OpenStack environment to test this thoroughly and some tests naturally failed because of that.

This PR is a result of the following commands (in a clean Go `1.11.5` environment):

```sh
go mod init
go get ./...
go mod tidy
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,19 @@
 language: go
 sudo: false
 install:
-- go get golang.org/x/crypto/ssh
-- go get -v -tags 'fixtures acceptance' ./...
-- go get github.com/wadey/gocovmerge
-- go get github.com/mattn/goveralls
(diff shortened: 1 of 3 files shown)


jtopjian:
recheck

jtopjian:
Just some notes for posterity:

Enabling go modules doesn't change current guidelines of trying not to include external dependencies (in other words: ensuring this library is as dependent-less as possible) as well as makes any decisions about versioning and version scheduling.

PR #1456, opened 2019-02-19
Update capsule data model

hongbin:
We are going to update the capsule data model in server side [1].
This is the corresponding change in gophercloud.

[1] https://review.openstack.org/#/c/636070/

For #[PUT ISSUE NUMBER HERE]

code in this PR:

[PUT URLS HERE]

### DIFF
--- a/acceptance/openstack/container/v1/capsules_test.go
+++ b/acceptance/openstack/container/v1/capsules_test.go
@@ -73,6 +73,9 @@ func TestCapsule(t *testing.T) {
 
 		for _, m := range CapsuleList {
 			capsuleUUID := m.UUID
+			if capsuleUUID != capsule.UUID {
+				continue
+			}
 			capsule, err := capsules.Get(client, capsuleUUID).Extract()
 
 			th.AssertNoErr(t, err)
--- a/openstack/container/v1/capsules/results.go
+++ b/openstack/container/v1/capsules/results.go
@@ -57,7 +57,7 @@ type Capsule struct {
 	Memory string `json:"memory"`
 
 	// The name of the capsule
-	MetaName string `json:"meta_name"`
+	MetaName string `json:"name"`
 
 	// Indicates whether capsule is currently operational.
 	Status string `json:"status"`
@@ -75,23 +75,14 @@ type Capsule struct {
 	// other APIs that might want a capsule reference.
 	Links []interface{} `json:"links"`
 
-	// The capsule version
-	CapsuleVersion string `json:"capsule_version"`
-
 	// The capsule restart policy
-	RestartPolicy string `json:"restart_policy"`
+	RestartPolicy map[string]string `json:"restart_policy"`
 
 	// The capsule metadata labels
-	MetaLabels map[string]string `json:"meta_labels"`
-
-	// The list of containers uuids inside capsule.
-	ContainersUUIDs []string `json:"containers_uuids"`
+	MetaLabels map[string]string `json:"labels"`
 
 	// The capsule IP addresses
 	Addresses map[string][]Address `json:"addresses"`
-
-	// The capsule volume attached information
-	VolumesInfo map[string][]string `json:"volumes_info"`
 
 	// The container object inside capsule
 	Containers []Container `json:"containers"`
--- a/openstack/container/v1/capsules/testing/fixtures.go
+++ b/openstack/container/v1/capsules/testing/fixtures.go
@@ -183,8 +183,8 @@ const CapsuleGetBody_OldTime = `
   "project_id": "6b8ffef2a0ac42ee87887b9cc98bdf68",
   "cpu": 1,
   "memory": "1024M",
-  "meta_name": "test",
-  "meta_labels": {"web": "app"},
+  "name": "test",
+  "labels": {"web": "app"},
   "created_at": "2018-01-12 09:37:25+00:00",
   "updated_at": "2018-01-12 09:37:26+00:00",
   "links": [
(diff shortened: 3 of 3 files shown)


hongbin:
recheck

hongbin:
@jtopjian It looks the CI failed from Octavia devstack plugin. Is anyone looking into the issue?

jtopjian:
Yes, I've opened an issue about it: https://github.com/theopenlab/openlab/issues/192

hongbin:
@jtopjian ping for review

jtopjian:
@hongbin This looks like a pretty big breaking change.Some of the field names are having their JSON tags renamed, which helps. But will the removal of certain fields cause problems with anyone using this in any OpenStack version prior to the upcoming release?

hongbin:
@jtopjian this does look like a really big change but it is unavoidable because we are consolidating the data model: https://review.openstack.org/#/c/636070/ . It is pretty much merge two database table into one so the change is big. It is very hard to maintain backward-compatibility while rolling out the change.

As far as I know, the only one who are using this API via gophercloud is: https://github.com/virtual-kubelet/virtual-kubelet/pull/239 . The PR is still open and I will patch that PR once we finish the data model merging.

hongbin:
@kevin-zhaoshuai FYI

jtopjian:
> As far as I know, the only one who are using this API via gophercloud is: [virtual-kubelet/virtual-kubelet#239](https://github.com/virtual-kubelet/virtual-kubelet/pull/239) 

Right, but wouldn't this put a limitation on anyone who uses `virtual-kubelet` to have to be running Zun master in production?

Similarly, if anyone wanted to make Terraform resources for Zun, only those running master+ would be able to use them. And if we already had Terraform resources, they would now be broke.

But maybe I'm not fully understanding the changes. If someone running Zun Stein were to use Gophercloud to obtain the information of a capsule, would they get any unmarshalling errors?

jtopjian:
Sorry, not Stein - Rocky.

hongbin:
We bump an API version for this series of change. So, a potential improvement is to add a check for the API version and process the response differently. If you prefer this approach, I need your help to show me some examples of how other projects are doing this.

jtopjian:
Ahh, OK, so is this part of a microversion update or a standard major version update?

hongbin:
It is part of a microversion update

jtopjian:
OK, cool. We have some general guidelines on how to implement microversions here: https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md.

Unfortunately this change feels like it falls under "Modified Response Fields" which is the most difficult to implement in Go (or any typed language). Let me think on this one for a day.

hongbin:
@jtopjian would you consider branching the gophercloud to match the openstack branching. For example, we could have gophercloud stable/rocky branch that matches to the corresponding openstack release.

jtopjian:
> would you consider branching the gophercloud to match the openstack branching

That's a topic which has been discussed but it doesn't seem like it's practical to do. There are still lots of unimplemented API calls in Gophercloud, so the branches would be incomplete and put a lot of responsibility on backporting releveant API calls.

Since we already have separate packages for each major version, it helps Gophercloud support actions across multiple releases.

Microversions, while admirable for what they're trying to do, help with handling breaking changes within a major version, but they are not type friendly. They rely on being able to dynamically change an API response and have that API response be able to dynamically parse it.

The Microversions doc I linked to is the current set of best practices we've come up with for handling them.

hongbin:
@jtopjian ACK, I will find a time to go through the Microversion doc.

Another thing I can offer is to add those removed fields back with placeholder value. Then, gophercloud can parse the response without error. What do you think?

jtopjian:
> Another thing I can offer is to add those removed fields back with placeholder value. Then, gophercloud can parse the response without error. What do you think?

I think that would be a last resort. Thinking about this in a larger context, if we had to do the same thing for other services, things would get a little messy.

I have an idea that I'm going to work on - I'll post a link when it's ready.

jtopjian:
@hongbin I have a possible solution here: https://github.com/gophercloud/gophercloud/compare/master...jtopjian:zun-capsule-microversion

The idea is that the application will either be hard-coded or somehow detect that the new microversion should be used and then do the following:

```go
client, err := openstack.NewContainerV1(providerClient, endpointOptions)
client.Microversion = "X.Y"

allPages, err := capsules.List(client, nil)
allCapsules, err := capsules.ExtractCapsulesVXY(allPages)
```

VXY will, obviously, need replaced with whatever microversion is the minimum required for this new model.

Let me know what you think.

hongbin:
Thanks, I will try that approach later.

hongbin:
@jtopjian ,

You approach should work. What would make it perfect is going one step further:

* In ExtractCapsules, read the API version header from the response
* Perform legacy or new capsule extraction based on the API version above
* If above extract a legacy capsule object, manually convert it to a up-to-date capsule object
* Return the up-to-date capsule object

If above is done, in theory, users should be able to use the same code to consume different API version without being aware of the differences.

jtopjian:
@hongbin That kind of functionality would be great, but unfortunately not possible in Go. The reason is because the function can only return one type of Capsule - not multiple types. I spent a long time trying to figure out clever ways around this, but wasn't able to find anything.

If you're having problems falling asleep, this very long thread will help: https://github.com/gophercloud/gophercloud/issues/441 :)

So at the moment, there unfortunately needs to be some kind of either/or situation: Either `ExtractCapsules` is called or `ExtractCapsulesVXY`. 

Thank you for taking a look. If you'd like, I can finish off this work and have it merged. The only remaining piece is to replace "VXY" with the actual microversion - do you know what the version will be?

hongbin:
Sure. The version is 1.32. Thanks for figuring this out.

jtopjian:
Thanks! I'll make the appropriate changes (including acceptance tests) tomorrow. 🙂

hongbin:
No problem. Just let me know if anything I can help.

hongbin:
In flavor of https://github.com/gophercloud/gophercloud/pull/1469

Issue #1458, opened 2019-02-20
Can not update cluster node_count when the magnum API is running on python 3


tghartland:
If the magnum API is running on python 3, when making a PATCH request to update `node_count` with a string as the value the API does not accept the value.
```
HTTP/1.1 400 Bad Request
Date: Wed, 20 Feb 2019 12:34:50 GMT
Server: Werkzeug/0.14.1 Python/3.5.2
OpenStack-API-Minimum-Version: container-infra 1.1
OpenStack-API-Maximum-Version: container-infra 1.6
OpenStack-API-Version: container-infra 1.6
vary: OpenStack-API-Version
Content-Type: application/json
Content-Length: 239
x-openstack-request-id: req-045c01a7-32ca-4f0d-b596-334a01d717f6
Connection: close

{"errors": [{"status": 400, "code": "client", "title": "Invalid input for field/attribute node_count", "detail": "Invalid input for field/attribute node_count. Value: '2'. unorderable types: str() < int()", "links": [], "request_id": ""}]}
```
The same request but without quotes around the `node_count` value in the request works properly:
```
HTTP/1.1 202 Accepted
Date: Wed, 20 Feb 2019 12:35:37 GMT
Server: Werkzeug/0.14.1 Python/3.5.2
OpenStack-API-Minimum-Version: container-infra 1.1
OpenStack-API-Maximum-Version: container-infra 1.6
OpenStack-API-Version: container-infra 1.6
vary: OpenStack-API-Version
Content-Length: 48
Content-Type: application/json
x-openstack-request-id: req-92b55b82-21cf-4a58-9903-f22b63e1b44e
Connection: close

{"uuid": "39a26b2c-74a5-4515-9bc8-9f8e71eba7b5"}
```
When the API is running on python 2 either way is accepted, it is only python 3 that is stricter.

The issue is that the cluster UpdateOpts only makes it possible to send a string:

https://github.com/gophercloud/gophercloud/blob/dcc6e84aef1b6713accea7d7d7252ad2bc0e7034/openstack/containerinfra/v1/clusters/requests.go#L138-L142

Providing an alternative UpdateOpts using an int as the value type works without any other modifications:

```go
type UpdateOptsInt struct {
	Op    clusters.UpdateOp `json:"op" required:"true"`
	Path  string            `json:"path" required:"true"`
	Value int               `json:"value,omitempty"`
}

[...]

PR #1463, opened 2019-02-21
Container Infra v1: Use interface for update value

jtopjian:
This commit changes the Value field of UpdateOpts to be an interface{}
instead of a string. This is to send type-specific values to the API
service.

For #1458 

/cc @tghartland would you be able to test this and confirm it's working?

### DIFF
--- a/acceptance/openstack/containerinfra/v1/clusters_test.go
+++ b/acceptance/openstack/containerinfra/v1/clusters_test.go
@@ -63,7 +63,7 @@ func TestClustersCRUD(t *testing.T) {
 	allPagesDetail, err := clusters.ListDetail(client, nil).AllPages()
 	th.AssertNoErr(t, err)
 
-	allClustersDetail, err := clusters.ExtractClusters(allPages)
+	allClustersDetail, err := clusters.ExtractClusters(allPagesDetail)
 	th.AssertNoErr(t, err)
 
 	var foundDetail bool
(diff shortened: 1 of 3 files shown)


jtopjian:
recheck

tghartland:
I've tested this, passing a string to UpdateOpts still causes the value to be rejected by the API (as expected), but now passing an int works.

jtopjian:
@tghartland Thank you for testing. Just to confirm: this is expected behavior for you? If the value really should be a string, then this change is still valid, right?

tghartland:
https://github.com/openstack/magnum/blob/9323da78199c7ea5e54d6ec155c32ba6e23968de/api-ref/source/parameters.yaml#L424-L430

The node_count parameter is supposed to be an integer. That it accepts a string when the API is run with python2 is probably unintended.

jtopjian:
Right, but what I mean is that when updating other attributes besides `node_count`, this patch is still valid, correct?

tghartland:
Ah I see what you mean now. I'm not even sure what other parameter I would use to test that.
Looking at the update documentation [here](https://docs.openstack.org/magnum/latest/user/#update) it seems as though `node_count` is the only thing that can be updated.
Given that the I still got the old error exactly as it was before the change when giving the value in the new UpdateOpts as a string again, I think it should still be fine.

jtopjian:
Sounds good. Thank you for looking into this.

Issue #1459, opened 2019-02-20
reauthentication gets stuck in a mutex deadlock


majewsky:
Since merging #1226, reauthentication gets stuck in a mutex deadlock. To reproduce:

```go
package main

import (
	"fmt"

	"github.com/gophercloud/gophercloud"
	"github.com/gophercloud/gophercloud/openstack"
	"github.com/gophercloud/gophercloud/openstack/identity/v3/projects"
	"github.com/gophercloud/utils/openstack/clientconfig"
)

func failIf(err error) {
	if err != nil {
		panic(err.Error())
	}
}

func main() {
	authOpts, err := clientconfig.AuthOptions(nil)
	failIf(err)
	authOpts.AllowReauth = true
	provider, err := openstack.AuthenticatedClient(*authOpts)
	failIf(err)

	//simulate expired token by setting a bogus token value
	provider.SetToken("this is not a valid token")

	client, err := openstack.NewIdentityV3(provider, gophercloud.EndpointOpts{})
	failIf(err)

	//do a random request -> this gets a 401 response, does reauth, and gets stuck during reauth
	pages, err := projects.List(client, nil).AllPages()
	failIf(err)
	projects, err := projects.ExtractProjects(pages)
	failIf(err)
	fmt.Printf("There are %d projects.\n", len(projects))
}
```

What happens, as far as I can see:

1. ProviderClient.Reauthenticate locks `client.mut`.
2. ProviderClient.Reauthenticate calls its ReauthFunc.
3. Upon receiving the authentication result, the ReauthFunc calls ProviderClient.SetTokenAndAuthResult.
4. ProviderClient.SetTokenAndAuthResult attempts to lock `client.mut` again, thus getting stuck in a deadlock.

I'm not sure though why this problem only appeared after #1226. It didn't substantially change the way that `client.mut` was used. However, I suspect that there is some interaction with #1408.

@zimnx @jrperritt Any ideas what's going on here? In particular: What are the semantics of `client.reauthmut` and `client.mut`, respectively? What are these mutexes supposed to guard exactly? Do we have some documentation on this somewhere?

@jtopjian Since this is a pretty severe issue (and it hit me very hard in head in production), I'm totally fine with you reverting #1226 if this cannot be resolved quickly.

majewsky:
[...]

PR #1464, opened 2019-02-21
Core: Fix deadlock with reauthentication

jtopjian:
This commit fixes a deadlock issue during reauthentication. It removes a
possibly unnecessary locking of the client's locking mechanism so the
reauthentication process can safely lock the client.

For #1459 

/cc @majewsky This resolves the problem for me. I think this fix also makes sense. I could be wrong :)

### DIFF
--- a/acceptance/openstack/identity/v3/reauth_test.go
+++ b/acceptance/openstack/identity/v3/reauth_test.go
@@ -1,0 +1,34 @@
+// +build acceptance
+
+package v3
+
+import (
+	"testing"
+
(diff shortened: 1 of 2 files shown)


kiwik:
recheck

jtopjian:
> If I understand it correctly, `client.mut` protects write access to auth attributes of `client` like `client.TokenID` and `client.authResult`, whereas `client.reauthmut` protects access to `client.ReauthFunc`, right?

That's my understanding as well and a pretty succinct way of describing it. I'll add some notes to the code.

Issue #1306, opened 2018-10-22
Computer V2: why don't we support "Create Multiple Servers" 


npu21:
As we can see, the struct  CreateOpts only support create one server one time,but there is a second kind of create call which can build multiple servers at once. This supports all the same parameters as create with a few additional attributes specific to multiple create. refer here please https://developer.openstack.org/api-ref/compute/?expanded=create-multiple-servers-detail
In some scenario, we need this ability and it's supposed to be more efficient.
So, do we have a plan to support this or maybe I miss some backgroud info here?

jtopjian:
I could have swore we already had this as an [extension](https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions) but it looks like we don't.

Looks like no one hasn't contributed it.

npu21:
@jtopjian ,yeah, just go through the  [https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions](url),but seems don't have this feature;
I could contribute a PR for this and hopes the github is ok then 🤔

jtopjian:
Implemented in #1471

PR #1471, opened 2019-02-23
Compute V2: Create multiple servers

code0x9:
For #1306 

API:
https://developer.openstack.org/api-ref/compute/?expanded=create-multiple-servers-detail#create-server

Schema:
https://github.com/openstack/nova/blob/5852bae034533129608dd24b58e33733e093e957/nova/api/openstack/compute/schemas/servers.py#L195

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -192,9 +192,6 @@ type CreateOpts struct {
 	// Max specifies Maximum number of servers to launch.
 	Max int `json:"max_count,omitempty"`
 
-	// ReturnReservationID sets respond with reservation_id only.
-	ReturnReservationID bool `json:"return_reservation_id,omitempty"`
-
 	// ServiceClient will allow calls to be made to retrieve an image or
 	// flavor ID by name.
 	ServiceClient *gophercloud.ServiceClient `json:"-"`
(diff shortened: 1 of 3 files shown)


openstack/compute/v2/servers/requests.go
```diff
@@ -183,9 +183,18 @@ type CreateOpts struct {
...
 	AccessIPv6 string `json:"accessIPv6,omitempty"`
 
+	// Min specifies Minimum number of servers to launch.
+	Min int `json:"min_count,omitempty"`
+
+	// Max specifies Maximum number of servers to launch.
+	Max int `json:"max_count,omitempty"`
+
+	// ReturnReservationID sets respond with reservation_id only.
+	ReturnReservationID bool `json:"return_reservation_id,omitempty"`
```
jtopjian: If I understand this correctly, setting this to `true` will cause the response body to be a different format? If so, then this will cause a JSON error when running `Create` and `Extract`. To get around this, we will need to create a new `Extract` method to handle this. This seems to be out of scope for the original issue this PR is fixing, so let's remove this.

code0x9: right. just removed that code. let's do it when it is needed.

jtopjian:
@code0x9 This looks good to me. Let me know if you're done working on this and I'll merge it. 

The OpenLab test failure can be ignored.

code0x9:
@jtopjian I'm done. please merge this.

PR #1480, opened 2019-02-25
support context.Context for outgoing HTTP requests

### LINKED ISSUE #1476: Feature request: support context.Context to cancel/time-out requests
IMHO it would be neat if one could pass an existing `context.Context` that would be propagated to the outgoing HTTP requests. Concrete example: Prometheus uses gophercloud for its [OpenStack service discovery](https://github.com/prometheus/prometheus/tree/master/discovery/openstack) (aka SD) which lists OpenStack instances at regular intervals. The SD goroutine gets a context that is canceled when it needs to exit (because the program stops or the configuration gets reloaded). Right now it means that when a request to the Nova API is in progress and the context is canceled, we need to wait for the request completion before effectively returning.

Implementation-wise, I could see it as an additional `Context` field to the [`gophercloud.ProviderClient`](https://godoc.org/github.com/gophercloud/gophercloud#ProviderClient) struct. If not nil, the context would be injected in the constructed [`http.Request`](https://github.com/gophercloud/gophercloud/blob/307b801773c3c1018ca059dd511b9a08714e97d6/provider_client.go#L313).

If this is something that is found helpful, I'm willing to start a PR.

simonpasquier:
Closes #1476

cc @jtopjian

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -2,6 +2,7 @@ package gophercloud
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"io"
@@ -75,6 +76,9 @@ type ProviderClient struct {
 	// Throwaway determines whether if this client is a throw-away client. It's a copy of user's provider client
 	// with the token and reauth func zeroed. Such client can be used to perform reauthorization.
 	Throwaway bool
+
+	// Context is the context passed to the HTTP request.
+	Context context.Context
 
 	// mut is a mutex for the client. It protects read and write access to client attributes such as getting
 	// and setting the TokenID.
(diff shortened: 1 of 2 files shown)


jtopjian:
@simonpasquier In general, this looks good to me. Before I sign-off on it, I wanted to make sure the same thing couldn't be done in a similar way as what we have outlined in the [FAQ](https://github.com/gophercloud/gophercloud/blob/master/docs/FAQ.md#implementing-default-logging-and-re-authentication-attempts)?

simonpasquier:
@jtopjian I thought about using a custom `RoundTripper` injecting a crafted context but it wouldn't apply to the TCP connection + TLS negotiation phases.

jtopjian:
@simonpasquier Good point. Thank you for implementing this!

simonpasquier:
Thanks for the quick feedback! It was a pleasure contributing :smiley:

PR #1481, opened 2019-02-25
Networking V2: Add DNS extension support

### LINKED ISSUE #776: Network V2: Add DNS Attributes to Network, Port and FloatingIP resources
The neutron API allows setting DNS attributes on some resources, which allows for auto provisioning of DNS Records, and better internal DNS resolution.

https://docs.openstack.org/mitaka/networking-guide/config-dns-int.html

kayrus:
Resolves #776

### DIFF
--- a/openstack/networking/v2/extensions/dns/doc.go
+++ b/openstack/networking/v2/extensions/dns/doc.go
@@ -1,0 +1,3 @@
+// Package portsbinding provides information and interaction with the port
+// binding extension for the OpenStack Networking service.
+package dns
--- a/openstack/networking/v2/extensions/dns/doc.go
+++ b/openstack/networking/v2/extensions/dns/doc.go
@@ -1,3 +1,0 @@
-// Package portsbinding provides information and interaction with the port
-// binding extension for the OpenStack Networking service.
-package dns
--- a/openstack/networking/v2/extensions/dns/requests.go
+++ b/openstack/networking/v2/extensions/dns/requests.go
@@ -59,7 +59,7 @@ func (opts PortCreateOptsExt) ToPortCreateMap() (map[string]interface{}, error) 
 	return base, nil
 }
 
-// PortUpdateOptsExt adds port binding options to the base ports.UpdateOpts
+// PortUpdateOptsExt adds DNS options to the base ports.UpdateOpts
 type PortUpdateOptsExt struct {
 	// UpdateOptsBuilder is the interface options structs have to satisfy in order
 	// to be used in the main Update operation in this package.
@@ -85,7 +85,7 @@ func (opts PortUpdateOptsExt) ToPortUpdateMap() (map[string]interface{}, error) 
 	return base, nil
 }
 
-// CreateOptsExt adds floating IP DNS options to the base floatingips.CreateOpts.
+// FloatingIPCreateOptsExt adds floating IP DNS options to the base floatingips.CreateOpts.
 type FloatingIPCreateOptsExt struct {
 	// CreateOptsBuilder is the interface options structs have to satisfy in order
 	// to be used in the main Create operation in this package.
@@ -118,13 +118,13 @@ func (opts FloatingIPCreateOptsExt) ToFloatingIPCreateMap() (map[string]interfac
 	return base, nil
 }
 
-// CreateOptsExt adds network DNS options to the base networks.CreateOpts.
+// NetworkCreateOptsExt adds network DNS options to the base networks.CreateOpts.
 type NetworkCreateOptsExt struct {
 	// CreateOptsBuilder is the interface options structs have to satisfy in order
 	// to be used in the main Create operation in this package.
 	networks.CreateOptsBuilder
 
-	// Set DNS domain to filter the networks
+	// Set DNS domain to the network
 	DNSDomain string `json:"dns_domain,omitempty"`
 }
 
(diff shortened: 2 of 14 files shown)


kayrus:
@jtopjian in terms of the package structure. Is it ok to have ports/floatingips/networks extensions in one package? Or do I need to split them into:

* portsdns
* floatingipsdns
* networksdns

?

jtopjian:
It can all be in the same package. Check out the `portsecurity` extension for an example.

kayrus:
Ok then. I already have a working PoC in Terraform. I just need to find some time to write tests here.

kayrus:
@jtopjian ready for review

jtopjian:
@kayrus Overall this looks good to me.

I've had issues with all of the acceptance tests, but it might be either due to the version I'm testing on (Queens), adding DNS integration _after_ initial setup, or something else.

* For the `TestDNSPortCRUDL` test, listing by `dns_name` returns all ports no matter the value. I've confirmed the query string looks correct, though.
* For the `TestDNSFloatingIPCRDL` test, the `dns_name` and `dns_domain` values are not being set on the floating IP at all.
* For the `TestDNSNetwork` test, listing by `dns_domain` returns a 501 Not Implemented error.

If you've confirmed all tests pass, that's good enough for me.

Out of curiosity, how was your environment configured to have all of these pass?

kayrus:
* Ports filtering by `dns_name` works fine in mitaka
* floating IPs are not working, since the external DNS service should be configured.
* Listing networks by `dns_domain` doesn't work in my env as well. Could you please confirm this issue? https://bugs.launchpad.net/neutron/+bug/1818318

> Out of curiosity, how was your environment configured to have all of these pass?

Only ports test does pass in my env :) Fortunately, ports is the only DNS feature I use.

jtopjian:
> Only ports test does pass in my env :) Fortunately, ports is the only DNS feature I use.

I can understand the difficulty of setting up an environment for external dns + floating IPs. I'm OK with including that code since I believe there is a strong indication that it will work.

But in general, I'm not comfortable merging code and tests that have not been confirmed to pass. Either through a personal devstack/dev environment or through OpenLab, the code needs to pass a test.

If listing networks by dns_domain doesn't work, then we need to remove it from here. Maybe it'll be supported in the future, but it's not appropriate to have code in anticipation of something working in the future. 

Listing ports by `dns_name` seems to either not be supported or it was broke after Mitaka. We'll then need to skip that test using [`SkipRelease`](https://github.com/gophercloud/gophercloud/blob/master/acceptance/clients/conditions.go#L71) for Newton to master.

kayrus:
@jtopjian you mean remove tests or `ListOpts`? Is it possible just to comment the code out?

jtopjian:
Let's just remove it. It's possible it'll be commented out indefinitely and there's no use having cruft like that around.

jtopjian:
Sorry, with regard to "tests or `ListOpts`", remove both. If it's not working, it shouldn't be included.

openstack/networking/v2/extensions/dns/requests.go
```diff
@@ -0,0 +1,171 @@
...
+	port := base["port"].(map[string]interface{})
+
+	if opts.DNSName != "" {
+		port["dns_name"] = opts.DNSName
+	}
+
+	return base, nil
+}
+
+// PortUpdateOptsExt adds port binding options to the base ports.UpdateOpts
```
jtopjian: `s/port binding/DNS/`

openstack/networking/v2/extensions/dns/requests.go
```diff
@@ -0,0 +1,171 @@
...
+	port := base["port"].(map[string]interface{})
+
+	if opts.DNSName != nil {
+		port["dns_name"] = *opts.DNSName
+	}
+
+	return base, nil
+}
+
+// CreateOptsExt adds floating IP DNS options to the base floatingips.CreateOpts.
```
jtopjian: FloatingIPCreateOptsExt

openstack/networking/v2/extensions/dns/requests.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	if opts.DNSDomain != "" {
+		floatingip["dns_domain"] = opts.DNSDomain
+	}
+
+	return base, nil
+}
+
+// CreateOptsExt adds network DNS options to the base networks.CreateOpts.
```
jtopjian: NetworkCreateOptsExt

openstack/networking/v2/extensions/dns/requests.go
```diff
@@ -0,0 +1,171 @@
...
+type NetworkCreateOptsExt struct {
+	// CreateOptsBuilder is the interface options structs have to satisfy in order
+	// to be used in the main Create operation in this package.
+	networks.CreateOptsBuilder
+
+	// Set DNS domain to filter the networks
```
jtopjian: Set DNS domain to the network

openstack/networking/v2/extensions/dns/results.go
```diff
@@ -0,0 +1,24 @@
+package dns
```
jtopjian: The comments in this file need updated.

openstack/networking/v2/extensions/dns/doc.go
```diff
@@ -0,0 +1,3 @@
+// Package portsbinding provides information and interaction with the port
+// binding extension for the OpenStack Networking service.
+package dns
```
jtopjian: The comments in this file need updated.

kayrus:
@jtopjian comments fixed

Issue #1485, opened 2019-02-27
Add support for Ironic baremetal Inspector API


elfosardo:
There is currently no support for Ironic Baremetal Inspector API in gophercloud.
I started working on it and will have a PR open shortly.

Reference https://developer.openstack.org/api-ref/baremetal-introspection/

Issue #1465, opened 2019-02-22
Changes in behavior of go fmt in 1.11


elfosardo:
Submitting a PR (#1461) we encountered an awkward issue when running the format script.
When running it with go1.11 everything looks good, on the other hand, running it with go1.10 will fail the test.

This is due to a documented change of behavior in go fmt between 1.10 and 1.11.
https://github.com/golang/go/issues/26228

Ideally the project should use only one version (the latest) of go to run the format script.

jtopjian:
@elfosardo rewriting the problematic structs to use newlines (which is the more common style anyway) should resolve this problem:

```diff
diff --git a/openstack/baremetal/v1/drivers/testing/fixtures.go b/openstack/baremetal/v1/drivers/testing/fixtures.go
index be080a9..01c55aa 100644
--- a/openstack/baremetal/v1/drivers/testing/fixtures.go
+++ b/openstack/baremetal/v1/drivers/testing/fixtures.go
@@ -218,19 +218,54 @@ const SingleDriverDiskProperties = `

 var (
        DriverAgentIpmitool = drivers.Driver{
-               Name:       "agent_ipmitool",
-               Type:       "classic",
-               Hosts:      []string{"897ab1dad809"},
-               Links:      []interface{}{map[string]interface{}{"href": "http://127.0.0.1:6385/v1/drivers/agent_ipmitool", "rel": "self"}, map[string]interface{}{"href": "http://127.0.0.1:6385/drivers/agent_ipmitool", "rel": "bookmark"}},
-               Properties: []interface{}{map[string]interface{}{"href": "http://127.0.0.1:6385/v1/drivers/agent_ipmitool/properties", "rel": "self"}, map[string]interface{}{"href": "http://127.0.0.1:6385/drivers/agent_ipmitool/properties", "rel": "bookmark"}},
+               Name:  "agent_ipmitool",
+               Type:  "classic",
+               Hosts: []string{"897ab1dad809"},
+               Links: []interface{}{
+                       map[string]interface{}{
+                               "href": "http://127.0.0.1:6385/v1/drivers/agent_ipmitool",
+                               "rel":  "self",
+                       },
+                       map[string]interface{}{
+                               "href": "http://127.0.0.1:6385/drivers/agent_ipmitool",
+                               "rel":  "bookmark",
+                       },
+               },
+               Properties: []interface{}{
+                       map[string]interface{}{
[...]

PR #1486, opened 2019-02-27
Baremetal Inspector Client Node Introspection

elfosardo:
For #1485

This patch adds the Node Introspection API calls according to [1]

Start Introspection (POST)
Get Introspection status (GET)
List all Introspection statuses (GET)
Abort Introspection (POST)
Get Introspection Data (GET)
Reapply Introspection on stored data (POST)

Source code reference:
https://github.com/openstack/ironic-inspector/tree/master/ironic_inspector

[1] https://developer.openstack.org/api-ref/baremetal-introspection/#node-introspection

### DIFF
--- a/acceptance/clients/clients.go
+++ b/acceptance/clients/clients.go
@@ -285,6 +285,27 @@ func NewBareMetalV1NoAuthClient() (*gophercloud.ServiceClient, error) {
 	})
 }
 
+// NewBareMetalIntrospectionV1Client returns a *ServiceClient for making calls
+// to the OpenStack Bare Metal Introspection v1 API. An error will be returned
+// if authentication or client creation was not possible.
+func NewBareMetalIntrospectionV1Client() (*gophercloud.ServiceClient, error) {
+	ao, err := openstack.AuthOptionsFromEnv()
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := openstack.AuthenticatedClient(ao)
+	if err != nil {
+		return nil, err
+	}
+
+	client = configureDebug(client)
+
+	return openstack.NewBareMetalIntrospectionV1(client, gophercloud.EndpointOpts{
+		Region: os.Getenv("OS_REGION_NAME"),
+	})
+}
+
 // NewDBV1Client returns a *ServiceClient for making calls
 // to the OpenStack Database v1 API. An error will be returned
 // if authentication or client creation was not possible.
--- a/openstack/baremetalintrospection/v1/introspection/requests.go
+++ b/openstack/baremetalintrospection/v1/introspection/requests.go
@@ -1,0 +1,116 @@
+package introspection
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// ListIntrospectionsOptsBuilder allows extensions to add additional parameters to the
+// ListIntrospections request.
+type ListIntrospectionsOptsBuilder interface {
+	ToIntrospectionsListQuery() (string, error)
+}
+
+// ListIntrospectionsOpts allows the filtering and sorting of paginated collections through
+// the Introspection API. Filtering is achieved by passing in struct field values that map to
+// the node attributes you want to see returned. Marker and Limit are used
+// for pagination.
+type ListIntrospectionsOpts struct {
+	// Requests a page size of items.
+	Limit int `q:"limit"`
+
+	// The ID of the last-seen item.
+	Marker string `q:"marker"`
+}
+
+// ToIntrospectionsListQuery formats a ListIntrospectionsOpts into a query string.
+func (opts ListIntrospectionsOpts) ToIntrospectionsListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
+// ListIntrospections makes a request against the Inspector API to list the current introspections.
+func ListIntrospections(client *gophercloud.ServiceClient, opts ListIntrospectionsOptsBuilder) pagination.Pager {
+	url := listIntrospectionsURL(client)
+	if opts != nil {
+		query, err := opts.ToIntrospectionsListQuery()
+		if err != nil {
+			return pagination.Pager{Err: err}
+		}
+		url += query
+	}
+	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
+		var rpage = IntrospectionPage{pagination.LinkedPageBase{PageResult: r}}
+		return rpage
+	})
+}
+
+// GetIntrospectionStatus makes a request against the Inspector API to get the
+// status of a single introspection.
+func GetIntrospectionStatus(client *gophercloud.ServiceClient, nodeID string) (r GetIntrospectionStatusResult) {
+	_, r.Err = client.Get(introspectionURL(client, nodeID), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	return
+}
+
+// StartOptsBuilder allows extensions to add additional parameters to the
+// Start request.
+type StartOptsBuilder interface {
+	ToStartIntrospectionQuery() (string, error)
+}
+
+// StartOpts represents options to start an introspection.
+type StartOpts struct {
+	// Whether the current installation of ironic-inspector can manage PXE booting of nodes.
+	ManageBoot *bool `q:"manage_boot"`
+}
+
+// ToStartIntrospectionQuery converts a StartOpts into a request.
+func (opts StartOpts) ToStartIntrospectionQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
+// StartIntrospection initiate hardware introspection for node NodeID .
+// All power management configuration for this node needs to be done prior to calling the endpoint.
+func StartIntrospection(client *gophercloud.ServiceClient, nodeID string, opts StartOptsBuilder) (r StartResult) {
+	_, err := opts.ToStartIntrospectionQuery()
+	if err != nil {
+		r.Err = err
+		return
+	}
+
+	_, r.Err = client.Post(introspectionURL(client, nodeID), nil, nil, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+	})
+
+	return
+}
+
+// AbortIntrospection abort running introspection.
+func AbortIntrospection(client *gophercloud.ServiceClient, nodeID string) (r AbortResult) {
+	_, r.Err = client.Post(abortIntrospectionURL(client, nodeID), nil, nil, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+	})
+
+	return
+}
+
+// GetIntrospectionData return stored data from successful introspection.
+func GetIntrospectionData(client *gophercloud.ServiceClient, nodeID string) (r DataResult) {
+	_, r.Err = client.Get(introspectionDataURL(client, nodeID), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	return
+}
+
+// ReApplyIntrospection triggers introspection on stored unprocessed data.
+// No data is allowed to be sent along with the request.
+func ReApplyIntrospection(client *gophercloud.ServiceClient, nodeID string) (r ApplyDataResult) {
+	_, r.Err = client.Post(introspectionUnprocessedDataURL(client, nodeID), nil, nil, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+	})
+
+	return
+}
--- a/openstack/baremetalintrospection/v1/introspection/results.go
+++ b/openstack/baremetalintrospection/v1/introspection/results.go
@@ -1,0 +1,260 @@
+package introspection
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+type introspectionResult struct {
+	gophercloud.Result
+}
+
+// Extract interprets any introspectionResult as an Introspection, if possible.
+func (r introspectionResult) Extract() (*Introspection, error) {
+	var s Introspection
+	err := r.ExtractInto(&s)
+	return &s, err
+}
+
+// ExtractInto will extract a response body into an Introspection struct.
+func (r introspectionResult) ExtractInto(v interface{}) error {
+	return r.Result.ExtractIntoStructPtr(v, "")
+}
+
+// ExtractIntrospectionsInto will extract a collection of introspectResult pages into a
+// slice of Introspection entities.
+func ExtractIntrospectionsInto(r pagination.Page, v interface{}) error {
+	return r.(IntrospectionPage).Result.ExtractIntoSlicePtr(v, "introspection")
+}
+
+// ExtractIntrospections interprets the results of a single page from a
+// ListIntrospections() call, producing a slice of Introspection entities.
+func ExtractIntrospections(r pagination.Page) ([]Introspection, error) {
+	var s []Introspection
+	err := ExtractIntrospectionsInto(r, &s)
+	return s, err
+}
+
+// IntrospectionPage abstracts the raw results of making a ListIntrospections()
+// request against the Inspector API. As OpenStack extensions may freely alter
+// the response bodies of structures returned to the client, you may only safely
+// access the data provided through the ExtractIntrospections call.
+type IntrospectionPage struct {
+	pagination.LinkedPageBase
+}
+
+// Introspection represents an introspection in the OpenStack Bare Metal Introspector API.
+type Introspection struct {
+	// Whether introspection is finished
+	Finished bool `json:"finished"`
+
+	// State of the introspection
+	State string `json:"state"`
+
+	// Error message, can be null; "Canceled by operator" in case introspection was aborted
+	Error string `json:"error"`
+
+	// UUID of the introspection
+	UUID string `json:"uuid"`
+
+	// UTC ISO8601 timestamp
+	StartedAt time.Time `json:"-"`
+
+	// UTC ISO8601 timestamp or null
+	FinishedAt time.Time `json:"-"`
+
+	// Link to the introspection URL
+	Links []interface{} `json:"links"`
+}
+
+// IsEmpty returns true if a page contains no Introspection results.
+func (r IntrospectionPage) IsEmpty() (bool, error) {
+	s, err := ExtractIntrospections(r)
+	return len(s) == 0, err
+}
+
+// NextPageURL uses the response's embedded link reference to navigate to the
+// next page of results.
+func (r IntrospectionPage) NextPageURL() (string, error) {
+	var s struct {
+		Links []gophercloud.Link `json:"introspection_links"`
+	}
+	err := r.ExtractInto(&s)
+	if err != nil {
+		return "", err
+	}
+	return gophercloud.ExtractNextURL(s.Links)
+}
+
+// UnmarshalJSON trie to convert values for started_at and finished_at from the
+// json response into RFC3339 standard. Since Introspection API can remove the
+// Z from the format, if the conversion fails, it falls back to an RFC3339
+// with no Z format supported by gophercloud.
+func (r *Introspection) UnmarshalJSON(b []byte) error {
+	type tmp Introspection
+	var s struct {
+		tmp
+		StartedAt  string `json:"started_at"`
+		FinishedAt string `json:"finished_at"`
+	}
+
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+
+	*r = Introspection(s.tmp)
+
+	if s.StartedAt != "" {
+		t, err := time.Parse(time.RFC3339, s.StartedAt)
+		if err != nil {
+			t, err = time.Parse(gophercloud.RFC3339NoZ, s.StartedAt)
+			if err != nil {
+				return err
+			}
+		}
+		r.StartedAt = t
+	}
+
+	if s.FinishedAt != "" {
+		t, err := time.Parse(time.RFC3339, s.FinishedAt)
+		if err != nil {
+			t, err = time.Parse(gophercloud.RFC3339NoZ, s.FinishedAt)
+			if err != nil {
+				return err
+			}
+		}
+		r.FinishedAt = t
+	}
+
+	return nil
+}
+
+// GetIntrospectionStatusResult is the response from a GetIntrospectionStatus operation.
+// Call its Extract method to interpret it as an Introspection.
+type GetIntrospectionStatusResult struct {
+	introspectionResult
+}
+
+// StartResult is the response from a StartIntrospection operation.
+// Call its ExtractErr method to determine if the call succeeded or failed.
+type StartResult struct {
+	gophercloud.ErrResult
+}
+
+// AbortResult is the response from an AbortIntrospection operation.
+// Call its ExtractErr method to determine if the call succeeded or failed.
+type AbortResult struct {
+	gophercloud.ErrResult
+}
+
+// Data represents the full introspection data collected.
+// The format and contents of the stored data depends on the ramdisk used
+// and plugins enabled both in the ramdisk and in inspector itself.
+// This structure has been provided for basic compatibility but it
+// will need extensions
+type Data struct {
+	AllInterfaces map[string]BaseInterfaceType `json:"all_interfaces"`
+	BootInterface string                       `json:"boot_interface"`
+	CPUArch       string                       `json:"cpu_arch"`
+	CPUs          int                          `json:"cpus"`
+	Error         string                       `json:"error"`
+	Interfaces    map[string]BaseInterfaceType `json:"interfaces"`
+	Inventory     InventoryType                `json:"inventory"`
+	IPMIAddress   string                       `json:"ipmi_address"`
+	LocalGB       int                          `json:"local_gb"`
+	MACs          []string                     `json:"macs"`
+	MemoryMB      int                          `json:"memory_mb"`
+	RootDisk      RootDiskType                 `json:"root_disk"`
+}
+
+// Sub Types defined under Data and deeper in the structure
+
+type BaseInterfaceType struct {
+	ClientID string `json:"client_id"`
+	IP       string `json:"ip"`
+	MAC      string `json:"mac"`
+	PXE      bool   `json:"pxe"`
+}
+
+type BootInfoType struct {
+	CurrentBootMode string `json:"current_boot_mode"`
+	PXEInterface    string `json:"pxe_interface"`
+}
+
+type CPUType struct {
+	Architecture string   `json:"architecture"`
+	Count        int      `json:"count"`
+	Flags        []string `json:"flags"`
+	Frequency    string   `json:"frequency"`
+	ModelName    string   `json:"model_name"`
+}
+
+type InterfaceType struct {
+	BIOSDevName string                 `json:"biosdevname"`
+	ClientID    string                 `json:"client_id"`
+	HasCarrier  bool                   `json:"has_carrier"`
+	IPV4Address string                 `json:"ipv4_address"`
+	IPV6Address string                 `json:"ipv6_address"`
+	Lldp        map[string]interface{} `json:"lldp"`
+	MACAddress  string                 `json:"mac_address"`
+	Name        string                 `json:"name"`
+	Product     string                 `json:"product"`
+	Vendor      string                 `json:"vendor"`
+}
+
+type InventoryType struct {
+	BmcAddress   string           `json:"bmc_address"`
+	Boot         BootInfoType     `json:"boot"`
+	CPU          CPUType          `json:"cpu"`
+	Disks        []RootDiskType   `json:"disks"`
+	Interfaces   []InterfaceType  `json:"interfaces"`
+	Memory       MemoryType       `json:"memory"`
+	SystemVendor SystemVendorType `json:"system_vendor"`
+}
+
+type MemoryType struct {
+	PhysicalMb int `json:"physical_mb"`
+	Total      int `json:"total"`
+}
+
+type RootDiskType struct {
+	Hctl               string `json:"hctl"`
+	Model              string `json:"model"`
+	Name               string `json:"name"`
+	Rotational         bool   `json:"rotational"`
+	Serial             string `json:"serial"`
+	Size               int    `json:"size"`
+	Vendor             string `json:"vendor"`
+	Wwn                string `json:"wwn"`
+	WwnVendorExtension string `json:"wwn_vendor_extension"`
+	WwnWithExtension   string `json:"wwn_with_extension"`
+}
+
+type SystemVendorType struct {
+	Manufacturer string `json:"manufacturer"`
+	ProductName  string `json:"product_name"`
+	SerialNumber string `json:"serial_number"`
+}
+
+// Extract interprets any IntrospectionDataResult as IntrospectionData, if possible.
+func (r DataResult) Extract() (*Data, error) {
+	var s Data
+	err := r.ExtractInto(&s)
+	return &s, err
+}
+
+// DataResult represents the response from a GetIntrospectionData operation.
+// Call its Extract method to interpret it as a Data.
+type DataResult struct {
+	gophercloud.Result
+}
+
+// ApplyDataResult is the response from an ApplyData operation.
+// Call its ExtractErr method to determine if the call succeeded or failed.
+type ApplyDataResult struct {
+	gophercloud.ErrResult
+}
--- a/openstack/baremetalintrospection/v1/introspection/testing/fixtures.go
+++ b/openstack/baremetalintrospection/v1/introspection/testing/fixtures.go
@@ -1,0 +1,374 @@
+package testing
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection"
+	th "github.com/gophercloud/gophercloud/testhelper"
+	"github.com/gophercloud/gophercloud/testhelper/client"
+)
+
+// IntrospectionListBody contains the canned body of a introspection.IntrospectionList response.
+const IntrospectionListBody = `
+{
+  "introspection": [
+    {
+      "error": null,
+      "finished": true,
+      "finished_at": "2017-08-17T11:36:16",
+      "links": [
+        {
+          "href": "http://127.0.0.1:5050/v1/introspection/05ccda19-581b-49bf-8f5a-6ded99701d87",
+          "rel": "self"
+        }
+      ],
+      "started_at": "2017-08-17T11:33:43",
+      "state": "finished",
+      "uuid": "05ccda19-581b-49bf-8f5a-6ded99701d87"
+    },
+    {
+      "error": null,
+      "finished": true,
+      "finished_at": "2017-08-16T12:24:30",
+      "links": [
+        {
+          "href": "http://127.0.0.1:5050/v1/introspection/c244557e-899f-46fa-a1ff-5b2c6718616b",
+          "rel": "self"
+        }
+      ],
+      "started_at": "2017-08-16T12:22:01",
+      "state": "finished",
+      "uuid": "c244557e-899f-46fa-a1ff-5b2c6718616b"
+    }
+  ]
+}
+`
+
+// IntrospectionStatus contains the respnse of a single introspection satus.
+const IntrospectionStatus = `
+{
+  "error": null,
+  "finished": true,
+  "finished_at": "2017-08-16T12:24:30",
+  "links": [
+    {
+      "href": "http://127.0.0.1:5050/v1/introspection/c244557e-899f-46fa-a1ff-5b2c6718616b",
+      "rel": "self"
+    }
+  ],
+  "started_at": "2017-08-16T12:22:01",
+  "state": "finished",
+  "uuid": "c244557e-899f-46fa-a1ff-5b2c6718616b"
+}
+`
+
+// IntrospectionDataJSONSample contains sample data reported by the introspection process.
+const IntrospectionDataJSONSample = `
+{
+   "cpu_arch":"x86_64",
+   "macs":[
+      "52:54:00:4e:3d:30"
+   ],
+   "root_disk":{
+      "rotational":true,
+      "vendor":"0x1af4",
+      "name":"/dev/vda",
+      "hctl":null,
+      "wwn_vendor_extension":null,
+      "wwn_with_extension":null,
+      "model":"",
+      "wwn":null,
+      "serial":null,
+      "size":13958643712
+   },
+   "interfaces": {
+      "eth0": {
+         "ip":"172.24.42.100",
+         "mac":"52:54:00:4e:3d:30",
+         "pxe":true,
+         "client_id":null
+     }
+   },
+   "cpus":2,
+   "boot_interface":"52:54:00:4e:3d:30",
+   "memory_mb":2048,
+   "ipmi_address":"192.167.2.134",
+   "inventory":{
+      "bmc_address":"192.167.2.134",
+      "interfaces":[
+         {
+            "lldp":null,
+            "product":"0x0001",
+            "vendor":"0x1af4",
+            "name":"eth1",
+            "has_carrier":true,
+            "ipv4_address":"172.24.42.101",
+            "client_id":null,
+            "mac_address":"52:54:00:47:20:4d"
+         },
+         {
+            "lldp":null,
+            "product":"0x0001",
+            "vendor":"0x1af4",
+            "name":"eth0",
+            "has_carrier":true,
+            "ipv4_address":"172.24.42.100",
+            "client_id":null,
+            "mac_address":"52:54:00:4e:3d:30"
+         }
+      ],
+      "disks":[
+         {
+            "rotational":true,
+            "vendor":"0x1af4",
+            "name":"/dev/vda",
+            "hctl":null,
+            "wwn_vendor_extension":null,
+            "wwn_with_extension":null,
+            "model":"",
+            "wwn":null,
+            "serial":null,
+            "size":13958643712
+         }
+      ],
+      "boot":{
+         "current_boot_mode":"bios",
+         "pxe_interface":"52:54:00:4e:3d:30"
+      },
+      "system_vendor":{
+         "serial_number":"Not Specified",
+         "product_name":"Bochs",
+         "manufacturer":"Bochs"
+      },
+      "memory":{
+         "physical_mb":2048,
+         "total":2105864192
+      },
+      "cpu":{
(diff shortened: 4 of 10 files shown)


elfosardo:
@jtopjian I'm seeing again the same issue mentioned in #1465 
any advice on how to avoid that here ?

jtopjian:
@elfosardo Try removing all occurrences of fields that are set to `interface{}(nil)`. The tests should still pass (omitting the field is the same as setting it to `interface{}(nil)`) and `gofmt` should work with all versions.

elfosardo:
@jtopjian I tried removing the occurrences but it was giving an error as it was expecting them with a value of `nil`
I checked the format script and I can see there are 2 files in an exclusion list that seem to have the same issue:
```
ignore_files=(
  "./openstack/compute/v2/extensions/quotasets/testing/fixtures.go"
  "./openstack/networking/v2/extensions/vpnaas/ikepolicies/testing/requests_test.go"
)
```
I think adding fixtures.go to the ignore list is reasonable at this point considering that the format is actually correct and it's just a problem between go versions.

jtopjian:
> I think adding fixtures.go to the ignore list is reasonable at this point considering that the format is actually correct and it's just a problem between go versions.

I think this is flagging a larger issue with the code.

If you look through Gophercloud, you'll see _very_ few occurrences of `interface{}(nil)`. Unless Ironic Inspector is an edge case, this is a flag that something isn't being modeled correctly.

I see that you're copying and pasting the following fixture: https://github.com/openstack/ironic-inspector/blob/054f300290c45035bb1cc1b222b2cc8cf26c3002/api-ref/source/samples/api-v1-data-introspection-response.json. 

Are you absolutely sure that fixture is representative of what a real response looks like? Are there really that many `null` values? And if there are `null` values, that means that Ironic Inspector is building a well-formed structure and not just sending a generic dict back in the response (which is what you're modeling when you declare Data as a generic `map[string]interface{}`).

For example, if I search for [`client_id`](https://github.com/openstack/ironic-inspector/search?q=client_id&unscoped_q=client_id), I find: https://github.com/openstack/ironic-inspector/blob/f7079e9acccfedc6b1bf5a1269869d39da80a700/ironic_inspector/plugins/standard.py#L170.

This tells us that there is a well-formed interfaces dict/struct that can be modeled appropriately in Gophercloud.

If I remove all `nil` values from the fixture struct and all `null` values from the fixture JSON, the test passes.

If Inspector was really sending back a generic dict, then the null values in the JSON response wouldn't exist. I recommend generating a real API response. If there are indeed null values, you'll need to start modeling how the data schema really looks. This way you can add the appropriate `omitempty` tags.

elfosardo:
@jtopjian thank you for your comment.
The sample json response I chose represents one of the possible responses that the GET introspection data request can retrieve, probably a most basic one.
The reason why I went for a generic dict representation is because of how the response of ironic-inspector is generated.
Ironic-inspector introspection operations are based on plugins, most of them are defined already in the ironic-inspector code itself, but there can be custom written plugins (e.g. third party vendors plugins), and the format and contents of the stored data depends on the plugins enabled both in the ramdisk and in inspector itself and also on the ramdisk used.
The fact that there are so many null values is because the structure itself can change and it's built with different structures put together, that may change based on the plugins enabled.
I believe you understand how it would be very hard to have a single model representation for this response.
Please let me know if you have any advice on how to face this problem.

openstack/baremetalinspector/v1/introspection/results.go
```diff
@@ -0,0 +1,226 @@
...
+	Model              string      `json:"model"`
+	Wwn                interface{} `json:"wwn"`
+	Serial             interface{} `json:"serial"`
+	Size               int         `json:"size"`
+}
+
+// InventoryType ...
+type InventoryType struct {
+	BmcAddress   string                 `json:"bmc_address"`
+	Interfaces   []interface{}          `json:"interfaces"`
```
jtopjian: I think we can create a more defined struct based off of this: 

* https://github.com/openstack/ironic-python-agent/blob/cfb62c87658e32af373278a8de655367f3b489af/ironic_python_agent/hardware.py#L635
* https://github.com/openstack/ironic-python-agent/blob/cfb62c87658e32af373278a8de655367f3b489af/ironic_python_agent/hardware.py#L291

```go
type Data struct {
  ...
  Interfaces []Interface `json:"interfaces"`
  ...
}

type Interface struct {
  Name string `json:"name"`
  MACAddress string `json:"mac_address"`
  ...
}
```

openstack/baremetalinspector/v1/introspection/results.go
```diff
@@ -0,0 +1,226 @@
...
+	Disks        []RootDiskType         `json:"disks"`
+	Boot         map[string]interface{} `json:"boot"`
+	SystemVendor map[string]interface{} `json:"system_vendor"`
```
jtopjian: Similarly, I think we can define a more formal `SystemVendor` struct: 

* https://github.com/openstack/ironic-python-agent/blob/cfb62c87658e32af373278a8de655367f3b489af/ironic_python_agent/hardware.py#L824
* https://github.com/openstack/ironic-python-agent/blob/cfb62c87658e32af373278a8de655367f3b489af/ironic_python_agent/hardware.py#L337

openstack/baremetalinspector/v1/introspection/results.go
```diff
@@ -0,0 +1,226 @@
...
+	Serial             interface{} `json:"serial"`
+	Size               int         `json:"size"`
+}
+
+// InventoryType ...
+type InventoryType struct {
+	BmcAddress   string                 `json:"bmc_address"`
+	Interfaces   []interface{}          `json:"interfaces"`
+	Disks        []RootDiskType         `json:"disks"`
+	Boot         map[string]interface{} `json:"boot"`
```
jtopjian: Similarly for `Boot`: https://github.com/openstack/ironic-python-agent/blob/cfb62c87658e32af373278a8de655367f3b489af/ironic_python_agent/hardware.py#L346

openstack/baremetalinspector/v1/introspection/results.go
```diff
@@ -0,0 +1,226 @@
...
+type RootDiskType struct {
+	Rotational         bool        `json:"rotational"`
+	Vendor             string      `json:"vendor"`
+	Name               string      `json:"name"`
+	Hctl               interface{} `json:"hctl"`
+	WwnVendorExtension interface{} `json:"wwn_vendor_extension"`
+	WwnWithExtension   interface{} `json:"wwn_with_extension"`
+	Model              string      `json:"model"`
+	Wwn                interface{} `json:"wwn"`
+	Serial             interface{} `json:"serial"`
```
jtopjian: For the above fields, it looks like you're choosing `interface{}` because the test fixture has the value set to `null`? We should try to figure out what kind of core type they are. 

`hctl` looks to be a string: https://github.com/openstack/ironic-python-agent/blob/cfb62c87658e32af373278a8de655367f3b489af/ironic_python_agent/hardware.py#L229-L230

```shell
# ls /sys/block/sdc/device/scsi_device/
2:0:0:1
```

The other fields, which I don't have native access to, also look like strings: https://github.com/openstack/ironic-python-agent/blob/cfb62c87658e32af373278a8de655367f3b489af/ironic_python_agent/tests/unit/test_hardware.py#L898-L906

openstack/baremetalinspector/v1/introspection/results.go
```diff
@@ -0,0 +1,226 @@
...
+// Data represents introspection data collected.
+// The format and contents of the stored data depends on the ramdisk used
+// and plugins enabled both in the ramdisk and in inspector itself.
+// This structure has been provided for basic compatibility but it
+// will need extensions
+type Data struct {
+	CPUArch       string                 `json:"cpu_arch"`
+	MACs          []string               `json:"macs"`
+	RootDisk      RootDiskType           `json:"root_disk"`
+	Interfaces    map[string]interface{} `json:"interfaces"`
```
jtopjian: `interfaces` and `all_interfaces` look like they are defined here: https://github.com/openstack/ironic-inspector/blob/f7079e9acccfedc6b1bf5a1269869d39da80a700/ironic_inspector/plugins/standard.py#L252-L255

openstack/baremetalinspector/v1/introspection/results.go
```diff
@@ -0,0 +1,226 @@
...
+// UnmarshalJSON trie to convert values for started_at and finished_at from the
+// json response into RFC3339 standard. Since Introspection API can remove the
+// Z from the format, if the conversion fails, it falls back to an RFC3339
+// with no Z format supported by gophercloud.
+func (r *Introspection) UnmarshalJSON(b []byte) error {
+	type tmp Introspection
+	var s struct {
+		tmp
+		StartedAt  string `json:"started_at"`
+		FinishedAt string `json:"finished_at"`
```
jtopjian: Are you certain that the time could be in two different formats? Or is this a precaution?

openstack/baremetalinspector/v1/introspection/results.go
juliakreger: These values can also simply be a null value on the JSON document or possibly even  not exist such as the values noted. See https://github.com/openstack/ironic-python-agent/blob/master/ironic_python_agent/hardware.py#L245

openstack/baremetalinspector/v1/introspection/results.go
juliakreger: So the class references your pointing to are part of IPA's internal programming model. Is it wise to try and mirror the programming model in a separate API's return data?

jtopjian: > Is it wise to try and mirror the programming model in a separate API's return data?

What I'm looking for is a more formal structure to the output of this API suite. As mentioned in the comments earlier, what triggered this for me is the amount of `null` values in the test fixture. This means that the field is being set to `null` _somewhere_. Contrast this to a generic dictionary that didn't have the key set at all - then the `null` fields simply wouldn't exist.

Put another way, `mac_address` is being set _somewhere_. It didn't magically appear out of thin air. So if Ironic Inspector is mirroring the internal programming model of the agent in its API response, then, yes, I think it's appropriate.

If this isn't appropriate, then the Ironic Inspector API should really have a more formal model of its API response. Returning an amorphous dictionary of data is difficult to work with - how can I be sure that a field will always exist?

openstack/baremetalinspector/v1/introspection/results.go
jtopjian: It's totally fine if the fields' values can sometimes be `null` / not set. The important part is that the API response is including the field in the response: https://github.com/openstack/ironic-inspector/blob/054f300290c45035bb1cc1b222b2cc8cf26c3002/api-ref/source/samples/api-v1-continue-request.json#L7

On the other hand, if the Ironic Inspector test fixtures are _not_ accurately portraying a _real_ API response (meaning, arbitrary fields are being added just for the sake of a unit test), then that's a bad test fixture and it should not be used in Gophercloud.

In other words: https://github.com/openstack/ironic-inspector/blob/054f300290c45035bb1cc1b222b2cc8cf26c3002/api-ref/source/samples/api-v1-continue-request.json has been used as the basis of the Gophercloud fixture. If this fixture is mocked and not accurate, then it's better to generate fixtures that have actually been pulled from real API calls -- that can make a world of different in how something is implemented in a typed language.

If Ironic Inspector _sometimes_ doesn't set the `wwn_vendor_extension` field in `root_disk`:

```json
  "root_disk": {
    "rotational": true,
    "vendor": "0x1af4",
    "name": "/dev/vda",
    "hctl": null,
    "wwn_with_extension": null,
    "model": "",
    "wwn": null,
    "serial": null,
    "size": 13958643712
  },
```

Then `root_disk` is _absolutely_ a generic `map[string]interface{}`. But because the fixtures are showing `null`, that tells me that Ironic Inspector or the Agent are setting the field somewhere, thus creating a well-formed response.

I hope that makes sense.

openstack/baremetalinspector/v1/introspection/results.go
jtopjian: One additional thought: I'm only assuming that Ironic Inspector is getting its values from `ironic-python-agent`. The values seem to match up. If this is incorrect and it's purely coincidence, then my apologies.

But as I've mentioned before, it's the existence of `null` values that has me believing that there really is a formal definition _somewhere_.

jtopjian:
@juliakreger @elfosardo Attempting to clarify my comments a little more:

First, I fully understand if I'm coming off as being overly picky with an addition of OCD. Can the entire result be combined into a generic `map[string]interface{}`? Yes, absolutely. But by doing that, we're removing a large part of why someone would choose Go over another language: the strict type implementation.

If Ironic Inspector doesn't actually have an official API response definition, okay... but that makes API guarantees difficult, not to mention actually using the result in production. If we were to use a generic `map[string]interface{}` for everything, production code implementing Gophercloud would need to look like:

```go
if v, ok := result["macs"]; ok {
  if v, ok := v.([]interface{}); ok {
    for _, i := range v {
      if mac, ok := i.(string); ok {
        fmt.Println(mac)
      }
    }
  }
}
```

Instead of:

```go
for _, mac := range result.MACs {
  fmt.Println(mac)
}
```

Based on the above, you can see how digging into a more nested piece of data such as `ipv4_address` would become much more difficult. There is a _lot_ of room for causing panics due to not checking if a key exists or not doing a type assertion.

This is why I place so much emphasis on data modeling.

Second, since I don't have access to an Ironic/Inspector environment, I can only go off of what the fixtures are telling me. Looking at the timestamps of the fixtures, I can tell that the fixtures bundled in this PR were not generated recently in a devstack or prod environment. They were pulled from here: 

* https://github.com/openstack/ironic-inspector/tree/master/api-ref/source/samples
* https://developer.openstack.org/api-ref/baremetal-introspection/

This means that I must trust that the fixtures are accurate and weren't amended for convenience of writing a quick unit test. This sounds trivial, but incorrect fixtures have occurred before. Go is the kind of language that is going to quickly let you know when a mocked fixture is not the same as what it sees in reality.

The fixtures are telling me that well-formed data structures are being returned. This is because of the amount of `null` values being returned. If it was truly a generic hash/dictionary/map being returned, then `null` keys wouldn't even be set and thus not be in the response.

The latest commit has done a very good job at modeling the API response into a series of well-defined structures. The additional comments I gave are an attempt to model the remaining generic `map[string]interfaces{}`. If that's not possible, _that is totally fine_. 

I apologize if I'm coming across as picking on this PR -- that is not my intention at all. This PR has the unfortunate goal of modeling a very structured set of data. If you look through the rest of Gophercloud, a lot of the work is done in the modeling of API responses.

acceptance/clients/clients.go
```diff
@@ -285,6 +285,27 @@ func NewBareMetalV1NoAuthClient() (*gophercloud.ServiceClient, error) {
 	})
 }
 
+// NewBareMetalInspectorV1Client returns a *ServiceClient for making calls
+// to the OpenStack Bare Metal Inspector v1 API. An error will be returned
+// if authentication or client creation was not possible.
+func NewBareMetalInspectorV1Client() (*gophercloud.ServiceClient, error) {
```
dtantsur: Comment on the whole patch: it's either "ironic inspector" OR "baremetal introspection", but not a mix of them.

openstack/baremetalinspector/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+
+// StartOptsBuilder ...
+type StartOptsBuilder interface {
+	ToStartIntrospectionQuery() (string, error)
+}
+
+// StartOpts ...
+type StartOpts struct {
+	// Whether the current installation of ironic-inspector can manage PXE booting of nodes.
+	ManageBoot bool `q:"manage_boot"`
```
dtantsur: What's going to be send by default? The correct answers are "nothing" or `true`, `false` is a specific mode.

openstack/baremetalinspector/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+func GetIntrospectionData(client *gophercloud.ServiceClient, NodeID string) (r dataResult) {
+	_, r.Err = client.Get(introspectionDataURL(client, NodeID), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	return
+}
+
+// ApplyIntrospectionData triggers introspection on stored unprocessed data.
+// No data is allowed to be sent along with the request.
+func ApplyIntrospectionData(client *gophercloud.ServiceClient, NodeID string) (r ApplyDataResult) {
```
dtantsur: Maybe `ReapplyIntrospection`? "Apply data" sounds a bit weird to me.

openstack/baremetalinspector/v1/introspection/results.go
```diff
@@ -0,0 +1,226 @@
...
+type AbortResult struct {
+	gophercloud.ErrResult
+}
+
+// Data represents introspection data collected.
+// The format and contents of the stored data depends on the ramdisk used
+// and plugins enabled both in the ramdisk and in inspector itself.
+// This structure has been provided for basic compatibility but it
+// will need extensions
+type Data struct {
```
dtantsur: I hope these fields are all optional, because we don't have any guarantees about it. Actually, maybe we should leave only `inventory` since it's the only thing that has a defined (in IPA) format.

openstack/baremetalinspector/v1/introspection/results.go
dtantsur: A word of caution: do not rely on `map[string]string` for these fields, since we don't guarantee not adding integers or more complex structures later. If we can have a structure with some fields, it may be fine (as long as Go is sane and does not blow up on unexpected fields).

acceptance/clients/clients.go
jtopjian: Good catch.

Please use Baremetal / BareMetal where appropriate. Gophercloud uses service names and not project/code names.

openstack/baremetalinspector/v1/introspection/requests.go
jtopjian: for `q` tags (these are internal to Gophercloud), zero-values are not sent. So if `ManageBoot` isn't specified, then `?manage_boot=false` won't be sent in the request.

If sending false is a valid thing, then this will need changed to `*bool`.

openstack/baremetalinspector/v1/introspection/results.go
jtopjian: Please see my other comments in this area. I'm hesitant to have `Data` as one large `map[string]interface{}` for reasons mentioned. I'm a bit surprised if this API has no guaranteed / defined structure at all. 

However, as I've also stated, I have no knowledge of any of these APIs, so I'm leaning on other peoples' expertise. If it's best to have generic `map[string]interfaces{}` in certain areas, okay. But using the results is going to be a headache.

Also, if a JSON response does not contain a field that a struct is expecting, all is fine. If a JSON response contains _extra_ fields that the struct does not account for, all is fine.

openstack/baremetalinspector/v1/introspection/requests.go
dtantsur: Sending `manage_boot=false` is a valid case indeed.

openstack/baremetalinspector/v1/introspection/results.go
dtantsur: The problem with this data is that it's a hybrid of more or less well defined `inventory` and whatever ironic-inspector plugins put there (which is poorly documented and often legacy). Is it possible in Go to have a hybrid approach as well? I know in `serde` (Rust serializatin library) you can have a defined structure with one mapping field serving as catch-all for everything that was not mapped to defined fields.

openstack/baremetalinspector/v1/introspection/results.go
```diff
@@ -0,0 +1,226 @@
...
+	RootDisk      RootDiskType           `json:"root_disk"`
+	Interfaces    map[string]interface{} `json:"interfaces"`
+	CPUs          int                    `json:"cpus"`
+	BootInterface string                 `json:"boot_interface"`
+	MemoryMB      int                    `json:"memory_mb"`
+	IPMIAddress   string                 `json:"ipmi_address"`
+	Inventory     InventoryType          `json:"inventory"`
+	Error         string                 `json:"error"`
+	LocalGB       int                    `json:"local_gb"`
+	AllInterfaces map[string]interface{} `json:"all_interfaces"`
```
dtantsur: Speaking of poorly defined interfaces, the `extra_hardware` plugin defines the `extra` key, which is a 3 level nested dict: https://github.com/openstack/ironic-inspector/blob/master/ironic_inspector/plugins/extra_hardware.py#L66

openstack/baremetalinspector/v1/introspection/results.go
jtopjian: Go doesn't have anything which natively does that (but it would be nice). We have an internal function that does it, though. We use it for things like Keystone where arbitrary extra key/values can be sent in the API response:

https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/users/results.go#L60-L74

I've never tested it with nested things, but it _should_ work.

If we can get a confirmed set of base/expected fields, I'm more than happy to work on options to catch the undefined/optional pieces.

openstack/baremetalinspector/v1/introspection/requests.go
elfosardo: changed

openstack/baremetalinspector/v1/introspection/requests.go
elfosardo: changed

acceptance/clients/clients.go
elfosardo: changed to baremetalintrospection

openstack/baremetalinspector/v1/introspection/results.go
elfosardo: this is more of a precaution

openstack/baremetalinspector/v1/introspection/testing/fixtures.go
```diff
@@ -0,0 +1,402 @@
...
+			Disks: []introspection.RootDiskType{
+				introspection.RootDiskType{
+					Rotational:         true,
+					Model:              "",
+					Name:               "/dev/vda",
+					WwnVendorExtension: interface{}(nil),
+					Wwn:                interface{}(nil),
+					WwnWithExtension:   interface{}(nil),
+					Hctl:               interface{}(nil),
+					Serial:             interface{}(nil),
```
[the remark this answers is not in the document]
elfosardo: removed

openstack/baremetalinspector/v1/introspection/testing/fixtures.go
```diff
@@ -0,0 +1,402 @@
...
+		MACs:    []string{"52:54:00:4e:3d:30"},
+		RootDisk: introspection.RootDiskType{
+			Rotational:         true,
+			Model:              "",
+			Name:               "/dev/vda",
+			WwnVendorExtension: interface{}(nil),
+			Wwn:                interface{}(nil),
+			WwnWithExtension:   interface{}(nil),
+			Hctl:               interface{}(nil),
+			Serial:             interface{}(nil),
```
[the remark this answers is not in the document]
elfosardo: removed

openstack/baremetalinspector/v1/introspection/results.go
elfosardo: My understanding is that it will not cause any issue if it's not present in the structure.
Of course it will be interpreted and not formally defined.

elfosardo:
@jtopjian @dtantsur @juliakreger I don't mind moving forward defining an even more formal definition of the data structure, or just refine what we already have.
I think it's clear that because of how the json data is built we need to consider basically the entire response made by optional fields.

openstack/baremetalintrospection/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+// GetIntrospectionStatus makes a request against the Inspector API to get the
+// status of a single introspection.
+func GetIntrospectionStatus(client *gophercloud.ServiceClient, NodeID string) (r GetIntrospectionStatusResult) {
+	_, r.Err = client.Get(introspectionURL(client, NodeID), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	return
+}
+
+// StartOptsBuilder ...
```
jtopjian:
```
// StartOptsBuilder allows extensions to add additional parameters to the
// Start request.
```

openstack/baremetalintrospection/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+type StartOptsBuilder interface {
+	ToStartIntrospectionQuery() (string, error)
+}
+
+// StartOpts ...
```
jtopjian:
```
// StartOpts represents options to start an introspection.
```

openstack/baremetalintrospection/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+type StartOpts struct {
+	// Whether the current installation of ironic-inspector can manage PXE booting of nodes.
+	ManageBoot *bool `q:"manage_boot"`
+}
+
+// ToStartIntrospectionQuery ...
```
jtopjian:
```
// ToStartIntrospectionQuery converts a StartOpts into a request.
```

openstack/baremetalintrospection/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
+		var rpage = IntrospectionPage{pagination.LinkedPageBase{PageResult: r}}
+		return rpage
+	})
+}
+
+// GetIntrospectionStatus makes a request against the Inspector API to get the
+// status of a single introspection.
+func GetIntrospectionStatus(client *gophercloud.ServiceClient, NodeID string) (r GetIntrospectionStatusResult) {
```
jtopjian: `s/NodeID/nodeID`

openstack/baremetalintrospection/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+
+// ToStartIntrospectionQuery ...
+func (opts StartOpts) ToStartIntrospectionQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
+// StartIntrospection initiate hardware introspection for node NodeID .
+// All power management configuration for this node needs to be done prior to calling the endpoint.
+func StartIntrospection(client *gophercloud.ServiceClient, NodeID string, opts StartOptsBuilder) (r StartResult) {
```
jtopjian: `s/NodeID/nodeID`

openstack/baremetalintrospection/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+
+	_, r.Err = client.Post(introspectionURL(client, NodeID), nil, nil, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+	})
+
+	return
+}
+
+// AbortIntrospection abort running introspection.
+func AbortIntrospection(client *gophercloud.ServiceClient, NodeID string) (r AbortResult) {
```
jtopjian: `s/NodeID/nodeID`

openstack/baremetalintrospection/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+	_, r.Err = client.Post(abortIntrospectionURL(client, NodeID), nil, nil, &gophercloud.RequestOpts{
+		OkCodes: []int{202},
+	})
+
+	return
+}
+
+// GetIntrospectionData return stored data from successful introspection.
+func GetIntrospectionData(client *gophercloud.ServiceClient, NodeID string) (r dataResult) {
```
jtopjian: `s/NodeID/nodeID`

also, `dataResult` should be `DataResult`.

Upper-case variables/methods/functions represent _exported_ items in the package. It's technically possible to have upper-case names in the function _parameters_ but since they are scoped to the package-local function, it doesn't make sense, so it's never done.

However, upper-case return variables are used to denote exported types. In this case, `dataResult` would never be accessible outside of the `introspection` package.

openstack/baremetalintrospection/v1/introspection/requests.go
```diff
@@ -0,0 +1,115 @@
...
+	_, r.Err = client.Get(introspectionDataURL(client, NodeID), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	return
+}
+
+// ReApplyIntrospection triggers introspection on stored unprocessed data.
+// No data is allowed to be sent along with the request.
+func ReApplyIntrospection(client *gophercloud.ServiceClient, NodeID string) (r ApplyDataResult) {
```
jtopjian: `s/NodeID/nodeID`

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// introspectionResult ...
```
jtopjian: I recommend removing this doc comment. `introspectionResult` is local to this package and not exposed to the user.

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+type introspectionResult struct {
+	gophercloud.Result
+}
+
+// Extract interprets any nodeResult as an Introspection, if possible.
```
jtopjian:
```
Extract interprets any introspectionResult as an Introspection, if possible.
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+func (r introspectionResult) Extract() (*Introspection, error) {
+	var s Introspection
+	err := r.ExtractInto(&s)
+	return &s, err
+}
+
+// ExtractInto ...
```
jtopjian:
```
// ExtractInto will extract a response body into an Introspection struct.
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+func (r introspectionResult) ExtractInto(v interface{}) error {
+	return r.Result.ExtractIntoStructPtr(v, "")
+}
+
+// ExtractIntrospectionsInto ...
```
jtopjian:
```
// ExtractIntrospectionsInto will extract a collection of introspectResult pages into a 
// slice of Introspection entities.
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+	StartedAt time.Time `json:"-"`
+
+	// UTC ISO8601 timestamp or null
+	FinishedAt time.Time `json:"-"`
+
+	// Link to the introspection URL
+	Links []interface{} `json:"links"`
+}
+
+// IsEmpty returns true if a page contains no Server results.
```
jtopjian:
```
// IsEmpty returns true if a page contains no Introspection results.
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+				return err
+			}
+		}
+		r.FinishedAt = t
+	}
+
+	return nil
+}
+
+// GetIntrospectionStatusResult is the response from a Get operation.
```
jtopjian: `s/Get/GetIntrospectionStatus`

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+// Call its Extract method to interpret it as an Introspection.
+type GetIntrospectionStatusResult struct {
+	introspectionResult
+}
+
+// StartResult ...
```
jtopjian:
```
// StartResult is the response from a StartIntrospection operation.
// Call its ExtractErr method to determine if the call succeeded or failed.
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+type StartResult struct {
+	gophercloud.ErrResult
+}
+
+// AbortResult ...
```
jtopjian:
```
// AbortResult is the response from an AbortIntrospection operation.
// Call its ExtractErr method to determine if the call succeeded or failed.
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+	IPMIAddress   string                 `json:"ipmi_address"`
+	LocalGB       int                    `json:"local_gb"`
+	MACs          []string               `json:"macs"`
+	MemoryMB      int                    `json:"memory_mb"`
+	RootDisk      RootDiskType           `json:"root_disk"`
+}
+
+// Sub Types defined under Data and deeper in the structure
+
+// BootInfoType ...
```
jtopjian: Either omit the doc comment or do something like:

```
// BootInfoType represents <short description>
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+type BootInfoType struct {
+	CurrentBootMode string `json:"current_boot_mode"`
+	PXEInterface    string `json:"pxe_interface"`
+}
+
+// CPUType ...
```
jtopjian: Either omit the doc comment or do something like:

```
// CPUType represents <short description>
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+type CPUType struct {
+	Architecture string   `json:"architecture"`
+	Count        int      `json:"count"`
+	Flags        []string `json:"flags"`
+	Frequency    string   `json:"frequency"`
+	ModelName    string   `json:"model_name"`
+}
+
+// InterfaceType ...
```
jtopjian: Either omit the doc comment or do something like:

```
// InterfaceType represents <short description>
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+	IPV4Address string `json:"ipv4_address"`
+	IPV6Address string `json:"ipv6_address"`
+	Lldp        string `json:"lldp"`
+	MACAddress  string `json:"mac_address"`
+	Name        string `json:"name"`
+	Product     string `json:"product"`
+	Vendor      string `json:"vendor"`
+}
+
+// InventoryType ...
```
jtopjian: Either omit the doc comment or do something like:

```
// InventoryType represents <short description>
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+	BmcAddress   string           `json:"bmc_address"`
+	Boot         BootInfoType     `json:"boot"`
+	CPU          CPUType          `json:"cpu"`
+	Disks        []RootDiskType   `json:"disks"`
+	Interfaces   []InterfaceType  `json:"interfaces"`
+	Memory       MemoryType       `json:"memory"`
+	SystemVendor SystemVendorType `json:"system_vendor"`
+}
+
+// MemoryType ...
```
jtopjian: Either omit the doc comment or do something like:

```
// MemoryType represents <short description>
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+type MemoryType struct {
+	PhysicalMb int `json:"physical_mb"`
+	Total      int `json:"total"`
+}
+
+// RootDiskType ...
```
jtopjian: Either omit the doc comment or do something like:

```
// RootDiskType represents <short description>
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+	Rotational         bool   `json:"rotational"`
+	Serial             string `json:"serial"`
+	Size               int    `json:"size"`
+	Vendor             string `json:"vendor"`
+	Wwn                string `json:"wwn"`
+	WwnVendorExtension string `json:"wwn_vendor_extension"`
+	WwnWithExtension   string `json:"wwn_with_extension"`
+}
+
+// SystemVendorType ...
```
jtopjian: Either omit the doc comment or do something like:

```
// SystemVendorType represents <short description>
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+
+// Extract interprets any IntrospectionDataResult as IntrospectionData, if possible.
+func (r dataResult) Extract() (*Data, error) {
+	var s Data
+	err := r.ExtractInto(&s)
+	return &s, err
+}
+
+// DataResult ...
+type dataResult struct {
```
jtopjian: `s/dataResult/DataResult`

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// Extract interprets any IntrospectionDataResult as IntrospectionData, if possible.
+func (r dataResult) Extract() (*Data, error) {
+	var s Data
+	err := r.ExtractInto(&s)
+	return &s, err
+}
+
+// DataResult ...
```
jtopjian:
```
// DataResult represents the response from a GetIntrospectionData operation.
// Call its Extract method to interpret it as a Data.
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,256 @@
...
+type dataResult struct {
+	gophercloud.Result
+}
+
+// ApplyDataResult ...
```
jtopjian:
```
// ApplyDataResult is the response from an ApplyData operation.
// Call its ExtractErr method to determine if the call succeeded or failed.
```

openstack/baremetalintrospection/v1/introspection/testing/fixtures.go
```diff
@@ -0,0 +1,380 @@
...
+      "rel": "self"
+    }
+  ],
+  "started_at": "2017-08-16T12:22:01",
+  "state": "finished",
+  "uuid": "c244557e-899f-46fa-a1ff-5b2c6718616b"
+}
+`
+
+// IntrospectionDataJSONSample ...
```
jtopjian: Comments which have `...` in this file can be safely removed unless you want to add more details to them.

openstack/baremetalintrospection/v1/introspection/testing/requests_test.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection"
+	"github.com/gophercloud/gophercloud/pagination"
+	th "github.com/gophercloud/gophercloud/testhelper"
+	"github.com/gophercloud/gophercloud/testhelper/client"
+)
+
+// TestListIntrospections ...
```
jtopjian: Comments that have `...` in this file can be safely removed unless you want to add more details to them.

openstack/baremetalintrospection/v1/introspection/requests.go
elfosardo: thanks, this was actually a typo, it was meant to be DataResult

openstack/baremetalinspector/v1/introspection/results.go
elfosardo: I created an InterfaceType for the interfaces in the inventory, which is the most important part.
I think the interfaces in the main dict can stay as this for now.

openstack/baremetalinspector/v1/introspection/results.go
elfosardo: Created a BootInfoType for this

openstack/baremetalinspector/v1/introspection/results.go
elfosardo: Created a SystemVendorType for this

openstack/baremetalinspector/v1/introspection/results.go
elfosardo: I had a better look at the code and some definitions are actually deprecated.
Changed code accordingly.

elfosardo:
@jtopjian @dtantsur @juliakreger I made some other changes to be as close as possible to the data defined in the base API code (both inspector and IPA) and modified the json data fixture as well since there were deprecated fields.
The schema should be now correct with no generic maps or interfaces.
Please let me know if I forgot anything, thanks!

jtopjian:
I'm going to merge this, but keep in mind this isn't written in stone - follow-up PRs are totally fine.

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,260 @@
...
+type AbortResult struct {
+	gophercloud.ErrResult
+}
+
+// Data represents the full introspection data collected.
+// The format and contents of the stored data depends on the ramdisk used
+// and plugins enabled both in the ramdisk and in inspector itself.
+// This structure has been provided for basic compatibility but it
+// will need extensions
+type Data struct {
```
dtantsur: I would repeat my recommendation to leave only `Inventory` here (and possibly add `Extra`). Everything else is not set in stone, with some items being legacy from quite ancient times. Instead, I'd provide a way for consumers to retrieve the whole data as `map[string]interface` and do whatever they want with it.

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,260 @@
...
+	ModelName    string   `json:"model_name"`
+}
+
+type InterfaceType struct {
+	BIOSDevName string                 `json:"biosdevname"`
+	ClientID    string                 `json:"client_id"`
+	HasCarrier  bool                   `json:"has_carrier"`
+	IPV4Address string                 `json:"ipv4_address"`
+	IPV6Address string                 `json:"ipv6_address"`
+	Lldp        map[string]interface{} `json:"lldp"`
```
dtantsur: Why `MAC` and `BIOS`, but `Lldp`?

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -0,0 +1,260 @@
...
+	Flags        []string `json:"flags"`
+	Frequency    string   `json:"frequency"`
+	ModelName    string   `json:"model_name"`
+}
+
+type InterfaceType struct {
+	BIOSDevName string                 `json:"biosdevname"`
+	ClientID    string                 `json:"client_id"`
+	HasCarrier  bool                   `json:"has_carrier"`
+	IPV4Address string                 `json:"ipv4_address"`
```
dtantsur: I think `IPv4Address` feels a bit more natural.

elfosardo:
@dtantsur thanks for the comments, I will do a follow-up with some minor changes and to define better lldp as it has a defined schema

PR #1460, opened 2019-02-20
Identity V3: Add credentials package (#1457)

himanshuvar:
* Identity V3: Add credentials package

For #1457

Adding the support for Keystone Credential package:

https://github.com/openstack/keystone/tree/master/keystone/credential

PR #1487, opened 2019-02-27
Fix typos in tags for ListOpts

kayrus:
found within #1460 code review


jtopjian:
Thanks for catching this - I'll merge it once tests have finished.

Issue #1489, opened 2019-03-01
Octavia: Support vip_port_id for load balancer creation


lingxiankong:
Currently, the `vip_port_id` is not supported.

jtopjian:
Implemented in #1489

PR #1490, opened 2019-03-01
Octavia: Support to specify vip_port_id for creating load balancer

lingxiankong:
For #1489

code in this PR:

- Source code: <https://github.com/openstack/octavia/blob/612b042e64d9fcffb983f958ed0d6de6a9f765d3/octavia/api/v2/types/load_balancer.py#L115>
- API doc: <https://developer.openstack.org/api-ref/load-balancer/v2/index.html#create-a-load-balancer>

### DIFF
--- a/openstack/loadbalancer/v2/loadbalancers/requests.go
+++ b/openstack/loadbalancer/v2/loadbalancers/requests.go
@@ -81,10 +81,16 @@ type CreateOpts struct {
 	// Human-readable description for the Loadbalancer.
 	Description string `json:"description,omitempty"`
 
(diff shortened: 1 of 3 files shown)


lingxiankong:
Hi @jtopjian, I need your help for the coverage, with the changes I made, I don't know how I could increase the test coverage.

jtopjian:
Oh, don't ever worry about the coverage. It's definitely a non-blocking test. And when coveralls reports numbers like 0.01%, it's more of a calculation oddity.

I'll talk a look at this tomorrow :)

Issue #1491, opened 2019-03-01
Possible to create invalid v3 auth if just project_name defined.


leigh-j:
When using v3 auth and using projects not in the default domain, specifying just project name is not valid according to the spec.
[From the spec](https://developer.openstack.org/api-ref/identity/v3/?expanded=password-authentication-with-scoped-authorization-detail#authenticating-with-an-application-credential)

> An ID is sufficient to uniquely identify a project but if a project is specified by name, then the domain of the project must also be specified in order to uniquely identify the project by name. A domain scope may be specified by either the domain’s ID or name with equivalent results.

This is a quirky corner case with the openstack api.

jtopjian:
Fixed in #1492

PR #1492, opened 2019-03-01
check for project id if project name specified

leigh-j:
Check extra params are set if project name is given

Setting project name is insufficient according to the spec for project scoped auth.

For #1491 

https://developer.openstack.org/api-ref/identity/v3/?expanded=password-authentication-with-scoped-authorization-detail#authenticating-with-an-application-credential

### DIFF
--- a/openstack/auth_env.go
+++ b/openstack/auth_env.go
@@ -13,15 +13,19 @@ AuthOptionsFromEnv fills out an identity.AuthOptions structure with the
 settings found on the various OpenStack OS_* environment variables.
 
 The following variables provide sources of truth: OS_AUTH_URL, OS_USERNAME,
-OS_PASSWORD, OS_TENANT_ID, and OS_TENANT_NAME.
+OS_PASSWORD and OS_PROJECT_ID.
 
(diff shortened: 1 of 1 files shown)


leigh-j:
Related #1460

jtopjian:
@leigh-j Thank you for working on this. Please let me know when it is ready for review.

leigh-j:
Its ready now,  The entire AuthOptionsFromEnv function has no tests so should I just add a new test package under https://github.com/gophercloud/gophercloud/tree/master/testing ? 

It looks like the v3 token code has the logic already though, I cant see any other changes required
https://github.com/gophercloud/gophercloud/blob/8788235096d743c9abd213ba53adca6b0e9d0b58/auth_options.go#L359

jtopjian:
@leigh-j I think this should be fine without a test - the change makes sense to me.

In general, `auth_env.go` is more of a convenience and is largely used to bootstrap testing in Gophercloud. There's a much more robust authentication package at https://github.com/gophercloud/utils/tree/master/openstack/clientconfig.

PR #1495, opened 2019-03-01
Networking V2: Add network MTU extension support

### LINKED ISSUE #1493: Networking V2: Add MTU extension support
Neutron supports MTU extension: https://github.com/openstack/neutron/blob/mitaka-eol/neutron/extensions/netmtu.py

And starting from Pike, MTU extension is writeable: https://github.com/openstack/neutron/blob/stable/pike/neutron/extensions/netmtu_writable.py

@jtopjian I need you suggestion on how to deal with this implementation. Is it required to introduce two extensions? Is it possible just to add an MTU field into the existing Network structure?

kayrus:
Resolves #1493

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/mtu/mtu_test.go
+++ b/acceptance/openstack/networking/v2/extensions/mtu/mtu_test.go
@@ -1,0 +1,136 @@
+// +build acceptance networking
+
+package mtu
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	networking "github.com/gophercloud/gophercloud/acceptance/openstack/networking/v2"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/common/extensions"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/mtu"
+	"github.com/gophercloud/gophercloud/openstack/networking/v2/networks"
+	"github.com/gophercloud/gophercloud/pagination"
+	th "github.com/gophercloud/gophercloud/testhelper"
+)
+
+func TestMTUNetworkCRUDL(t *testing.T) {
+	clients.RequireAdmin(t)
+
+	client, err := clients.NewNetworkV2Client()
+	th.AssertNoErr(t, err)
+
+	extension, err := extensions.Get(client, "net-mtu").Extract()
+	if err != nil {
+		t.Skip("This test requires dns-integration Neutron extension")
+	}
+	tools.PrintResource(t, extension)
+
+	mtuWritable, _ := extensions.Get(client, "net-mtu-writable").Extract()
+	tools.PrintResource(t, mtuWritable)
+
+	// Create Network
+	var networkMTU int
+	if mtuWritable != nil {
+		networkMTU = 1450
+	}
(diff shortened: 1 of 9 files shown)


kayrus:
@jtopjian could you please check it in your environment? It works fine in mitaka.

```
Bad request with: [GET http://192.168.200.182:9696/v2.0/networks?mtu=1450&name=TESTACC-ZtUqAVkR]:
{"NeutronError": {"message": "[u'mtu'] is invalid attribute for filtering", "type": "HTTPBadRequest", "detail": ""}}
```

kayrus:
This is a bug. [Read-only](https://github.com/openstack/neutron-lib/blob/fc2a81058bfd3ba9fd3501660156c71ff1c8129c/neutron_lib/api/definitions/network_mtu.py#L51..L52) MTU supports listing, [writable](https://github.com/openstack/neutron-lib/blob/fc2a81058bfd3ba9fd3501660156c71ff1c8129c/neutron_lib/api/definitions/network_mtu_writable.py#L54..L56) doesn't support. Filed a [bug report](https://bugs.launchpad.net/neutron/+bug/1818317) and disabled the listing.

kayrus:
@jtopjian resolved conflicts

acceptance/openstack/networking/v2/extensions/mtu/mtu_test.go
```diff
@@ -0,0 +1,136 @@
...
+
+func TestMTUNetworkCRUDL(t *testing.T) {
+	clients.RequireAdmin(t)
+
+	client, err := clients.NewNetworkV2Client()
+	th.AssertNoErr(t, err)
+
+	extension, err := extensions.Get(client, "net-mtu").Extract()
+	if err != nil {
+		t.Skip("This test requires dns-integration Neutron extension")
```
jtopjian: nit: `s/dns-integration/mtu`

kayrus:
> Additionally, my personal test environment has both mtu and mtu-writeable enabled. With both enabled, listing works. I wonder if both extensions are supposed to be used together?

Probably, then there should be a request to OpenLab team.

> I recommend changing the mtu from 1450 to something non-standard like 1449. That way you can remove filtering by name+mtu and only filter on the non-standard mtu number. I tried this and it works.

ok, let's try.

acceptance/openstack/networking/v2/extensions/mtu/mtu_test.go
kayrus: done

kayrus:
recheck

jtopjian:
recheck

jtopjian:
The failure is due to a panic caused by the code.

The good news is that the code is reporting that both extensions are loaded - so it's still possible for this to work.

kayrus:
Well, filtering doesn't work:

```
convenience.go:35: Failure in mtu_test.go, line 68: unexpected error "Bad request with: [GET http://192.168.1.13:9696/v2.0/networks?mtu=1449], error message: {\"NeutronError\": {\"message\": \"[u'mtu'] is invalid attribute for filtering\", \"type\": \"HTTPBadRequest\", \"detail\": \"\"}}"
```

I'll keep the condition.

kayrus:
@jtopjian kindly ping

jtopjian:
Testing with filtering was worth a shot - thank you! I guess the filtering bug must have been recently introduced since it's working for me.

Issue #1388, opened 2019-01-09
Compute V2: support for strings hypervisors IDs


ozerovandrei:
OpenStack Compute V2 API Hypervisors response contains string representation for id field starting from 2.53 microversion: [2.53 os-hypervisors](https://docs.openstack.org/nova/latest/reference/api-microversion-history.html#maximum-in-pike).

Currently Gophercloud supports only int representation for this field: [hypervisors/results.go#L75](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/hypervisors/results.go#L75) and it should be fixed.

[API reference](https://developer.openstack.org/api-ref/compute/?expanded=#list-hypervisors)
[Gophercloud Microversions discussion](https://github.com/gophercloud/gophercloud/issues/441)

jtopjian:
Implemented in #1500

PR #1499, opened 2019-03-09
Compute V2: add support for hypervisor string UUID

ozerovandrei:
This commit adds support for the Compute API 2.53 microversion
hypervisors implementation by adding string UUIDs in the basic and
uptime hypervisor result structure.

It also updates unit and acceptance tests and extends the hypervisor
package documentation.

For #1388

code in this PR:

https://github.com/openstack/nova/blob/stable/rocky/nova/api/openstack/compute/hypervisors.py#L60
https://github.com/openstack/nova/blob/stable/rocky/nova/api/openstack/compute/hypervisors.py#L75
https://github.com/openstack/nova/blob/stable/rocky/nova/api/openstack/compute/hypervisors.py#L252


ozerovandrei:
recheck

jtopjian:
Implemented as part of #1500

Issue #1370, opened 2018-12-20
Compute V2: support for strings services IDs


ozerovandrei:
OpenStack Compute V2 API Services response contains string representation for `id` field starting from `2.53` microversion: [2.53 os-services](https://docs.openstack.org/nova/latest/reference/api-microversion-history.html#maximum-in-pike).

Currently Gophercloud supports only int representation for this field: [services/results.go#L23](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/services/results.go#L23) and it should be fixed.

[API reference](https://developer.openstack.org/api-ref/compute/?expanded=list-compute-services-detail#list-compute-services)
[Gophercloud Microversions discussion](https://github.com/gophercloud/gophercloud/issues/441)

jtopjian:
Implemented in #1500

PR #1500, opened 2019-03-09
Compute V2: add support for service string UUID

ozerovandrei:
This commit adds suport Compute API 2.53 microversion
services implementation by adding string UUIDs in the basic service
result structure.

It also updates unit tests.

For #1370 

https://github.com/openstack/nova/blob/stable/rocky/nova/api/openstack/compute/services.py#L84

### DIFF
--- a/openstack/compute/v2/extensions/hypervisors/doc.go
+++ b/openstack/compute/v2/extensions/hypervisors/doc.go
@@ -12,7 +12,7 @@ Example of Show Hypervisor Details
 
 	fmt.Printf("%+v\n", hypervisor)
 
-Example of Show Hypervisor Details with Compute API microversion older that 2.53
+Example of Show Hypervisor Details with Compute API microversion greater that 2.53
 
     hypervisorID := "c48f6247-abe4-4a24-824e-ea39e108874f"
     hypervisor, err := hypervisors.Get(computeClient, hypervisorID).Extract()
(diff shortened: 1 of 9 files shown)


ozerovandrei:
recheck

openstack/compute/v2/extensions/hypervisors/doc.go
```diff
@@ -4,14 +4,24 @@ and shows summary statistics for all hypervisors over all compute nodes in the O
...
-	hypervisor, err := hypervisors.Get(computeClient, 42).Extract()
+	hypervisorID := "42"
+	hypervisor, err := hypervisors.Get(computeClient, hypervisorID).Extract()
 	if err != nil {
 		panic(err)
 	}
 
 	fmt.Printf("%+v\n", hypervisor)
 
+Example of Show Hypervisor Details with Compute API microversion older that 2.53
```
jtopjian: nit: let's word it as "microversion greater than 2.53"

openstack/compute/v2/extensions/hypervisors/doc.go
```diff
@@ -37,15 +47,24 @@ Example of Show Hypervisor Statistics
 
 	fmt.Printf("%+v\n", hypervisorsStatistics)
 
-Example of Show Hypervisor Uptime
+Example of Show Hypervisor Uptime with Compute API microversion older that 2.53
```
jtopjian: nit: let's word it as "microversion greater than 2.53"

PR #1509, opened 2019-03-14
Change UpdatedAt and CreatedAt to time.Time

iurygregory:
For #1429

This PR updates the type of the variables `UpdatedAt` and `CreatedAt` to time.Time
and adds the `UnmarshalJSON` function

### DIFF
--- a/openstack/baremetal/v1/ports/results.go
+++ b/openstack/baremetal/v1/ports/results.go
@@ -1,6 +1,9 @@
 package ports
 
 import (
+	"encoding/json"
+	"time"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
@@ -59,17 +62,37 @@ type Port struct {
 	Extra map[string]interface{} `json:"extra"`
 
 	// The UTC date and time when the resource was created, ISO 8601 format.
-	CreatedAt string `json:"created_at"`
+	CreatedAt time.Time `json:"-"`
 
 	// The UTC date and time when the resource was updated, ISO 8601 format.
 	// May be “null”.
-	UpdatedAt string `json:"updated_at"`
+	UpdatedAt time.Time `json:"-"`
 
 	// A list of relative links. Includes the self and bookmark links.
 	Links []interface{} `json:"links"`
 
 	// Indicates whether the Port is a Smart NIC port.
 	IsSmartNIC bool `json:"is_smartnic"`
+}
+
+// UnmarshalJSON to override default
+func (r *Port) UnmarshalJSON(b []byte) error {
+	type tmp Port
+	var s struct {
+		tmp
+		CreatedAt gophercloud.JSONRFC3339MilliNoZ `json:"created_at"`
+		UpdatedAt gophercloud.JSONRFC3339MilliNoZ `json:"updated_at"`
+	}
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*r = Port(s.tmp)
+
+	r.CreatedAt = time.Time(s.CreatedAt)
+	r.UpdatedAt = time.Time(s.UpdatedAt)
+
+	return nil
 }
 
 // PortPage abstracts the raw results of making a List() request against
--- a/openstack/baremetal/v1/ports/results.go
+++ b/openstack/baremetal/v1/ports/results.go
@@ -80,8 +80,8 @@ func (r *Port) UnmarshalJSON(b []byte) error {
 	type tmp Port
 	var s struct {
 		tmp
-		CreatedAt gophercloud.JSONRFC3339MilliNoZ `json:"created_at"`
-		UpdatedAt gophercloud.JSONRFC3339MilliNoZ `json:"updated_at"`
+		CreatedAt gophercloud.JSONRFC3339ZNoT `json:"created_at"`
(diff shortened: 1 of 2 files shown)


iurygregory:
@jtopjian what would be the correct parser for the ISO8061? The API returns an string like "2019-02-15T09:52:24+00:00" not sure how I would handle the parser, any ideas? =)

jtopjian:
@iurygregory The cause of the errors is that you need to create the time fixture as:

```diff
- fooCreated, _ = time.Parse(gophercloud.JSONRFC3339ZNoT, "2019-02-15T09:52:24+00:00")
+ fooCreated, _ = time.Parse(gophercloud.RFC3339ZNoT, "2019-02-15T09:52:24+00:00")
```

`gophercloud.RFC3339ZNoT` is the string constant that defines the time format. It's a little confusing and I'm sorry about that.

But I've got some good news: The time format that Ironic is using is a "proper" RFC3339 format. You don't see this much (I just assumed it wasn't the standard format and never checked). This means that we don't need any custom `UnmarshalJSON` method at all. 

Instead, we can just use Go's standard time parsing and everything will just work. Try the following changes:

```diff
diff --git a/openstack/baremetal/v1/ports/results.go b/openstack/baremetal/v1/ports/results.go
index fe53455..506b6c6 100644
--- a/openstack/baremetal/v1/ports/results.go
+++ b/openstack/baremetal/v1/ports/results.go
@@ -1,7 +1,6 @@
 package ports

 import (
-       "encoding/json"
        "time"

        "github.com/gophercloud/gophercloud"
@@ -62,11 +61,11 @@ type Port struct {
        Extra map[string]interface{} `json:"extra"`

        // The UTC date and time when the resource was created, ISO 8601 format.
-       CreatedAt time.Time `json:"-"`
+       CreatedAt time.Time `json:"created_at"`

        // The UTC date and time when the resource was updated, ISO 8601 format.
        // May be “null”.
-       UpdatedAt time.Time `json:"-"`
+       UpdatedAt time.Time `json:"updated_at"`

        // A list of relative links. Includes the self and bookmark links.
        Links []interface{} `json:"links"`
@@ -75,26 +74,6 @@ type Port struct {
        IsSmartNIC bool `json:"is_smartnic"`
 }

-// UnmarshalJSON to override default
-func (r *Port) UnmarshalJSON(b []byte) error {
-       type tmp Port
-       var s struct {
-               tmp
-               CreatedAt gophercloud.JSONRFC3339ZNoT `json:"created_at"`
-               UpdatedAt gophercloud.JSONRFC3339ZNoT `json:"updated_at"`
-       }
-       err := json.Unmarshal(b, &s)
-       if err != nil {
-               return err
-       }
-       *r = Port(s.tmp)
-
-       r.CreatedAt = time.Time(s.CreatedAt)
-       r.UpdatedAt = time.Time(s.UpdatedAt)
-
-       return nil
-}
-
 // PortPage abstracts the raw results of making a List() request against
 // the API.
 type PortPage struct {
diff --git a/openstack/baremetal/v1/ports/testing/fixtures.go b/openstack/baremetal/v1/ports/testing/fixtures.go
index cd1e3d8..dafeed8 100644
--- a/openstack/baremetal/v1/ports/testing/fixtures.go
+++ b/openstack/baremetal/v1/ports/testing/fixtures.go
@@ -6,7 +6,6 @@ import (
        "testing"
        "time"

-       "github.com/gophercloud/gophercloud"
        "github.com/gophercloud/gophercloud/openstack/baremetal/v1/ports"
        th "github.com/gophercloud/gophercloud/testhelper"
        "github.com/gophercloud/gophercloud/testhelper/client"
@@ -138,10 +137,10 @@ const SinglePortBody = `
 `

 var (
-       fooCreated, _ = time.Parse(gophercloud.JSONRFC3339ZNoT, "2019-02-15T09:52:24+00:00")
-       fooUpdated, _ = time.Parse(gophercloud.JSONRFC3339ZNoT, "2019-02-15T09:55:19+00:00")
-       BarCreated, _ = time.Parse(gophercloud.JSONRFC3339ZNoT, "2019-02-15T09:52:23+00:00")
-       BarUpdated, _ = time.Parse(gophercloud.JSONRFC3339ZNoT, "2019-02-15T09:55:19+00:00")
+       fooCreated, _ = time.Parse(time.RFC3339, "2019-02-15T09:52:24+00:00")
+       fooUpdated, _ = time.Parse(time.RFC3339, "2019-02-15T09:55:19+00:00")
+       BarCreated, _ = time.Parse(time.RFC3339, "2019-02-15T09:52:23+00:00")
+       BarUpdated, _ = time.Parse(time.RFC3339, "2019-02-15T09:55:19+00:00")
        PortFoo       = ports.Port{
                UUID:                "f2845e11-dbd4-4728-a8c0-30d19f48924a",
                NodeUUID:            "ddd06a60-b91e-4ab4-a6e7-56c0b25b6086",
```

What I've done is removed the custom `UnmarshalJSON` method and changed the time fields' JSON tag back to `created_at` and `updated_at` (but the types are kept as `time.Time`). This makes Go process and parse the time fields in the internal JSON unmarshal method.

In a perfect world, all JSON timestamps would look like how Ironic has formatted them and no custom `UnmarshalJSON` method would need to be used. But, for a variety of legitimate reasons, different formats are used and so when the "standard" format is used, it's actually an outlier :)

iurygregory:
recheck

iurygregory:
@jtopjian Thank you very much for the help =D

Issue #1515, opened 2019-03-21
Support Magnum cluster resize API


openstacker:
Recently in Magnum we introduced a new API v1/clusters/clusterID/actions/resize [1] to enable the capability that consumer can resize the node count of the cluster and meanwhile the API allows to pass in the node instance ID so as to delete particular nodes. That's very useful for the auto-healing and auto-scaling cases which Magnum will leverage k8s cluster Autoscaler to achieve that.

Before that, the consumer will have to call Heat API firstly to remove the particular to remove the nodes and then call Magnum API again to update the node count, which is not ideal obviously. Hence the new API was proposed. More information please refer [2] [3].

[1] https://review.openstack.org/638572
[2] https://review.openstack.org/631378
[3] https://github.com/kubernetes/autoscaler/pull/1690

PR #1516, opened 2019-03-21
Support Magnum resize API

openstacker:
Recently in Magnum we introduced a new API v1/clusters/clusterID/actions/resize [1]
to enable the capability that consumer can resize the node count of the
cluster and meanwhile the API allows to pass in the node instance ID
so as to delete particular nodes. That's very useful for the auto-healing
and auto-scaling cases which Magnum will leverage k8s cluster Autoscaler
to achieve that.

Without this new API, the consumer has to call Heat API firstly to remove
the particular to remove the nodes and then call Magnum API again to
update the node count, which is not ideal obviously. Hence the new API
was proposed. More information please refer [2] [3].

For #1515

[1] https://review.openstack.org/638572
[2] https://review.openstack.org/631378
[3] kubernetes/autoscaler#1690

### DIFF
--- a/openstack/containerinfra/v1/clusters/requests.go
+++ b/openstack/containerinfra/v1/clusters/requests.go
@@ -175,3 +175,40 @@ func Update(client *gophercloud.ServiceClient, id string, opts []UpdateOptsBuild
 	}
 	return
 }
+
+// ResizeOptsBuilder allows extensions to add additional parameters to the
+// Resize request.
+type ResizeOptsBuilder interface {
+	ToClusterResizeMap() (map[string]interface{}, error)
+}
+
+// ResizeOpts params
+type ResizeOpts struct {
+	NodeCount     *int     `json:"node_count" required:"true"`
+	NodesToRemove []string `json:"nodes_to_remove,omitempty"`
+	NodeGroup     string   `json:"nodegroup,omitempty"`
+}
+
+// ToClusterResizeMap constructs a request body from ResizeOpts.
+func (opts ResizeOpts) ToClusterResizeMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "")
+}
+
+// Resize an existing cluster node count.
+func Resize(client *gophercloud.ServiceClient, id string, opts ResizeOptsBuilder) (r ResizeResult) {
+	b, err := opts.ToClusterResizeMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+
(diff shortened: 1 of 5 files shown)


jtopjian:
@openstacker Thanks! This looks good to me. I understand that doing acceptance tests for Magnum can be a little hard because of the amount of resources and time is required. Can you confirm that you've tested this with an actual cluster?

openstacker:
Hi @jtopjian  Yep, I have tested this with a real cluster and it worked for me. After this, I'm going to leverage this in Cluster Autoscaler and that will test this again and I'm happy to fix bugs if there is any. Thanks for reviewing.

openstack/containerinfra/v1/clusters/requests.go
```diff
@@ -175,3 +175,40 @@ func Update(client *gophercloud.ServiceClient, id string, opts []UpdateOptsBuild
...
+// ResizeOptsBuilder allows extensions to add additional parameters to the
+// Resize request.
+type ResizeOptsBuilder interface {
+	ToClusterResizeMap() (map[string]interface{}, error)
+}
+
+// ResizeOpts params
+type ResizeOpts struct {
+	NodeCount     *int     `json:"node_count" required:"true"`
+	NodesToRemove []string `json:"discovery_url,omitempty"`
```
ozerovandrei: It seems that it should have the `json:"nodes_to_remove,omitempty"` tag given changes from [#638572](https://review.openstack.org/#/c/638572/22/magnum/api/controllers/v1/cluster_actions.py)

ozerovandrei: Maybe it's also better to provider an additional unit test for this parameter?

jtopjian: > It seems that it should have the json:"nodes_to_remove,omitempty"

Good catch - thank you!

openstacker: @ozerovandrei  Yes, the NodesToRemove is also optional. Thanks.

jtopjian: @openstacker Just to clarify, @ozerovandrei meant that the JSON tag `discovery_url` does not match `NodesToRemove`. `discovery_url` needs to be changed to `nodes_to_remove`.

openstacker: Ah, yes, sorry, that's a copy-paste issue. I just fixed it. Thanks.

PR #1517, opened 2019-03-22
Fix OK codes in keypair creation

pamburus:
Added code 201 (accepted) to ok codes in keypair creation.
Without this fix, keypair creation fails with compute microversion "2.60" because it gets status 201.

For #1518

### DIFF
--- a/openstack/compute/v2/extensions/keypairs/requests.go
+++ b/openstack/compute/v2/extensions/keypairs/requests.go
@@ -68,7 +68,7 @@ func Create(client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r Create
 		return
 	}
 	_, r.Err = client.Post(createURL(client), b, &r.Body, &gophercloud.RequestOpts{
-		OkCodes: []int{200},
+		OkCodes: []int{200, 201},
 	})
 	return
 }


jtopjian:
Hi @pamburus! Thank you for submitting this.

> Without this fix, keypair creation fails with compute microversion "2.60" because it gets status 201.

Can you point to the API code which enforces this restriction?

pamburus:
> Hi @pamburus! Thank you for submitting this.
> 
> > Without this fix, keypair creation fails with compute microversion "2.60" because it gets status 201.
> 
> Can you point to the API code which enforces this restriction?

Which restriction do you mean?

Here is the change in API: since version 2.2 statuses are 200 and 201
https://developer.openstack.org/api-ref/compute/?expanded=create-server-detail,create-or-import-keypair-detail#create-or-import-keypair

jtopjian:
In order to implement API communication changes, we need to see where in the Nova API python code (https://github.com/openstack/nova) a 201 is returned.

The [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial) has a section on this here: https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md

pamburus:
> In order to implement API communication changes, we need to see where in the Nova API python code (https://github.com/openstack/nova) a 201 is returned.
> 
> The [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial) has a section on this here: https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md

Here is the code in nova:
https://github.com/openstack/nova/blob/f5c8c4f42e4a8c5bfa3a872c0cf010d9dceb059b/nova/api/openstack/compute/keypairs.py#L76

jtopjian:
> Here is the code in nova:

Yep - perfect. Thanks!

PR #1519, opened 2019-03-22
Acc Tests: Re-enable TestVolumeActionsUploadImageDestroy

jtopjian:
For https://github.com/theopenlab/openlab/issues/222

### DIFF
--- a/acceptance/openstack/blockstorage/extensions/volumeactions_test.go
+++ b/acceptance/openstack/blockstorage/extensions/volumeactions_test.go
@@ -14,8 +14,6 @@ import (
 )
 
 func TestVolumeActionsUploadImageDestroy(t *testing.T) {
-	t.Skip("This test is diabled because it sometimes fails in OpenLab")
-
 	blockClient, err := clients.NewBlockStorageV2Client()
 	th.AssertNoErr(t, err)
 


bzhaoopenstack:
recheck

bzhaoopenstack:
recheck

bzhaoopenstack:
Hi @jtopjian,
I login the test VM, I found the whole process in API layer is OK, but the issue happened during upload image through a cinder volume. The whole worklist will reach swift, (as it's a glance backend). The swift service will fail as No space on the device. 
.0.1 - - [25/Mar/2019:07:23:05 +0000] "PUT /sdb1/102/AUTH_4f74106463ab458e8ea34e5f5527c3b4/glance/bb2e8bc6-1cff-49ca-b14e-01b1466ca652-00005" 201 - "PUT http://192.168.0.22
 __call__ error with PUT /sdb1/402/AUTH_4f74106463ab458e8ea34e5f5527c3b4/glance/bb2e8bc6-1cff-49ca-b14e-01b1466ca652-00006 : [Errno 28] No space left on device (txn: tx90bd
 object-server: ERROR __call__ error with PUT /sdb1/402/AUTH_4f74106463ab458e8ea34e5f5527c3b4/glance/bb2e8bc6-1cff-49ca-b14e-01b1466ca652-00006 : [Errno 28] No space left o

So I check the local space usage like below.
root@ubuntu-xenial-chameleon-openlab-0000008346:/opt/stack/new/devstack# df --human
Filesystem      Size  Used Avail Use% Mounted on
udev            3.9G     0  3.9G   0% /dev
tmpfs           799M  1.3M  798M   1% /run
/dev/vda1        76G   22G   51G  31% /
tmpfs           3.9G   72K  3.9G   1% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/sr0        410K  410K     0 100% /mnt/config
tmpfs           799M     0  799M   0% /run/user/1000
tmpfs           512M  123M  390M  24% /opt/stack/data/etcd
/dev/loop0      2.0G  1.1G 1002M  51% /opt/stack/data/swift/drives/sdb1

As I reproduce the issue, I create a size = 1(1G) volume for test, but we can see the last line(which the device used by swift) is only available less than 1G, so the upload image process will fail. The good way to fix it is increase the device size which used by swift. I just found the configuration in swift(/etc/swift/object-server/1.conf), and you can set the "SWIFT_LOOPBACK_DISK_SIZE" in devstack local.conf for fixing it. Thanks.

huangtianhua:
@jtopjian Sorry to disturb, I opened a issue https://github.com/theopenlab/openlab/issues/302  seems some feature are not support in stable/pike, so it better to skip the tests if the stable branch of openstack is not support?

jtopjian:
@huangtianhua 

> so it better to skip the tests if the stable branch of openstack is not support?

Most likely, yes. Please let me know which tests should be disabled and I can review.

huangtianhua:
@jtopjian you can see the jobs log, and there are some tests you have to looking into, and maybe you can open an issue to fix it?

jtopjian:
Thanks. I'll take a look.

jtopjian:
recheck

Issue #1523, opened 2019-03-25
Support listing loadbalancer API versions


MaysaMacedo:
The OpenStack Octavia API includes, in its 2.5 version, the support for tagging octavia resources.
When trying to consume this new API support, already added in Gophercloud, the version of Octavia API needs to be verified, because if tags is not supported in the applied version, another filtering solution must be used.

This issue is about adding apiversions package for load-balancer described in here:
`
https://developer.openstack.org/api-ref/load-balancer/?expanded=#`

jtopjian:
@MaysaMacedo You're in luck 🙂 

This PR should help you with this: https://github.com/gophercloud/gophercloud/pull/1521

lingxiankong:
Hi @MaysaMacedo, for double check, if you don't specify tags param when creating the resource, do you still see the error?

MaysaMacedo:
Thank you, @jtopjian  :)

Hi @lingxiankong, I've only had the opportunity to test in an env that octavia API did not have tag support. Thus, when creating a resource and then filtering with tags to delete specific resources, would result in an error, which I believe to be the expected behavior.

ozerovandrei:
Implemented in #1524, thank you @MaysaMacedo !

PR #1524, opened 2019-03-25
Add support to Load Balancer api version requests

MaysaMacedo:
This PR adds apiversion package for load balancer resources, enabling API consumers to discover available versions. 

For #1523

[Octavia API version support ](https://developer.openstack.org/api-ref/load-balancer/?expanded=#)

### DIFF
--- a/openstack/loadbalancer/v2/apiversions/requests.go
+++ b/openstack/loadbalancer/v2/apiversions/requests.go
@@ -5,7 +5,7 @@ import (
 	"github.com/gophercloud/gophercloud/pagination"
 )
 
-// lists all the load balancer API versions available to end-users.
+// List lists all the load balancer API versions available to end-users.
 func List(c *gophercloud.ServiceClient) pagination.Pager {
 	return pagination.NewPager(c, apiVersionsURL(c), func(r pagination.PageResult) pagination.Page {
 		return APIVersionPage{pagination.SinglePageBase(r)}
--- a/openstack/loadbalancer/v2/apiversions/requests.go
+++ b/openstack/loadbalancer/v2/apiversions/requests.go
@@ -1,0 +1,13 @@
+package apiversions
+
+import (
+	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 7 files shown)


openstack/loadbalancer/v2/apiversions/requests.go
```diff
@@ -0,0 +1,13 @@
+package apiversions
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// ListVersions lists all the load balancer API versions available to end-users.
+func ListVersions(c *gophercloud.ServiceClient) pagination.Pager {
```
ozerovandrei: It's not in any docs or anything but it seems that it will be better if we try to use the common `List` name for this particular exported function.

We already have it in [sharedfilesystems](https://github.com/gophercloud/gophercloud/blob/master/openstack/sharedfilesystems/apiversions/requests.go#L9), [containerinfra](https://github.com/gophercloud/gophercloud/blob/master/openstack/containerinfra/apiversions/requests.go#L9) packages.

MaysaMacedo:
Thanks for the review, @ozerovandrei .

openstack/loadbalancer/v2/apiversions/requests.go
```diff
@@ -0,0 +1,13 @@
+package apiversions
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// lists all the load balancer API versions available to end-users.
```
ozerovandrei: This comment should start with `List` so it will look good on godoc.org:
`// List lists all the API versions available to end-users.`

MaysaMacedo:
@ozerovandrei right, thank you.

MaysaMacedo:
recheck

jtopjian:
@MaysaMacedo It looks like OpenLab is having a problem at the moment. You can ignore the failures. In addition, this PR is not adding any acceptance tests, so there's nothing new being tested.

PR #1527, opened 2019-03-27
Keymanager: add "Expiration" into the secrets CreateOpts

### LINKED ISSUE #1526: Keymanager: add "Expiration" into the secrets CreateOpts
It is useful to control the expiration date of the certificate, therefore I suggest to add an Expiration field into the barbican secret resource. refs to https://github.com/terraform-providers/terraform-provider-openstack/pull/650
The fix is very simple and I already verified it works.

source code refs: https://github.com/openstack/barbican/blob/4c0ddda941289fba8e5ec4341b5d02d155d46162/barbican/objects/secret.py#L32
https://github.com/openstack/barbican/blob/4c0ddda941289fba8e5ec4341b5d02d155d46162/barbican/common/validators.py#L162
https://github.com/openstack/barbican/blob/4c0ddda941289fba8e5ec4341b5d02d155d46162/barbican/api/controllers/secrets.py#L424
https://github.com/openstack/barbican/blob/4c0ddda941289fba8e5ec4341b5d02d155d46162/barbican/api/controllers/secrets.py#L435
https://github.com/openstack/barbican/blob/4c0ddda941289fba8e5ec4341b5d02d155d46162/barbican/api/controllers/secrets.py#L442
https://github.com/openstack/barbican/blob/4c0ddda941289fba8e5ec4341b5d02d155d46162/barbican/plugin/resources.py#L359..L361
https://github.com/openstack/barbican/blob/4c0ddda941289fba8e5ec4341b5d02d155d46162/barbican/model/repositories.py#L391

The `secret_model` is passed through the function with the `expiration` field.

kayrus:
Resolves #1526

P.S. @jtopjian  Probably I'm mistaken, but the `Payload` element in the `UpdateOpts` is useless.

UPD. Yes, I'm mistaken. Left the payload as is.

### DIFF
--- a/openstack/keymanager/v1/secrets/requests.go
+++ b/openstack/keymanager/v1/secrets/requests.go
@@ -223,6 +223,9 @@ type CreateOpts struct {
 
 	// SecretType is the type of secret.
 	SecretType SecretType `json:"secret_type,omitempty"`
+
+	// If set, the secret will not be available after this time.
+	Expiration string `json:"expiration,omitempty"`
 }
 
 // ToSecretCreateMap formats a CreateOpts into a create request.


jtopjian:
I'm in favor of this, certainly, but two things:

1. Can we use `time.Time` as the field? 
2. Either in discussion here or in #1526, we need to dig up the API code that allows this (which might help answer the above).

kayrus:
> 1.

I referred to the application credential implementation. Probably you've missed that there was a string field, and you merged these changes: https://github.com/gophercloud/gophercloud/blob/2e9e55040ce562d7c410e677a2ed93c7e2bb973a/openstack/identity/v3/applicationcredentials/requests.go#L70

I can change the type in both packages if you wish.

2. I've added more references in the #1526.

jtopjian:
> Probably you've missed that there was a string field, and you merged these changes:

Yes, totally my mistake on that. We should get that changed to `time.Time` at some point.

Issue #1529, opened 2019-03-27
Fix API Versions Packages


jtopjian:
There has been a lot of activity around implementing `apiversions` in different services. Coincidentally I implemented the Shared Filesystem `apiversions` package in Terraform and ran into an issue.

The URL to query is generated incorrectly. It assumes a certain format when the endpoint URL could be one of [several different formats](https://github.com/gophercloud/gophercloud/blob/master/openstack/utils/testing/base_endpoint_test.go).

Fortunately we already went through this to determine the base Identity Endpoint and created a [reusable function](https://github.com/gophercloud/gophercloud/blob/master/openstack/utils/base_endpoint.go). This fix has been confirmed here: https://github.com/terraform-providers/terraform-provider-openstack/pull/715#issuecomment-477344827.

I'll open a PR to implement this fix across the various `apiversions` packages.

jtopjian:
Fixed in #1530

PR #1530, opened 2019-03-28
Fix apiversions base URLs

jtopjian:
For #1529 

This commit updates the `apiversions` packages to account for a variety of base URLs which have been known to exist in the Keystone catalog: https://github.com/gophercloud/gophercloud/blob/master/openstack/utils/testing/base_endpoint_test.go

These packages should also be housed under a versionless directory. For example, `networking/v2/apiversions` _should_ be `networking/apiversions` since it should be possible to use this package to retrieve the API versions for _all_ major versions.

However, there is a catch: services such as Block Storage changed the format of the API results a while back. this is why `blockstorage/v1/apiversions` exists. I fell down a rabbit hole looking into this a while back (https://github.com/gophercloud/gophercloud/pull/458). I haven't reviewed all of these packages to determine if there are other occurrences like this, so for now, the packages will stay where they are.

### DIFF
--- a/acceptance/openstack/networking/v2/apiversion_test.go
+++ b/acceptance/openstack/networking/v2/apiversion_test.go
@@ -39,7 +39,7 @@ func TestAPIResourcesList(t *testing.T) {
 
 	allPages, err := apiversions.ListVersionResources(client, "v2.0").AllPages()
 	if err != nil {
-		t.Fatalf("Unable to list api version reosources: %v", err)
+		t.Fatalf("Unable to list api version resources: %v", err)
 	}
 
 	allVersionResources, err := apiversions.ExtractVersionResources(allPages)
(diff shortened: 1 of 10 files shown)


jtopjian:
@openstacker @MaysaMacedo: Total coincidence that you both contributed an `apiversions` package shortly before I discovered this bug. I recommend updating your application's Gophercloud library once this is merged.

It's very possible that you won't see this bug in environments that you have access to, but I'm confident that you will eventually get reports of failures. The following test has a list for all supported URL formats once this is merged: https://github.com/gophercloud/gophercloud/blob/master/openstack/utils/testing/base_endpoint_test.go

MaysaMacedo:
@jtopjian okay, thanks for letting us know.

jtopjian:
@ozerovandrei whoops - ok, should be good to go once OpenLab finishes.

jtopjian:
@ozerovandrei Thanks!

PR #1531, opened 2019-04-01
Baremetal Inspector noauth support

zaneb:
For #1485

Add noauth support for the baremetal introspection (ironic-inspector)
service.

This is heavily based on the equivalent noauth support already present for the main Ironic API.

Source reference:
https://github.com/openstack/ironic-inspector/tree/master/ironic_inspector

### DIFF
--- a/openstack/baremetalintrospection/noauth/doc.go
+++ b/openstack/baremetalintrospection/noauth/doc.go
@@ -1,0 +1,16 @@
+package noauth
+
+/*
+Package noauth provides support for noauth bare metal introspection endpoints.
+
+	Example of obtaining and using a client:
+
+	client, err := noauth.NewBareMetalIntrospectionNoAuth(noauth.EndpointOpts{
+		IronicInspectorEndpoint: "http://localhost:5050/v1/",
+	})
+	if err != nil {
+		panic(err)
+	}
+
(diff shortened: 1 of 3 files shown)


jtopjian:
This looks good to me, though I have no way of verifying it first-hand. If it works for you, I'm happy to merge this.

zaneb:
Fixed an inconsistency between the docs and the code. I can confirm that this is working for me.

openstack/baremetalintrospection/noauth/doc.go
```diff
@@ -0,0 +1,16 @@
+package noauth
```
ozerovandrei: We're using the Go packaging format when the `package something` directive goes after the `/* Package something...` comment.
See [containerinfra/v1/clusters/doc.go](https://github.com/gophercloud/gophercloud/blob/master/openstack/containerinfra/v1/clusters/doc.go) for the example.
Also there is an example in [godoc-documenting-go-code](https://blog.golang.org/godoc-documenting-go-code).

ozerovandrei: Oh, I see that it will be bad for consistency with the other baremetal stuff already merged into Gophercloud, so ignore my previous comment 😄

jtopjian: Ah, interesting - I never noticed this before. I can get this cleaned up soon.

Issue #1533, opened 2019-04-05
Add extra params for Heat list stack query


ntk148v:
GopherCloud Orchestration list stack function is [lack of parameters](https://github.com/gophercloud/gophercloud/blob/master/openstack/orchestration/v1/stacks/requests.go#L220), compare to [the original List stack API](https://developer.openstack.org/api-ref/orchestration/v1/?expanded=list-stacks-detail).

jtopjian:
Implemented in #1534

PR #1534, opened 2019-04-05
Add extra Heat list stacks function parameters

ntk148v:
For #1533 

code in this PR:

https://developer.openstack.org/api-ref/orchestration/v1/?expanded=list-stacks-detail

Signed-off-by: Kien Nguyen <kiennt2609@gmail.com>


jtopjian:
For reference, the code implementing this can be found here: https://github.com/openstack/heat/blob/37336c0faf5df57ea878e104ae6cc7fcd01509c1/heat/objects/stack.py#L137

PR #1537, opened 2019-04-06
KeyManager v1: Adding Expiration to Secrets

jtopjian:
For #1526 
Supersedes #1527 

/cc @kayrus


kayrus:
Looks like expiration date won't be printed in debug mode, because it is omitted in json (`-`). Correct me if I'm wrong.

jtopjian:
You're correct. It's because `tools.PrintResource` is just doing a simple `json.MarshalIndent`. There was discussion a few years ago about using something like spew, but we're trying really hard to limit dependencies.

There are a few acceptance tests where a second `tools.PrintResource` is used to explicitly print the omitted field:

```go
tools.PrintResource(t, secret.Expiration)
```

PR #1542, opened 2019-04-07
Remove imagePullPolicy from acceptance test

hongbin:
Zun is going to require admin privilege [1] to create resource
with this field specified. Gophercloud acceptance test should
use non-admin credential to test so we remove this field from
acceptance test.

[1] https://review.openstack.org/#/c/650585/

For #1545


hongbin:
@jtopjian ping for review

PR #1548, opened 2019-04-17
Baremetal introspection: correct handling of LLDP data

zaneb:
The LLDP data in the inventory doesn't appear as a JSON object, but as
a JSON array of [type, value] pairs such as:

[[1, "04112233aabbcc"], [2, "07373334"], ...]

(There can be multiple fields with the same type, which is why they are
not returned as a JSON object.)

There is a place where the raw LLDP data is processed into a JSON object
- under the `lldp_processed` key in the top-level `all_interfaces` list
(i.e. not in the inventory), here represented by BaseInterfaceType.

This patch corrects handling for LLDP data in the inventory (i.e. in
InterfaceType) and adds handling for it in BaseInterfaceType

For #1485 

code in this PR:

https://github.com/openstack/ironic-inspector/blob/8.2.0/ironic_inspector/test/unit/test_plugins_lldp_basic.py#L171-L196

### DIFF
--- a/openstack/baremetalintrospection/v1/introspection/results.go
+++ b/openstack/baremetalintrospection/v1/introspection/results.go
@@ -199,7 +199,7 @@ type InterfaceType struct {
 	HasCarrier  bool                   `json:"has_carrier"`
 	IPV4Address string                 `json:"ipv4_address"`
 	IPV6Address string                 `json:"ipv6_address"`
-	Lldp        map[string]interface{} `json:"lldp"`
+	LLDP        map[string]interface{} `json:"lldp"`
 	MACAddress  string                 `json:"mac_address"`
 	Name        string                 `json:"name"`
 	Product     string                 `json:"product"`
--- a/openstack/baremetalintrospection/v1/introspection/results.go
+++ b/openstack/baremetalintrospection/v1/introspection/results.go
@@ -2,6 +2,7 @@ package introspection
 
 import (
 	"encoding/json"
+	"fmt"
 	"time"
 
 	"github.com/gophercloud/gophercloud"
@@ -174,10 +175,11 @@ type Data struct {
 // Sub Types defined under Data and deeper in the structure
 
 type BaseInterfaceType struct {
-	ClientID string `json:"client_id"`
-	IP       string `json:"ip"`
-	MAC      string `json:"mac"`
-	PXE      bool   `json:"pxe"`
+	ClientID      string                 `json:"client_id"`
+	IP            string                 `json:"ip"`
+	MAC           string                 `json:"mac"`
+	PXE           bool                   `json:"pxe"`
+	LLDPProcessed map[string]interface{} `json:"lldp_processed"`
 }
 
 type BootInfoType struct {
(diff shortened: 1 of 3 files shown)


openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -174,10 +175,11 @@ type Data struct {
...
 type BaseInterfaceType struct {
-	ClientID string `json:"client_id"`
-	IP       string `json:"ip"`
-	MAC      string `json:"mac"`
-	PXE      bool   `json:"pxe"`
+	ClientID      string                 `json:"client_id"`
+	IP            string                 `json:"ip"`
+	MAC           string                 `json:"mac"`
+	PXE           bool                   `json:"pxe"`
+	LldpProcessed map[string]interface{} `json:"lldp_processed"`
```
jtopjian: nit: `LLDPProcessed`?

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -193,17 +195,22 @@ type CPUType struct {
 	ModelName    string   `json:"model_name"`
 }
 
+type LldpTlvType struct {
```
jtopjian: nit: `LLDPTLV`?

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -193,17 +195,22 @@ type CPUType struct {
...
-	MACAddress  string                 `json:"mac_address"`
-	Name        string                 `json:"name"`
-	Product     string                 `json:"product"`
-	Vendor      string                 `json:"vendor"`
+	BIOSDevName string        `json:"biosdevname"`
+	ClientID    string        `json:"client_id"`
+	HasCarrier  bool          `json:"has_carrier"`
+	IPV4Address string        `json:"ipv4_address"`
+	IPV6Address string        `json:"ipv6_address"`
+	Lldp        []LldpTlvType `json:"lldp"`
```
jtopjian: nit: `LLDP`

I realize this was originally `Lldp`, but it was missed during a prior review. Acronyms should be capitalized.

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -240,6 +247,28 @@ type SystemVendorType struct {
 	SerialNumber string `json:"serial_number"`
 }
 
+// UnmarshalJson interprets an LLDP TLV [key, value] pair as an LldpTlv structure
+func (tlv *LldpTlvType) UnmarshalJSON(data []byte) error {
```
jtopjian: nit: `s/tlv/r` 

method receivers have an internal coding style of `r`

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -240,6 +247,28 @@ type SystemVendorType struct {
...
+	var list []interface{}
+	if err := json.Unmarshal(data, &list); err != nil {
+		return err
+	}
+
+	fieldtype, ok := list[0].(float64)
```
jtopjian: It might be a good idea to do a length check first. Maybe something like:

```go
if len(list) != 2 {
    // either return nil and let the field be empty
    // or return an error stating that the JSON structure was incorrect.
    // I'd opt for the former, but up to you.
}
```

openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -240,6 +247,28 @@ type SystemVendorType struct {
...
+	if !ok {
+		return errors.New("LLDP TLV key is not number")
```
jtopjian: nit: `fmt.Errorf` instead of `errors.New`. (There are a handful of `errors.New` in Gophercloud, but they should be changed to `fmt.Errorf` at some point.)

The two are virtually identical - it's more just an internal coding style.

openstack/baremetalintrospection/v1/introspection/results.go
dtantsur: Yeah, if we do `ID`, we should do `LLDP`

openstack/baremetalintrospection/v1/introspection/results.go
zaneb: Oh, cool. I prefer that but was trying to match the existing code. Will switch it to LLDP.

zaneb:
Thanks for the review! I agree with all your comments and will update the patch today.

Issue #1547, opened 2019-04-09
Add additional fields at struct UpdateOpts same as Volume struct


unixzen:
Before starting a PR, please read our [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial).

unixzen:
Hello!

I try make snapshot of volume and get error message

```
error message: {"badRequest": {"message": "Invalid volume: Volume 91cedfd1-1d8b-4d62-bdb1-4414a21e01db status must be available, but current status is: in-use.", "code": 400}}
```

But i try change status of volume to available for making snapshot, but i get another error:

```
./snapshot_create.go:38:41: unknown field 'Status' in struct literal of type volumes.UpdateOpts 
```
I think that problem
https://godoc.org/github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes#UpdateOpts

At struct UpdatOpts there are 3 fields, but at the same time at Volume struct about 20. Can you fix this? Or may be i can try make PR. Thanks!

jtopjian:
The block storage API only allows a subset of fields to be updated: https://developer.openstack.org/api-ref/block-storage/v3/index.html?expanded=update-a-volume-detail#update-a-volume

If your environment requires you to detach a volume before snapshotting, perhaps the [`volumeattach`](https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions/volumeattach) package can help?

unixzen:
At my environment i can't detach volume. I found flag force (https://godoc.org/github.com/gophercloud/gophercloud/openstack/blockstorage/v2/snapshots#CreateOpts), i could be using it. But i saw that PR (https://github.com/gophercloud/gophercloud/pull/960) still not merge to master branch. And it means gophercloud still not support cinder backup possibility?

unixzen:
> At my environment i can't detach volume. I found flag force (https://godoc.org/github.com/gophercloud/gophercloud/openstack/blockstorage/v2/snapshots#CreateOpts), i could be using it. But i saw that PR (#960) still not merge to master branch. And it means gophercloud still not support cinder backup possibility?

What about this @jtopjian :point_up: , i mean merge request about backup possibility?

jtopjian:
@unixzen I'll work on getting #960 merged. It needs to be cleaned up / reorganized as well as have acceptance tests added.

unixzen:
[...]

PR #1552, opened 2019-04-21
Block Storage: Backups

jtopjian:
For #903 
For #1547 
Supersedes #960

* https://github.com/openstack/cinder/blob/master/cinder/backup/api.py
* https://github.com/openstack/cinder/blob/master/cinder/api/schemas/backups.py
* https://github.com/openstack/cinder/blob/master/cinder/objects/backup.py


h00130372:
recheck stable/stein

jtopjian:
recheck stable/stein

jtopjian:
A note for posterity before I merge this: I had some trouble with the "update" action but the Cinder code and API docs seem to say the fields are able to be updating it, so it's probably just an issue with my testing environment. But maybe I'm wrong.

Issue #1566, opened 2019-05-14
Auth: introduce token auth


kayrus:
gophercloud should handle token auth

requirements:
* scope should be set in order to fetch catalog
* add support to `AuthOptionsFromEnv` func
* add `OS_AUTH_TYPE` env support

PR #1567, opened 2019-05-14
Don't print token on token auth method

kayrus:
it is a part of the #1566

this PR prevents token to be printed. Same should be done for the terraform openstack provider in future.


kayrus:
@jtopjian ready for the review

Issue #1569, opened 2019-05-16
identity/v3/tokens.Get() swallows connection errors


majewsky:
PR coming up in a second.

*Edit:* Maybe to provide a bit more context, I was seeing an entirely empty `GetResult` returned from a call to `tokens.Get()`, because the endpoint I was using was not reachable from my development machine. With the fix in #1570, I'm now seeing an appropriate error in `result.Err`.

jtopjian:
Fixed in #1570

PR #1570, opened 2019-05-16
do not swallow connection errors in tokens.Get()

majewsky:
For #1569.

It might be worthwhile to check if this problem also exists in other request methods in other packages.


jtopjian:
> It might be worthwhile to check if this problem also exists in other request methods in other packages.

This looks to be the only occurrence.

jtopjian:
@majewsky Just to confirm: this PR is good to go?

majewsky:
Sure, go ahead.

PR #1571, opened 2019-05-16
Hide application credential auth while debug

kayrus:
Relates to #1567. There is no need to replicate it in terraform, it is already there.


jtopjian:
@kayrus Just to confirm: this is good to go?

kayrus:
@jtopjian yes. last time I forgot to take a look on the terraform code and compare.

Issue #1573, opened 2019-05-20
Support mark unhealthy for Heat stack resource


openstacker:
Heat does support mark the specified resource in the stack as unhealthy, then by calling Heat stack update, the resource will be rebuilt. That's a very useful feature for auto-healing. Especially for the use case that the failure needs to be detected from inside. 

[1] https://developer.openstack.org/api-ref/orchestration/v1/index.html#mark-a-resource-as-unhealthy
[2] https://github.com/openstack/heat/blob/98a6d988df1ea4825dac3dec08de10eda148fe77/heat/engine/service.py#L1914
[3] https://github.com/openstack/heat/blob/3ec69a96f6872463cf446fc242af83bcab1abe70/heat/api/openstack/v1/resources.py#L168

jtopjian:
Implemented in #1574

PR #1574, opened 2019-05-20
Support mark unhealthy action for stack resource

openstacker:
Heat supports mark a resource as unhealthy via API, which is very
useful for auto healing use case. After marked a resource unhealthy,
the status of the resource will be 'CHECK_FAILED', then the resource
will be rebuilt as long as a Heat stack update triggered.

For #1573

Here is the API reference for resource mark unhealthy:

https://developer.openstack.org/api-ref/orchestration/v1/index.html?expanded=mark-a-resource-as-unhealthy-detail#mark-a-resource-as-unhealthy


openstacker:
@lingxiankong @newptone For your reference in case you're interested.

openstacker:
@jtopjian Could you please help review this one? Thanks.

Issue #1575, opened 2019-05-20
Add parent_resource attribute for Heat stack resource


openstacker:
The attribute `parent_resource` could be helpful to go through a stack from bottom to top.

jtopjian:
Implemented in #1576

PR #1576, opened 2019-05-20
Add parent_resource attribute for Heat stack resource

openstacker:
The attribute `parent_resource` is helpful when building a reverse mapping between the resource and their parents.

[1] https://github.com/openstack/heat/blob/master/heat/rpc/api.py#L76

For #1575


openstacker:
@jtopjian And please help review this one. Cheers.

PR #1579, opened 2019-05-22
Proper port binding unset

### LINKED ISSUE #1578: Networking V2: proper update handling for the port bindings
Initial issue is described here: https://github.com/terraform-providers/terraform-provider-openstack/issues/756. The actual bug was found during the additional tests.

When port is updated and bindings are unset, there should be a way to declare whether bindings parameters should be ignored or unset.

kayrus:
Resolves #1578

### DIFF
--- a/acceptance/openstack/networking/v2/extensions/portsbinding/portsbinding.go
+++ b/acceptance/openstack/networking/v2/extensions/portsbinding/portsbinding.go
@@ -18,7 +18,7 @@ type PortWithBindingExt struct {
 
 // CreatePortsbinding will create a port on the specified subnet. An error will be
 // returned if the port could not be created.
-func CreatePortsbinding(t *testing.T, client *gophercloud.ServiceClient, networkID, subnetID, hostID string) (PortWithBindingExt, error) {
+func CreatePortsbinding(t *testing.T, client *gophercloud.ServiceClient, networkID, subnetID, hostID string, profile map[string]interface{}) (PortWithBindingExt, error) {
 	portName := tools.RandomString("TESTACC-", 8)
 	portDescription := tools.RandomString("TESTACC-PORT-DESC-", 8)
 	iFalse := false
@@ -36,6 +36,7 @@ func CreatePortsbinding(t *testing.T, client *gophercloud.ServiceClient, network
 	createOpts := portsbinding.CreateOptsExt{
 		CreateOptsBuilder: portCreateOpts,
 		HostID:            hostID,
+		Profile:           profile,
 	}
 
 	var s PortWithBindingExt
--- a/acceptance/openstack/networking/v2/extensions/portsbinding/portsbinding_test.go
+++ b/acceptance/openstack/networking/v2/extensions/portsbinding/portsbinding_test.go
@@ -31,20 +31,23 @@ func TestPortsbindingCRUD(t *testing.T) {
 
 	// Define a host
 	hostID := "localhost"
+	profile := map[string]interface{}{"foo": "bar"}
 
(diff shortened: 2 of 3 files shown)


kayrus:
@jtopjian ready for review

jtopjian:
This is an interesting situation.

Part 1:

The core reason this is happening is because of how Gophercloud is handling API extensions. We already have `omitempty` set [here](https://github.com/gophercloud/gophercloud/blob/946411b1b473c6298535ccb94a96e86f7bc34e54/openstack/networking/v2/extensions/portsbinding/requests.go#L66) but because we manually build the request body [here](https://github.com/gophercloud/gophercloud/blob/946411b1b473c6298535ccb94a96e86f7bc34e54/openstack/networking/v2/extensions/portsbinding/requests.go#L70), the JSON tags don't actually serve a purpose.

Because of how we build request bodies with API extensions, we check if something is nil or not to emulate `omitempty`.

If the portsbinding fields were part of the main port `CreateOpts`, this problem would not exist.

Part 2:

Sending both `{}` and `nil/null` have the same result.

An alternative solution to this is by making the following change in Terraform:

```diff
--- a/openstack/resource_openstack_networking_port_v2.go
+++ b/openstack/resource_openstack_networking_port_v2.go
@@ -345,13 +345,17 @@ func resourceNetworkingPortV2Create(d *schema.ResourceData, meta interface{}) er
 
                        // Convert raw string into the map
                        rawProfile := binding["profile"].(string)
-                       if len(rawProfile) > 0 {
+                       if rawProfile != "" {
                                err := json.Unmarshal([]byte(rawProfile), &profile)
                                if err != nil {
                                        return fmt.Errorf("Failed to unmarshal the JSON: %s", err)
                                }
                        }
 
+                       if len(profile) == 0 {
+                               profile = nil
+                       }
+
                        finalCreateOpts = portsbinding.CreateOptsExt{
                                CreateOptsBuilder: finalCreateOpts,
                                HostID:            binding["host_id"].(string),
```

However, that's pushing a quirk of Gophercloud to the application which I don't think is fair.

I think changing `{}` to `null` is safe to do:

https://github.com/openstack/neutron/blob/f12176c97cddae1082a0c43cec6b243aaab1420a/neutron/db/portbindings_db.py#L56-L66

But we should keep this change in mind if someone reports otherwise (in this case, the change will need to be pushed to Terraform).

@kayrus I'm in favor of this PR, but let me know if any of the above changes your mind about this implementation. I'll merge depending on your decision :)

kayrus:
@jtopjian create issue could be solved in an easier way:

```diff
@@ -341,7 +341,7 @@ func resourceNetworkingPortV2Create(d *schema.ResourceData, meta interface{}) er
        if v, ok := d.GetOkExists("binding"); ok {
                for _, raw := range v.([]interface{}) {
                        binding := raw.(map[string]interface{})
-                       profile := map[string]interface{}{}
+                       var profile map[string]interface{}
 
                        // Convert raw string into the map
                        rawProfile := binding["profile"].(string)
```

What is important is handle an update. Please have a look on [this branch](https://github.com/kayrus/terraform-provider-openstack/tree/omit-binding-profile). I tried to cover all possible use cases, e.g.
* omit profile, when it is not set or changed
* unset profile, when it is changed to be empty
* allow only json objects or null (don't allow arrays `[]`)

I might miss something and probably it doesn't require so many changes.

jtopjian:
> create issue could be solved in an easier way:

👍 

> I might miss something and probably it doesn't require so many changes.

I think these Terraform changes look good to me.

I'll go ahead and merge this.

Issue #1027, opened 2018-06-12
Network Port QoS Policy API


Twister915:
Adding implementations for the following routes to the client would be desirable:

Specifically, these routes, under network/v2.0/qos (list here, as well: https://developer.openstack.org/api-ref/network/v2/index.html#quality-of-service)

**QoS Extenstion common CRUD**
- [x] `GET /rule-types` #1030
- [x] `GET /rule-types/{rule_type}` #1625
- [x] `GET /policies` #1591
- [x] `POST /policies` #1595
- [x] `GET /policies/{policy_id}` #1593
- [x] `PUT /policies/{policy_id}` #1601
- [x] `DELETE /policies/{policy_id}` #1602
- [x] `GET /policies/{policy_id}/bandwidth_limit_rules` #1582
- [x] `POST /policies/{policy_id}/bandwidth_limit_rules` #1584
- [x] `GET /policies/{policy_id}/bandwidth_limit_rules/{rule_id}` #1583
- [x] `PUT /policies/{policy_id}/bandwidth_limit_rules/{rule_id}` #1588
- [x] `DELETE /policies/{policy_id}/bandwidth_limit_rules/{rule_id}` #1590
- [x] `GET /policies/{policy_id}/dscp_marking_rules ` #1594
- [x] `POST /policies/{policy_id}/dscp_marking_rules ` #1604
- [x] `GET /policies/{policy_id}/dscp_marking_rules/{rule_id}` #1607
- [x] `PUT /policies/{policy_id}/dscp_marking_rules/{rule_id}` #1605
- [x] `DELETE /policies/{policy_id}/dscp_marking_rules/{rule_id}` #1608
- [x] `GET /policies/{policy_id}/minimum_bandwidth_rules `  #1613
- [x] `POST /policies/{policy_id}/minimum_bandwidth_rules ` #1615
- [x] `GET /policies/{policy_id}/minimum_bandwidth_rules/{rule_id}` #1614
- [x] `PUT /policies/{policy_id}/minimum_bandwidth_rules/{rule_id}` #1622
- [x] `DELETE /policies/{policy_id}/minimum_bandwidth_rules/{rule_id}` #1623

**QoS Extension Ports CRUD**
- [x] `GET /v2.0/ports/{port_id}` #1580
- [x] `POST /v2.0/ports/{port_id}` #1580
- [x] `PUT /v2.0/ports/{port_id}` #1580

**QoS Extension Networks CRUD**
[...]

PR #1580, opened 2019-05-24
Networking V2: add QoS policy extension for Port

ozerovandrei:
Allow to Create a Port with the QoS policy.

Allow to add or delete a QoS policy from an existing Port.

Allow to retrieve a Port with QoS policy from the API.

For #1027

code in this PR:

https://github.com/openstack/neutron-lib/blob/stable/stein/neutron_lib/api/definitions/qos.py#L91


ozerovandrei:
Note: this PR doesn't introduce any new acceptance tests because we need to support common QoS CRUD to to that.

PR #1583, opened 2019-05-27
Networking V2: add bw limit rules Get

ozerovandrei:
Allow to get a specific bandwidth limit rule.

For #1027

code in this PR:

https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/qos.py#L30
https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/qos.py#L115


ozerovandrei:
Merged in #1584

Issue #1585, opened 2019-05-30
Support for Server Diagnostics


nareiber:
Nova has some diagnostic information for instances. The format and structure is dependent on nova api microversion[1], so I was just going to return a gophercloud result that callers can extract into their own structures depending on their nova api microversion. Also, my cloud's nova api is < 2.48 so I cannot test the new response in a real environment :)

Use case:  we dump stats on the computes and the instances residing on said compute into graphite/grafana to easily spot when an instance is being gluttonous and risking the computes health.

[1] https://developer.openstack.org/api-ref/compute/?expanded=show-server-diagnostics-detail

jtopjian:
@nareiber Just a heads up that I've changed the return type to a map instead of a pointer to map here: https://github.com/gophercloud/gophercloud/pull/1596/commits/e15bc592646bc9139350e5f29453fa47df0084b8

jtopjian:
Implemented in #1592 and #1596

PR #1586, opened 2019-05-30
[WIP] Diagnostics for servers

nareiber:
For #1585

I can only test with <2.48 of the nova api, so I didn't implement any custom extraction functions or properly model the response, as my available version just returns something easily extracted into a map. 

https://developer.openstack.org/api-ref/compute/?expanded=show-server-diagnostics-detail

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -810,3 +810,9 @@ func ShowConsoleOutput(client *gophercloud.ServiceClient, id string, opts ShowCo
 	})
 	return
 }
+
+// Diagnostics
+func GetServerDiagnostics(client *gophercloud.ServiceClient, serverId string) (r ServerDiagnosticsResult) {
+	_, r.Err = client.Get(serverDiagnosticsURL(client, serverId), &r.Body, nil)
+	return
+}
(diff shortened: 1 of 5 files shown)


jtopjian:
@nareiber Thank you for submitting this!

A few quick notes:

1. I think this should be a separate package under `openstack/compute/v2/extensions/diagnostics`:

```
$ nova list-extensions |  grep -i diag
```

2. The API docs are helpful, but we'll need to reference the implementation to the actual Python code. There's more information about this [here](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md)

3. 

> so I didn't implement any custom extraction functions or properly model the response, as my available version just returns something easily extracted into a map.

This is why we want to look at the actual Python code - so we can model the data properly. In theory, _anything_ can be extracted into a map. 🙂 

If you can only test the base diagnostics functionality, that's fine. We have methods of amending results for [microversions](https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md).

nareiber:
Sounds good. I'll get on it this weekend! Thanks!

nareiber:
Closing, re-opened as https://github.com/gophercloud/gophercloud/pull/1592

PR #1587, opened 2019-06-01
Networking V2: rename qos bw rules List call

ozerovandrei:
Use ListBandwidthLimitRules as a method name for the List call.

For #1027


ozerovandrei:
Merged with #1589

PR #1588, opened 2019-06-01
Networking V2: add bw limit rule Update

ozerovandrei:
For #1027

Allow to update a bandwidth limit rule.

code in this PR:

https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/qos.py#L30
https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/qos.py#L115


ozerovandrei:
Merged with #1589

PR #1592, opened 2019-06-01
Diagnostics Compute Extension

nareiber:
For #1585

The call for diagnostics comes into the nova api[1], is sent over rpc to the compute manager[2], and finally ends up depending on the driver[3]. My env uses libvirt, which builds and returns a 'single-level' python dict[4]. The response is hard to map to a struct in this scenario because there are possible extra values for extra vnics and disks.

Future work needs to be done on adding diagnostics/microversions.go to provide extract methods for a better structured, 'multi-level' dict response that is available with nova microversion 2.48 or higher[1][5], but I don't feel comfortable working on that without having it available in my environment to fully test against.

[1] https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/server_diagnostics.py#L42-L47
[2] https://github.com/openstack/nova/blob/master/nova/compute/api.py#L3681-L3683
[3] https://github.com/openstack/nova/blob/master/nova/compute/manager.py#L4930-L4935
[4] https://github.com/openstack/nova/blob/master/nova/virt/libvirt/driver.py#L9183-L9241
[5] https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/views/server_diagnostics.py

### DIFF
--- a/openstack/compute/v2/extensions/diagnostics/doc.go
+++ b/openstack/compute/v2/extensions/diagnostics/doc.go
@@ -1,0 +1,14 @@
+/*
+Package diagnostics returns details about a nova instance diagnostics
+
+Example of Show Diagnostics
+
+	diags, err := diagnostics.Get(computeClient, serverId).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Printf("%+v\n", diags)
+
+*/
+package diagnostics
--- a/openstack/compute/v2/extensions/diagnostics/requests.go
+++ b/openstack/compute/v2/extensions/diagnostics/requests.go
@@ -1,0 +1,11 @@
+package diagnostics
+
+import (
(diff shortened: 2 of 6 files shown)


jtopjian:
> https://github.com/openstack/nova/blob/master/nova/virt/libvirt/driver.py#L9183-L9241

This truly looks like a structure that we can't deterministically model. Thank you for the research here.

This PR looks good to me. Was there any additional work you wanted to do or is this ready to be merged?

nareiber:
Yeah, with the different virt platforms it's pretty gnarly stuff to work with. It's probably better to let the consumer who knows what they're using(microversion, too) struct it out if they need to. 

This is ok to merge from my end. For now, this is all the work my project needs/produced. If I do find an improvement or add more functionality I can open a PR later. Thanks for all your help btw!

PR #1601, opened 2019-06-08
Networking V2: add QoS policy Update call

ozerovandrei:
Implement QoS policy update call.

For #1027

code in this PR:

https://github.com/openstack/neutron-lib/blob/stable/stein/neutron_lib/api/definitions/qos.py#L54
https://github.com/openstack/neutron/blob/stable/stein/neutron/extensions/qos.py#L43


ozerovandrei:
Merged in #1603

PR #1603, opened 2019-06-08
Networking V2: add QoS policies acceptance tests

ozerovandrei:
Add TestPoliciesCRUD acceptance test and CreateQoSPolicy,
DeleteQoSPolicy helpers.

For #1027


ozerovandrei:
Looks like openlabtesting is down, it's better to recheck this PR after some time.

ozerovandrei:
Can't really find it in Neutron's code but Update call should return http.StatusOK (200) instead of http.StatusAccepted (202). It works similar to other Neutron extensions but its code is very difficult 😞

PR #1611, opened 2019-06-10
Baremetal introspection: Read extra_hardware data

zaneb:
If the `extra_hardware` processing hook is enabled in ironic-inspector,
then additional data is provided as a top-level object named `"extra"`,
with a number of sections, each of which is also an object containing
objects with various key/value data.

For #1485

code in this PR:

https://opendev.org/openstack/ironic-inspector/src/branch/master/ironic_inspector/plugins/extra_hardware.py

### DIFF
--- a/openstack/baremetalintrospection/v1/introspection/results.go
+++ b/openstack/baremetalintrospection/v1/introspection/results.go
@@ -171,6 +171,7 @@ type Data struct {
 	MemoryMB      int                          `json:"memory_mb"`
 	RootDisk      RootDiskType                 `json:"root_disk"`
 	Hostname      string                       `json:"hostname"`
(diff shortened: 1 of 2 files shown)


jtopjian:
This looks good to me and I trust that it works.

If it's not too much trouble, would you be able to add a unit test or amend an existing one? I think it would also serve as a form of documentation for those who dug into the tests. The only reason I raise this is because mentally building a `map[string]map[string]interface{}` in one's head might be a challenge.

> Every OpenStack API designer should be required to describe the client-side of their API in a statically typed language before proceeding with implementation.

:slightly_smiling_face:

zaneb:
> mentally building a map[string]map[string]interface{} in one's head might be a challenge

I assumed you gophers would be used to that ;)

zaneb:
recheck

Issue #1616, opened 2019-06-11
Baremetal Introspection: Add hostname to introspection data


bcrochet:
https://review.opendev.org/#/c/663991/ adds a hostname field to the introspection data. A patch is needed to expose it via Gophercloud.

jtopjian:
Implemented in #1617

PR #1617, opened 2019-06-11
Baremetal Introspection: Add hostname to introspection data

bcrochet:
A patch is being made to the Ironic Python Agent to return the hostname
that the introspected machine thinks it has. This enables Gophercloud
to pass that along.

For #1616 

https://review.opendev.org/#/c/663991/


openstack/baremetalintrospection/v1/introspection/results.go
```diff
@@ -170,6 +170,7 @@ type Data struct {
 	MACs          []string                     `json:"macs"`
 	MemoryMB      int                          `json:"memory_mb"`
 	RootDisk      RootDiskType                 `json:"root_disk"`
+	Hostname      string                       `json:"hostname,omitempty"`
```
jtopjian: nit: `omitempty` is only needed for structs that will send a JSON body in an API request. Since this is a result, it's not needed. I'll fix this post-merge.

PR #1625, opened 2019-06-13
Networking V2: add QoS rule type Get call

ozerovandrei:
Implement GetRuleType function with unit test.

Update QoS rule types documentation.

Update QoS rule types acceptance test.

Extend RuleType struct with additional fields. Unfortunately the
ParameterValues field is implemented as interface{} given it's dynamic
behavior (it can be map or slice in different situtations).

For #1027

code in this PR:

https://github.com/openstack/neutron-lib/blob/stable/stein/neutron_lib/api/definitions/qos_rule_type_details.py
https://github.com/openstack/neutron/blob/stable/stein/neutron/services/qos/drivers/manager.py#L88
https://github.com/openstack/neutron/blob/stable/stein/neutron/services/qos/qos_plugin.py#L369


ozerovandrei:
recheck

Issue #1626, opened 2019-06-14
Baremetal Introspection: Hostname is not in the correct type


bcrochet:
The Hostname field for baremetal introspection was added as a top-level field, when it should have been in the InventoryType.

jtopjian:
Fixed in #1627

PR #1627, opened 2019-06-14
Baremetal Introspection: Move hostname to InventoryType

bcrochet:
The IPA patch added the hostname to the Inventory field, not as a top-level
field. This moves the hostname into the InventoryType where it belongs.

For #1626


jtopjian:
@bcrochet Given how this was initially placed incorrectly, can you add a unit test to ensure it's correct? Please either use actual API output from a real query or base a fixture on a real query.

PR #1629, opened 2019-06-21
Skip some tests for the certain releases of OpenStack

huangtianhua:
This changes:
1. refactor the func clients.SkipRelease
2. skip some of lb and container tests which are supportted after stein

Related-Bug: theopenlab/openlab#305

### DIFF
--- a/acceptance/clients/conditions.go
+++ b/acceptance/clients/conditions.go
@@ -66,10 +66,13 @@ func RequireNovaNetwork(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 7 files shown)


jtopjian:
@huangtianhua Thank you for your work on this, but I'm not sure I want to modify the `SkipRelease` function -- I'd prefer to keep it as-is so it kind of mimics a Python decorator.

I've opened #1630 to look into the testing issues - let's see if we can try to resolve everything there?

Issue #1632, opened 2019-06-25
Easily get HTTP Error status of gophercloud errors


snigle:
Hello, in some case I want to redirect the status code of openstack to my API directly.
I don't want to maintains the enormous switch case with all errors types.
I would like a generic error with the http code. It's almost already implemented but I need an interface to cast to the generic error.

jtopjian:
Implemented in #1820

PR #1633, opened 2019-06-25
Generic error interface to get status code For #1632

snigle:
For #1632

Permit to handle gopher error easily of need to just return a status code.
Eg : 
```go
gophError := err.(gophercloud.GenericError)
fmt.Println(gophError.GetStatusCode()) 
```

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -92,6 +92,17 @@ func (e ErrUnexpectedResponseCode) Error() string {
 	return e.choseErrString()
 }
 
+// GetStatusCode returns the actual status code of the error.
+func (e ErrUnexpectedResponseCode) GetStatusCode() int {
+	return e.Actual
+}
+
+// GenericError can be used to easily get the error status code without doing switch case on all error types.
+type GenericError interface {
+	Error() string
+	GetStatusCode() int
+}
+
 // ErrDefault400 is the default error type returned on a 400 HTTP response code.
 type ErrDefault400 struct {
 	ErrUnexpectedResponseCode
--- a/testing/errors_test.go
+++ b/testing/errors_test.go
@@ -1,0 +1,24 @@
+package testing
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud"
+	th "github.com/gophercloud/gophercloud/testhelper"
+)
+
+func TestGetResponseCode(t *testing.T) {
+	respErr := gophercloud.ErrUnexpectedResponseCode{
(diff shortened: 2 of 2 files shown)


jtopjian:
@snigle Thank you for working on this.

Are you sure an interface is needed? Since all `ErrDefaultNNN` error types inherit from `ErrUnexpectedResponseCode`, then just having the `GetStatusCode` method will work without the need for an interface.

Also, would you be able to add some unit tests for this PR? Something like the following under a new file called `gophercloud/testing/errors_test.go` is a good start:

```go
package testing

import (
  "testing"

)

func TestGetResponseCode(t *testing.T) {
  respErr := gophercloud.ErrUnexpectedResponseCode{
    URL:      "http://example.com",
    Method:   "GET",
    Expected: []int{200},
    Actual:   404,
    Body:     nil,
  }

  err404 := gophercloud.ErrDefault404{respErr}

  th.AssertEquals(t, err404.GetStatusCode(), 404)
}
```

snigle:
> @snigle Thank you for working on this.
> 
> Are you sure an interface is needed? Since all `ErrDefaultNNN` error types inherit from `ErrUnexpectedResponseCode`, then just having the `GetStatusCode` method will work without the need for an interface.
> 
> Also, would you be able to add some unit tests for this PR? Something like the following under a new file called `gophercloud/testing/errors_test.go` is a good start:
> 
> ```go
> package testing
> 
> import (
>   "testing"
> 
>   "github.com/gophercloud/gophercloud"
>   th "github.com/gophercloud/gophercloud/testhelper"
> )
> 
> func TestGetResponseCode(t *testing.T) {
>   respErr := gophercloud.ErrUnexpectedResponseCode{
>     URL:      "http://example.com",
>     Method:   "GET",
>     Expected: []int{200},
>     Actual:   404,
>     Body:     nil,
>   }
> 
>   err404 := gophercloud.ErrDefault404{respErr}
> 
>   th.AssertEquals(t, err404.GetStatusCode(), 404)
> }
> ```

I need the interface because gophercloud return error interface and not ErrDefault404 type. Added the test file thanks.

jtopjian:
> I need the interface because gophercloud return error interface and not ErrDefault404 type

Right, but you can cast any error to the common `ErrUnexpectedResponseCode` which would act the same as an interface. If this won't work for you, can you provide some example code that uses the interface so I can better understand the situation?

jtopjian:
Superseded by #1820

Issue #1600, opened 2019-06-06
Compute V2: servergroups policy and rules fields


ozerovandrei:
There have been some changes for Compute V2 Server Groups API: https://docs.openstack.org/nova/stein/reference/api-microversion-history.html#id58

Gophercloud should support `policy` and optional `rules` fields in structs of the [compute/v2/extensions/servergroups](https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions/servergroups) package.

Docs: https://developer.openstack.org/api-ref/compute/#server-groups-os-server-groups

PR #1635, opened 2019-07-01
Compute V2: "policy", "rules" for servergroups GET

ozerovandrei:
Add "policy" and "rules" extract methods to servergroups package.
Those methods are available in 2.64 microversion.

For #1600

code in this PR:

https://github.com/openstack/nova/blob/stable/stein/nova/api/openstack/compute/server_groups.py#L42
https://github.com/openstack/nova/blob/stable/stein/nova/api/openstack/compute/server_groups.py#L105
https://github.com/openstack/nova/blob/stable/stein/nova/api/openstack/compute/schemas/server_groups.py#L59
https://github.com/openstack/nova/blob/stable/stein/nova/api/openstack/compute/schemas/server_groups.py#L65

### DIFF
--- a/openstack/compute/v2/extensions/servergroups/doc.go
+++ b/openstack/compute/v2/extensions/servergroups/doc.go
@@ -36,5 +36,23 @@ Example to Delete a Server Group
 	if err != nil {
 		panic(err)
 	}
+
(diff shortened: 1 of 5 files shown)


ozerovandrei:
Merged in #1636

PR #1638, opened 2019-07-04
Baremetal API: Update CleanStep struct to support manual cleaning

longkb:
Currently, the **Args** property in CleanStep struct has **map[string]string that
does not match with manual cleaning steps in Ironic. It should be dictionary in python[1][2] and interface{} in golang . For example:

BIOS setting for a node:

```
[
        {
            "interface": "bios",
            "step": "apply_configuration",
            "args": {
                "settings": [{"name": "hyper_threading_enabled", "value": "false"}]
            }
        }
]
```

This PR aims to update **Args** in **CleanStep** struct.
[1] https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/node.py#L76
[2] https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/node.py#L624

Signed-off-by: Kim Bao Long <longkb@vn.fujitsu.com>

### DIFF
--- a/openstack/baremetal/v1/nodes/requests.go
+++ b/openstack/baremetal/v1/nodes/requests.go
@@ -388,9 +388,9 @@ func GetSupportedBootDevices(client *gophercloud.ServiceClient, id string) (r Su
 // the value for ‘args’ is a keyword variable argument dictionary that is passed to the cleaning step
 // method.
 type CleanStep struct {
-	Interface string            `json:"interface" required:"true"`
-	Step      string            `json:"step" required:"true"`
-	Args      map[string]string `json:"args,omitempty"`
+	Interface string                 `json:"interface" required:"true"`
+	Step      string                 `json:"step" required:"true"`
+	Args      map[string]interface{} `json:"args,omitempty"`
 }
 
 // ProvisionStateOptsBuilder allows extensions to add additional parameters to the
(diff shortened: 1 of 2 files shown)


Yikun:
recheck

jtopjian:
@longkb Thank you for submitting this. Would you be able to read over our [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial)? 

Notably, can you link to the Ironic server-side code that allows this to be a `map[string]interface{}` instead of a `map[string]string`? And would you be able to add some unit tests to validate this change?

longkb:
@jtopjian : Thanks for your comment. 
I have update the commit message with the reference to ironic side. The **args** should be dictionary instead on string as description in [1]. Please check it :)

[1] https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/node.py#L624

jtopjian:
> From _CLEAN_STEPS_SCHEMA [1], args is declare as object type, and it is also metioned as dictionary instead on string in description[2].

That's perfect - thank you!

Technically, we should also update the unit tests to account for this change, too. The unit tests are valid, but it would be better to use a more detailed `map[string]interface{}`. I'm okay with leaving things as they are. This comment is more for posterity in case this topic comes up again.

Thank you for your work on this.

jtopjian:
recheck

longkb:
Wow, it's merged! 👍 
Thanks @jtopjian :)
I am a new comer in Gopher Cloud, so your comment is very useful for me. I hope that I will have more contribution in Gopher Cloud :) 
Thank you!

Issue #1639, opened 2019-07-11
Octavia health monitor should not require URLPath to be set for HTTP/HTTPS monitors


dulek:
Currently gophercloud requires setting `URLPath` when creating an [Octavia health monitor with type HTTP or HTTPS](https://github.com/gophercloud/gophercloud/blob/master/openstack/loadbalancer/v2/monitors/requests.go#L148-L151). According to [Octavia API reference](https://developer.openstack.org/api-ref/load-balancer/v2/?expanded=create-health-monitor-detail#create-health-monitor) it should not be enforced as `url_path` defaults to "/" (and code confirms it for all Queens-Train Octavia releases).

This is really annoying because in Rocky's Octavia there's a bug preventing creation of HTTPS health monitors with url_path set. And when I try to not set it I get error from gopherclou.

jtopjian:
@dulek Thank you for reporting this. This should be fixed once #1640 is merged.

It looks like the behavior you're seeing was only added to Octavia in June 2018 while the initial Gophercloud Octavia implementation most likely targeted toward a release/behavior before that. This might explain the results you're seeing currently.

dulek:
@jtopjian: That was fast, thanks! Left a minor comment on the patch, but otherwise this would help us.

jtopjian:
This has been merged. Please let me know if you're still seeing issues.

PR #1640, opened 2019-07-11
Load Balancer v2: Remove Health Monitor Validation

jtopjian:
For #1639 

Octavia now supports defaults for when the url_path and expected_codes
are not supplied in the request body. This commit removes the local
validation for these fields to allow the API defaults to be used.

Users working with earlier versions of Octavia will need to be mindful
that they cannot rely on defaults being provided by the API and
therefore must provide explicit values.

https://github.com/openstack/octavia/blob/b2d40c1120727c68b570955413516290ad85f809/octavia/api/v2/controllers/health_monitor.py#L134-L143

### DIFF
--- a/openstack/loadbalancer/v2/monitors/requests.go
+++ b/openstack/loadbalancer/v2/monitors/requests.go
@@ -108,7 +108,6 @@ type CreateOpts struct {
 	MaxRetries int `json:"max_retries" required:"true"`
 
 	// URI path that will be accessed if Monitor type is HTTP or HTTPS.
-	// Required for HTTP(S) types.
 	URLPath string `json:"url_path,omitempty"`
 
 	// The HTTP method used for requests by the Monitor. If this attribute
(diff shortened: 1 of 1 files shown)


openstack/loadbalancer/v2/monitors/requests.go
```diff
@@ -108,16 +108,14 @@ type CreateOpts struct {
...
 	URLPath string `json:"url_path,omitempty"`
 
 	// The HTTP method used for requests by the Monitor. If this attribute
 	// is not specified, it defaults to "GET". Required for HTTP(S) types.
 	HTTPMethod string `json:"http_method,omitempty"`
 
 	// Expected HTTP codes for a passing HTTP(S) Monitor. You can either specify
-	// a single status like "200", or a range like "200-202". Required for HTTP(S)
-	// types.
+	// a single status like "200", or a range like "200-202".
```
dulek: Well, just for the sake of correctness - such strings work with Octavia as well:

"200, 202"
"200-202, 401"

jtopjian: Thanks! This has been updated.

Issue #1642, opened 2019-07-15
Identity V3: Trusts API support


ozerovandrei:
Docs: [OS-TRUST-API](https://developer.openstack.org/api-ref/identity/v3-ext/?expanded=create-trust-detail,list-trusts-detail#os-trust-api)

Progress:

- [x] Create #1643
- [x] Delete #1644
- [x] List #1855 
- [x] Get #1855 
- [ ] List roles delegated by a trust
- [ ] Check if a role is delegated by a trust
- [ ] Get role delegated by a trust

PR #1643, opened 2019-07-15
Identity V3: add Trust Create and Extract funcions

ozerovandrei:
Implement Create and Extract functions for v3/extenstions/trusts package
with unit tests and docs.

This commit also adds Role struct to this package.

For #1642

code in this PR:

https://github.com/openstack/keystone/blob/stable/stein/keystone/trust/core.py#L135
https://github.com/openstack/keystone/blob/stable/stein/keystone/api/trusts.py#L204


ozerovandrei:
This has been merged in #1644

Issue #1646, opened 2019-07-24
networking/v2/extensions/layer3/floatingips lack of `Time.time Created/Updated` fields


naumvd95:
Need to add `Time.time Created/Updated`  fields inside results.go,  mapped from `json:"updated_at"`/`json:"created_at"`

General API reference: https://developer.openstack.org/api-ref/network/v2/?expanded=list-floating-ips-detail,show-floating-ip-details-detail#show-floating-ip-details

that fields defined in standard_attr.py , that means that it should be applied to all neutron-resources though decorator:
https://github.com/openstack/neutron/blob/041203f1bb1a9cb814ab58c6088d0229cfe9d9bb/neutron/db/standard_attr.py#L181
https://github.com/openstack/neutron/blob/041203f1bb1a9cb814ab58c6088d0229cfe9d9bb/neutron/db/models/l3.py#L70-L71

naumvd95:
I wanted to implement it by myself, don't know how to assign it properly

jtopjian:
@naumvd95 I think adding a comment like you've done is the best way to go. 🙂

naumvd95:
@jtopjian its ready for review , thanks in advance)

jtopjian:
Implemented in #1647

naumvd95:
@jtopjian thanks a lot ;)

PR #1647, opened 2019-07-25
[Neutron][FloatingIp] updated_at/created_at fields

naumvd95:
* time.Time object
* mapped to existing json fields

For #1646 

General API reference: https://developer.openstack.org/api-ref/network/v2/?expanded=list-floating-ips-detail,show-floating-ip-details-detail#show-floating-ip-details

that fields defined in standard_attr.py , that means that it should be applied to all neutron-resources though decorator:
https://github.com/openstack/neutron/blob/041203f1bb1a9cb814ab58c6088d0229cfe9d9bb/neutron/db/standard_attr.py#L181
https://github.com/openstack/neutron/blob/041203f1bb1a9cb814ab58c6088d0229cfe9d9bb/neutron/db/models/l3.py#L70-L71


naumvd95:
@jtopjian its ready for review , thanks in advance)

jtopjian:
oops - I read too quickly. The fields should be `CreatedAt` and `UpdatedAt` to reflect the JSON tags. I will fix this.

Issue #1631, opened 2019-06-24
ContainerInfra: Cluster resize error decoding response


tghartland:
When using the clusters.Resize method for the new resize API, the cluster is successfully resized but the returned result is always an error with just the message "EOF".

It comes from [here](https://github.com/gophercloud/gophercloud/blob/157432124aab520975efedb5e54b95287785578d/provider_client.go#L479) where the response body is decoded. The reason is that cluster resize does not return a body when the request is successful.

tghartland:
@openstacker can you confirm if the resize response should have a body or not? In #1516 you tested the code and it worked, but now we are running magnum stein and by the CLI client / curl / gophercloud we do not get a body in the resize request, which causes gophercloud to fail as there is nothing to decode.

With `openstack coe cluster resize --debug`:
```
REQ: curl -g -i -X POST https://openstack.cern.ch:9511/v1/clusters/424b2152-c417-40e0-ba75-73748963ebab/actions/resize -H "Accept: application/json" -H "Content-Type: application/json" -H "OpenStack-API-Version: container-infra latest" -H "User-Agent: None" -H "X-Auth-Token: {SHA256}xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx" -d '{"node_count": 6}'
https://openstack.cern.ch:9511 "POST /v1/clusters/424b2152-c417-40e0-ba75-73748963ebab/actions/resize HTTP/1.1" 202 0
RESP: [202] Connection: keep-alive Content-Length: 0 Date: Mon, 24 Jun 2019 08:40:39 GMT OpenStack-API-Maximum-Version: container-infra 1.7 OpenStack-API-Minimum-Version: container-infra 1.1 OpenStack-API-Version: container-infra 1.7 Server: Werkzeug/0.14.1 Python/2.7.5 Strict-Transport-Security: max-age=15768000 vary: OpenStack-API-Version x-openstack-request-id: req-e916c398-d41c-4938-8b45-0ee8eca45d57
RESP BODY: Omitted, Content-Type is set to None. Only application/json responses have their bodies logged.
POST call to container-infra for https://openstack.cern.ch:9511/v1/clusters/424b2152-c417-40e0-ba75-73748963ebab/actions/resize used request id req-e916c398-d41c-4938-8b45-0ee8eca45d57
clean_up ResizeCluster: 
END return value: 0
[...]

PR #1649, opened 2019-07-25
WIP ContainerInfra: extract ID from cluster resize result

tghartland:
The cluster resize result should contain just the cluster's ID, so it needs its own `Extract` method. Currently magnum is sending an empty response which is causing a different error in gophercloud, this will be fixed in magnum with [1].

Will keep as WIP until the fix is merged in magnum and I have been able to test with it.

For #1631 

code in this PR:

[1] https://review.opendev.org/#/c/672600/

### DIFF
--- a/openstack/containerinfra/v1/clusters/results.go
+++ b/openstack/containerinfra/v1/clusters/results.go
@@ -60,6 +60,14 @@ func (r UpdateResult) Extract() (string, error) {
 	return s.UUID, err
 }
 
+func (r ResizeResult) Extract() (string, error) {
+	var s struct {
+		UUID string
+	}
+	err := r.ExtractInto(&s)
+	return s.UUID, err
+}
+
 type Cluster struct {
 	APIAddress        string             `json:"api_address"`
 	COEVersion        string             `json:"coe_version"`
--- a/openstack/containerinfra/v1/clusters/testing/fixtures.go
+++ b/openstack/containerinfra/v1/clusters/testing/fixtures.go
@@ -289,8 +289,7 @@ func HandleDeleteClusterSuccessfully(t *testing.T) {
 
 var ResizeResponse = fmt.Sprintf(`
 {
-	"uuid": "%s",
-	"node_count": 2
+	"uuid": "%s"
 }`, clusterUUID)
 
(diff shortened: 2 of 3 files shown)


jtopjian:
recheck stable/newton

jtopjian:
oops - wrong PR. Ignore those test results :)

jtopjian:
@tghartland It looks like the upstream patch has been merged - is it ready for review? If so, can you confirm that the UUID is the only field being returned now (you're removing the `node_count` field in the test fixtures).

tghartland:
@jtopjian Yes the fix is merged upstream, I wanted to wait until we had it deployed here so I could test this change and be 100% certain it works (but it's taking longer than expected to get the magnum fix deployed).

This can be reviewed now and I can test later.

Since this changes the results API for extracting the ResizeResult, is there anything special that has to be done or should it just be documented whenever the next release is made?

jtopjian:
@tghartland I'm totally fine waiting until you're able to confirm functionality. We can then also verify if the response body has changed.

If the response has changed, we might want to change how the new extraction is done, similar to what's being done [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/container/v1/capsules/microversions.go).

tghartland:
@jtopjian the response should just be the UUID of the cluster
```
{
   "uuid":"746e779a-751a-456b-a3e9-c883d734946f"
}
```
The code in this PR should handle that already, is the code you linked doing something else?

jtopjian:
Ah, yeah, I see what's going on.

Prior to this PR, the result was being extracted into a generic `Cluster` struct. This PR implements a more fine-tuned extraction of just the UUID. So it's quite possible that `node_count` was never returned in the first place. 

Since the results of the extraction method are being changed, we'll want to document it - I'll do that in the changelog.

But again, this can all wait until you've confirmed functionality.

tghartland:
@jtopjian sorry it's been so long with no activity here. I've now tested this on the magnum master branch and it does work. (The fix has been released in magnum Train now as well).

jtopjian:
@tghartland No problem at all. I'm going to go ahead and merge this. Since Magnum isn't being actively tested in OpenLab, the test results won't really matter.

Thank you for your work on this.

Issue #1648, opened 2019-07-25
openstack/networking/v2/extensions/security/groups lack of `Time.time CreatedAt/UpdatedAt` fields


naumvd95:
Need to add Time.time Created/Updated fields inside results.go, mapped from json:"updated_at"/json:"created_at"

General API reference: https://developer.openstack.org/api-ref/network/v2/?expanded=show-security-group-detail#show-security-group

that fields defined in standard_attr.py , that means that it should be applied to all neutron-resources though decorator:
https://github.com/openstack/neutron/blob/041203f1bb1a9cb814ab58c6088d0229cfe9d9bb/neutron/db/standard_attr.py#L181
https://github.com/openstack/neutron/blob/041203f1bb1a9cb814ab58c6088d0229cfe9d9bb/neutron/db/models/securitygroup.py#L26

naumvd95:
ill do it

naumvd95:
@jtopjian ready for review

jtopjian:
Implemented in #1654

PR #1654, opened 2019-07-26
[Neutron][SecurityGroup] updated_at/created_at fields

naumvd95:
for: #1648

* time.Time object
* mapped to existing json fields

General API reference: https://developer.openstack.org/api-ref/network/v2/?expanded=show-security-group-detail#show-security-group

that fields defined in standard_attr.py , that means that it should be applied to all neutron-resources though decorator:
https://github.com/openstack/neutron/blob/041203f1bb1a9cb814ab58c6088d0229cfe9d9bb/neutron/db/standard_attr.py#L181
https://github.com/openstack/neutron/blob/041203f1bb1a9cb814ab58c6088d0229cfe9d9bb/neutron/db/models/securitygroup.py#L26

### DIFF
--- a/openstack/networking/v2/extensions/security/groups/results.go
+++ b/openstack/networking/v2/extensions/security/groups/results.go
@@ -1,6 +1,8 @@
 package groups
 
 import (
+	"time"
+
 	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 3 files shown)


naumvd95:
@jtopjian ready for review

Issue #1653, opened 2019-07-26
Manila: add support for share metadata


gman0:
gophercloud's `sharedfilesystems` package is lacking support for modifying metadata of existing Manila shares. I'll work on a PR.

jtopjian:
Implemented in #1656

PR #1656, opened 2019-07-26
Manila: add support for share metadata modification

gman0:
For #1653

Manila API reference for Share metadata: https://docs.openstack.org/api-ref/shared-file-system/#share-metadata

* [Show metadata implementation](https://github.com/openstack/manila/blob/cec13b8057fa025fa1b1c3bef180687a394d6fd6/manila/api/v1/share_metadata.py#L42-L45)
* [Set metadata implementation](https://github.com/openstack/manila/blob/master/manila/api/v1/share_metadata.py#L98-L119)
* [Update metadata implementation](https://github.com/openstack/manila/blob/master/manila/api/v1/share_metadata.py#L98-L119)
* [Unset metadata implementation](https://github.com/openstack/manila/blob/master/manila/share/api.py#L1805-L1807)

### DIFF
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -406,8 +406,10 @@ type SetMetadataOptsBuilder interface {
 	ToSetMetadataMap() (map[string]interface{}, error)
 }
 
-// SetMetadata sets metadata of the specified share. To extract the updated
-// metadata from the response, call the Extract method on the SetMetadataResult.
+// SetMetadata sets metadata of the specified share.
+// Existing metadata items are either kept or overwritten by the metadata from the request.
+// To extract the updated metadata from the response, call the Extract
+// method on the SetMetadataResult.
 func SetMetadata(client *gophercloud.ServiceClient, id string, opts SetMetadataOptsBuilder) (r SetMetadataResult) {
 	b, err := opts.ToSetMetadataMap()
 	if err != nil {
@@ -441,8 +443,10 @@ type UpdateMetadataOptsBuilder interface {
 	ToUpdateMetadataMap() (map[string]interface{}, error)
 }
 
-// UpdateMetadata updates metadata of the specified share. To extract the updated
-// metadata from the response, call the Extract method on the SetMetadataResult.
+// UpdateMetadata updates metadata of the specified share.
+// All existing metadata items are discarded and replaced by the metadata from the request.
+// To extract the updated metadata from the response, call the Extract
+// method on the SetMetadataResult.
 func UpdateMetadata(client *gophercloud.ServiceClient, id string, opts UpdateMetadataOptsBuilder) (r UpdateMetadataResult) {
 	b, err := opts.ToUpdateMetadataMap()
 	if err != nil {
--- a/openstack/sharedfilesystems/v2/shares/requests.go
+++ b/openstack/sharedfilesystems/v2/shares/requests.go
@@ -379,3 +379,89 @@ func Update(client *gophercloud.ServiceClient, id string, opts UpdateOptsBuilder
 	})
 	return
 }
+
+// ShowMetadata retrieves metadata of the specified share. To extract the retrieved
+// metadata from the response, call the Extract method on the ShowMetadataResult.
+func ShowMetadata(client *gophercloud.ServiceClient, id string) (r ShowMetadataResult) {
+	_, r.Err = client.Get(showMetadataURL(client, id), &r.Body, nil)
+	return
+}
+
+// SetMetadataOpts contains options for setting share metadata.
+// For more information about these parameters, please, refer to the shared file systems API v2,
+// Share Metadata, Show share metadata documentation.
+type SetMetadataOpts struct {
+	Metadata map[string]string `json:"metadata"`
+}
+
+// ToSetMetadataMap assembles a request body based on the contents of an
+// SetMetadataOpts.
+func (opts SetMetadataOpts) ToSetMetadataMap() (map[string]interface{}, error) {
+	return gophercloud.BuildRequestBody(opts, "")
+}
+
+// SetMetadataOptsBuilder allows extensions to add additional parameters to the
+// SetMetadata request.
+type SetMetadataOptsBuilder interface {
+	ToSetMetadataMap() (map[string]interface{}, error)
+}
+
+// SetMetadata sets metadata of the specified share. To extract the updated
+// metadata from the response, call the Extract method on the SetMetadataResult.
+func SetMetadata(client *gophercloud.ServiceClient, id string, opts SetMetadataOptsBuilder) (r SetMetadataResult) {
+	b, err := opts.ToSetMetadataMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+
+	_, r.Err = client.Post(setMetadataURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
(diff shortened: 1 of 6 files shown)


gman0:
PTAL @jtopjian @gouthampacha

jtopjian:
recheck

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -379,3 +379,89 @@ func Update(client *gophercloud.ServiceClient, id string, opts UpdateOptsBuilder
...
+}
+
+// SetMetadataOptsBuilder allows extensions to add additional parameters to the
+// SetMetadata request.
+type SetMetadataOptsBuilder interface {
+	ToSetMetadataMap() (map[string]interface{}, error)
+}
+
+// SetMetadata sets metadata of the specified share. To extract the updated
+// metadata from the response, call the Extract method on the SetMetadataResult.
```
gouthampacha: Worth calling out the difference between SetMetadata and UpdateMetadata, as a comment for future maintainers:

SetMetadata allows adding new metadata items or updating values of existing metadata items; however, UpdateMetadata will drop existing metadata and replace it with new metadata key=value pairs as specified.

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -379,3 +379,89 @@ func Update(client *gophercloud.ServiceClient, id string, opts UpdateOptsBuilder
 	})
 	return
 }
+
+// ShowMetadata retrieves metadata of the specified share. To extract the retrieved
+// metadata from the response, call the Extract method on the ShowMetadataResult.
+func ShowMetadata(client *gophercloud.ServiceClient, id string) (r ShowMetadataResult) {
```
jtopjian: It looks like the Manila code has support to retrieve a single metadata/metadatum: https://github.com/openstack/manila/blob/cec13b8057fa025fa1b1c3bef180687a394d6fd6/manila/api/v1/share_metadata.py#L121


To prevent confusion, let's change this to either `Metadata` or `GetMetadata`. Both have been used in Gophercloud before.

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -379,3 +379,89 @@ func Update(client *gophercloud.ServiceClient, id string, opts UpdateOptsBuilder
...
+
+	_, r.Err = client.Post(updateMetadataURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+
+	return
+}
+
+// UnsetMetadata deletes a single key-value pair from the metadata of the specified share.
+func UnsetMetadata(client *gophercloud.ServiceClient, id, key string) (r UnsetMetadataResult) {
```
jtopjian: Because this API call is only for a single piece of metadata/metadatum, let's change this to either `UnsetMetadatum` or `DeleteMetadatum`. "Metadatum" is used throughout Gophercloud to denote a single piece of metadata.

openstack/sharedfilesystems/v2/shares/requests.go
gouthampacha: +1, this can be GetMetadatum, perhaps?

jtopjian: If that API call is exposed, I agree. The API documentation doesn't have it listed, but that might just mean it's undocumented :)

gouthampacha: Yep! fixing the docs here: https://review.opendev.org/#/c/673112/

gman0:
@jtopjian @gouthampacha thank you guys for the feedback and guidance!! All your comments are  addressed in separate commits, I can squash them later if needed once the review is finished.

openstack/sharedfilesystems/v2/shares/requests.go
gman0: done in https://github.com/gophercloud/gophercloud/pull/1656/commits/5c226fec8324d00d2ecfeeb05cb247fa58e6a293

openstack/sharedfilesystems/v2/shares/requests.go
gman0: I went for `DeleteMetadatum`: https://github.com/gophercloud/gophercloud/pull/1656/commits/6a073de8a3b4a4cba704b77bae968bc226306237

openstack/sharedfilesystems/v2/shares/requests.go
gman0: refactored `ShowMetadata` into `GetMetadata`: https://github.com/gophercloud/gophercloud/pull/1656/commits/da281eda3863bfe81db0c8b0160b0abaf307aced

added `GetMetadatum`: https://github.com/gophercloud/gophercloud/pull/1656/commits/66bd2c0febfe8a1c1f4a02d3f8080739d3c66046

jtopjian:
recheck

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -324,3 +324,65 @@ type ExtendResult struct {
...
+		Metadata map[string]string `json:"metadata"`
+	}
+	err := r.ExtractInto(&s)
+	return s.Metadata, err
+}
+
+// GetMetadatumResult contains the response body and error from a GetMetadatum request.
+type GetMetadatumResult struct {
+	gophercloud.Result
+	key string
```
jtopjian: I understand the purpose of this, but let's use the style that has been previously used here: https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/results.go#L346-L352

This returns a `map[string]string` so the user/dev can do:

```go
result, err := shares.GetMetadatum(c, shareID, "foo").Extract()
fmt.Println(result["foo"])
```

Once it's updated like this, it looks like GetMetadata, GetMetadatum, SetMetadata, and UpdateMetadata can all share a single result type:

```go
type MetadataResult struct {
  gophercloud.Result
}

func (r MetadataResult) Extract() (map[string]string, error) {
 ...
}
```

Let me know if this makes sense or if you have any questions.

gman0: Actually `GetMetadatum` would still need a separate result type because it uses `meta` parent in its JSON response instead of `metadata` like the other methods use (see example responses for [show-metadata](https://github.com/openstack/manila/blob/master/api-ref/source/samples/share-show-metadata-response.json) VS [show-metadata-item](https://github.com/openstack/manila/blob/master/api-ref/source/samples/share-show-metadata-item-response.json)). I'll merge the result types for the rest of the methods.

gman0: Should `GetMetadatumResult` then still return `map[string]string` like you're suggesting, or rather keep things as they are in this PR and return `string`? I'm totally fine with either, especially if it's to keep things consistent...although the latter seems a bit more user-friendly by not having to repeat the metadata key.

jtopjian: > Actually GetMetadatum would still need a separate result type because it uses meta parent in its JSON response instead of metadata like the other methods use 

Good catch. 

> Should GetMetadatumResult then still return map[string]string like you're suggesting,

Yes, please.

I agree that it can be more intuitive to have only the value returned, but using this method keeps things consistent with other metadata areas in Gophercloud. In addition, it more closely models the API response (since the API response is a dict/object/hash).

gman0:
PTAL @jtopjian

PR #1659, opened 2019-07-29
Allow modification of the container secret refs

### LINKED ISSUE #1658: Key Manager V1: Allow modification of the container secret references
Barbican API supports omitting the SecretRefs dict and modifying it later on.

https://docs.openstack.org/barbican/stein/api/reference/containers.html

https://github.com/openstack/barbican/blob/3c2a99604a4eeca9cd37d3d1986276dcbed8b455/barbican/common/validators.py#L781
https://github.com/openstack/barbican/blob/3c2a99604a4eeca9cd37d3d1986276dcbed8b455/barbican/common/validators.py#L753

kayrus:
Resolves #1658

### DIFF
--- a/openstack/keymanager/v1/containers/results.go
+++ b/openstack/keymanager/v1/containers/results.go
@@ -230,3 +230,27 @@ func ExtractConsumers(r pagination.Page) ([]Consumer, error) {
 	err := (r.(ConsumerPage)).ExtractInto(&s)
 	return s.Consumers, err
 }
+
+type secretRefResult struct {
+	gophercloud.Result
+}
+
+// Extract interprets any secretRefResult as a SecretRef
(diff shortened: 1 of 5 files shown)


kayrus:
@jtopjian ready for review

openstack/keymanager/v1/containers/results.go
```diff
@@ -230,3 +230,27 @@ func ExtractConsumers(r pagination.Page) ([]Consumer, error) {
 	err := (r.(ConsumerPage)).ExtractInto(&s)
 	return s.Consumers, err
 }
+
+type secretRefResult struct {
```
jtopjian: Where is this being used?

kayrus: You're right, this was a copy-paste. Will remove it.

PR #1662, opened 2019-07-31
manila share-types: add IDFromName convenience function

gman0:
For #1663 

This PR adds `IDFromName` convenience function for Manila share types.


gman0:
PTAL @jtopjian

jtopjian:
recheck

PR #1664, opened 2019-08-02
Identity V3: Fix RegionID type

kayrus:
For #1665 

Region ID is a string, not an int

https://docs.openstack.org/api-ref/identity/v3/?expanded=list-endpoints-detail#service-catalog-and-endpoints


kayrus:
@jtopjian ready for review

jtopjian:
@kayrus Thank you for the fix. If you can, please follow the correct process of opening an issue and specifying the service API code that proves the implementation is correct.

kayrus:
@jtopjian sorry, I somehow forgot to open an issue at this time.

Issue #1670, opened 2019-08-07
Networking v2: Issue with Time Fields


jtopjian:
The Networking resources which include `CreatedAt` and `UpdatedAt` fields cause time parsing errors with older releases of OpenStack. This is because the fields are currently set to `time.Time` (RFC 3339) but older OpenStack releases send the date back as an RFC 3339 "no Z" format.

The structs which include these fields need to be updated with a custom UnmarshalJSON method to handle both time formats.

jtopjian:
Fixed in #1671

PR #1671, opened 2019-08-07
Networking v2: Support two time formats for network resources

jtopjian:
For #1670


jtopjian:
recheck stable/mitaka

jtopjian:
That error is okay - there are still some failing Mitaka tests. The important part is that there were tests which previously failed due to time parsing issue but those failures are no longer in the results.

Issue #1673, opened 2019-08-12
ContainerInfra: Support fixed_network/fixed_subnet/floating_ip_enabled for cluster creation


lingxiankong:
- The corresponding patch in OpenStack Magnum: https://review.opendev.org/#/c/671182/
- The three new fields were added in https://review.opendev.org/#/c/671182/6/magnum/api/controllers/v1/cluster.py

- [x] fixed_network
- [x] fixed_subnet
- [x] floating_ip_enabled

jtopjian:
@lingxiankong Just a heads up that I'm going to be on vacation from Wednesday to Sunday, so if you submit any more PRs, they might have to wait until Sunday to get merged.

For `floating_ip_enabled`, according to the code, it's defined as:

```
floating_ip_enabled = wsme.wsattr(types.boolean, default=True)
```

By default, it is `true`, which means users will specify it in `CreateOpts` when they want to disable floating IPs (`false`). Therefore, we need to define it as:

```
FloatingIPEnabled *bool `json:"floating_ip_enabled,omitempty"`
```

Let me know if you have any problems.

lingxiankong:
@jtopjian thanks for the reminder. Yeah, I totally understand what you said, actually I've added all those three fields in one PR originally, but according to the contributor guide, it was split into 3 separate ones. Will submit there other 2 hopefully before you go :-)

jtopjian:
> I've added all those three fields in one PR originally, but according to the contributor guide, it was split into 3 separate ones.

As an aside, I need to update the guide - the current rules were created for the early days of this project, when it didn't have a lot of features and contributors wanted to get their features added in bulk very quickly.  As long as the PR is small-ish, please feel free to combine more than one feature.

PR #1676, opened 2019-08-12
Containerinfra: Allow setting fixed_subnet when creating cluster

lingxiankong:
For #1673

code in this PR:

https://github.com/openstack/magnum/blob/04fd0470ad3b35c4af5212eee82431b1be43d64c/magnum/api/controllers/v1/cluster.py#L180


lingxiankong:
@jtopjian this PR is ready for review :-)

Issue #1669, opened 2019-08-06
Compute V2: Tags API


ozerovandrei:
Gophercloud currently only supports setting tags while creating a server or getting tags from an existing server but there is no support to perform updates on server tags.

For: [terraform-provider-openstack#722](https://github.com/terraform-providers/terraform-provider-openstack/issues/722)

Docs: https://docs.openstack.org/api-ref/compute/?expanded=#server-tags-servers-tags

- [x] List tags #1678
- [x] Check tag existence #1679
- [x] Add single tag #1695
- [x] Replace all tags #1694
- [x] Delete a single tag #1699
- [x] Delete all tags #1700
- [x] Tags in server body #1734
- [x] List / filter servers by tags #1759

PR #1678, opened 2019-08-14
Compute V2: add tags extension with List call

ozerovandrei:
Add the new tags extension with the List method.

Provide unit test and documentation.

For #1669

code in this PR:

https://github.com/openstack/nova/blob/stable/stein/nova/api/openstack/compute/server_tags.py#L86
https://github.com/openstack/nova/blob/stable/stein/nova/objects/tag.py#L65


huangtianhua:
recheck

ozerovandrei:
Merged in #1679

Issue #1680, opened 2019-08-19
openstack/loadbalancer/v2/loadbalancers lack of `Time.time CreatedAt/UpdatedAt` fields


naumvd95:
Need to add Time.time Created/Updated fields inside results.go, mapped from json:"updated_at"/json:"created_at"

General API reference:https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=show-load-balancer-details-detail#show-load-balancer-details

that fields defined in:
https://github.com/openstack/octavia/blob/22df2e0f484b3560cbb47d33b7f2f62cb57519fd/octavia/api/v2/types/load_balancer.py#L46

naumvd95:
will do by myself

naumvd95:
faced with issue during PR checking, time format from octavia response seems slightly differs and cannot be parsed by sane-default time.Time configuration
```
parsing time ""2019-07-15T15:57:42"" as ""2006-01-02T15:04:05Z07:00"": cannot parse """ as "Z07:00"
```

naumvd95:
related go issue https://github.com/golang/go/issues/21990

jtopjian:
@naumvd95 Please see https://github.com/gophercloud/gophercloud/pull/1671

naumvd95:
@jtopjian  Awesome !! 
so I'll refactor my PR and add similar custom struct

naumvd95:
@jtopjian btw, seems  in fixtures u left some `Z` cases like in networks
https://github.com/gophercloud/gophercloud/pull/1671/files#diff-1e0b8676b22caba7b7d81baf122157daR198
its ok?

jtopjian:
Yes, per #1670, #1671 is meant to support _both_ timezone formats. In similar manner, you'll want to include fixtures for both formats.

naumvd95:
@jtopjian ready for review

jtopjian:
Implemented in #1681

PR #1681, opened 2019-08-19
 [Octavia][Loadbalancers] updated_at/created_at fields

naumvd95:
for: #1680

time.Time object
mapped to existing json fields

General API reference:https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=show-load-balancer-details-detail#show-load-balancer-details

that fields defined in:
https://github.com/openstack/octavia/blob/22df2e0f484b3560cbb47d33b7f2f62cb57519fd/octavia/api/v2/types/load_balancer.py#L46

### DIFF
--- a/openstack/loadbalancer/v2/loadbalancers/results.go
+++ b/openstack/loadbalancer/v2/loadbalancers/results.go
@@ -1,6 +1,9 @@
 package loadbalancers
 
 import (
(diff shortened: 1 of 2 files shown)


naumvd95:
@jtopjian ready for review

Issue #514, opened 2017-09-14
Feature request: Add support for FWaaS v2.0


Crapworks:
Hi There!

My OpenStack provider only supports FWaaS v2.0 and the v1.0 endpoint is disabled. Since v1.0 is also flagged as deprecated, Gophercloud should support the new FWaaS v2.0 endpoint:

https://developer.openstack.org/api-ref/network/v2/index.html#fwaas-v2-0-current-fwaas-firewall-groups-firewall-policies-firewall-rules

Cheers,
Christian

khdegraaf:
I am working on implementing this right now (for Huawei Cloud providers like OTC) in a private fork.  When this is finished and tested, I'll do a Pull request.

jtopjian:
@khdegraaf Thanks for your interest in this. Just a heads up that for large features such as this, you'll want to open one PR for each API request method.

For example, one PR for `GET firewall_group`, one for `DELETE firewall_group`, etc. 

Let me know if you have any questions.

khdegraaf:
[...]

PR #1685, opened 2019-08-28
[wip] FWaaS v2.0 Rules Get/List

Elethiomel:
For  #514 

https://github.com/openstack/neutron-fwaas/blob/stable/queens/neutron_fwaas/extensions/firewall_v2.py#L153
https://github.com/openstack/neutron-fwaas/blob/stable/queens/neutron_fwaas/extensions/firewall_v2.py#L156

https://docs.openstack.org/api-ref/network/v2/#id378
https://docs.openstack.org/api-ref/network/v2/#id374


Elethiomel:
Drat, pushed delete to master, rather than its own branch. :(

Elethiomel:
How should I handle this? Git reset?

jtopjian:
@Elethiomel This looks good, but we'll need to see better API references in the Python code. For example, how do we know that `source_ip_address` is a valid list option? It might be listed in the API documentation, but the docs aren't guaranteed to be correct. See [here](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md) for more details on what we need.

For Neutron, from past research, we know that all fields of a resource are valid list options. Therefore, we need to find a model of a firewall rule:

https://github.com/openstack/neutron-fwaas/blob/27906d0acff349c9e5ac7955d203fb2233c295db/neutron_fwaas/db/firewall/v2/firewall_db_v2.py#L68-L82

It looks like a few fields are missing and a few fields might be invalid. Limit, Marker, SortKey, and SortDir should be standard fields that can stay.

The easiest way to validate these is to create acceptance tests for them. The v1 [fwaas](https://github.com/gophercloud/gophercloud/tree/master/acceptance/openstack/networking/v2/extensions/fwaas) tests might be able to help. You'll have to create a rule outside of Gophercloud (possibly using the `neutron` or `openstack` command-line tools) but you should be able to create a test that validates listing rules. 

More information on running acceptance tests is [here](https://github.com/gophercloud/gophercloud/tree/master/acceptance). You can run the acceptance tests with the `OS_DEBUG=1` environment variable set and that will print out the actual JSON API requests and responses -- this will help validate your code.

I recommend spending some time getting the acceptance tests to run. For tests where you have to create resources outside of Gophercloud, it doesn't make sense to include them in the PR, but as more and more of the FWaaS v2 stuff gets merged, you can begin including tests. But more importantly, it'll give you a way to quickly validate and iterate on the code you're working on.

Elethiomel:
i'm going to restart this from scratch. Closing PR. Will open a new one soon

Issue #1637, opened 2019-07-01
Add support for port forwarding for floating IPs


simonre:
Port forwarding for floating IPs is not supported in gophercloud but it is a feature in OpenStack. This allows to reuse public IPs.

**API:**
 https://developer.openstack.org/api-ref/network/v2/#show-port-forwarding

**Source**: https://github.com/openstack/neutron/blob/master/neutron/objects/port_forwarding.py

**Schema:** https://github.com/openstack/neutron/blob/master/neutron/db/models/port_forwarding.py

- [x] Create
- [x] List
- [x] Get
- [x] Update
- [x] Delete

simonre:
This is fully implemented now. ( #1651 , #1652 , #1688, #1698 , #1703 )

jtopjian:
@simonre Thank you for your work on this! I apologize if it got off to a rough start. Your work is very much appreciated!

simonre:
@jtopjian No worries at all, I think it's very enjoyable working with you! I'm just a bit slow with my PRs sometimes as I sometimes don't have too much time to work on them

PR #1688, opened 2019-09-06
List Port Forwardings

simonre:

For #1637 

code in this PR:

Source: 
https://github.com/openstack/neutron/blob/f21444f247108212dc835036c98e354d6e6f7ace/neutron/services/portforwarding/pf_plugin.py#L487

API:
https://docs.openstack.org/api-ref/network/v2/?expanded=create-port-forwarding-detail#list-floating-ip-port-forwardings

### DIFF
--- a/openstack/networking/v2/extensions/layer3/portforwarding/requests.go
+++ b/openstack/networking/v2/extensions/layer3/portforwarding/requests.go
@@ -1,6 +1,53 @@
 package portforwarding
 
-import "github.com/gophercloud/gophercloud"
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+type ListOptsBuilder interface {
+	ToPortForwardingListQuery() (string, error)
+}
+
+// ListOpts allows the filtering and sorting of paginated collections through
+// the API. Filtering is achieved by passing in struct field values that map to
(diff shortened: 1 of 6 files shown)


simonre:

openstack/networking/v2/extensions/layer3/portforwarding/requests.go
```diff
@@ -1,6 +1,53 @@
...
+type ListOptsBuilder interface {
+	ToPortForwardingListQuery() (string, error)
+}
+
+// ListOpts allows the filtering and sorting of paginated collections through
+// the API. Filtering is achieved by passing in struct field values that map to
+// the port forwarding attributes you want to see returned. SortKey allows you to
+// sort by a particular network attribute. SortDir sets the direction, and is
+// either `asc' or `desc'. Marker and Limit are used for pagination.
+type ListOpts struct {
```
jtopjian: We need to add `limit` (int) and `marker` (string) here.

Gophercloud's paging automatically sets limit and marker when paging internally, but if the user wanted to limit the results themselves, we need that in `ListOpts`.

simonre:
@jtopjian I added the `limit` and `marker` fields

jtopjian:
@simonre sorry for the delay - I'm out of the office. Thank you!

Issue #1689, opened 2019-09-07
Compute v2: Volume Type support in Boot from Volume


jtopjian:
Per discussion here (https://github.com/terraform-providers/terraform-provider-openstack/issues/736) support for a volume type needs to be added to the `bootfromvolume` extension.

jtopjian:
Implemented in #1690

PR #1690, opened 2019-09-07
Compute v2: Adding VolumeType to bootfromvolume

jtopjian:
For #1689 

* https://github.com/openstack/nova/blob/324db786c86eeb69278736c8e9db6d22f68080e6/nova/objects/block_device.py#L96
* https://github.com/openstack/nova/blob/c883f2201026b9237a1fba9a0a55ff607d57e787/nova/block_device.py#L44-L48
* https://github.com/openstack/nova/blob/b879f1b6dfe285c143dfee41dbb0c8d697a775b4/nova/api/openstack/compute/schemas/servers.py#L354-L356


jtopjian:
It's difficult to add this to the existing acceptance test suite since the Compute tests need cleaned up for microversions beyond 2.1. However, I have tested it offline and can confirm that it works.

PR #1694, opened 2019-09-13
Compute V2: add Replace method for Tags

ozerovandrei:
Implement compute/v2/extensions/tags.Replace.

For #1669

code in this PR:

https://github.com/openstack/nova/blob/stable/stein/nova/api/openstack/compute/server_tags.py#L156
https://github.com/openstack/nova/blob/stable/stein/nova/objects/tag.py#L70


ozerovandrei:
Merged in #1696

PR #1698, opened 2019-09-13
Get Port Forwarding

simonre:
For #1637 

code in this PR:

Source:
https://github.com/openstack/neutron/blob/f21444f247108212dc835036c98e354d6e6f7ace/neutron/services/portforwarding/pf_plugin.py#L469

API:
https://docs.openstack.org/api-ref/network/v2/?expanded=list-floating-ip-port-forwardings-detail#list-floating-ip-port-forwardings


simonre:
@jtopjian Ready for review

PR #1699, opened 2019-09-15
Compute V2: add Delete method for Tags

ozerovandrei:
Implement compute/v2/extensions/tags.Delete.

For #1669

code in this PR:

https://github.com/openstack/nova/blob/stable/stein/nova/api/openstack/compute/server_tags.py#L180


ozerovandrei:
Merged in #1701

PR #1703, opened 2019-09-16
Update port forwarding

simonre:
For #1637 

code in this PR:

Source:
https://github.com/openstack/neutron/blob/f21444f247108212dc835036c98e354d6e6f7ace/neutron/services/portforwarding/pf_plugin.py#L368

https://github.com/openstack/neutron-lib/blob/56033ba643812a30577f6ab17648806c2ee494ba/neutron_lib/api/definitions/floating_ip_port_forwarding.py

API:
https://docs.openstack.org/api-ref/network/v2/?expanded=update-a-port-forwarding-detail#update-a-port-forwarding


simonre:
@jtopjian I don't think the two failures here are due to problems with my code, this should be ready to review

jtopjian:
recheck

Issue #1704, opened 2019-09-17
Neutron V2: fix lbaas v2 update function options parameter 


ozerovandrei:
Type of options parameter should be `UpdateOptsBuilder` interface instead of specific `UpdateOpts` struct type:
- https://github.com/gophercloud/gophercloud/blob/v0.4.0/openstack/networking/v2/extensions/lbaas_v2/listeners/requests.go#L196

- https://github.com/gophercloud/gophercloud/blob/v0.4.0/openstack/networking/v2/extensions/lbaas_v2/loadbalancers/requests.go#L161

 
Similar to other Update functions across Gophercloud.

For https://github.com/terraform-providers/terraform-provider-openstack/issues/867, https://github.com/terraform-providers/terraform-provider-openstack/issues/805

PR #1706, opened 2019-09-17
Neutron LB V2: use UpdateOptsBuilder

ozerovandrei:
Use the UpdateOptsBuilder instead of the specific UpdateOpts struct
type in extensions/lbaas_v2/loadbalancers/requests.Update.

For #1704


ozerovandrei:
I'm going to merge this because here are no issues with loadbalancers tests and those changes doesn't break existing API.

PR #1723, opened 2019-10-01
LBaaSv2: Properly handle update TLS references

### LINKED ISSUE #1722: Lbaas: Listener's SniContainerRefs cannot be unset
`UpdateOpts` struct for the listeners contains:

```
SniContainerRefs []string `json:"sni_container_refs,omitempty"`
```

Which means that `SniContainerRefs` cannot be set to an empty list.

I suggest to add the:

```
SniContainerRefs *[]string `json:"sni_container_refs,omitempty"`
```

UPD: the same has to be done for the `DefaultTlsContainerRef` and `AllowedCIDRs`.

UPD2: If the protocol is `TERMINATED_HTTPS`, then `DefaultTlsContainerRef` is required and cannot be unset.

kayrus:
Resolves #1722


kayrus:
@jtopjian ready for review.

Issue #1709, opened 2019-09-18
Octavia: Support listener allowed_cidrs


lingxiankong:
API: https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-listener-detail#id34

jtopjian:
@lingxiankong After merging #1710, I think I see an issue:

With `UpdateOpts`, I don't think it's currently possible to remove all CIDRs from a listener. Empty slices with `omitempty` will cause the field to be omitted from the PUT request.

Therefore, I think `[]string` needs to be changed to `*[]string` so that `&allowedCIDRs` (which would be an empty string slice) can be passed.

Can you test this and confirm?

Or is removing all CIDRs not a valid use-case?

lingxiankong:
@jtopjian yes, you are right, I didn't test this edge case, I will do a quick fix.

lingxiankong:
@jtopjian after a second thought, I think it's OK in gophercloud to omit `allowed_cidrs` when updating a listener with `allowedCIDRs=[]string{}`. From a user's perspective, it doesn't make sense to pass an empty list to update a listener, which means the load balancer can't be accessed from anywhere.

It's like you can't create a listener with empty `allowed_cidrs` list, because the default behavior is to allow all, e.g.:
```
$ cat <<EOF | http post ${octavia_prefix}/lbaas/listeners X-Auth-Token:$token
> {
>     "listener": {
>         "loadbalancer_id": "b136383c-fab0-4816-9575-411015f2a9d6",
>         "protocol": "HTTP",
>         "protocol_port": "81",
>         "allowed_cidrs": []
>     }
> }
> EOF
HTTP/1.1 201 Created
Connection: close
Content-Length: 858
Content-Type: application/json
Date: Thu, 19 Sep 2019 03:37:36 GMT
Server: Apache/2.4.18 (Ubuntu)
x-openstack-request-id: req-995b97fc-8b53-417b-b4b9-3752ab6d5a60

{
    "listener": {
        "admin_state_up": true,
        "allowed_cidrs": null,
        "client_authentication": "NONE",
        "client_ca_tls_container_ref": null,
        "client_crl_container_ref": null,
        "connection_limit": -1,
        "created_at": "2019-09-19T03:37:36",
        "default_pool_id": null,
        "default_tls_container_ref": null,
        "description": "",
        "id": "fe076dca-dcf8-4df1-97a5-2de7d604413e",
        "insert_headers": {},
[...]

PR #1727, opened 2019-10-04
loadbalancer: Fix AllowedCIDRs for listener update

lingxiankong:
For #1709

Octavia code: https://github.com/openstack/octavia/blob/master/octavia/api/v2/types/listener.py#L189

When using `*[]string{}`, the `listener.AllowedCIDRs` could be updated to `[]`


jtopjian:
recheck

lingxiankong:
@jtopjian This PR is ready for review, not sure the failed continuous-integration job is related to this change.

jtopjian:
@lingxiankong Yes, the Travis test has been fixed in another PR. Thank you for following-up with this!

Issue #1731, opened 2019-10-05
Add AttachedVolumes to Servers


sbueringer:
Hi,

it would be nice if the AttachedVolumes field could be added to the parsed servers.

See also https://docs.openstack.org/api-ref/compute/?expanded=list-servers-detailed-detail,list-servers-detail#list-servers-detailed

jtopjian:
Implemented in #1732

PR #1732, opened 2019-10-05
added attachments

sbueringer:
For #1731

code in this PR:
* volume attachments are added to the server object here: https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/views/servers.py#L669-L688


sbueringer:
What should I do to not decrease the coverage? Not sure what the problem is :)

jtopjian:
@sbueringer The coverage can be ignored - sometimes it calculates things oddly.

However, we do need to see the actual Nova python code where this feature is implemented. See [here](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md) for more details and let me know if you have any questions.

sbueringer:
@jtopjian Okay thx. I added the link to the code where the os volume attachments are added above

openstack/compute/v2/servers/results.go
```diff
@@ -212,6 +212,9 @@ type Server struct {
 	// to it.
 	SecurityGroups []map[string]interface{} `json:"security_groups"`
 
+	// AttachedVolumes includes the volume attachments of this instance
+	AttachedVolumes []map[string]string `json:"os-extended-volumes:volumes_attached"`
```
jtopjian: Instead of `[]map[string]string`, how about:

```go
type AttachedVolume struct {
  ID string `json:"id"`
}
```

and then

```go
// AttachedVolumes includes the volume attachments of this instance
AttachedVolumes []AttachedVolume `json:"os-extended-volumes:volumes_attached"`
```

According to the code, `id` is the only field except if microversion 2.3+ is used, and then there is a `delete_on_termination` field. But since that's a microservice addition, we'll have to handle that differently, so let's not worry about that in this PR.

sbueringer:
@jtopjian PTAL :)

sbueringer:
Thx for the fast review ;)

PR #1734, opened 2019-10-06
Compute V2: allow to read Tags from Server body

ozerovandrei:
Add compute/v2/extensions/tags.ServerTagsExt to allow read tags right
from a Server body.

Fix compute/v2/extensions/tags to use the valid microversion.

Add ServerTagsExt usage example to compute/v2/extensions/tags docs.

Add ServerTagsExt into ServerTags acceptance test.

For #1669

code in this PR:

https://github.com/openstack/nova/blob/stable/stein/nova/api/openstack/compute/servers.py#L353


ozerovandrei:
recheck

PR #1738, opened 2019-10-10
Octavia LBaaS v2: Allow unsetting the member's subnet id in batch mem…

### LINKED ISSUE #1737: Octavia LBaaS v2: Allow unsetting the member's subnet id in batch members update
`PUT /members`  - https://github.com/openstack/octavia/blob/e6cd43d6cc2710266a1990c952201c5714a03393/octavia/api/v2/types/member.py#L111

it represents the `MemberPOST` class - https://github.com/openstack/octavia/blob/e6cd43d6cc2710266a1990c952201c5714a03393/octavia/api/v2/types/member.py#L68, therefor the subnet_id update is possible there.

kayrus:
Resolves #1737

### DIFF
--- a/openstack/loadbalancer/v2/pools/requests.go
+++ b/openstack/loadbalancer/v2/pools/requests.go
@@ -345,11 +345,49 @@ type BatchUpdateMemberOptsBuilder interface {
 	ToBatchMemberUpdateMap() (map[string]interface{}, error)
 }
 
-type BatchUpdateMemberOpts CreateMemberOpts
+// BatchUpdateMemberOpts is the common options struct used in this package's BatchUpdateMembers
+// operation.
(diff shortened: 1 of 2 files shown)


kayrus:
@jtopjian ready for review

jtopjian:
`BatchUpdateMemberOpts` is an alias to `CreateMemberOpts`. This was most likely done out of convenience because the two shared fields.

But if we're now changing the types of the update struct, we shouldn't do this by modifying the create struct. There are other areas on Gophercloud where create and update look the same except update has pointers for types to account for unsetting values.

I suggest we unalias `BatchUpdateMemberOpts` and make it a first-class struct.

kayrus:
@jtopjian good point

PR #1746, opened 2019-10-15
avoid reauthentication loops

### LINKED ISSUE #1745: ProviderClient.Request should not perform multiple reauthentications in a row
So I just got this error message:
```
Successfully re-authenticated, but got error executing request: Successfully
re-authenticated, but got error executing request: Successfully re-authenticated, but
got error executing request: Successfully re-authenticated, but got error executing
request: Successfully re-authenticated, but got error executing request: Successfully
re-authenticated, but got error executing request: Successfully re-authenticated, but
got error executing request: Successfully re-authenticated, but got error executing
request: Successfully re-authenticated, but got error executing request: Successfully
re-authenticated, but got error executing request: Successfully re-authenticated, but
got error executing request: Successfully re-authenticated, but got error executing
request: Successfully re-authenticated, but got error executing request: Successfully
re-authenticated, but got error executing request: Successfully re-authenticated, but
got error executing request: Successfully re-authenticated, but got error executing
request: Successfully re-authenticated, but got error executing request: Successfully
re-authenticated, but got error executing request: Successfully re-authenticated, but
got error executing request: Get https://service.example.com/some/endpoint/: dial tcp
12.34.56.78:443: connect: connection refused
```

I omitted the remaining 1109 instances of `Successfully re-authenticated, but got error executing request` because the point is clear. The backend service has a bug that causes it to respond with 401 even if the token is fresh and perfectly valid, which causes ProviderClient.Request to reauthenticate ad infinitum, or in this case, until a different error occurs.

I don't think it makes much sense for ProviderClient.Request to reauthenticate more than once for a single request. If the endpoint returns 401 for a second time, ProviderClient.Request should abort and give an `ErrDefault401` to the user.

majewsky:
It does not make sense to reauthenticate multiple times for a single request. When the backing service returns 401 for a brand-new token, reauthenticating again will just send us into an infinite loop.

Closes #1745.

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -94,9 +94,10 @@ type ProviderClient struct {
 // reauthlock represents a set of attributes used to help in the reauthentication process.
 type reauthlock struct {
 	sync.RWMutex
-	reauthing    bool
-	reauthingErr error
-	done         *sync.Cond
+	//This channel is non-nil during reauthentication. It can be used to ask the
+	//goroutine doing Reauthenticate() for its result. Look at the implementation
+	//of Reauthenticate() for details.
+	ongoing chan<- (chan<- error)
 }
 
 // AuthenticatedHeaders returns a map of HTTP headers that are common for all
@@ -106,11 +107,15 @@ func (client *ProviderClient) AuthenticatedHeaders() (m map[string]string) {
 		return
 	}
 	if client.reauthmut != nil {
+		//if a Reauthenticate is in progress, wait for it to complete
 		client.reauthmut.Lock()
-		for client.reauthmut.reauthing {
-			client.reauthmut.done.Wait()
-		}
+		ongoing := client.reauthmut.ongoing
 		client.reauthmut.Unlock()
+		if ongoing != nil {
+			responseChannel := make(chan error)
+			ongoing <- responseChannel
+			_ = <-responseChannel
+		}
 	}
 	t := client.Token()
 	if t == "" {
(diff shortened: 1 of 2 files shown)


majewsky:
Okay, so this got way worse than I expected.

The test failure is in `go test ./testing -run TestConcurrentReauth`. ProviderClient.Reauthenticate has several places right now where it unlocks the reauth mutex before locking it again. This allows concurrent goroutines to glitch through the reauthentication without the token having changed at all.

I tried to replace the `hasReauthenticated` flag with a counter so that I can permit more than one reauthentication per Request call. It turns out that I need to allow 250 (!) reauthentications per Request call to make the test pass mostly reliably. (With a reduced number of concurrent requests, that number goes even higher.)

There is definitely something wrong with Reauthenticate(). I have some ideas and will poke around a bit.

majewsky:
Okay, so this PR also got way bigger than I expected. I threw away most of the mutexes in `type reauthlock` and moved to an implementation that uses channels to coordinate concurrent Reauthenticate calls. This implementation is way more readable and more idiomatic, and also passes all the tests on the first try.

jtopjian:
@majewsky Thanks so much for looking into this.

I think I understand the implementation - it looks like you have implemented correctly what the removed code was trying to achieve. I can also confirm the existing tests are passing, even doing a count of 50 rounds. And `-race` isn't reporting any issues, either.

Is it possible to add an additional unit test to reproduce the issue you discovered?

majewsky:
I added a testcase. You can verify that the testcase detects the fixed bug by removing the `&& !state.hasReauthenticated` check from `provider_client.go:432`.

Issue #1666, opened 2019-08-05
Networking V2: Quotas Get/Update support


ozerovandrei:
Docs: https://docs.openstack.org/api-ref/network/v2/index.html#quotas-extension-quotas

For: [terraform-provider-openstack#657](https://github.com/terraform-providers/terraform-provider-openstack/issues/657)

ozerovandrei:
I'm going to close this since all needed features for terraform-provider-openstack#657 are implemented.

I think that all the remaining stuff should be linked to an other issue since I won't implement them in any near future. So I'm closing it because I don't want it to be opened for ages.

PR #1747, opened 2019-10-15
Networking V2: add quotas.Update method

ozerovandrei:
Add openstack/networking/v2/extensions/quotas.Update.

Add unit and acceptance tests.

Update package docs.

For #1666

code in this PR:

https://github.com/openstack/neutron/blob/stable/train/neutron/extensions/quotasv2.py#L116


ozerovandrei:
recheck

PR #1752, opened 2019-10-17
Auth: Allow TokenID passthrough when the scope is not set

### LINKED ISSUE #1566: Auth: introduce token auth

kayrus:
Resolves #1566

Reference:
https://github.com/terraform-providers/terraform-provider-openstack/pull/861#issuecomment-531261745

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -187,16 +187,53 @@ func v3auth(client *gophercloud.ProviderClient, endpoint string, opts tokens3.Au
 		v3Client.Endpoint = endpoint
 	}
 
-	result := tokens3.Create(v3Client, opts)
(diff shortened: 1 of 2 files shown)


kayrus:
recheck

jtopjian:
I think there's an issue with OpenLab building the ironic environment. I think I've pinned it down - just need to find a minute to review and open an issue :)

kayrus:
recheck

kayrus:
@jtopjian ready for review

jtopjian:
@kayrus Thank you for looking into this.

Issue #1748, opened 2019-10-16
Core client: debug round-tripper as a part of gophercloud package


kayrus:
@jtopjian currently debug logging is supported only in https://github.com/gophercloud/gophercloud/blob/6f9cb195db5276cf596293ac30b08c93acd12851/acceptance/clients/http.go package. OpenStack Terraform provider just copies this logic: https://github.com/terraform-providers/terraform-provider-openstack/blob/493f93252e8071f592f03ad2da3bdf4288dc365f/openstack/types.go#L26..L33

It would be nice to see the debug RoundTripper in gophercloud, therefore every app, which requires debug logging can reuse it. I'd like to create the "gophercloud/debug" package containing the LogRoundTripper struct with methods.

In addition it should support custom loggers, e.g. https://github.com/sirupsen/logrus or https://github.com/kubernetes/klog

jtopjian:
We can look at adding this, but it would be better if it was added to https://github.com/gophercloud/utils/

kayrus:
Can you move this issue into the https://github.com/gophercloud/utils/ project?

PR #1756, opened 2019-10-20
Acceptance tests: use gophercloud/utils/client debug client

kayrus:
Part of the #1748


jtopjian:
I understand that implementing this removes duplication, but I'd rather leave it in place.

I'd prefer not to have Gophercloud core depend on the utils repo/package - it creates a circular dependency that might cause issues later on.

kayrus:
ok, not so important for the acceptance tests

PR #1758, opened 2019-10-22
Identity V3: Add access rules support for application credentials

### LINKED ISSUE #1757: Identity V3: Add access rules field support for application credentials
There is a new feature released in Train:
https://docs.openstack.org/api-ref/identity/v3/#what-s-new-in-version-3-13-train

https://github.com/openstack/keystone/blob/stable/train/keystone/application_credential/schema.py#L32..L66

@jtopjian I wonder how exactly new options should be introduced: within an extension or directly inside the "applicationcredentials" package?

kayrus:
Resolves #1757 

acceptance tests will be added later

@jtopjian not sure whether I need to create a separate struct to create access rules


jtopjian:
> not sure whether I need to create a separate struct to create access rules

Do you mean because you are currently re-using `AccessRule` from `results.go`? 

Technically, we can't use `Links` for the creation of an access rule, so that might be reason enough to create a struct in `requests.go`. 

On the other hand, this kind of re-use is done in in a few other places already. In addition, there are a lot of `Links` fields missing in other structs. To date, no one has reported an issue with this and pagination automatically handles the `Links` fields when they're missing (just like how you have `NextPageURL` right now. Here's another [example](https://github.com/gophercloud/gophercloud/blob/0caa4986e4a4a25b0c9fca247cc3291cdbfc76ad/openstack/compute/v2/extensions/usage/results.go#L131-L140)).

So there are two options:

1. If you don't have a need to use `Links`, you can remove it from the result struct and re-use that struct for creation, but we'll have to be mindful that this _might_ need changed in the future.

2. Create an `AccessRuleCreateOpts` struct that looks like `AccessRule` but does not include `Links`.

Thoughts?

kayrus:
recheck

kayrus:
@jtopjian probably the first option is the best choice. I don't have a use case to fetch and process links attribute for an individual access rule and even an application credential.

huangtianhua:
recheck

kayrus:
@jtopjian ready for review

openstack/identity/v3/applicationcredentials/results.go
```diff
@@ -39,8 +53,8 @@ type ApplicationCredential struct {
 	Roles []Role `json:"roles"`
 	// The expiration time of the application credential, if one was specified.
 	ExpiresAt time.Time `json:"-"`
-	// Links contains referencing links to the application credential.
-	Links map[string]interface{} `json:"links"`
```
jtopjian: Let's add this back. I mentioned that no one has noticed when these fields are missing, but I'd prefer to leave the existing ones in place just in case they're being used somewhere.

PR #1759, opened 2019-10-23
Compute V2: implement server tags filtering support

Fedosin:
Server tags support was added in Nova API since 2.26, so we can implement it here as well.

For #1669

code in this PR:

https://github.com/openstack/nova/blob/3b54979ae02cc74bcf215713c16916bbf4131443/nova/api/openstack/compute/schemas/servers.py#L645-L651
https://docs.openstack.org/api-ref/compute/?expanded=list-servers-detail,show-server-details-detail#list-servers

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -55,6 +55,22 @@ type ListOpts struct {
 	// TenantID lists servers for a particular tenant.
 	// Setting "AllTenants = true" is required.
 	TenantID string `q:"tenant_id"`
+
+	// This requires the client to be set to microversion 2.26 or later.
+	// Tags filters on specific server tags. All tags must be present for the server.
+	Tags string `q:"tags"`
+
+	// This requires the client to be set to microversion 2.26 or later.
+	// TagsAny filters on specific server tags. At least one of the tags must be present for the server.
+	TagsAny string `q:"tags-any"`
+
+	// This requires the client to be set to microversion 2.26 or later.
+	// NotTags filters on specific server tags. All tags must be absent for the server.
+	NotTags string `q:"not-tags"`
+
+	// This requires the client to be set to microversion 2.26 or later.
+	// NotTagsAny filters on specific server tags. At least one of the tags must be absent for the server.
+	NotTagsAny string `q:"not-tags-any"`
 }
 
 // ToServerListQuery formats a ListOpts into a query string.


jtopjian:
@Fedosin Thank you for submitting this. The referenced issue #1669 has been closed, but I'm OK with re-opening it for this PR. However, there are a few things that need worked out.

Per our [contributor tutorial](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md) we need to see links to the actual API service implementation of listing tags, not just documentation. To save some time, [here's](https://github.com/openstack/nova/blob/3b54979ae02cc74bcf215713c16916bbf4131443/nova/api/openstack/compute/schemas/servers.py#L645-L651) the needed information.

Next, you are adding a field of `Tags` to the `servers.Server` result struct. Since tags was added in a microversion, it needs to be handled separately. We already have support for extracting tags [here](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/servers/microversions.go). Please see our documentation on [microversions](https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md) for more details. It would be best if this was reverted.

@ozerovandrei on that note, I believe [this](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/tags/results.go#L5-L9) can be removed since we already have this functionality in the servers' `microversions.go` file. I can open a PR to fix this and then this PR can be rebased once that's done.

On the core topic of this PR, the additions to the `ListOpts` is correct and follows our microversion guidelines. We're not able to use `omitempty` with `ListOpts`, so the current implementation is fine.

I'll open a PR to fix the tags extension and then this PR can be rebased. Once that's done, this should be ready to go.

Please let me know if anyone has any questions.

acceptance/openstack/compute/v2/servers_test.go
```diff
@@ -515,13 +515,7 @@ func TestServersTags(t *testing.T) {
...
 	client.Microversion = "2.26"
 
-	// Check server tags in body.
-	type serverWithTagsExt struct {
-		servers.Server
-		tags.ServerTagsExt
-	}
-
-	serverWithTags := serverWithTagsExt{}
+	serverWithTags := servers.Server{}
```
jtopjian: Per other comments, please revert this change and rebase. After the rebase, I believe this file can go unmodified. Let me know if you have any questions.

openstack/compute/v2/servers/requests.go
```diff
@@ -55,6 +55,18 @@ type ListOpts struct {
...
+	Tags []string `q:"tags"`
+
+	// TagsAny filters on specific server tags. At least one of the tags must be present for the server.
+	TagsAny []string `q:"tags-any"`
+
+	// NotTags filters on specific server tags. All tags must be absent for the server.
+	NotTags []string `q:"not-tags"`
+
+	// NotTagsAny filters on specific server tags. At least one of the tags must be absent for the server.
+	NotTagsAny []string `q:"not-tags-any"`
```
jtopjian: Just one nit with these in-line comments: can you add:

```
// This requires the client to be set to microversion 2.26 or later.
```

To each field?

openstack/compute/v2/servers/results.go
```diff
@@ -217,6 +217,10 @@ type Server struct {
 
 	// Fault contains failure information about a server.
 	Fault Fault `json:"fault"`
+
+	// Tags is a list of server tags. Tags are arbitrarily defined strings
+	// attached to a server.
+	Tags []string `json:"tags"`
```
jtopjian: Per the other comments/discussion, please revert this as there's already a way to achieve this. Let me know if you have any questions with the current implementation, though.

Fedosin:
I made this as an extension to ListOpts, plus added a test. 
We can always avoid making the extension and put these parameters directly in ListOpts, but as you said it doesn't comply with the rules.

jtopjian:
> We can always avoid making the extension and put these parameters directly in ListOpts

I'm so sorry for the confusion. Adding the parameters directly in `ListOpts` is the right solution. This is outlined under the "New Request Fields" section in the [microversions](https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md#new-request-fields) doc, but it is terribly confusing.

There are three forms of an API amendment/extension that we try to deal with:

1. New API functionality that is added via an extension.

Nova used to have more well-defined/first-class extensions but collapsed them all into the core code, but other services (such as [Neutron](https://github.com/gophercloud/gophercloud/tree/master/openstack/networking/v2/extensions/portsecurity)) still have a concept of extensions. So between that and the existing `openstack/compute/v2/extensions` packages, new API features that don't modify the existing request and response bodies are considered an extension.

2. New _response_ fields that were amended via a microversion

These go under `microversions.go` in a way that doesn't modify the API response. The reason for this is because of Go's concept of a "zero-value". 

We _could_ add the field `Tags` to the `servers.Server` struct, but there is no way of determining if the server response doesn't support tags or if there are no tags in the response body. For "tags" specifically, this isn't a huge deal, but there might be other fields where this distinction is important.

So we use the concept of composing new structs and the developer can add their own conditional logic on when they want to do this, with the thinking that the developer will know when they need to support a microversion and when the response should have a microversion-amended response.

3. New _request_ options that were amended via a microversion

Because we have control over how the request body is sent, we can omit fields that are sent in a request body. So it's possible for us to support microversion amendments in-line of the original structs and functions in the `requests.go` files.

It would be great if we could do this for responses, but as mentioned above, the zero-value makes it indeterminable, so that's why there's the extra work involved.

You can see some other examples of existing microversion amendments here:

* https://github.com/gophercloud/gophercloud/blob/32225edcb95e34f926834ae229ccf2d88de83c68/openstack/compute/v2/extensions/extendedserverattributes/microversions.go
* https://github.com/gophercloud/gophercloud/blob/3b05c006441c40ed2ed2654b04913f0b18343a52/openstack/compute/v2/extensions/servergroups/requests.go#L30-L36

Long story short, the way you originally had the ListOpts is correct. Just add an in-line comment that it requires the `client` to be set to a specific microversion. If no tag options are specified, their default value is an empty string and it won't be sent in the request body.

Adding the fields to ListOpts is the only change needed to add the functionality you want, so the PR can be trimmed to just that.

I'm really sorry about the confusion and the extra work you put in because of it.

jtopjian:
recheck

jtopjian:
The test failures are not related to this PR.

Fedosin:
My pleasure :) thanks for your support!

PR #1760, opened 2019-10-24
Compute v2: Move ServerTagsExt to servers.TagsExt

jtopjian:
This commit moves the tags extension's ServerTagsExt to
the servers package. While this is slighly nuanced, the tags
extension can be thought of to work directly on a server's
tags and the ability to extract a server's tags can be part
of the servers package since the original ExtractTags
microversion support already exists.

For #1669

### DIFF
--- a/acceptance/openstack/compute/v2/servers_test.go
+++ b/acceptance/openstack/compute/v2/servers_test.go
@@ -518,7 +518,7 @@ func TestServersTags(t *testing.T) {
 	// Check server tags in body.
 	type serverWithTagsExt struct {
 		servers.Server
-		tags.ServerTagsExt
+		servers.TagsExt
 	}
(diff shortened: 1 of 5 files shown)


jtopjian:
@ozerovandrei after looking at https://github.com/gophercloud/gophercloud/blob/0907b320e0ac3a4de50c0fbda5c67064086eecee/openstack/compute/v2/extensions/tags/results.go#L5-L9

I see the benefit to having a struct extension like you made since there is not an easy way to see the tags from a list of servers. But since we already have the ability to extract tags from a single server result in `microversions.go`, I think the struct extension is better located in the `servers` package. 

Like the commit message says, this is slightly nuanced / nit picking, so I'm open to discussion on this.

PR #1772, opened 2019-11-19
[wip] FWaaS v2.0 Rule Get (depends on PR 1771)

Elethiomel:
For #514 

https://github.com/openstack/neutron-fwaas/blob/27906d0acff349c9e5ac7955d203fb2233c295db/neutron_fwaas/db/firewall/v2/firewall_db_v2.py#L68-L82
--


Elethiomel:
@jtopjian I'm not 100% sure what happened here. It looks like the testing bombed out due to another reason. My acceptance tests are passing locally.

jtopjian:
recheck

jtopjian:
@Elethiomel Yeah, it's either a transient error or a change in the OpenStack API causing a test to fail. If it happens again, we'll ignore it for this PR.

Issue #1767, opened 2019-11-11
Support containerinfra node group API


tghartland:
* [x] Create #1800
* [x] Get / List #1774
* [x] Update #1800
* [x] Delete #1800

NodeGroup API representation:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L58-L127

Returned fields limited to these:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L147-L149

List method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L226-L262

Get method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L265-L280

Create method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L283-L323

Available fields when creating node group:
https://github.com/openstack/python-magnumclient/blob/c0d368360831cd29affa47e77eba4e81e4c6f0f8/magnumclient/v1/nodegroups.py#L21-L23

Update method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L326-L338

Delete method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L341-L355

tghartland:
@jtopjian I have the code for all API methods written now [1], but there is a Magnum fix waiting to be merged [2]. Is it fine to start reviewing / merging here?

[1] https://github.com/tghartland/gophercloud/tree/nodegroups-delete/openstack/containerinfra/v1/nodegroups and other branches
[2] https://review.opendev.org/#/c/694230/

tghartland:
[...]

PR #1774, opened 2019-11-19
ContainerInfra: nodegroups Get 

tghartland:
For #1767

NodeGroup API representation:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L58-L127

List method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L226-L262

Get method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L265-L280

Waiting on https://review.opendev.org/#/c/694230/ to be merged to fix listing with limited number of results per page.

### DIFF
--- a/acceptance/openstack/containerinfra/v1/nodegroups_test.go
+++ b/acceptance/openstack/containerinfra/v1/nodegroups_test.go
@@ -1,0 +1,73 @@
+// +build acceptance containerinfra
+
+package v1
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/openstack/containerinfra/v1/nodegroups"
+	th "github.com/gophercloud/gophercloud/testhelper"
+)
+
+func TestNodeGroupsCRUD(t *testing.T) {
+	// API not available until Magnum train
+	clients.SkipRelease(t, "stable/mitaka")
+	clients.SkipRelease(t, "stable/newton")
+	clients.SkipRelease(t, "stable/ocata")
+	clients.SkipRelease(t, "stable/pike")
+	clients.SkipRelease(t, "stable/queens")
+	clients.SkipRelease(t, "stable/rocky")
+	clients.SkipRelease(t, "stable/stein")
+
+	client, err := clients.NewContainerInfraV1Client()
+	th.AssertNoErr(t, err)
+
+	client.Microversion = "1.9"
+
+	clusterTemplate, err := CreateKubernetesClusterTemplate(t, client)
+	th.AssertNoErr(t, err)
+	defer DeleteClusterTemplate(t, client, clusterTemplate.UUID)
+
+	clusterID, err := CreateKubernetesCluster(t, client, clusterTemplate.UUID)
+	th.AssertNoErr(t, err)
+	defer DeleteCluster(t, client, clusterID)
+
+	t.Run("list", func(t *testing.T) { testNodeGroupsList(t, client, clusterID) })
+	t.Run("listone-get", func(t *testing.T) { testNodeGroupGet(t, client, clusterID) })
+}
+
+func testNodeGroupsList(t *testing.T, client *gophercloud.ServiceClient, clusterID string) {
+	allPages, err := nodegroups.List(client, clusterID, nil).AllPages()
+	th.AssertNoErr(t, err)
+
+	allNodeGroups, err := nodegroups.ExtractNodeGroups(allPages)
+	th.AssertNoErr(t, err)
+
+	// By default two node groups should be created
+	th.AssertEquals(t, 2, len(allNodeGroups))
+}
+
+func testNodeGroupGet(t *testing.T, client *gophercloud.ServiceClient, clusterID string) {
+	listOpts := nodegroups.ListOpts{
+		Role: "worker",
+	}
+	allPages, err := nodegroups.List(client, clusterID, listOpts).AllPages()
+	th.AssertNoErr(t, err)
+
+	allNodeGroups, err := nodegroups.ExtractNodeGroups(allPages)
+	th.AssertNoErr(t, err)
+
+	// Should be one worker node group
+	th.AssertEquals(t, 1, len(allNodeGroups))
+
+	ngID := allNodeGroups[0].UUID
+
+	ng, err := nodegroups.Get(client, clusterID, ngID).Extract()
+	th.AssertNoErr(t, err)
+
+	// Should have got the same node group as from the list
+	th.AssertEquals(t, ngID, ng.UUID)
+	th.AssertEquals(t, "worker", ng.Role)
+}
--- a/openstack/containerinfra/v1/nodegroups/doc.go
+++ b/openstack/containerinfra/v1/nodegroups/doc.go
@@ -1,0 +1,57 @@
+/*
+Package nodegroups provides methods for interacting with the Magnum node group API.
+
+All node group actions must be performed on a specific cluster,
+so the cluster UUID/name is required as a parameter in each method.
+
+
+Create a client to use:
+
+    opts, err := openstack.AuthOptionsFromEnv()
+    if err != nil {
+        panic(err)
+    }
+
+    provider, err := openstack.AuthenticatedClient(opts)
+    if err != nil {
+        panic(err)
+    }
+
+    client, err := openstack.NewContainerInfraV1(provider, gophercloud.EndpointOpts{Region: os.Getenv("OS_REGION_NAME")})
+    if err != nil {
+        panic(err)
+    }
+
+    client.Microversion = "1.9"
+
+
+Example of Getting a node group:
+
+    ng, err := nodegroups.Get(client, clusterUUID, nodeGroupUUID).Extract()
+    if err != nil {
+    	panic(err)
+    }
+    fmt.Printf("%#v\n", ng)
+
+
+Example of Listing node groups:
+
+    listOpts := nodegroup.ListOpts{
+    	Role: "worker",
+    }
+
+    allPages, err := nodegroups.List(client, clusterUUID, listOpts).AllPages()
+    if err != nil {
+    	panic(err)
+    }
+
+    ngs, err := nodegroups.ExtractNodeGroups(allPages)
+    if err != nil {
+    	panic(err)
+    }
+
+    for _, ng := range ngs {
+    	fmt.Printf("%#v\n", ng)
+    }
+*/
+package nodegroups
--- a/acceptance/openstack/containerinfra/v1/containerinfra.go
+++ b/acceptance/openstack/containerinfra/v1/containerinfra.go
@@ -17,20 +17,20 @@ import (
 	th "github.com/gophercloud/gophercloud/testhelper"
 )
 
-// CreateClusterTemplate will create a random cluster tempalte. An error will be returned if the
-// cluster-template could not be created.
-func CreateClusterTemplate(t *testing.T, client *gophercloud.ServiceClient) (*clustertemplates.ClusterTemplate, error) {
+// CreateClusterTemplateCOE will create a random cluster template for the specified orchestration engine.
+// An error will be returned if the cluster template could not be created.
+func CreateClusterTemplateCOE(t *testing.T, client *gophercloud.ServiceClient, coe string) (*clustertemplates.ClusterTemplate, error) {
 	choices, err := clients.AcceptanceTestChoicesFromEnv()
 	if err != nil {
 		return nil, err
 	}
 
 	name := tools.RandomString("TESTACC-", 8)
-	t.Logf("Attempting to create cluster template: %s", name)
+	t.Logf("Attempting to create %s cluster template: %s", coe, name)
 
 	boolFalse := false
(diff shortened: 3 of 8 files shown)


tghartland:
I have added acceptance tests for Get/List, but in my devstack environment I had to make a change in the general cluster setup here
https://github.com/gophercloud/gophercloud/blob/bee62241ad2c5b15b27906b87e7920a87b7652f4/acceptance/openstack/containerinfra/v1/containerinfra.go#L30
to add
```go
three := 3
createOpts := clustertemplates.CreateOpts{ 
    DockerVolumeSize := &three,
```
otherwise I would get the error
```
Failure in nodegroups_test.go, line 31:
unexpected error "Bad request with: [POST http://128.141.193.156:9511/v1/clustertemplates], error message:
{\"errors\": [{\"status\": 400, \"code\": \"client\", \"links\": [],
               \"title\": \"docker volume size None GB is not valid, expecting minimum value 3GB for devicemapper storage driver\",
               \"detail\": \"docker volume size None GB is not valid, expecting minimum value 3GB for devicemapper storage driver.\",
               \"request_id\": \"\"}]}"
```
I don't know if this is specific to my devstack or not. @jtopjian do the containerinfra acceptance tests not run in the openlab testing? I don't see them in the logs/results.

tghartland:
No this is the same on our production stein deployment.
```
$ openstack coe cluster template create test --coe kubernetes --image 2886e9bf-7702-4c06-99c1-39a9e7bd5951 --external-network CERN_NETWORK
docker volume size None GB is not valid, expecting minimum value 3GB for devicemapper storage driver (HTTP 400)
```
It's only a problem when using the devicemapper storage driver, setting docker_storage_driver to overlayfs allows the volume size to be unset.

I think this needs a separate fix to the acceptance test to set the volume size, as long as it doesn't break anything else in the testing environment.

tghartland:
The issue with the storagedriver was part of a larger problem, that moving on to the Create acceptance tests helped me figure out.

To begin with, the docker swarm template that is created [here](https://github.com/gophercloud/gophercloud/blob/bee62241ad2c5b15b27906b87e7920a87b7652f4/acceptance/openstack/containerinfra/v1/containerinfra.go#L31) is not compatible for using node groups with (the full functionality of node groups requires a kubernetes cluster).

Since the node group tests will have to create their own template using kubernetes anyway, that template can use overlay2 as the storage driver which fixes that issue. There was also a problem with the template dns_nameserver which was preventing nodes from starting in our environment. I'm running with it pointing to our internal DNS for now and I'll leave it as 8.8.8.8 when I push the code here.

Next was an issue with timeouts. The acceptance tests for containerinfra/v1/clusters do not check this error here https://github.com/gophercloud/gophercloud/blob/bee62241ad2c5b15b27906b87e7920a87b7652f4/acceptance/openstack/containerinfra/v1/clusters_test.go#L22
but since I had added an assert for this error in the node group tests I was seeing a time out from the WaitFor function that it uses after creating the cluster, which has a hard-coded timeout of 300 seconds
https://github.com/gophercloud/gophercloud/blob/bee62241ad2c5b15b27906b87e7920a87b7652f4/acceptance/tools/tools.go#L16-L18

(this means that the containerinfra/v1/clusters tests start running on a cluster that is still in CREATE_IN_PROGRESS if the 300 second timeout is reached).

Cluster creation of a kubernetes cluster takes ~500 seconds in my devstack, so I have also had to duplicate the cluster creation function. In this I will manually loop and check the status instead of using tools.WaitFor so that I can extend the timeout.

I'll move the cluster and cluster template changes back from my Create branch to this one once I've made sure that everything is working properly.

jtopjian:
@tghartland Thank you for all of the work investigating these issues. Please feel free to update anything you see as needing fixed.

tghartland:
Acceptance test updated to use its own createTemplate and createCluster functions.

```
        nodegroups_test.go:126: Attempting to create cluster: TESTACC-94gWjZoc using template dd01f685-3d9d-45d8-b439-8ad73fad9ac9
        nodegroups_test.go:151: Cluster Create Request ID: req-5ca32995-eccf-4a6b-b8c8-e39d685d0a7e
        nodegroups_test.go:159: Cluster created: 9dc1e2a5-d826-4a95-9364-bbf02a6e552b
        nodegroups_test.go:179: Successfully created cluster: TESTACC-94gWjZoc id: 9dc1e2a5-d826-4a95-9364-bbf02a6e552b
    --- PASS: TestNodeGroupsCRUD/list (0.15s)
    --- PASS: TestNodeGroupsCRUD/listone-get (0.30s)
        containerinfra.go:137: Attempting to delete cluster: 9dc1e2a5-d826-4a95-9364-bbf02a6e552b
        containerinfra.go:155: Successfully deleted cluster: 9dc1e2a5-d826-4a95-9364-bbf02a6e552b
        containerinfra.go:76: Attempting to delete cluster-template: dd01f685-3d9d-45d8-b439-8ad73fad9ac9
        containerinfra.go:83: Successfully deleted cluster-template: dd01f685-3d9d-45d8-b439-8ad73fad9ac9
PASS
ok      github.com/gophercloud/gophercloud/acceptance/openstack/containerinfra/v1       561.073s
```

tghartland:
@jtopjian I've finished the acceptance tests for create/update/delete and there's nothing else that would need to be propagated back to here, so it's ready for review.

jtopjian:
@tghartland Thanks! I'm going to see about getting Magnum/ContainerInfra re-enabled in OpenLab so we can have these tested regularly.

I might need you to rebase this if I'm able to get that working.

tghartland:
@jtopjian I can rebase whenever you're ready.

Is there a timeout for the tests in OpenLab? For my testing I had to add `-timeout=20m` to the script provided [here](https://github.com/gophercloud/gophercloud/tree/master/acceptance#3-run-the-test-suite). The full test with creating a cluster plus creating another node group took 15 minutes in devstack.

jtopjian:
I know OpenLab has a 3 hour global time limit set. I'll look at increasing the individual test timeout for the Magnum tests - thanks for the heads up.

And just to confirm: did you have to make any changes to the devstack environment or are the tests able to run in a "vanilla" environment?

tghartland:
There aren't any major changes, the only thing that might be an issue is that the devstack I'm using is a little out of date
```
commit 7840b6e291b5e28ba96ce871b5e01d66c14d9201 (HEAD -> master, origin/master, origin/HEAD)
Merge: 56f23cc5 c67a689f
Author: Zuul <zuul@review.opendev.org>
Date:   Tue Oct 8 19:33:53 2019 +0000
```
It also has [this fix](https://review.opendev.org/#/c/694230/) which isn't merged yet but the List tests don't use any paginated responses so that shouldn't be a problem.

acceptance/openstack/containerinfra/v1/nodegroups_test.go
```diff
@@ -0,0 +1,181 @@
...
+
+	ng, err := nodegroups.Get(client, clusterID, ngID).Extract()
+	th.AssertNoErr(t, err)
+
+	// Should have got the same node group as from the list
+	th.AssertEquals(t, ngID, ng.UUID)
+	th.AssertEquals(t, "worker", ng.Role)
+}
+
+func createKubernetesClusterTemplate(t *testing.T, client *gophercloud.ServiceClient) (*clustertemplates.ClusterTemplate, error) {
```
jtopjian: nit: Can you move this to `containerinfra.go` as `CreateKubernetesClusterTemplate`? This way it can be reusable for other tests

acceptance/openstack/containerinfra/v1/nodegroups_test.go
```diff
@@ -0,0 +1,181 @@
...
+	tools.PrintResource(t, clusterTemplate.CreatedAt)
+
+	th.AssertEquals(t, name, clusterTemplate.Name)
+	th.AssertEquals(t, choices.ExternalNetworkID, clusterTemplate.ExternalNetworkID)
+	th.AssertEquals(t, choices.MagnumImageID, clusterTemplate.ImageID)
+
+	return clusterTemplate, nil
+}
+
+func createKubernetesCluster(t *testing.T, client *gophercloud.ServiceClient, clusterTemplateID string) (string, error) {
```
jtopjian: Same here

openstack/containerinfra/v1/nodegroups/doc.go
```diff
@@ -0,0 +1,68 @@
...
+
+
+Example of Listing node groups:
+
+    // Options:
+    // Marker:  pagination marker for large data sets. (UUID field from node group).
+    // Limit:   maximum number of resources to return in a single result.
+    // SortKey: column to sort results by. Default: id.
+    // SortDir: direction to sort. "asc" or "desc". Default: asc.
+    // Role:    list all nodegroups with the specified role.
```
jtopjian: nit: The above docs can be removed. The ListOpts struct with comments will be formatted in a readable way in godoc.

openstack/containerinfra/v1/nodegroups/doc.go
```diff
@@ -0,0 +1,68 @@
...
+    	panic(err)
+    }
+
+    for _, ng := range ngs {
+    	fmt.Printf("%#v\n", ng)
+    }
+
+    // When listing node groups, only these fields are returned
+    // "uuid", "name", "flavor_id", "node_count", "role",
+    // "is_default", "image_id", "status", "stack_id"
```
jtopjian: nit: This can be removed, too. The result struct will be rendered in a readable way. If you'd like add annotation comments to the relevant fields, that might work better:

```go
// UUID is the unique ID of the node group.
// This is returned both in List and Get.
UUID string `json:"uuid"`
```

acceptance/openstack/containerinfra/v1/nodegroups_test.go
tghartland: In that case rather than just duplicating the code and changing a few lines for the timeout I'd prefer to make it work properly.
I will make a separate PR for [these changes](https://github.com/tghartland/gophercloud/compare/master...tghartland:containerinfra-acceptance-timeout) which I can then use here.

I can't test it yet because a power cut broke my devstack, but I'll open that pull request as soon as possible.

tghartland: I've been able to test the changes I want to make in devstack on a virtual machine, it took a lot longer than the physical machine I was using so there will definitely need to be some tweaking to get the timeouts right for the openlab test environment.

The tests took 40 minutes to complete on a virtual machine devstack, if it's anything like that in the CI then it might be worth changing the existing cluster tests so that the node group tests are done half way through. This prevents having to create a second cluster just for the node group tests, but would sacrifice having clear separation of tests for the two separate APIs. 

To get the cluster tests working on Magnum master branch I also had to change the cluster template, for the same `docker_storage_driver` issue. Hopefully that won't cause problems with old magnum versions.

I have my physical machine devstack back now, the clusters CRUD test runs in 4 minutes thankfully.

I've opened a PR for the changes here: https://github.com/gophercloud/gophercloud/pull/1794.

jtopjian: >  if it's anything like that in the CI then it might be worth changing the existing cluster tests so that the node group tests are done half way through.

My goal is to have the cluster tests run as a separate job, like the ironic tests are currently. That should help keep all testing to around 60-70 minutes.

> Hopefully that won't cause problems with old magnum versions.

If it does, I'll try to manage that. The current Magnum support has all been done ad-hoc against any cloud we could find, and then aggregating the results, so it's always been a little hacky.

Thanks so much for working on this.

acceptance/openstack/containerinfra/v1/nodegroups_test.go
tghartland: Done, and since I had changed the template storage driver to overlay2 in #1794 it was possible to create the template in a common function `CreateClusterTemplateCOE`, with wrappers for creating swarm or kubernetes (keeping the default as swarm).

acceptance/openstack/containerinfra/v1/nodegroups_test.go
tghartland: Done using the changes in #1794 to change the timeout for creating a kubernetes cluster. The timeouts will probably need to be adjusted once we get the results from the CI.

openstack/containerinfra/v1/nodegroups/doc.go
tghartland: Done.

openstack/containerinfra/v1/nodegroups/doc.go
tghartland: I've moved this information to the documentation of `List` itself. I think it's more readable that way than spreading it out over multiple lines of the NodeGroup struct.

https://github.com/gophercloud/gophercloud/pull/1774/commits/f911976aad959fa0b3ba9984657b43f4af11e67f#diff-781e55e8261bfc534c6345810f539b08R55-R59

tghartland:
@jtopjian thanks for your review. I've just pushed the suggested changes along with a rebase onto master for the changes in #1794.

PR #1794, opened 2019-12-05
Containerinfra: acceptance test timeout changes and fixes

tghartland:
For #1767

Adds configurable timeouts to cluster creation which will be needed for kubernetes clusters, to be able to test node groups. A kubernetes cluster needs longer than 5 minutes to create.

There are also some fixes for the existing clusters API acceptance tests.
* Check timeout error when creating the cluster
* Fix update opts value type for replacing node count
* Change template storage driver

I have tried to make as few changes and additions as possible to the existing code, by making the existing functions just call to the versions that accept a timeout.

With these fixes, the cluster acceptance tests pass in my devstack on magnum master branch.
For the tests in CI, the timeouts may need to be tweaked.

### DIFF
--- a/acceptance/openstack/containerinfra/v1/clusters_test.go
+++ b/acceptance/openstack/containerinfra/v1/clusters_test.go
@@ -42,7 +42,7 @@ func TestClustersCRUD(t *testing.T) {
 		clusters.UpdateOpts{
 			Op:    clusters.ReplaceOp,
 			Path:  "/node_count",
-			Value: "2",
+			Value: 2,
 		},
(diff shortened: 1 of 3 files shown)


jtopjian:
recheck

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
@tghartland I'm going to merge this one, too. Once we get Magnum integrated in OpenLab, we can look at fixing any errors, if needed.

Thank you very much for your work on this.

PR #1795, opened 2019-12-09
[WIP]Switch to install devstack on ubuntu 18.04

huangtianhua:
This change switch to install devstack on ubuntu
bionic, because openstack/devstack has dropped
xenial support.

Related-Bug: theopenlab/openlab#412

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -12,6 +12,7 @@
     description: |
       Run gophercloud acceptance test on master branch
     run: .zuul/playbooks/gophercloud-acceptance-test/run.yaml
+    nodeset: ubuntu-bionic
 
 - job:
     name: gophercloud-acceptance-test-ironic
@@ -19,6 +20,7 @@
     description: |
       Run gophercloud ironic acceptance test on master branch
     run: .zuul/playbooks/gophercloud-acceptance-test-ironic/run.yaml
+    nodeset: ubuntu-bionic
 
 - job:
     name: gophercloud-acceptance-test-stein
--- a/.zuul/playbooks/gophercloud-acceptance-test/run.yaml
+++ b/.zuul/playbooks/gophercloud-acceptance-test/run.yaml
@@ -19,7 +19,7 @@
           set -o pipefail
           set -x
           echo $(export |grep OS_BRANCH)
-
+          export OS_DEBUG=1
           go get ./... || true
           ./script/acceptancetest -v 2>&1 | tee $TEST_RESULTS_TXT
         executable: /bin/bash
--- a/acceptance/openstack/sharedfilesystems/v2/shares_test.go
+++ b/acceptance/openstack/sharedfilesystems/v2/shares_test.go
@@ -118,7 +118,7 @@ func TestGrantAndRevokeAccess(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unable to create a shared file system client: %v", err)
 	}
-	client.Microversion = "2.51"
+	client.Microversion = "2.49"
 
 	share, err := CreateShare(t, client)
 	if err != nil {
--- a/acceptance/openstack/sharedfilesystems/v2/shares_test.go
+++ b/acceptance/openstack/sharedfilesystems/v2/shares_test.go
@@ -118,7 +118,7 @@ func TestGrantAndRevokeAccess(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Unable to create a shared file system client: %v", err)
 	}
-	client.Microversion = "2.7"
+	client.Microversion = "2.51"
 
 	share, err := CreateShare(t, client)
 	if err != nil {


tghartland:
recheck

huangtianhua:
recheck

jtopjian:
The test failure is from Manila and should be a transient failure. The important part is that the tests are running again!

jtopjian:
recheck

jtopjian:
recheck

jtopjian:
If this fails again, I might have to look at the failing test in more detail.

huangtianhua:
Seems the test failed:
2019-12-11 06:31:23.492599 | ubuntu-bionic | --- FAIL: TestGrantAndRevokeAccess (6.99s)
2019-12-11 06:31:23.492696 | ubuntu-bionic |     tools.go:80: {
2019-12-11 06:31:23.492756 | ubuntu-bionic |           "share_id": "708092b7-9445-4e00-9811-bdfc7f0d231e",
2019-12-11 06:31:23.492792 | ubuntu-bionic |           "access_type": "ip",
2019-12-11 06:31:23.492830 | ubuntu-bionic |           "access_to": "0.0.0.0/32",
2019-12-11 06:31:23.492865 | ubuntu-bionic |           "access_level": "ro",
2019-12-11 06:31:23.492896 | ubuntu-bionic |           "state": "new",
2019-12-11 06:31:23.492944 | ubuntu-bionic |           "id": "6c6c5365-4c92-48e1-8cbd-a7a9731714f4"
2019-12-11 06:31:23.492965 | ubuntu-bionic |         }
2019-12-11 06:31:23.493071 | ubuntu-bionic |     shares.go:93: Failed to wait for 'deleted' status for 708092b7-9445-4e00-9811-bdfc7f0d231e share: An error occurred, wrong status: error_deleting

jtopjian:
@huangtianhua Right. I'm not sure what the core issue is. Manila is one of the more harder services for me to test outside of OpenLab. I'm not sure if the root cause is that the share is being deleted before access is revoked or if it is something new being caused by Ubuntu 18.04.

If this is the only thing preventing us from merging this PR, then I'll merge this PR and disable the test for the time being.

Is there anything left for OpenLab to do in order to finish the Ubuntu 18.04 support or is this ready to go?

huangtianhua:
@jtopjian, OK, I will have a look.

huangtianhua:
recheck

tghartland:
> I'm not sure if the root cause is that the share is being deleted before access is revoked or if it is something new being caused by Ubuntu 18.04.

I tried a few things to try and cause a race condition like this, running the test against our production deployment (api version 2.49, stein). I had to change `CreateShare` to use our cephfs share type and create a non-public share, but without making any changes to `TestGrantAndRevokeAccess` it passed every time.

I tried changing the `defer DeleteShare` to `go DeleteShare`, in which case the error is 
```
{"badRequest": {"message": "Access rules cannot be denied while the share, any of its replicas or migration copies lacks a valid host or is in an invalid state.", "code": 400}}
```
Changing it to just `DeleteShare` so that it is deleted even before the grant access:
```
Unable to grant access: Resource not found
```
Using `go DeleteShare` and moving it to be just before the revoke access either passed or failed with the same error as the first change.

Using `go ...` for `DeleteShare` and for `RevokeAccess` (calling them in either order) mostly passed and sometimes failed with
```
Unable to revoke access: Resource not found
```

By having a `defer DeleteShare` and calling `DeleteShare` at the end of the function so that it runs twice:
```
Unable to delete share 54332ea3-649f-4c7b-bd52-0489b75f53d1: Resource not found
```

Adding an `ExtendShare` to the end of the test, right before the deferred deletion
```
shares.go:88: Unable to delete share 77c5f4df-90a3-4b7b-8ce7-27b44ee1f4c3: Request forbidden: error message: {"forbidden": {"message": "Invalid share: Share status must be one of ('available', 'error', 'inactive').", "code": 403}}
shares.go:93: Failed to wait for 'deleted' status for 77c5f4df-90a3-4b7b-8ce7-27b44ee1f4c3 share: A timeout occurred
```
This shouldn't have gone to a timeout, there should be a return in this error check here
https://github.com/gophercloud/gophercloud/blob/f940f50ff1f774107ecff13425265c5361682f74/acceptance/openstack/sharedfilesystems/v2/shares.go#L86-L89
but that's unrelated.

None of these match the error in OpenLab, so it probably is something introduced by 18.04, or maybe something about the share type `NFS` that I'm not able to use.

tghartland:
@huangtianhua good idea unsetting the microversion. I can replicate that result against my production deployment, getting exactly the same error.
```
--- FAIL: TestGrantAndRevokeAccess (30.45s)
    shares_test.go:132: Unable to grant access: Resource not found
```
Adding a sleep after the share creation did not help, but by setting the microversion to 2.49 the test passed successfully. Please try that next in the CI.

jtopjian:
Thank you both for looking into this issue. 

I'm going to merge this PR and then disable the failing tests. At least that will allow us to have the majority of the tests being run again.

I'll then open another Pull Request where we can further troubleshoot these failing tests.

Issue #1798, opened 2019-12-10
Support for SHELVE/UNSHELVE


qhua948:
After reading through the document, I believe shelve/shelve action are currently not supported.

Can this be added?

jtopjian:
Implemented in #1799

PR #1799, opened 2019-12-10
Added Shelve, Shelve-offload and Unshelve API

qhua948:

For #1798 

code in this PR:

https://docs.openstack.org/api-guide/compute/server_concepts.html
https://docs.openstack.org/api-ref/compute/?expanded=shelve-server-shelve-action-detail,unshelve-restore-shelved-server-unshelve-action-detail,shelf-offload-remove-server-shelveoffload-action-detail

### Controllers
https://github.com/openstack/nova/blob/17b5a1ab85c358a1096788371849f0cccfe84972/nova/api/openstack/compute/shelve.py#L40
https://github.com/openstack/nova/blob/17b5a1ab85c358a1096788371849f0cccfe84972/nova/api/openstack/compute/shelve.py#L60
https://github.com/openstack/nova/blob/17b5a1ab85c358a1096788371849f0cccfe84972/nova/api/openstack/compute/shelve.py#L78

https://github.com/openstack/nova/blob/c6218428e9b29a2c52808ec7d27b4b21aadc0299/nova/compute/instance_actions.py#49

### APIs
https://github.com/openstack/nova/blob/6695b5633ba34b758b7f742985bc21122acb2637/nova/compute/api.py#L3906
https://github.com/openstack/nova/blob/6695b5633ba34b758b7f742985bc21122acb2637/nova/compute/api.py#L3930
https://github.com/openstack/nova/blob/6695b5633ba34b758b7f742985bc21122acb2637/nova/compute/api.py#L3994

https://github.com/openstack/nova/blob/c45d9da8d1c9b9e40a43965ea757a74015272ded/nova/compute/rpcapi.py#L1286

### Schema
https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/schemas/shelve.py

### DIFF
--- a/openstack/compute/v2/extensions/shelveunshelve/requests.go
+++ b/openstack/compute/v2/extensions/shelveunshelve/requests.go
@@ -23,14 +23,14 @@ type UnshelveOptsBuilder interface {
 	ToUnshelveMap() (map[string]interface{}, error)
 }
 
-// ShelveOffloadOpts specifies parameters of shelve-offload action.
-type UnshelvedOpts struct {
+// UnshelveOpts specifies parameters of shelve-offload action.
+type UnshelveOpts struct {
 	// Sets the availability zone to unshelve a server
 	// Available only after nova 2.77
 	AvailabilityZone string `json:"availability_zone,omitempty"`
 }
 
-func (opts UnshelvedOpts) ToUnshelveMap() (map[string]interface{}, error) {
+func (opts UnshelveOpts) ToUnshelveMap() (map[string]interface{}, error) {
 	// Key 'availabilty_zone' is required if the unshelve action is an object
 	// i.e {"unshelve": {}} will be rejected
 	b, err := gophercloud.BuildRequestBody(opts, "unshelve")
--- a/openstack/compute/v2/extensions/shelveunshelve/requests.go
+++ b/openstack/compute/v2/extensions/shelveunshelve/requests.go
@@ -17,8 +17,41 @@ func ShelveOffload(client *gophercloud.ServiceClient, id string) (r ShelveOffloa
 	return
 }
 
+// UnshelveOptsBuilder allows extensions to add additional parameters to the
+// Unshelve request.
+type UnshelveOptsBuilder interface {
+	ToUnshelveMap() (map[string]interface{}, error)
+}
+
+// ShelveOffloadOpts specifies parameters of shelve-offload action.
+type UnshelvedOpts struct {
+	// Sets the availability zone to unshelve a server
+	// Available only after nova 2.77
+	AvailabilityZone string `json:"availability_zone,omitempty"`
+}
+
+func (opts UnshelvedOpts) ToUnshelveMap() (map[string]interface{}, error) {
+	// Key 'availabilty_zone' is required if the unshelve action is an object
+	// i.e {"unshelve": {}} will be rejected
+	b, err := gophercloud.BuildRequestBody(opts, "unshelve")
+	if err != nil {
+		return nil, err
+	}
+
+	if _, ok := b["unshelve"].(map[string]interface{})["availability_zone"]; !ok {
+		b["unshelve"] = nil
+	}
+
+	return b, err
+}
+
 // Unshelve is the operation responsible for unshelve a Compute server.
-func Unshelve(client *gophercloud.ServiceClient, id string) (r UnshelveResult) {
-	_, r.Err = client.Post(extensions.ActionURL(client, id), map[string]interface{}{"unshelve": nil}, nil, nil)
+func Unshelve(client *gophercloud.ServiceClient, id string, opts UnshelveOptsBuilder) (r UnshelveResult) {
+	b, err := opts.ToUnshelveMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	_, r.Err = client.Post(extensions.ActionURL(client, id), b, nil, nil)
 	return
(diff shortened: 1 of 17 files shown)


qhua948:
recheck

openstack/compute/v2/extensions/shelveunshelve/requests.go
```diff
@@ -0,0 +1,25 @@
+package shelveunshelve
+
+import "github.com/gophercloud/gophercloud"
+
+func actionURL(client *gophercloud.ServiceClient, id string) string {
```
jtopjian: This should be moved to a new file called `urls.go`.

openstack/compute/v2/extensions/shelveunshelve/testing/doc.go
```diff
@@ -0,0 +1,2 @@
+// shelveunshelve unit tests
+package testing
```
jtopjian: This file isn't needed and can be safely omitted.

qhua948:
@jtopjian Thanks, I'll work on this later.

openstack/compute/v2/extensions/shelveunshelve/requests.go
qhua948: This is how it is it currently done in some other extensions, do you want this to be a big refractor?

jtopjian: I don't see how moving a function to a new file called `urls.go` is a big refactor. 

You're correct in that a few extensions are missing a `urls.go` file, but the vast majority of packages in this repo follow that pattern. I'd rather catch this inconsistency ahead of time.

If you don't want to, that's fine - I'll just do it later.

qhua948: I could do that as part of this PR, not a problem.

jtopjian: Just to clarify: you'd only need to add `urls.go` for this new package - not fix up any of the other packages. That can be done during some general cleanup at another time.

jtopjian:
@qhua948 Thank you for digging up the API code - it's a pain, but it can sometimes identify areas that aren't reflected in the docs. For this PR, the docs match the API code, so the exercise feels moot, but there's been enough occasions worth the effort.

It looks like the unshelve action can take an optional parameter of `availability_zone`: 

https://github.com/openstack/nova/blob/17b5a1ab85c358a1096788371849f0cccfe84972/nova/api/openstack/compute/shelve.py#L79-L81

We'll then need to add a struct called `UnshelveOpts` with `AvailabilityZone` field. Then we'll need to make sure the struct is conditionally sent only if it's specified, and null when not, in order to be backwards compatible. 

No doubt that's a bit more work than you anticipated by wanting to add support for basic un/shelving. I can add the . `AvailabilityZone` feature after this PR is merged. Just a heads up that the function signature for `Unshelve` will be changed shortly after this is merged, in case you wanted to use this feature right away.

qhua948:
Thanks for pointing it out, for the sake of completeness, I should add that in as well.

openstack/compute/v2/extensions/shelveunshelve/requests.go
qhua948: Since I've refactored it already, would you prefer to keep it this way or I can pull it out.

jtopjian: Let's go ahead and keep it - thanks for doing that :)

jtopjian:
@qhua948 let me know if you need any help. A rough draft of the conditional logic is going to be something like:

```go
func (opts UnshelveOpts) ToShelveUnshelveMap() (map[string]interface{}, error) {
  b, err := gophercloud.BuildRequestBody(opts, "")
  if err != nil {
    return nil, err
  }

  unshelveOpts := make(map[string]interface{})
  if b["availability_zone"] != "" {
    unshelveOpts["availability_zone"] = b["availability_zone"]
  }

  return unshelveOpts
}

func Unshelve(client *gophercloud.ServiceClient, id string, opts ShelveUnshelveOptsBuilder) (r UnshelveResult) {
  b, err := opts.ToShelveUnshelveOpts()
  if err != nil {
    r.Err = err
    return
  }

  _, r.Err = client.Post(extensions.ActionURL(client, id), map[string]interface{}{"unshelve": b}, nil, nil)
  return
}
```

I might be wrong about `make`, though - a unit test will confirm if that will end up sending `nil` or `{}`.

qhua948:
### Schema

https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/schemas/shelve.py

openstack/compute/v2/extensions/shelveunshelve/requests.go
qhua948: Cheers

openstack/compute/v2/extensions/shelveunshelve/requests.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// UnshelveOptsBuilder allows extensions to add additional parameters to the
+// Unshelve request.
+type UnshelveOptsBuilder interface {
+	ToUnshelveMap() (map[string]interface{}, error)
+}
+
+// ShelveOffloadOpts specifies parameters of shelve-offload action.
+type UnshelvedOpts struct {
```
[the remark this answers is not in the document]
qhua948: Indeed!

PR #1800, opened 2019-12-13
Containerinfra: nodegroups Create/Update/Delete

tghartland:
For #1767 

Create method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L283-L323

Available fields when creating node group:
https://github.com/openstack/python-magnumclient/blob/c0d368360831cd29affa47e77eba4e81e4c6f0f8/magnumclient/v1/nodegroups.py#L21-L23

Update method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L326-L338

Delete method:
https://github.com/openstack/magnum/blob/9c7cf5d6eb050dbbcdcb036bd0af2b6586070c0c/magnum/api/controllers/v1/nodegroup.py#L341-L355

### DIFF
--- a/acceptance/openstack/containerinfra/v1/nodegroups_test.go
+++ b/acceptance/openstack/containerinfra/v1/nodegroups_test.go
@@ -54,6 +54,8 @@ func TestNodeGroupsCRUD(t *testing.T) {
 		return (ng.Status == "CREATE_COMPLETE"), nil
 	}, 900*time.Second)
 	th.AssertNoErr(t, err)
+
+	t.Run("update", func(t *testing.T) { testNodeGroupUpdate(t, client, clusterID, nodeGroupID) })
 }
 
 func testNodeGroupsList(t *testing.T, client *gophercloud.ServiceClient, clusterID string) {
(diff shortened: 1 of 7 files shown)


tghartland:
Another thing to consider for the acceptance tests running in CI, these tests require creating four VMs.
* One master node
* One default worker node
* One node in the created node group
* And an extra in the created node group to give room to update the min_node_count, which must be <= the actual node count

I don't know how much memory the testing environment has available, if it can't support four VMs then I can change to not update the min_node_count and have three VMs in total. Lower than three is not possible.

jtopjian:
> I don't know how much memory the testing environment has available, if it can't support four VMs then I can change to not update the min_node_count and have three VMs in total. Lower than three is not possible.

That sounds good. In addition, we can always disable this test if we find it is too resource intensive. Having the acceptance test included, even if disabled, is still very useful for both documentation as well as being able to run the test manually from time to time.

Issue #1801, opened 2019-12-14
Acc Tests: Issues with Shared Filesystem Access Tests


jtopjian:
Since the acceptance test environment was moved to Ubuntu 18.04, the Shared Filesystem acceptance tests relating to "access rights" have begun to fail.

jtopjian:
Fixed in #1803

PR #1802, opened 2019-12-14
Acc Tests: Disable Shared Filesystem access tests

jtopjian:
For #1801


jtopjian:
recheck

jtopjian:
:roll_eyes:

jtopjian:
It looks like the tests are still running on the xenial node groups instead of bionic.

jtopjian:
recheck

jtopjian:
I'm not sure why the tests are still being run on xenial nodes instead of bionic, but I'm going to merge this anyway for in preparation of it being resolved.

kayrus:
recheck

PR #1803, opened 2019-12-15
Acc Tests: Enable Shared Filesystem access tests

jtopjian:
For #1801 

This PR is to troubleshoot the issues seen with the Shared Filesystem access tests.


jtopjian:
@huangtianhua even though I merged #1795, it looks like the tests are still being run on xenial nodes.

huangtianhua:
recheck

kayrus:
recheck

PR #1816, opened 2020-01-10
Barbican V1: Allow multiple ACL types in one request

### LINKED ISSUE #1763: Barbican V1: Allow multiple ACL types in one request
Current implementation supports one call per type. Even though it is not yet supported in Barbican to have an ACL type other than "read", for the future it would be nice to handle multiple ACL types at one call and avoid complex logic in dependent applications. I expect gophercloud to construct json request like:

```json
{
  "read": {
    "project-access": false,
    "users": [
      "GG27dVwR9gBMnsOaRoJ1DFJmZfdVjIdW"
    ]
  },
  "write": {
    "project-access": true,
    "users": [
      "35t0SzHNIpfiCWppFNpcyDmPQCsESiB1"
    ]
  }
}`
```

I propose the following modification to https://github.com/gophercloud/gophercloud/blob/6ad562af8c1fd8a69397a13caf58fa184d115b56/openstack/keymanager/v1/acls/requests.go#L26

```go
// SetOpts represents options to set an ACL on a resource.
type SetOpts []SetOpt

// ToACLSetMap formats a SetOpts into a set request.
func (opts SetOpts) ToACLSetMap() (map[string]interface{}, error) {
        b := make(map[string]interface{})
        for _, v := range opts {
                m, err := gophercloud.BuildRequestBody(v, v.Type)
                if err != nil {
                        return nil, err
                }
                b[v.Type] = m[v.Type]
        }
        return b, nil
}
```

https://github.com/openstack/barbican/blob/c40ef6f7d865858a26b934be52e193ab2fb3fd26/barbican/common/validators.py#L45

kayrus:
Resolves #1763

### DIFF
--- a/acceptance/openstack/keymanager/v1/acls_test.go
+++ b/acceptance/openstack/keymanager/v1/acls_test.go
@@ -30,9 +30,11 @@ func TestACLCRUD(t *testing.T) {
 	users := []string{user}
 	iFalse := false
 	setOpts := acls.SetOpts{
-		Type:          "read",
-		Users:         &users,
-		ProjectAccess: &iFalse,
+		acls.SetOpt{
+			Type:          "read",
+			Users:         &users,
+			ProjectAccess: &iFalse,
+		},
 	}
 
 	aclRef, err := acls.SetSecretACL(client, secretID, setOpts).Extract()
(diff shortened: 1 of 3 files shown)


kayrus:
@jtopjian ready for review

Issue #1834, opened 2020-02-02
LoadBalancer V2: lack of InsertHeaders in Listener.UpdateOpts 


ozerovandrei:
Octavia V2 API supports the update of "insert_headers" value: https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=update-a-listener-detail#update-a-listener

Gophercloud only supports this option in the Create and Read methods.

For https://github.com/terraform-providers/terraform-provider-openstack/issues/954

jtopjian:
@ozerovandrei If there are 2 headers currently set and someone wants to remove one, I'm assuming that they will set `UpdateOpts.InsertHeaders` to only have 1 entry.

But what if they want to remove all headers? Should it be `*map[string]string` instead?

ozerovandrei:
@jtopjian yeah you're right.

PR #1835, opened 2020-02-02
LoadBalancer V2: add InsertHeaders into UpdateOpts

ozerovandrei:
Add openstack/loadbalancer/v2/listeners.UpdateOpts.InsertHeaders.

For #1834

code in this PR:

https://github.com/openstack/octavia/blob/stable/train/octavia/api/v2/types/listener.py#L170


ozerovandrei:
recheck

Issue #1839, opened 2020-02-06
Malformed JSON Struct Tags


alrs:
The `Op` and `Path` fields of the `UpdateOperation` struct in `baremetal/v1/ports/requests` both have the JSON tag argument `required`, which [does not exist](https://github.com/golang/go/issues/16426) in the Go JSON library.

The `VirtualIP` member of an anonymous struct in `networking/v2/extensions/lbaas/vips` is tagged with the same JSON tag twice.

jtopjian:
@alrs Thanks.

The `required` tag should actually be:

```
`required:"true"`
```

Instead of part of the `json` tag.

https://github.com/gophercloud/gophercloud/blob/master/params.go#L64

jtopjian:
Fixed in #1840

PR #1840, opened 2020-02-06
Fix json struct tags

alrs:
For #1839 

This fixes malformed JSON struct tags in `networking/v2/extensions/lbaas/vips` and `baremetal/v1/ports`.


alrs:
Did a fixup to put `required:"true"` on `Op` and `Path` in `UpdateOperation{}`.

Issue #526, opened 2017-09-24
Support the Placement API


jaypipes:
The Placement API was [introduced](https://github.com/openstack/nova/commit/2ae10ce5225f593dd6b0d784adafa34939c26260#diff-fcca11e34c1b5fce52a4ddbc418aa2d5) to OpenStack in the Newton release.

It allows **administrative** users to request and modify data about resource providers, inventory records, allocation records and grouping of providers (aggregates).

While the Placement API is currently a directory within the nova source tree, it is an entirely independent service from the OpenStack Compute API. It has its own service endpoint in the Keystone catalog and has its own [REST API microversion history](https://github.com/openstack/nova/blob/9d6acfaf7cd09710accbbae4486114d1707515d0/nova/api/openstack/placement/rest_api_version_history.rst).

It would be great to add support to `gophercloud` for the placement API service.

I'm eager to contribute this if cores give me the go-ahead.

jtopjian:
> I'm eager to contribute this if cores give me the go-ahead.

Sure - go for it :)

For entire feature sets such as this, we like to break things up into per-method PRs. Picking something arbitrary from the link you have, each of [these](https://github.com/openstack/nova/blob/9d6acfaf7cd09710accbbae4486114d1707515d0/nova/api/openstack/placement/rest_api_version_history.rst#16-traits-api) would be a separate PR.

You can bundle them up into a single Issue like how it was done in https://github.com/gophercloud/gophercloud/issues/287.

Usually the first PR, GET, will be the largest since it'll include a lot of boilerplate code and the resource's resulting structure. You can "chain" branches/PRs together since the previous branch/PR will naturally use code from the first. If you submit PRs this way, prefix the title such as:

```
Placement: GET traits
[pending #above] Placement: PUT traits
```

This can cause a rebase chain if a parent PR needed changes. It's admittedly a nuisance, but it does help keep things small.

Let me know if you have any questions or need any help.

jtopjian:
@jaypipes Quick question: You mentioned the word "aggregates". Does the Placement API modify the functionality of the [os-aggregates](https://developer.openstack.org/api-ref/compute/#host-aggregates-os-aggregates) API calls?

I was just about to open an issue to add `os-aggregates` functionality and wanted to know if there will be some overlap. No doubt `os-aggregates` will still be applicable for pre-placement environments, but future behavior will be good to note.

jaypipes:
[...]

PR #1841, opened 2020-02-06
Add resource provider create API for placement.

dkt26111:
For #526

code in this PR:

API doc:
https://docs.openstack.org/api-ref/placement/#create-resource-provider

API code:
https://github.com/openstack/placement/blob/master/placement/handlers/resource_provider.py#L77


dkt26111:

jtopjian:
@dkt26111 This all looks good. Would you be able to add an example to `doc.go`?

Issue #1845, opened 2020-02-11
Implem compute servers instance_actions sdk


snigle:
Hi, this call api is not in the SDK : 
https://docs.openstack.org/api-ref/compute/?expanded=list-actions-for-server-detail

jtopjian:
@snigle You're right - thanks! 

This should go under "extensions": https://github.com/gophercloud/gophercloud/tree/master/openstack/compute/v2/extensions

reenjii:
Correct link to documentation: https://docs.openstack.org/api-ref/compute/?expanded=list-actions-for-server-detail#list-actions-for-server

jtopjian:
Implemented in #1848

PR #1846, opened 2020-02-11
add openstack server os-instance-actions sdk For #1845

snigle:
For #1845

code in this PR:

https://docs.openstack.org/api-ref/compute/?expanded=list-actions-for-server-detail
https://github.com/openstack/nova/blob/51ed40a6a5fc046cef35337980a1fc5ad704a421/nova/api/openstack/compute/instance_actions.py#L83


snigle:
will reopen with complete os-instance-action sdk

PR #1848, opened 2020-02-12
List instance actions

snigle:
For gophercloud#1845

code in this PR:

https://docs.openstack.org/api-ref/compute/?expanded=list-actions-for-server-detail
https://github.com/openstack/nova/blob/51ed40a6a5fc046cef35337980a1fc5ad704a421/nova/api/openstack/compute/instance_actions.py#L83

### DIFF
--- a/openstack/compute/v2/extensions/instanceactions/results.go
+++ b/openstack/compute/v2/extensions/instanceactions/results.go
@@ -1,0 +1,78 @@
+package instanceactions
+
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// InstanceAction represents an instance action.
+type InstanceAction struct {
+	Action       string    `json:"action"`
+	InstanceUUID string    `json:"instance_uuid"`
+	Message      string    `json:"message"`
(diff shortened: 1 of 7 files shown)


openstack/compute/v2/extensions/instanceactions/results.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+// InstanceAction represents an instance action.
+type InstanceAction struct {
+	Action       string    `json:"action"`
+	InstanceUUID string    `json:"instance_uuid"`
+	Message      *string   `json:"message"`
```
jtopjian: Having this as `*string` will cause the value to be `nil` if there is no message. Is there a reason to do this instead of having an empty value be `""`? If there is, I'm OK with keeping this as `*string`.

snigle: It permits to say to the dev that the field be empty.
If it's a pointer is easier for the dev to konw the field is optional and must be checked before used.

snigle:
Commit of review squashed : 
https://github.com/gophercloud/gophercloud/commit/70a4f48196204bf07e4f434275ac80a0dd3d9952

PR #1849, opened 2020-02-12
add fixedIPs filter on network port

snigle:
For gophercloud#1847

code in this PR:

https://docs.openstack.org/api-ref/network/v2/?expanded=list-ports-detail

https://github.com/openstack/neutron-lib/blob/478502b3dfb8c3ff66318169377251826563a398/neutron_lib/api/validators/__init__.py#L514-L551

https://github.com/openstack/neutron-lib/blob/56033ba643812a30577f6ab17648806c2ee494ba/neutron_lib/api/definitions/port.py#L62-L69

https://github.com/openstack/neutron/blob/79e6da0004a0e4e6996d2443121fdf84ae0eb34d/neutron/plugins/ml2/plugin.py#L2241

### DIFF
--- a/openstack/networking/v2/ports/requests.go
+++ b/openstack/networking/v2/ports/requests.go
@@ -1,6 +1,10 @@
 package ports
 
 import (
+	"fmt"
+	"net/url"
+	"strings"
+
 	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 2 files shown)


jtopjian:
recheck

Issue #1853, opened 2020-02-15
Microversions: Simplify Implementation of additional result fields


jtopjian:
As microversions become more widespread throughout Gophercloud, it appears that our initial guideline of providing a dedicated `Extract` function will soon become unmaintainable. I propose treating _result_ fields that have been added by a microversion the same way we treat _request_ fields.

ozerovandrei:
I agree.

jtopjian:
Implemented in #1854

PR #1854, opened 2020-02-15
Microversions: Simplify Result Fields

jtopjian:
For #1853


jtopjian:
@ozerovandrei Thanks!

PR #1858, opened 2020-02-19
Compute v2: InstanceActions ListOpts

jtopjian:
For #1845 

This commit adds support for list params in the instanceactions extension. It also does some documentation cleanup and adds acceptance tests.

* https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/instance_actions.py#L106

* https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/instance_actions.py#L110

* https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/instance_actions.py#L120

/cc @snigle FYI on the breaking function signature for `List`.


jtopjian:
I thought I merged this one a while ago...

PR #1861, opened 2020-02-27
Add resource provider usages API for placement

dkt26111:
For #526

code in this PR:

API doc:
https://docs.openstack.org/api-ref/placement/#list-resource-provider-usages

API code:
https://github.com/openstack/placement/blob/master/placement/handlers/usage.py#L38


dkt26111:

openstack/placement/v1/resourceproviders/results.go
```diff
@@ -33,6 +33,17 @@ type ResourceProvider struct {
...
 
+type Usage struct {
+	DiskGB   int `json:"DISK_GB"`
+	MemoryMB int `json:"MEMORY_MB"`
+	VCPU     int `json:"VCPU"`
+}
+
+type ResourceProviderUsage struct {
+	ResourceProviderGeneration int   `json:"resource_provider_generation"`
+	Usages                     Usage `json:"usages"`
```
jtopjian: I'm thinking this should be `map[string]int`? 

I don't see any reference to `DISK_GB` etc in the placement code and the API docs mention:

>  The value is a dictionary of resource classes paired with the sum of the allocations of that resource class for this resource provider.

Given it's possible to [create arbitrary resource classes](https://docs.openstack.org/api-ref/placement/#create-resource-class), I think the result of `Usages` is going to be highly dynamic.

Thoughts?

dkt26111: Sounds good to me.  I'll make the change.

jtopjian:
Implemented in #1862

PR #1862, opened 2020-02-27
Add resource provider inventories API for placement

dkt26111:
For #526

code in this PR:

API doc:
https://docs.openstack.org/api-ref/placement/#list-resource-provider-inventories

API code:
https://github.com/openstack/placement/blob/master/placement/handlers/inventory.py#L259


dkt26111:
This PR is based on top of https://github.com/gophercloud/gophercloud/pull/1861 so it contains the resource provider usage changes as well.

dkt26111:
@jtopjian This is ready for review as well.

openstack/placement/v1/resourceproviders/results.go
```diff
@@ -33,6 +33,37 @@ type ResourceProvider struct {
...
+	MemoryMB int `json:"MEMORY_MB"`
+	VCPU     int `json:"VCPU"`
+}
+
+type ResourceProviderUsage struct {
+	ResourceProviderGeneration int   `json:"resource_provider_generation"`
+	Usages                     Usage `json:"usages"`
+}
+
+type Inventory struct {
```
jtopjian: Similar to my comment in #1861:

This appears to be well defined: https://github.com/openstack/placement/blob/55635d918ce32667352fb94e4ae3e388069dd9c1/placement/schemas/inventory.py

So maybe this would work better?

```go
type ResourceProviderInventories struct {
	ResourceProviderGeneration int                  `json:"resource_provider_generation"`
	Inventories                map[string]Inventory `json:"inventories"`
}
```

dkt26111: I agree.  I'll make that change.

jtopjian:
I'm pretty sure the OpenLab error is transient/temporary.

PR #1873, opened 2020-03-02
Cinder V3: Support "protected" and "visibility" parameters in "UploadImageOpts

### LINKED ISSUE #1872: Cinder V3: Support "protected" and "visibility" parameters in "UploadImageOpts"
These fields were introduced since 3.1 microversion in OpenStack Newton: https://docs.openstack.org/cinder/latest/contributor/api_microversion_history.html#id1

https://github.com/openstack/cinder/blob/e921b3ae6c462f5f844485f7da7164e194efd00c/cinder/api/contrib/volume_actions.py#L232..L240

kayrus:
Resolves #1872

### DIFF
--- a/openstack/blockstorage/extensions/volumeactions/requests.go
+++ b/openstack/blockstorage/extensions/volumeactions/requests.go
@@ -2,6 +2,7 @@ package volumeactions
 
 import (
 	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 )
 
 // AttachOptsBuilder allows extensions to add additional parameters to the
@@ -241,6 +242,14 @@ type UploadImageOpts struct {
 
 	// Force image creation, usable if volume attached to instance.
 	Force bool `json:"force,omitempty"`
+
+	// Visibility defines who can see/use the image.
+	// supported since 3.1 microversion
+	Visibility images.ImageVisibility `json:"visibility,omitempty"`
+
+	// whether the image is not deletable.
+	// supported since 3.1 microversion
+	Protected bool `json:"protected,omitempty"`
 }
 
 // ToVolumeUploadImageMap assembles a request body based on the contents of a
--- a/openstack/blockstorage/extensions/volumeactions/requests.go
+++ b/openstack/blockstorage/extensions/volumeactions/requests.go
@@ -2,7 +2,6 @@ package volumeactions
 
 import (
 	"github.com/gophercloud/gophercloud"
-	"github.com/gophercloud/gophercloud/openstack/imageservice/v2/images"
 )
 
 // AttachOptsBuilder allows extensions to add additional parameters to the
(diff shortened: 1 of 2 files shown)


kayrus:
@jtopjian ready for review

openstack/blockstorage/extensions/volumeactions/results.go
```diff
@@ -157,6 +158,14 @@ type VolumeImage struct {
 	// Current status of the volume.
 	Status string `json:"status"`
 
+	// Visibility defines who can see/use the image.
+	// supported since 3.1 microversion
+	Visibility images.ImageVisibility `json:"visibility"`
```
jtopjian: Let's change this to just `string`. All results should be basic Golang types. This helps prevent any issue if the returned value is something custom or something we haven't accounted for.

(there might be some non-basic types hanging around other `results.go` files - these were either added prior to this decision being made or have slipped by)

openstack/blockstorage/extensions/volumeactions/requests.go
```diff
@@ -241,6 +242,14 @@ type UploadImageOpts struct {
 
 	// Force image creation, usable if volume attached to instance.
 	Force bool `json:"force,omitempty"`
+
+	// Visibility defines who can see/use the image.
+	// supported since 3.1 microversion
+	Visibility images.ImageVisibility `json:"visibility,omitempty"`
```
jtopjian: For posterity:

This field was added with a microversion, but Cinder v3 support was merged using microversion 3.27 as the "base". So I consider anything prior to that to be grandfathered, such as this field.

Otherwise, this field would be a pointer, as recently changed here: https://github.com/gophercloud/gophercloud/pull/1854

openstack/blockstorage/extensions/volumeactions/requests.go
```diff
@@ -241,6 +242,14 @@ type UploadImageOpts struct {
...
+
+	// whether the image is not deletable.
+	// supported since 3.1 microversion
+	Protected bool `json:"protected,omitempty"`
```
jtopjian: Same here.

openstack/blockstorage/extensions/volumeactions/requests.go
kayrus: Are you sure about this? According to the latest code it is still checks for the explicit 3.1 microversion: https://github.com/openstack/cinder/blob/e921b3ae6c462f5f844485f7da7164e194efd00c/cinder/api/contrib/volume_actions.py#L232..L240

openstack/blockstorage/extensions/volumeactions/results.go
kayrus: done

openstack/blockstorage/extensions/volumeactions/requests.go
jtopjian: Sorry - to clarify: Cinder v3 support was merged into Gophercloud using microversion 3.27 as the "base".

Issue #1876, opened 2020-03-02
Two Dropped Errors in openstack/networking/v2/extensions/trunks/testing


alrs:
There are two dropped `err` variables in the package `openstack/networking/v2/extensions/trunks/testing`.

jtopjian:
Fixed in #1877

PR #1877, opened 2020-03-02
openstack/networking/v2/extensions/trunks/testing: Fix Two Dropped Errors

alrs:
For #1876

This fixes two dropped errors in `openstack/networking/v2/extensions/trunks/testing`.


jtopjian:
The OpenLab failure can be ignored - it's not related to this PR.

PR #1881, opened 2020-03-03
Compute V2: Fix userdata result extension

### LINKED ISSUE #1880: Compute V2: Server user data attribute has a typo
userdata attribute has a typo:

https://github.com/gophercloud/gophercloud/blob/a87527328216a750678e3bc100b8c0e4cd607667/openstack/compute/v2/extensions/extendedserverattributes/results.go#L43

https://github.com/openstack/nova/blob/72224775fb6a2eda0ceec51dfcb971fc50d4ea02/nova/api/openstack/compute/views/servers.py#L325

kayrus:
Resolves #1880

### DIFF
--- a/openstack/compute/v2/extensions/extendedserverattributes/results.go
+++ b/openstack/compute/v2/extensions/extendedserverattributes/results.go
@@ -1,5 +1,10 @@
 package extendedserverattributes
 
+import (
+	"encoding/base64"
+	"encoding/json"
+)
+
 // ServerAttributesExt represents basic OS-EXT-SRV-ATTR server response fields.
 // You should use extract methods from microversions.go to retrieve additional
 // fields.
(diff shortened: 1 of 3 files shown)


jtopjian:
Renaming `userdata` to `user_data` looks correct, but this PR is also introducing a feature where Gophercloud will encode the data on behalf of the user. I don't think we need to include that.

kayrus:
@jtopjian create server opts struct method does encode `user_data` to base64 automatically:
https://github.com/gophercloud/gophercloud/blob/675de7268e2618a557195c92ac49d1cc6d5e0174/openstack/compute/v2/servers/requests.go#L230..L238

jtopjian:
> create server opts struct method does encode user_data to base64 automatically:

Ahh right. I knew we had support for this, but similar to a few other legacy areas you've been catching, this is ripe for removal. I checked the Terraform provider and it looks like it's passing in a string... so it's leaning on Gophercloud's encoding capability.

Either way, let's keep the result as being encoded. It's not consistent, but IMO it's the right way for Gophercloud to handle it.

PR #1884, opened 2020-03-03
Compute V2: Support VMs without a network

### LINKED ISSUE #1883: Compute V2: Allow VMs without a network.
Starting with microversion 2.37, `network` object in servers.CreateOpts can be a string: `auto` or `none`. `none` allows to create a VM without a network.

> Starting with microversion 2.37, this field is required and the special string values auto and none can be specified for networks. auto tells the Compute service to use a network that is available to the project, if one exists. If one does not exist, the Compute service will attempt to automatically allocate a network for the project (if possible). none tells the Compute service to not allocate a network for the instance. The auto and none values cannot be used with any other network values, including other network uuids, ports, fixed IPs or device tags. These are requested as strings for the networks value, not in a list. See the associated example.

(c) https://docs.openstack.org/api-ref/compute/?expanded=create-server-detail#create-server

refs:
* https://github.com/openstack/nova/blob/72224775fb6a2eda0ceec51dfcb971fc50d4ea02/nova/objects/network_request.py#L25..L26
* https://github.com/openstack/nova/blob/72224775fb6a2eda0ceec51dfcb971fc50d4ea02/nova/api/openstack/compute/servers.py#L398..L445

kayrus:
Resolves #1883

### DIFF
--- a/acceptance/openstack/compute/v2/compute.go
+++ b/acceptance/openstack/compute/v2/compute.go
@@ -1198,3 +1198,55 @@ func CreateRemoteConsole(t *testing.T, client *gophercloud.ServiceClient, server
 	t.Logf("Successfully created console: %s", remoteConsole.URL)
 	return remoteConsole, nil
 }
+
+// CreateNoNetworkServer creates a basic instance with a randomly generated name.
+// The flavor of the instance will be the value of the OS_FLAVOR_ID environment variable.
+// The image will be the value of the OS_IMAGE_ID environment variable.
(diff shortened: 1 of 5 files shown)


jtopjian:
I'd like to see an acceptance test added to this. If needed, we can wait until OpenLab is working again.

kayrus:
recheck

kayrus:
@jtopjian done

kayrus:
@jtopjian no problem at all

Issue #1889, opened 2020-03-11
Identity v3: add support for OS-EP-FILTER (endpoint groups) extension


pgaxatte:
I'd like to add support for this Keystone extension: https://docs.openstack.org/api-ref/identity/v3-ext/index.html#os-ep-filter-api

I'm not sure how to name the package and was thinking `endpointgroups` (path would be `openstack/identity/v3/extensions/endpointgroups`) since the extension is mostly endpoint groups CRUD and project/endpoint association.

This would lead to the following functions in `openstack/identity/v3/extensions/endpointgroups/requests.go` (taken in order of the Keystone documentation):

```
package endpointgroups

func Create() {
    // Create Endpoint Group
    // POST /v3/OS-EP-FILTER/endpoint_groups
}

func Get() {
    // Get Endpoint Group
    // GET /v3/OS-EP-FILTER/endpoint_groups/{endpoint_group_id}
}

// Not implemented:
// Check Endpoint Group
// HEAD /v3/OS-EP-FILTER/endpoint_groups/{endpoint_group_id}

func Update() {
    // Update Endpoint Group
    // PATCH /v3/OS-EP-FILTER/endpoint_groups/{endpoint_group_id}
}

func Delete() {
    // Delete Endpoint Group
    // DELETE /v3/OS-EP-FILTER/endpoint_groups/{endpoint_group_id}
}

func List() {
    // List Endpoint Groups
    // GET /v3/OS-EP-FILTER/endpoint_groups
}

func AssociateProjectToEndpoint() {
    // Create Association
    // PUT /v3/OS-EP-FILTER/projects/{project_id}/endpoints/{endpoint_id}
}

// Not implemented:
// Check Association
// HEAD /v3/OS-EP-FILTER/projects/{project_id}/endpoints/{endpoint_id}


func DisassociateProjectToEndpoint() {
    // Delete Association
    // DELETE /v3/OS-EP-FILTER/projects/{project_id}/endpoints/{endpoint_id}
}

func ListEndpointsAssociatedToProject() {
    // List Associations by Project
    // GET /v3/OS-EP-FILTER/projects/{project_id}/endpoints
}

func ListProjectsAssociatedToEndpoint() {
    // List Associations by Endpoint
    // GET /v3/OS-EP-FILTER/endpoints/{endpoint_id}/projects
}

func Associate() {
    // Create Endpoint Group to Project Association
[...]

PR #1893, opened 2020-03-12
Identity v3: OS-EP-FILTERS endpoint group list/get

pgaxatte:
For #1889

API doc:
https://docs.openstack.org/api-ref/identity/v3-ext/index.html#os-ep-filter-api

API calls:
https://github.com/openstack/keystone/blob/ba2e4b83e88d5c1576e9d2cd6a760367deef3a62/keystone/api/os_ep_filter.py#L68-L79

Structures: https://github.com/openstack/keystone/blob/ba2e4b83e88d5c1576e9d2cd6a760367deef3a62/keystone/catalog/backends/sql.py#L605

EndpointGroupFilter structure: https://github.com/openstack/keystone/blob/ba2e4b83e88d5c1576e9d2cd6a760367deef3a62/keystone/api/os_ep_filter.py#L54

### DIFF
--- a/openstack/identity/v3/extensions/endpointgroups/results.go
+++ b/openstack/identity/v3/extensions/endpointgroups/results.go
@@ -37,9 +37,6 @@ type EndpointFilter struct {
 
 	// RegionID is the ID of the region to filter
 	RegionID string `json:"region_id,omitempty"`
-
-	// Enabled filters on enabled or disabled endpoints
-	Enabled *bool `json:"enabled,omitempty"`
 }
 
 // EndpointGroup represents a group of endpoints matching one or several filter
--- a/openstack/identity/v3/extensions/endpointgroups/results.go
+++ b/openstack/identity/v3/extensions/endpointgroups/results.go
@@ -1,0 +1,79 @@
+package endpointgroups
+
+import (
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+type commonResult struct {
+	gophercloud.Result
+}
+
+// Extract interprets a GetResult, CreateResult or UpdateResult as a concrete
+// EndpointGroup. An error is returned if the original call or the extraction
+// failed.
+func (r commonResult) Extract() (*EndpointGroup, error) {
+	var s struct {
+		EndpointGroup *EndpointGroup `json:"endpoint_group"`
+	}
+	err := r.ExtractInto(&s)
+	return s.EndpointGroup, err
+}
+
+// GetResult is the response from a Get operation. Call its Extract method
+// to interpret it as an EndpointGroup.
+type GetResult struct {
+	commonResult
+}
+
+// EndpointFilter represents a set of one or several criteria to match endpoints.
+type EndpointFilter struct {
+	// Availability is the interface type to filter (admin, internal,
+	// or public), referenced by the gophercloud.Availability type.
+	Availability gophercloud.Availability `json:"interface,omitempty"`
+
+	// ServiceID is the ID of the service to filter
+	ServiceID string `json:"service_id,omitempty"`
+
+	// RegionID is the ID of the region to filter
+	RegionID string `json:"region_id,omitempty"`
+
+	// Enabled filters on enabled or disabled endpoints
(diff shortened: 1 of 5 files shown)


jtopjian:
@pgaxatte This is a good start 🙂 

One important thing that we'll need to reference is where the struct fields are defined in the Keystone code. For example, we can see that the `EndpointGroup` struct is defined [here](https://github.com/openstack/keystone/blob/ba2e4b83e88d5c1576e9d2cd6a760367deef3a62/keystone/catalog/backends/sql.py#L605), but where are the fields in the `EndpointFilter` struct coming from? This might be a little difficult to determine, but it could uncover some undocumented behavior about Keystone that we need to account for in a typed language. Let me know if you have any problems.

A second note is a small nit about the spacing: if you check your GoDoc and JSON fixtures, you'll see some misaligned spacing from a mix of tabs and spaces. This is a normal issue with Go.

pgaxatte:
@jtopjian Thanks for the review!

I fix the spacing in the doc.
I also found where the fields of the endpoint group filters are defined but it seems not to be exactly compliant with the documentation: there is no `enabled` field apparently

jtopjian:
> I also found where the fields of the endpoint group filters are defined but it seems not to be exactly compliant with the documentation: there is no enabled field apparently

When there's discrepancies between the code and the docs, the code is the authoritative answer. If the code is too confusing to untangle, then running the `openstack` command with `--debug` is a good way of confirming fields.

This PR is looking good. Let me know if you'd like me to do a full review.

pgaxatte:
Hello @jtopjian 

Thanks for your patience on this PR.
I removed the Enabled field as it does not appear in keystone's sources.

Is there something I should do about the 0.02% drop in coverage? Should I add tests for the pagination?

Could you do a full review on this?

On your approval, I'll squash my commits.

openstack/identity/v3/extensions/endpointgroups/results.go
```diff
@@ -0,0 +1,76 @@
...
+// criteria.
+type EndpointGroup struct {
+	// ID is the unique ID of the endpoint group.
+	ID string `json:"id"`
+
+	// Name is the name of the new endpoint group.
+	Name string `json:"name"`
+
+	// Filters is an EndpointFilter type describing the filter criteria
+	Filters EndpointFilter `json:"filters"`
```
jtopjian, line 52: I'm thinking we want to do a `map[string]interface{}` here instead of a well-defined `EndpointFilter`. 

The filter keys you linked to (https://github.com/openstack/keystone/blob/ba2e4b83e88d5c1576e9d2cd6a760367deef3a62/keystone/api/os_ep_filter.py#L54) appear to be for POST/creation validation and not defining what the actual data structure is.

The schema looks to be defining this field as a generic JSON blob (https://github.com/openstack/keystone/blob/ba2e4b83e88d5c1576e9d2cd6a760367deef3a62/keystone/catalog/backends/sql.py#L614).

It's very possible that right now only the `valid_filter_keys` are the only 3 entries, but since the schema is defining this as a generic JSON blob, I think we should just use `map[string]interface{}` to reflect that.

Thoughts?

openstack/identity/v3/extensions/endpointgroups/results.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// GetResult is the response from a Get operation. Call its Extract method
+// to interpret it as an EndpointGroup.
+type GetResult struct {
+	commonResult
+}
+
+// EndpointFilter represents a set of one or several criteria to match endpoints.
+type EndpointFilter struct {
```
jtopjian, line 30: If the above comment is true, we can get rid of this struct. When POST/create is implemented, we can do something like [this](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/flavors/requests.go#L22-L35) to help with the input validation.

EmilienM:
@pgaxatte Salut Pierre, any update on this one? do you need help to get this going?

Thanks for your PR btw.

pgaxatte:
I am abandoning this change (as I don't work on this subject anymore) and totally forfeit my code and changes to @garbage-collector who took over this subject in our team.

@garbage-collector PR is here: https://github.com/gophercloud/gophercloud/pull/2316

PR #1894, opened 2020-03-12
Cinder V3: Support a backup record export and import

### LINKED ISSUE #1892: Cinder V3: Support a backup record export and import
Export a backup: https://docs.openstack.org/api-ref/block-storage/v3/?expanded=export-a-backup-detail#export-a-backup
https://github.com/openstack/cinder/blob/c29285e7dcda7f5300c5a3375294dfa4aeb56430/cinder/api/contrib/backups.py#L239..L253

Import a backup: https://docs.openstack.org/api-ref/block-storage/v3/?expanded=import-a-backup-detail#import-a-backup
https://github.com/openstack/cinder/blob/c29285e7dcda7f5300c5a3375294dfa4aeb56430/cinder/api/contrib/backups.py#L257..L280

kayrus:
Resolves #1892

### DIFF
--- a/openstack/blockstorage/extensions/backups/doc.go
+++ b/openstack/blockstorage/extensions/backups/doc.go
@@ -83,20 +83,27 @@ Example to Export a Backup
 
 Example to Import a Backup
 
+	status := "available"
 	availabilityZone := "region1b"
+	host := "cinder-backup-host1"
+	serviceMetadata := "volume_cf9bc6fa-c5bc-41f6-bc4e-6e76c0bea959/20200311192855/az_regionb_backup_b87bb1e5-0d4e-445e-a548-5ae742562bac"
+	size := 1
+	objectCount := 2
+	container := "my-test-backup"
+	service := "cinder.backup.drivers.swift.SwiftBackupDriver"
 	backupURL, _ := json.Marshal(backups.ImportBackup{
 		ID:               "d32019d3-bc6e-4319-9c1d-6722fc136a22",
-		Status:           "available",
+		Status:           &status,
 		AvailabilityZone: &availabilityZone,
 		VolumeID:         "cf9bc6fa-c5bc-41f6-bc4e-6e76c0bea959",
 		UpdatedAt:        time.Date(2020, 3, 11, 19, 29, 8, 0, time.UTC),
-		Host:             "cinder-backup-host1",
+		Host:             &host,
 		UserID:           "93514e04-a026-4f60-8176-395c859501dd",
-		ServiceMetadata:  "volume_cf9bc6fa-c5bc-41f6-bc4e-6e76c0bea959/20200311192855/az_regionb_backup_b87bb1e5-0d4e-445e-a548-5ae742562bac",
-		Size:             1,
-		ObjectCount:      2,
-		Container:        "my-test-backup",
-		Service:          "cinder.backup.drivers.swift.SwiftBackupDriver",
+		ServiceMetadata:  &serviceMetadata,
+		Size:             &size,
+		ObjectCount:      &objectCount,
+		Container:        &container,
+		Service:          &service,
 		CreatedAt:        time.Date(2020, 3, 11, 19, 25, 24, 0, time.UTC),
 		DataTimestamp:    time.Date(2020, 3, 11, 19, 25, 24, 0, time.UTC),
 		ProjectID:        "14f1c1f5d12b4755b94edef78ff8b325",
--- a/openstack/blockstorage/extensions/backups/doc.go
+++ b/openstack/blockstorage/extensions/backups/doc.go
@@ -71,5 +71,32 @@ Example to Delete a Backup
 	if err != nil {
 		panic(err)
 	}
+
+Example to Export a Backup
+
+	export, err := backups.Export(client, "uuid").Extract()
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(export)
+
+Example to Import a Backup
+
+	backupURL, _ := json.Marshal(map[string]string{
+		"foo": "bar",
+	})
+
+	options := backups.ImportOpts{
+		BackupService: "cinder.backup.drivers.swift.SwiftBackupDriver",
+		BackupURL:     backupURL,
+	}
+
+	backup, err := backups.Import(client, options).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+	fmt.Println(backup)
 */
 package backups
--- a/openstack/blockstorage/extensions/backups/doc.go
+++ b/openstack/blockstorage/extensions/backups/doc.go
@@ -83,8 +83,23 @@ Example to Export a Backup
 
 Example to Import a Backup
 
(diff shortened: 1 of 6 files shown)


kayrus:
@jtopjian ready for review. I also added a stuct to represent the `backup_url` data

openstack/blockstorage/extensions/backups/requests.go
```diff
@@ -213,3 +213,35 @@ func RestoreFromBackup(client *gophercloud.ServiceClient, id string, opts Restor
...
+	return
+}
+
+// ImportOpts contains options for importing a Backup. This object is passed to
+// the backups.ImportBackup function.
+type ImportOpts BackupRecord
+
+// ToImportMap assembles a request body based on the contents of a
+// ImportOpts.
+func (opts ImportOpts) ToImportMap() (map[string]interface{}, error) {
```
jtopjian: nit: `ToBackupImportMap()`. I forget if I mentioned in the previous PR that the naming was OK because `ToBackupBackup` would have been really awkward.

openstack/blockstorage/extensions/backups/results.go
```diff
@@ -187,3 +187,130 @@ func (r RestoreResult) Extract() (*Restore, error) {
...
+	return r.Result.ExtractIntoStructPtr(v, "backup-record")
+}
+
+// ImportResult contains the response body and error from an import request.
+type ImportResult struct {
+	commonResult
+}
+
+// ImportBackup contains all the information to import a Cinder Backup.
+type ImportBackup struct {
```
jtopjian: Where is this struct defined in Cinder?

openstack/blockstorage/extensions/backups/results.go
```diff
@@ -187,3 +187,130 @@ func (r RestoreResult) Extract() (*Restore, error) {
...
+	return &s, err
+}
+
+func (r ExportResult) ExtractInto(v interface{}) error {
+	return r.Result.ExtractIntoStructPtr(v, "backup-record")
+}
+
+// ImportResult contains the response body and error from an import request.
+type ImportResult struct {
+	commonResult
```
jtopjian: commonResult's Extract method returns a Backup struct - is this correct?

openstack/blockstorage/extensions/backups/doc.go
```diff
@@ -71,5 +71,47 @@ Example to Delete a Backup
...
+		Container:        "my-test-backup",
+		Service:          "cinder.backup.drivers.swift.SwiftBackupDriver",
+		CreatedAt:        time.Date(2020, 3, 11, 19, 25, 24, 0, time.UTC),
+		DataTimestamp:    time.Date(2020, 3, 11, 19, 25, 24, 0, time.UTC),
+		ProjectID:        "14f1c1f5d12b4755b94edef78ff8b325",
+	})
+
+	options := backups.ImportOpts{
+		BackupService: "cinder.backup.drivers.swift.SwiftBackupDriver",
+		BackupURL:     backupURL,
```
jtopjian: This needs to be base64 encoded, right?

kayrus: If you provide `[]byte` to a json encoder, it automatically encodes/decodes it to base64:

> Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.

https://golang.org/pkg/encoding/json/

openstack/blockstorage/extensions/backups/results.go
kayrus: Import result returns backup struct with only two fields: `id` and `name`. At least in my case. I decided to reuse a `Backup` struct for this. I don't see any cons.

openstack/blockstorage/extensions/backups/results.go
kayrus: I mentioned it in #1892, see https://github.com/openstack/cinder/blob/c29285e7dcda7f5300c5a3375294dfa4aeb56430/cinder/objects/backup.py#L32..L84

openstack/blockstorage/extensions/backups/doc.go
jtopjian: Well I learned something today 🙂 Thanks!

kayrus:
> Also, is it possible to implement acceptance tests? Or are these API calls too difficult?

It is possible, but you need to:
* create a volume
* create a backup
* wait for backup to be done
* copy swift container objects to a new container
* export an old backup
* delete an old backup and ensure old swift container is removed
* change a container name of an exported backup struct to a new container name
* import backup
* restore a new volume from a backup
* wait for its status
* check the volume contents somehow...

openstack/blockstorage/extensions/backups/requests.go
kayrus: Do I also need to rename `Export` -> `BackupExport` and `Import` -> `BackupImport`?

jtopjian: The function names? No - those can stay.

It's just the interface method name that needs to change to `To-Resource/Package-Action-ResultType` If you browse the other packages, you'll see that the interface names are verbose like this while the function/action names can be simple verbs.

I believe the reason behind this naming convention was to ensure that there is never a conflict with implementing a custom struct that can do many different things (implement many different interfaces).

kayrus: Ok, I'll rename only this method name.

openstack/blockstorage/extensions/backups/doc.go
kayrus: I learned this, when I could not understand why I get base64 string in acceptance tests. It appeared that I forgot to place `w.Header().Add("Content-Type", "application/json")` into a mock list func: https://github.com/gophercloud/gophercloud/blob/965816e95e604e19c77b1a086c9737164dc3a4c5/openstack/blockstorage/extensions/volumetransfers/testing/fixtures.go#L199
Just try to comment this string out and run tests ;)

openstack/blockstorage/extensions/backups/results.go
jtopjian: I think returning a Volume struct can be misleading. As an exaggerated example, we could technically have an Identity Project struct returned since it also has an id and name :slightly_smiling_face: 

Let's make `ImportResult` a small struct with an ID and Name field.

openstack/blockstorage/extensions/backups/results.go
jtopjian: I apologize - I completely missed that.

jtopjian:
> It is possible, but you need to:

OK, let's skip acceptance tests for this, then.

Have you been able to use this outside of an acceptance test with success?

kayrus:
> Have you been able to use this outside of an acceptance test with success?

Yes, it worked fine.

openstack/blockstorage/extensions/backups/results.go
kayrus: Done

Issue #1896, opened 2020-03-15
Remove IDFromName Functions


jtopjian:
Gophercloud has a bunch of `IDFromName` convenience functions throughout the various packages. While these functions are helpful, they aren't really appropriate for this repository. Instead, they should be moved to https://github.com/gophercloud/utils.

jtopjian:
Implemented in #1897

PR #1897, opened 2020-03-15
Removing IDFromName Functions

jtopjian:
For #1896


openstack/compute/v2/servers/requests.go
```diff
@@ -148,24 +146,14 @@ type CreateOpts struct {
...
-	// full URL to the image that contains the server's OS and initial state.
+	// ImageRef is the ID or full URL to the image that contains the
+	// server's OS and initial state.
 	// Also optional if using the boot-from-volume extension.
 	ImageRef string `json:"imageRef"`
 
-	// ImageName [optional; required if ImageRef is not provided] is the name of
-	// the image that contains the server's OS and initial state.
-	// Also optional if using the boot-from-volume extension.
-	ImageName string `json:"-"`
```
andreykaipov, line 159: Was removing the `ImageName` and `FlavorName` options intended? It was a nice convenience to be able to pass the service client and not have to worry about calling `IDFromName(...)` ourselves. If so, it might be worthwhile to remove the `ServiceClient` as an option from the `CreateOpts` here and possibly other packages as it serves no purpose anymore.

jtopjian: Good catch. Yes, it was intended though I do agree that it removes some well-received convenience.

PR #1899, opened 2020-03-19
Add resource provider traits API for placement

dkt26111:
For #526

code in this PR:

API doc:
https://docs.openstack.org/api-ref/placement/#list-resource-provider-traits

API code:
https://github.com/openstack/placement/blob/master/placement/handlers/trait.py#L184


dkt26111:
@jtopjian This is ready for review.  Thanks.

PR #1900, opened 2020-03-21
Auth: introduce ec2 credentials auth support

### LINKED ISSUE #1898: Identity V3: Introduce ec2 auth support
Hence it is stated that it is deprecated, it is still used by [Swift S3 middleware](https://docs.openstack.org/swift/latest/middleware.html#module-swift.common.middleware.s3api.s3api) or [Nova EC2](https://docs.openstack.org/ec2-api/latest/) emulation:

Auth is perfomed here:
* https://github.com/openstack/keystone/blob/62cd34d40d85af85b34146f9827da3414c09418f/keystone/api/_shared/EC2_S3_Resource.py#L39
* https://github.com/openstack/keystone/blob/62cd34d40d85af85b34146f9827da3414c09418f/keystone/api/ec2tokens.py#L36
* https://github.com/openstack/python-keystoneclient/blob/526350eaeeadef8c1ca5f80547c01f94342a6653/keystoneclient/contrib/ec2/utils.py#L80..L98

An example of the ec2 auth request: https://docs.openstack.org/api-ref/identity/v2-ext/?expanded=ec2-authentication-detail#ec2-authentication

kayrus:
Resolves #1898
@jtopjian before I introduce unit tests please take a look on this and let me know your opinion. It is hard to introduce unit tests, since token signature changes every call. I'd prefer to introduce an interface to create, list, delete ec2 credentials, then integrate them with ec2 auth.

### DIFF
--- a/openstack/identity/v3/extensions/ec2tokens/requests.go
+++ b/openstack/identity/v3/extensions/ec2tokens/requests.go
@@ -58,9 +58,7 @@ type AuthOptions struct {
 	Region string `json:"-"`
 	// Service is a service name to calculate an AWS signature V4. Optional.
 	Service string `json:"-"`
-	// Params is a map of parameters, used to configure an AWS signature
-	// calculation method. Defaults to AWS signature V4 parameters.
-	// Optional.
+	// Params is a map of GET method parameters. Optional.
 	Params map[string]string `json:"params"`
 	// AllowReauth allows Gophercloud to re-authenticate automatically
 	// if/when your token expires.
@@ -97,7 +95,7 @@ func EC2CredentialsBuildCanonicalQueryStringV2(params map[string]string) string 
 // EC2CredentialsBuildStringToSignV2 builds a string to sign an AWS signature
 // V2.
 // https://github.com/openstack/python-keystoneclient/blob/stable/train/keystoneclient/contrib/ec2/utils.py#L148
-func EC2CredentialsBuildStringToSignV2(opts AuthOptions, params map[string]string) string {
+func EC2CredentialsBuildStringToSignV2(opts AuthOptions) string {
 	stringToSign := strings.Join([]string{
 		opts.Verb,
 		opts.Host,
@@ -106,7 +104,7 @@ func EC2CredentialsBuildStringToSignV2(opts AuthOptions, params map[string]strin
 
 	return strings.Join([]string{
 		stringToSign,
-		EC2CredentialsBuildCanonicalQueryStringV2(params),
+		EC2CredentialsBuildCanonicalQueryStringV2(opts.Params),
 	}, "\n")
 }
 
@@ -152,7 +150,7 @@ func EC2CredentialsBuildSignatureKeyV4(secret, date, region, service string) []b
 // EC2CredentialsBuildSignatureV4 builds an AWS v4 signature based on input
 // parameters.
 // https://github.com/openstack/python-keystoneclient/blob/stable/train/keystoneclient/contrib/ec2/utils.py#L251
-func EC2CredentialsBuildSignatureV4(opts AuthOptions, params map[string]string, date time.Time, bodyHash string) string {
+func EC2CredentialsBuildSignatureV4(opts AuthOptions, signedHeaders string, date time.Time, bodyHash string) string {
 	scope := strings.Join([]string{
 		date.Format(EC2CredentialsDateFormatV4),
 		opts.Region,
@@ -163,9 +161,9 @@ func EC2CredentialsBuildSignatureV4(opts AuthOptions, params map[string]string, 
 	canonicalRequest := strings.Join([]string{
 		opts.Verb,
 		opts.Path,
-		EC2CredentialsBuildCanonicalQueryStringV4(opts.Verb, params),
-		EC2CredentialsBuildCanonicalHeadersV4(opts.Headers, params["X-Amz-SignedHeaders"]),
-		params["X-Amz-SignedHeaders"],
+		EC2CredentialsBuildCanonicalQueryStringV4(opts.Verb, opts.Params),
+		EC2CredentialsBuildCanonicalHeadersV4(opts.Headers, signedHeaders),
+		signedHeaders,
 		bodyHash,
 	}, "\n")
 	hash := sha256.Sum256([]byte(canonicalRequest))
@@ -205,7 +203,8 @@ func (opts *AuthOptions) ToTokenV3CreateMap(map[string]interface{}) (map[string]
 
 	// calculate signature, when it is not set
 	c, _ := b["credentials"].(map[string]interface{})
-	p := paramsToMap(c)
+	h := interfaceToMap(c, "headers")
+	p := interfaceToMap(c, "params")
 
 	// detect and process a signature v2
 	if v, ok := p["SignatureVersion"]; ok && v == "2" {
@@ -217,7 +216,7 @@ func (opts *AuthOptions) ToTokenV3CreateMap(map[string]interface{}) (map[string]
 		}
 		if v, ok := p["SignatureMethod"]; ok {
 			// params is a map of strings
-			strToSign := EC2CredentialsBuildStringToSignV2(*opts, p)
+			strToSign := EC2CredentialsBuildStringToSignV2(*opts)
 			switch v {
 			case EC2CredentialsHmacSha1V2:
 				// keystone uses this method only when HmacSHA256 is not available on the server side
@@ -244,24 +243,20 @@ func (opts *AuthOptions) ToTokenV3CreateMap(map[string]interface{}) (map[string]
 		// when body_hash is not set, generate a random one
 		c["body_hash"] = randomBodyHash()
 	}
-	if v, _ := c["headers"]; v == nil {
-		// when headers is not set, make an empty map
-		c["headers"] = make(map[string]string)
-	}
-	if _, ok := p["X-Amz-SignedHeaders"]; !ok {
-		// when X-Amz-SignedHeaders is not set, make an empty string
-		p["X-Amz-SignedHeaders"] = ""
-	}
-	p["X-Amz-Date"] = date.Format(EC2CredentialsTimestampFormatV4)
-	p["X-Amz-Algorithm"] = EC2CredentialsAwsHmacV4
-	p["X-Amz-Credential"] = strings.Join([]string{
+
+	signedHeaders, _ := h["X-Amz-SignedHeaders"]
+	c["signature"] = EC2CredentialsBuildSignatureV4(*opts, signedHeaders, date, c["body_hash"].(string))
+
+	h["X-Amz-Date"] = date.Format(EC2CredentialsTimestampFormatV4)
+	h["Authorization"] = fmt.Sprintf("%s Credential=%s/%s/%s/%s/%s, SignedHeaders=%s, Signature=%s",
+		EC2CredentialsAwsHmacV4,
 		opts.Access,
 		date.Format(EC2CredentialsDateFormatV4),
 		opts.Region,
 		opts.Service,
 		EC2CredentialsAwsRequestV4,
-	}, "/")
-	c["signature"] = EC2CredentialsBuildSignatureV4(*opts, p, date, c["body_hash"].(string))
+		signedHeaders,
+		c["signature"])
 
 	return b, nil
 }
@@ -308,18 +303,18 @@ func randomBodyHash() string {
 	return hex.EncodeToString(h)
 }
 
-// paramsToMap is a func used to represent a "credentials" map "params" value
-// as a "map[string]string"
-func paramsToMap(c map[string]interface{}) map[string]string {
+// interfaceToMap is a func used to represent a "credentials" map element as a
+// "map[string]string"
+func interfaceToMap(c map[string]interface{}, key string) map[string]string {
 	// convert map[string]interface{} to map[string]string
-	p := make(map[string]string)
-	if v, _ := c["params"].(map[string]interface{}); v != nil {
+	m := make(map[string]string)
+	if v, _ := c[key].(map[string]interface{}); v != nil {
 		for k, v := range v {
-			p[k] = v.(string)
+			m[k] = v.(string)
 		}
 	}
 
-	c["params"] = p
+	c[key] = m
 
-	return p
+	return m
 }
--- a/openstack/identity/v3/extensions/ec2tokens/requests.go
+++ b/openstack/identity/v3/extensions/ec2tokens/requests.go
@@ -1,0 +1,323 @@
+package ec2tokens
+
+import (
+	"crypto/hmac"
+	"crypto/sha1"
+	"crypto/sha256"
+	"encoding/hex"
+	"fmt"
+	"math/rand"
+	"net/url"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
+)
+
+const (
+	// aws4Request is a constant, used to generate AWS Credential V4.
+	aws4Request = "aws4_request"
+	// v2AlgoSha1 is a HMAC SHA1 signature method. Used to generate AWS
+	// Credential V2.
+	v2AlgoSha1 = "HmacSHA1"
+	// v2AlgoSha1 is a HMAC SHA256 signature method. Used to generate AWS
+	// Credential V2.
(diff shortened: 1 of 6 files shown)


jtopjian:
@kayrus This is looking good - I don't see any issues so far. As for unit tests, you can always test the various functions and methods without using mock http requests. For example, make direct calls to `canonicalQsV2` and compare the actual output with the expected out put. 

As you begin to finalize the code, it might be useful to add further documentation about what each function is doing. You can also link directly to the keystoneclient code as reference.

kayrus:
@jtopjian I added an ability to define a timestamp and a random hash, this allowed to perform unit tests. Also it is possible passthrough auth parameters with a signature without knowing the secret key.

I think the code is ready for review now. As for the acceptance tests, I'll add them after I introduce ec2 credentials management. You can test ec2 auth if you create them manually using `openstack ec2 credentials create`.

kayrus:
I forgot that credentials logic was already introduced in #1460

There are two endpoints available in keystone to manage ec2 credentials:
* `/v3/users/<userid>/credentials/OS-EC2`
* `/v3/credentials`

Creating ec2 credentials via `/v3/credentials` allows to specify more options like access ID and secret, while `/v3/users/<userid>/credentials/OS-EC2` automatically generates access ID and secret.

I'll create acceptance tests soon.

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+
+	var pairs []string
+	for _, k := range keys {
+		pairs = append(pairs, fmt.Sprintf("%s=%s", k, url.QueryEscape(params[k])))
+	}
+
+	return strings.Join(pairs, "&")
+}
+
+func stringToSignV2(opts AuthOptions, params map[string]string) string {
```
jtopjian: We should name this in an exportable and testable way along with including a doc string and unit test:

```
// EC2CredentialsBuildStringToSignV2 builds ...
```

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+		opts.Path,
+	}, "\n")
+
+	return strings.Join([]string{
+		stringToSign,
+		canonicalQsV2(params),
+	}, "\n")
+}
+
+func canonicalQsV4(verb string, params map[string]string) string {
```
jtopjian: We should name this in an exportable and testable way along with including a doc string and unit test:

```
// EC2CredentialsBuildCanonicalQueryStringV4 builds ...
// https://github.com/openstack/python-keystoneclient/blob/526350eaeeadef8c1ca5f80547c01f94342a6653/keystoneclient/contrib/ec2/utils.py#L244
```

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+	if verb == "POST" {
+		return ""
+	}
+	return canonicalQsV2(params)
+}
+
+func canonicalHeadersV4(headers map[string]string, signedHeaders string) string {
```
jtopjian: We should name this in an exportable and testable way along with including a doc string and unit test:

```
// EC2CredentialsBuildCanonicalHeaderStringV4 builds...
// https://github.com/openstack/python-keystoneclient/blob/526350eaeeadef8c1ca5f80547c01f94342a6653/keystoneclient/contrib/ec2/utils.py#L216
```

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+	for _, h := range strings.Split(signedHeaders, ";") {
+		if v, ok := headersLower[h]; ok {
+			headersList = append(headersList, h+":"+v)
+		}
+	}
+
+	return strings.Join(headersList, "\n") + "\n"
+}
+
+func sumHMAC1(key []byte, data []byte) []byte {
```
jtopjian: This function seems simple enough that I think we can forgo renaming, doc strings, unit tests, etc. However, I recommend moving this, and the other helper functions, to the very bottom of this file to denote that it's more of a simple helper function along with a general comment saying something along the lines of "The following are small helper functions used to help build the signature".

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+	hash := hmac.New(sha1.New, key)
+	hash.Write(data)
+	return hash.Sum(nil)
+}
+
+func sumHMAC256(key []byte, data []byte) []byte {
```
jtopjian: This function seems simple enough that I think we can forgo renaming, doc strings, unit tests, etc. However, I recommend moving this, and the other helper functions, to the very bottom of this file to denote that it's more of a simple helper function along with a general comment saying something along the lines of "The following are small helper functions used to help build the signature".

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+	hash := hmac.New(sha256.New, key)
+	hash.Write(data)
+	return hash.Sum(nil)
+}
+
+func signatureKeyV4(secret, date, region, service string) []byte {
```
jtopjian: We should name this in an exportable and testable way along with including a doc string and unit test:

```
// EC2CredentialsBuildSignatureKeyV4 builds...
// https://github.com/openstack/python-keystoneclient/blob/526350eaeeadef8c1ca5f80547c01f94342a6653/keystoneclient/contrib/ec2/utils.py#L169
```

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+	kDate := sumHMAC256([]byte("AWS4"+secret), []byte(date))
+	kRegion := sumHMAC256(kDate, []byte(region))
+	kService := sumHMAC256(kRegion, []byte(service))
+	return sumHMAC256(kService, []byte(aws4Request))
+}
+
+func signatureV4(opts AuthOptions, params map[string]string, date time.Time, bodyHash string) string {
```
jtopjian: We should name this in an exportable and testable way along with including a doc string and unit test:

```
// EC2CredentialsBuildSignatureV4 builds...
// https://github.com/openstack/python-keystoneclient/blob/526350eaeeadef8c1ca5f80547c01f94342a6653/keystoneclient/contrib/ec2/utils.py#L162
```

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+		scope,
+		hex.EncodeToString(hash[:]),
+	}, "\n")
+
+	key := signatureKeyV4(opts.Secret, date.Format(yyyymmdd), opts.Region, opts.Service)
+
+	return hex.EncodeToString(sumHMAC256(key, []byte(strToSign)))
+}
+
+func randomBodyHash() string {
```
jtopjian: This function seems simple enough that I think we can forgo renaming, doc strings, unit tests, etc. However, I recommend moving this, and the other helper functions, to the very bottom of this file to denote that it's more of a simple helper function along with a general comment saying something along the lines of "The following are small helper functions used to help build the signature".

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+	h := make([]byte, 64)
+	rand.Read(h)
+	return hex.EncodeToString(h)
+}
+
+func paramsToMap(c map[string]interface{}) map[string]string {
```
jtopjian: This function seems simple enough that I think we can forgo renaming, doc strings, unit tests, etc. However, I recommend moving this, and the other helper functions, to the very bottom of this file to denote that it's more of a simple helper function along with a general comment saying something along the lines of "The following are small helper functions used to help build the signature".

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+	b, err := gophercloud.BuildRequestBody(opts, "credentials")
+	if err != nil {
+		return nil, err
+	}
+
+	if opts.Signature != nil {
+		return b, nil
+	}
+
+	// calculate signature, when it is not set
```
jtopjian: Is there any benefit to refactoring this `ToTokenV3CreateMap` method to look more like this?

https://github.com/openstack/python-keystoneclient/blob/526350eaeeadef8c1ca5f80547c01f94342a6653/keystoneclient/contrib/ec2/utils.py#L87-L104

kayrus:
> With regard to unit tests and the random generation, isn't it possible to seed random with a constant so that the outcome is the same each time?

it is already done with an empty string:

https://github.com/kayrus/gophercloud/commit/9f7ec65b6dd2ce858ac954c651133956476d379f#diff-e00b786e8cb6f9ca02483223960d5978R83

I can add a static `foo`, if you want.

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+	// string. You can set the singature explicitly, when you already know
+	// it.
+	Signature interface{} `json:"signature"`
+	// Optional random hash, when signature is nil.
+	BodyHash *string `json:"body_hash"`
+	// Optional timestamp to calculate a V4 signature.
+	Timestamp *time.Time `json:"-"`
+}
+
+func canonicalQsV2(params map[string]string) string {
```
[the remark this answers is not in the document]
kayrus: done

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: done

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: done

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: done

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: Done.

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: done

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: done

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: done

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: done

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: done

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: my aim was to reduce an amount of if/then/else. added more comments.

kayrus:
@jtopjian code review acked.

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,279 @@
...
+	"net/url"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/tokens"
+)
+
+const (
```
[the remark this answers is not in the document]
kayrus: You were right. I have a use case, when I need to use these constants. Renamed them and exported.

kayrus:
Hold on, there is an issue with GET requests. I had not to use the "Params", they are used only for GET requests.

kayrus:
@jtopjian now it should be fine

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -0,0 +1,320 @@
...
+	if v, _ := c["body_hash"]; v == nil {
+		// when body_hash is not set, generate a random one
+		c["body_hash"] = randomBodyHash()
+	}
+
+	signedHeaders, _ := h["X-Amz-SignedHeaders"]
+	c["signature"] = EC2CredentialsBuildSignatureV4(*opts, signedHeaders, date, c["body_hash"].(string))
+
+	h["X-Amz-Date"] = date.Format(EC2CredentialsTimestampFormatV4)
+	h["Authorization"] = fmt.Sprintf("%s Credential=%s/%s/%s/%s/%s, SignedHeaders=%s, Signature=%s",
```
kayrus: @jtopjian what if we move `Authorization` header generation into a func?

jtopjian: If I'm understanding correctly, all this is doing is building a string, right? And a function for this would have to have several parameters passed into it. Unless I'm misunderstanding the idea you have, I think this is fine the way it is.

kayrus:
```diff
diff --git a/openstack/identity/v3/extensions/ec2tokens/requests.go b/openstack/identity/v3/extensions/ec2tokens/requests.go
index 9be866d1..233f00a3 100644
--- a/openstack/identity/v3/extensions/ec2tokens/requests.go
+++ b/openstack/identity/v3/extensions/ec2tokens/requests.go
@@ -180,6 +180,20 @@ func EC2CredentialsBuildSignatureV4(opts AuthOptions, signedHeaders string, date
        return hex.EncodeToString(sumHMAC256(key, []byte(strToSign)))
 }
 
+// EC2CredentialsBuildAuthorizationHeaderV4 builds an AWS v4 Authorization
+// header based on auth parameters, date and signature
+func EC2CredentialsBuildAuthorizationHeaderV4(opts AuthOptions, signedHeaders string, signature string, date time.Time) string {
+       return fmt.Sprintf("%s Credential=%s/%s/%s/%s/%s, SignedHeaders=%s, Signature=%s",
+               EC2CredentialsAwsHmacV4,
+               opts.Access,
+               date.Format(EC2CredentialsDateFormatV4),
+               opts.Region,
+               opts.Service,
+               EC2CredentialsAwsRequestV4,
+               signedHeaders,
+               signature)
+}
+
 // ToTokenV3ScopeMap is a dummy method to satisfy tokens.AuthOptionsBuilder interface
 func (opts *AuthOptions) ToTokenV3ScopeMap() (map[string]interface{}, error) {
        return nil, nil
@@ -239,24 +253,15 @@ func (opts *AuthOptions) ToTokenV3CreateMap(map[string]interface{}) (map[string]
        if opts.Timestamp != nil {
                date = *opts.Timestamp
        }
+       h["X-Amz-Date"] = date.Format(EC2CredentialsTimestampFormatV4)
        if v, _ := c["body_hash"]; v == nil {
                // when body_hash is not set, generate a random one
                c["body_hash"] = randomBodyHash()
        }
-
        signedHeaders, _ := h["X-Amz-SignedHeaders"]
-       c["signature"] = EC2CredentialsBuildSignatureV4(*opts, signedHeaders, date, c["body_hash"].(string))
-
-       h["X-Amz-Date"] = date.Format(EC2CredentialsTimestampFormatV4)
-       h["Authorization"] = fmt.Sprintf("%s Credential=%s/%s/%s/%s/%s, SignedHeaders=%s, Signature=%s",
-               EC2CredentialsAwsHmacV4,
-               opts.Access,
-               date.Format(EC2CredentialsDateFormatV4),
-               opts.Region,
-               opts.Service,
-               EC2CredentialsAwsRequestV4,
-               signedHeaders,
-               c["signature"])
+       signature := EC2CredentialsBuildSignatureV4(*opts, signedHeaders, date, c["body_hash"].(string))
+       c["signature"] = signature
+       h["Authorization"] = EC2CredentialsBuildAuthorizationHeaderV4(*opts, signedHeaders, signature, date)
 
        return b, nil
 }
```

jtopjian: Yeah, a single-line function feels unnecessary to me.

Issue #1901, opened 2020-03-23
Add support for update of Compute services


dkt26111:
There is currently no implementation for updating Compute services to either enable or disable a service:

https://docs.openstack.org/api-ref/compute/#update-compute-service

I will propose a PR shortly.

kayrus:
@jtopjian this issue is fixed by #1902

PR #1902, opened 2020-03-23
Add support for update of Compute service

dkt26111:
For #1901 

code in this PR:

API doc:
https://docs.openstack.org/api-ref/compute/#update-compute-service

API code:
https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/services.py#L199


dkt26111:

openstack/compute/v2/extensions/services/requests.go
```diff
@@ -11,3 +11,34 @@ func List(client *gophercloud.ServiceClient) pagination.Pager {
 		return ServicePage{pagination.SinglePageBase(r)}
 	})
 }
+
+// UpdateOpts specifies the base attributes that may be updated on a service.
+type UpdateOpts struct {
+	// Status represents the new service status. One of enabled or disabled.
+	Status string `json:"status,omitempty"`
```
jtopjian: Since this is input (part of a request) and there are finite choices, we can use a defined type:

```go
type ServiceStatus string

const (
  // ServiceEnabled is used to mark a service as being enabled.
  ServiceEnabled ServiceStatus = "enabled"

  // ServiceDisabled is used to mark a service as being disabled.
  ServiceDisabled ServiceStatus = "disabled"
)

type UpdateOpts struct {
  ...
  Status ServiceStatus `json:"status,omitempty"`
  ...
}
```

The `Status` field in the `results.go` will stay a string, though.

openstack/compute/v2/extensions/services/requests.go
```diff
@@ -11,3 +11,34 @@ func List(client *gophercloud.ServiceClient) pagination.Pager {
...
+
+	// DisabledReason represents the reason for disabling a service.
+	DisabledReason string `json:"disabled_reason,omitempty"`
+
+	// ForcedDown is a manual override to tell nova that the service in question
+	// has been fenced manually by the operations team.
+	ForcedDown bool `json:"forced_down,omitempty"`
```
jtopjian: The combination of `bool` and `omitempty` will cause values of false to never be sent. Just to be sure: there is absolutely no reason to ever send this field with a value of false, correct?

dkt26111: Correct, that is my understanding.

openstack/compute/v2/extensions/services/requests.go
dkt26111: Yea, that looks better.  I'll make that change.

Issue #1903, opened 2020-03-23
Missing list options to Compute API services list call


dkt26111:
The Compute API for listing services allows the specification of query parameters for binary or host name: https://docs.openstack.org/api-ref/compute/#list-compute-services

The current implementation of the List call does not allow for the specification of those parameters:
https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/services/requests.go#L9

I will put a PR to add options to the List API.

jtopjian:
Implemented in #1904

PR #1904, opened 2020-03-23
Adding query parameter to Compute list services API

dkt26111:
For #1903 

code in this PR:

API doc:
https://docs.openstack.org/api-ref/compute/#list-compute-services

API code:
https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/services.py#L345
https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/services.py#L79


dkt26111:
@jtopjian This is ready for review. Thanks.

PR #1906, opened 2020-03-26
Keystone V3: Support S3 tokens signature validation

### LINKED ISSUE #1905: Keystone V3: Add support for S3 tokens signature verification
It has a similar to ec2tokens logic, but doesn't return TokenID. Used for verification only.

https://github.com/openstack/keystone/blob/stable/train/keystone/api/s3tokens.py#L95

kayrus:
Resolves #1905
@jtopjian this PR is a bit hacky, because if I put s3tokens into a separate package, I need to duplicate a lot of logic.

### DIFF
--- a/openstack/identity/v3/extensions/ec2tokens/requests.go
+++ b/openstack/identity/v3/extensions/ec2tokens/requests.go
@@ -72,8 +72,10 @@ type AuthOptions struct {
 	BodyHash *string `json:"body_hash"`
 	// Timestamp is a timestamp to calculate a V4 signature. Optional.
 	Timestamp *time.Time `json:"-"`
-	// Token is a []byte string (encoded to base64 automatically) which was signed
-	// by an EC2 secret key. Used by S3 tokens for validation only.
+	// Token is a []byte string (encoded to base64 automatically) which was
+	// signed by an EC2 secret key. Used by S3 tokens for validation only.
+	// Token must be set with a Signature. If a Signature is not provided,
+	// a Token will be generated automatically along with a Signature.
 	Token []byte `json:"token,omitempty"`
 }
 
@@ -272,6 +274,7 @@ func (opts *AuthOptions) ToTokenV3CreateMap(map[string]interface{}) (map[string]
 	h["X-Amz-Date"] = date.Format(EC2CredentialsTimestampFormatV4)
 	h["Authorization"] = EC2CredentialsBuildAuthorizationHeaderV4(*opts, signedHeaders, c["signature"].(string), date)
 
+	// token is only used for S3 tokens validation and will be removed when using EC2 validation
 	c["token"] = stringToSign
 
 	return b, nil
--- a/openstack/identity/v3/extensions/ec2tokens/requests.go
+++ b/openstack/identity/v3/extensions/ec2tokens/requests.go
@@ -72,6 +72,9 @@ type AuthOptions struct {
 	BodyHash *string `json:"body_hash"`
 	// Timestamp is a timestamp to calculate a V4 signature. Optional.
 	Timestamp *time.Time `json:"-"`
+	// Token is a []byte string (encoded to base64 automatically) which was signed
+	// by an EC2 secret key. Used by S3 tokens for validation only.
+	Token []byte `json:"token,omitempty"`
 }
 
 // EC2CredentialsBuildCanonicalQueryStringV2 builds a canonical query string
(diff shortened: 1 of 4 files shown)


openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -72,6 +72,9 @@ type AuthOptions struct {
 	BodyHash *string `json:"body_hash"`
 	// Timestamp is a timestamp to calculate a V4 signature. Optional.
 	Timestamp *time.Time `json:"-"`
+	// Token is a []byte string (encoded to base64 automatically) which was signed
+	// by an EC2 secret key. Used by S3 tokens for validation only.
+	Token []byte `json:"token,omitempty"`
```
jtopjian: Just so I understand: this field is for someone to supply a pre-generated signed string?

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -245,18 +265,14 @@ func (opts *AuthOptions) ToTokenV3CreateMap(map[string]interface{}) (map[string]
...
-		opts.Access,
-		date.Format(EC2CredentialsDateFormatV4),
-		opts.Region,
-		opts.Service,
-		EC2CredentialsAwsRequestV4,
-		signedHeaders,
-		c["signature"])
+	h["Authorization"] = EC2CredentialsBuildAuthorizationHeaderV4(*opts, signedHeaders, c["signature"].(string), date)
+
+	c["token"] = stringToSign
```
jtopjian: If my previous comment is true, should there be some kind of check to make sure this isn't already set so it's not clobbered?

Additionally, let's add a comment above this line to say that while `token` is being set, it is only used for S3 and will be removed when using EC2 validation.

openstack/identity/v3/extensions/ec2tokens/requests.go
```diff
@@ -280,6 +299,29 @@ func Create(c *gophercloud.ServiceClient, opts tokens.AuthOptionsBuilder) (r tok
 	return
 }
 
+// ValidateS3Token authenticates an S3 request using EC2 credentials. Doesn't
+// generate a new token ID, but returns a tokens.CreateResult.
+func ValidateS3Token(c *gophercloud.ServiceClient, opts tokens.AuthOptionsBuilder) (r tokens.CreateResult) {
```
jtopjian: How about doing something similar to `tokens.Validate` where `(bool, error)` is returned: https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/tokens/requests.go#L145

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: Exactly.

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: Token must be set with a signature. If signature was not provided, token will be generated as well.

openstack/identity/v3/extensions/ec2tokens/requests.go
kayrus: What if we need to fetch user/project/domain info from the token response?

jtopjian: Are you sure that's included? The debug output doesn't show it.

kayrus: It returns the same token body as usual, when keystone auths a user. The only difference is it doesn't return a token.

kayrus: see https://github.com/openstack/keystone/blob/40b7de87e04fbc49ab3a96b5114a9ab09c2734b7/keystone/api/s3tokens.py#L100..L105 vs https://github.com/openstack/keystone/blob/40b7de87e04fbc49ab3a96b5114a9ab09c2734b7/keystone/api/ec2tokens.py#L67..L73

jtopjian: Oh, right - the output is being redacted/removed in the acceptance tests. OK, I don't want to bike shed this one. Perhaps the `Validate` function I linked to should be renamed to `IsValid` at some point in the future.

kayrus: Indeed debug doesn't show the json response, but content length is big and it should include the token data:

```
2020/03/27 13:58:13 [DEBUG] OpenStack Response Code: 200
2020/03/27 13:58:13 [DEBUG] OpenStack Response Headers: 
Content-Length: 6751
Date: Fri, 27 Mar 2020 13:58:13 GMT
Server: Apache/2.4.29 (Ubuntu)
Vary: X-Auth-Token
X-Openstack-Request-Id: req-07cf74fe-a252-4cbb-8107-0d263dc54b3e
```

Issue #1907, opened 2020-03-27
Support auth with system scope


zhsj:
https://docs.openstack.org/api-ref/identity/v3/?expanded=password-authentication-with-scoped-authorization-detail#system-scoped-example

``` json
{
    "auth": {
        "identity": {
            "methods": [
                "password"
            ],
            "password": {
                "user": {
                    "id": "ee4dfb6e5540447cb3741905149d9b6e",
                    "password": "devstacker"
                }
            }
        },
        "scope": {
            "system": {
                "all": true
            }
        }
    }
}
```

jtopjian:
Implemented in #1908

PR #1908, opened 2020-03-27
Support system scope in v3 auth

zhsj:
For #1907

https://docs.openstack.org/api-ref/identity/v3/?expanded=password-authentication-with-scoped-authorization-detail#system-scoped-example

https://github.com/openstack/keystone/blob/f5c9b094af8442219e75265a87551873e058003b/keystone/auth/schema.py#L97-L102

### DIFF
--- a/openstack/identity/v3/tokens/testing/requests_test.go
+++ b/openstack/identity/v3/tokens/testing/requests_test.go
@@ -275,6 +275,31 @@ func TestCreateProjectNameAndDomainNameScope(t *testing.T) {
 	`)
 }
 
+func TestCreateSystemScope(t *testing.T) {
+	options := tokens.AuthOptions{UserID: "fenris", Password: "g0t0h311"}
+	scope := &tokens.Scope{System: true}
+	authTokenPost(t, options, scope, `
+		{
+			"auth": {
+				"identity": {
+					"methods": ["password"],
+					"password": {
+						"user": {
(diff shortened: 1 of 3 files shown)


openstack/identity/v3/tokens/testing/requests_test.go
```diff
@@ -248,6 +248,31 @@ func TestCreateProjectNameAndDomainIDScope(t *testing.T) {
...
+					"system": {
+						"all": true
+					}
+				}
+			}
+		}
+	`)
+}
+
+func TestCreateSystemScope(t *testing.T) {
```
jtopjian: @zhsj I think you want the body of the above test in this function?

jtopjian:
@zhsj Thank you for submitting this. For future reference, we don't use the API docs as the only reference - we also need the actual OpenStack service code that implements the API feature. I've modified this PR description for you to include it.

openstack/identity/v3/tokens/testing/requests_test.go
zhsj: I copied the `TestCreateProjectNameAndDomainNameScope` to `TestCreateSystemScope`, but I ended up modifying the `TestCreateProjectNameAndDomainNameScope` func...

Now fixed.

jtopjian:
The OpenLab error is ignorable.

Issue #1914, opened 2020-03-31
orchestration templates fail to parse when using get_file


andrein:
When get_file is used in a template and the referenced file is not json/yaml, gophercloud throws the error: "Data in neither json nor yaml format." 

It looks like this happens because ```get_file``` and ```type``` are treated exactly the same way, but ```get_file``` should not fail if the content is not json/yaml

jtopjian:
Fixedin #1915

PR #1915, opened 2020-04-01
Fix orchestration template parsing

andrein:
When get_file is used in a template and the referenced file is not json/yaml,
gophercloud throws a cryptic error: "Data in neither json nor yaml format."

For #1914

### DIFF
--- a/openstack/orchestration/v1/stacks/utils.go
+++ b/openstack/orchestration/v1/stacks/utils.go
@@ -122,11 +122,6 @@ func (t *TE) Parse() error {
 	return nil
 }
 
-// Validate validates the contents of TE
-func (t *TE) Validate() error {
-	return nil
-}
-
 // igfunc is a parameter used by GetFileContents and GetRRFileContents to check
 // for valid URL's.
 type igFunc func(string, interface{}) bool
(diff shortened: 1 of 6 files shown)


openstack/orchestration/v1/stacks/utils.go
```diff
@@ -116,7 +119,7 @@ func (t *TE) Parse() error {
 			return ErrInvalidDataFormat{}
 		}
 	}
-	return t.Validate()
+	return t.Validate() // FIXME This always calls TE.Validate, not Template.Validate/Environment.Validate
```
jtopjian, line 122: I think the easiest way to deal with this would be to change this line here to `return nil` and then do something like:

```go
if err := childTemplate.Parse(); err == nil {
  if err := childTemplate.Validate(); err == nil {
    if err := childTemplate.getFileContents(childTemplate.Parsed, ignoreIf, recurse); err != nil {
      return err
    }
  }
}
```

andrein: Ok, in that case, shall I also remove ```func (t *TE) Validate() error```?

jtopjian: Yes, let's go ahead and remove it.

andrein: Ok, done!

andrein:
recheck

jtopjian:
@andrein Thank you for fixing this issue!

PR #1916, opened 2020-04-01
Keystone V3: Support EC2 credentials CRD

### LINKED ISSUE #1913: Keystone V3: Add support for EC2 credentials CRD
There are two endpoints available in keystone to manage ec2 credentials:
* `/v3/users/<userid>/credentials/OS-EC2` - https://docs.openstack.org/api-ref/identity/v2-ext/ , https://github.com/openstack/keystone/blob/stable/train/keystone/api/users.py#L311..L411
* https://github.com/openstack/keystone/blob/40b7de87e04fbc49ab3a96b5114a9ab09c2734b7/keystone/api/users.py#L50-L63
* `/v3/credentials` support was introduced in #1460 

Creating ec2 credentials via `/v3/credentials` allows to specify more options like access ID and secret, while `/v3/users/<userid>/credentials/OS-EC2` automatically generates access ID and secret.

kayrus:
resolves #1913

### DIFF
--- a/acceptance/openstack/identity/v3/ec2credentials_test.go
+++ b/acceptance/openstack/identity/v3/ec2credentials_test.go
@@ -1,0 +1,96 @@
+// +build acceptance
+
+package v3
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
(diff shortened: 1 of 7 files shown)


kayrus:
@jtopjian ready for review

jtopjian:
@kayrus This looks good - nice work! Can you add an example to `doc.go`?

PR #1922, opened 2020-04-02
Keystone V3: Introduce TOTP (two factor) auth support

### LINKED ISSUE #1917: Keystone V3: Add TOTP (two factor) auth method support
https://docs.openstack.org/keystoneauth/latest/api/keystoneauth1.identity.v3.html#keystoneauth1.identity.v3.TOTPMethod
https://docs.openstack.org/api-ref/identity/v3/?expanded=multi-step-authentication-2-factor-password-and-totp-example-detail#multi-step-authentication-2-factor-password-and-totp-example

https://github.com/openstack/keystone/blob/stable/train/keystone/auth/plugins/totp.py#L92

kayrus:
Resolves #1917

@jtopjian  I didn't find a way to perform tests in multi-step approach https://docs.openstack.org/api-ref/identity/v3/?expanded=multi-step-authentication-2-factor-password-and-totp-example-detail#multi-step-authentication-2-factor-password-and-totp-example

In my case passcodes are generated using a dongle and I can enter them in advance, therefore the auth request sends both password and totp auth requests in one call.

I'll try to add a receipt response handler in further PRs.

### DIFF
--- a/auth_options.go
+++ b/auth_options.go
@@ -181,7 +181,7 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]interface{}) (map[s
 		Password              *passwordReq              `json:"password,omitempty"`
 		Token                 *tokenReq                 `json:"token,omitempty"`
 		ApplicationCredential *applicationCredentialReq `json:"application_credential,omitempty"`
-		Totp                  *totpReq                  `json:"totp,omitempty"`
+		TOTP                  *totpReq                  `json:"totp,omitempty"`
 	}
 
 	type authReq struct {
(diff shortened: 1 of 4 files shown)


jtopjian:
>  I didn't find a way to perform tests in multi-step approach 

I'm OK with skipping tests on this one - I understand it can be difficult. As long as you've confirmed it works in some environment, that's good enough for me.

One style request for this PR: let's use `TOTP` and `totp`.

jtopjian:
@kayrus This looks good to me. Are you still working on it or is it ready to be merged?

kayrus:
@jtopjian ready

PR #1930, opened 2020-04-10
Swift: Add "bulk-delete" URL argument support

### LINKED ISSUE #1721: object storage bulk delete
Is [bulk delete supported](https://docs.openstack.org/ocata/user-guide/cli-swift-bulk-delete.html) on the [object storage API](https://godoc.org/github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects#Delete) ?

kayrus:
Resolves #1721

https://github.com/openstack/swift/blob/a495f1e327fac17246a296721fb1eda8470e4773/swift/common/middleware/bulk.py#L740

Response dict: https://github.com/openstack/swift/blob/a495f1e327fac17246a296721fb1eda8470e4773/swift/common/middleware/bulk.py#L413

### DIFF
--- a/acceptance/openstack/objectstorage/v1/containers_test.go
+++ b/acceptance/openstack/objectstorage/v1/containers_test.go
@@ -161,3 +161,40 @@ func TestListAllContainers(t *testing.T) {
 	}
 	th.AssertEquals(t, numContainers, len(containerNamesList))
 }
+
+func TestBulkDeleteContainers(t *testing.T) {
+	client, err := clients.NewObjectStorageV1Client()
+	if err != nil {
+		t.Fatalf("Unable to create client: %v", err)
+	}
+
+	numContainers := 20
+
+	// Create a slice of random container names.
+	cNames := make([]string, numContainers)
+	for i := 0; i < numContainers; i++ {
+		cNames[i] = tools.RandomString("test&happy?-", 8)
+	}
+
+	// Create numContainers containers.
+	for i := 0; i < len(cNames); i++ {
+		res := containers.Create(client, cNames[i], nil)
+		th.AssertNoErr(t, res.Err)
+	}
+
+	expectedResp := containers.BulkDeleteResponse{
+		ResponseStatus: "200 OK",
+		Errors:         [][]string{},
+		NumberDeleted:  numContainers,
+	}
+
+	resp, err := containers.BulkDelete(client, cNames).Extract()
+	th.AssertNoErr(t, err)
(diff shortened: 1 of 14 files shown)


kayrus:
@jtopjian ready for review

jtopjian:
We should be able to easily add an acceptance test for this, right? Let me know if you need help.

kayrus:
@jtopjian just noticed that bulk-delete is also valid for containers. It would be great if you help me with acceptance tests.

kayrus:
According to https://github.com/openstack/swift/blob/a495f1e327fac17246a296721fb1eda8470e4773/swift/common/middleware/bulk.py#L156..L194, bulk delete should also contain a body with filenames separated by a newline.

jtopjian:
> bulk delete should have a Content-Type: text/plain header and contain a body with filenames separated by a newline.

Right. Now I recall that I looked into this in the past and silently backed away 🙂 

This means that we'll need to modify `UpdateOpts` and `DeleteOps` for both containers and objects to include an option for a body, and then pass that body in the actual `Update` and `Delete` calls.

In addition, an actual structured response body is sent back, which differs from a normal Swift response where the bulk of the info is in the headers: https://docs.openstack.org/ocata/user-guide/cli-swift-bulk-delete.html

In my opinion, a better approach might be to make entirely new functions/actions called `BulkDelete` with their own `BulkDeleteOpts`. I believe Swift treats `POST` and `DELETE` equally, so let's just standardize on using `POST` (including a body with a `DELETE` call feels wrong).

`BulkDeleteOpts` can have a `Containers` or `Objects` field of type `[]string` where the user can specify the list of containers/objects to be bulk deleted. in the `To<Container/Object>BulkDeleteQuery` method, we can return a plaintext string of `[]string` joined by a newline, so that it can be passed to the `client.Post` method.

Finally, I think a 409 response (I think this means that only _some_ of the requested containers/objects were deleted) might be a little common and we should either consider treating it as valid response code... but I'm not 100% sure on this, yet.

Thoughts?

kayrus:
@jtopjian no concerns, though at first glance it looked like a low hanging fruit.

kayrus:
@jtopjian I reworked the PR. I also discovered and fixed redundant pointers usage and unencoded container/object names.

kayrus:
@jtopjian can you add a release note that container and object names are now urlencoded?

jtopjian:
Sure - but we should try to have things like this in separate PRs.

PR #1934, opened 2020-04-10
Cinder V3: Add support for Cinders Attachment API

### LINKED ISSUE #414: No support for new V3.27 Cinder Attachment API
With V3.27 Cinder introduced a standalone Attachment API, this API allows attachments to be created/updated/deleted etc in lieu of Nova and are now treated as independent resources in Cinder.  

We should add V3 support and implement micro versions for Cinder (Issue #413) then add the attachment API impl.

kayrus:
Resolves #414


kayrus:
recheck

kayrus:
@jtopjian ready for review

PR #1935, opened 2020-04-13
Keystone V3: Introduce OAuth1 auth support

### LINKED ISSUE #1418: Auth: add OAuth authentication method
[API reference](https://developer.openstack.org/api-ref/identity/v3-ext/?expanded=authenticate-with-identity-api-detail#os-oauth1-api)

Keystone code:
- https://github.com/openstack/keystone/blob/master/keystone/auth/plugins/oauth1.py
- https://github.com/openstack/keystone/blob/master/keystone/api/os_oauth1.py

Even using the generated token, there is an issue:

> `*terraform.EvalOpFilter, err: Request forbidden: [POST http://identity:5000/v3/auth/tokens], error message: {"error": {"message": "You are not authorized to perform the requested action: Using OAuth-scoped token to create another token. Create a new OAuth-scoped token instead.", "code": 403, "title": "Forbidden"}}`

```
2019/01/29 12:57:36 [DEBUG] OpenStack Request URL: POST https://identity/v3/auth/tokens
2019/01/29 12:57:36 [DEBUG] OpenStack Request Headers:
Accept: application/json
User-Agent: Terraform/0.11.7 gophercloud/2.0.0
2019/01/29 12:57:36 [DEBUG] OpenStack Request Body: {
  "auth": {
    "identity": {
      "methods": [
        "token"
      ],
      "token": {
        "id": "***"
      }
    }
  }
}
2019/01/29 12:57:37 [DEBUG] OpenStack Response Code: 403
2019/01/29 12:57:37 [DEBUG] OpenStack Response Headers:
Content-Length: 207
Date: Tue, 29 Jan 2019 11:57:37 GMT
Server: nginx/1.15.3
Strict-Transport-Security: max-age=15724800; includeSubDomains
Vary: X-Auth-Token
X-Openstack-Request-Id: req-89efad60-a071-4287-83c5-fd3cd8eca1a7

2019/01/29 12:57:37 [DEBUG] OpenStack Response Body: {
  "error": {
    "code": 403,
    "message": "You are not authorized to perform the requested action: Using OAuth-scoped token to create another token. Create a new OAuth-scoped token instead.",
    "title": "Forbidden"
  }
}
```

Ref: #1214

kayrus:
Resolves #1418 

@jtopjian let me know if you find something wrong in this code before I start code cleanup.

### DIFF
--- a/openstack/identity/v3/extensions/oauth1/requests.go
+++ b/openstack/identity/v3/extensions/oauth1/requests.go
@@ -430,7 +430,7 @@ func GetAccessTokenRole(client *gophercloud.ServiceClient, userID string, id str
 // The following are small helper functions used to help build the signature.
 
 // tokenParams builds a URLEncoded parameters string.
-func tokenParams(query url.Values, timestamp *time.Time, callback string) string {
+func tokenParams(query url.Values, timestamp *time.Time, callback string) url.Values {
 	if timestamp != nil {
 		// use provided timestamp
 		query.Set("oauth_timestamp", strconv.FormatInt(timestamp.Unix(), 10))
@@ -449,12 +449,11 @@ func tokenParams(query url.Values, timestamp *time.Time, callback string) string
 	}
 	query.Set("oauth_version", "1.0")
 
-	// sorted by key
-	return query.Encode()
+	return query
 }
 
 // stringToSign builds a string to be signed.
-func stringToSign(method string, u string, params string) []byte {
+func stringToSign(method string, u string, query url.Values) []byte {
 	parsedURL, _ := url.Parse(u)
 	p := parsedURL.Port()
 	s := parsedURL.Scheme
@@ -467,7 +466,7 @@ func stringToSign(method string, u string, params string) []byte {
 	return []byte(strings.Join([]string{
 		method,
 		url.QueryEscape(parsedURL.String()),
-		url.QueryEscape(params),
+		url.QueryEscape(query.Encode()), // sorted by key
 	}, "&"))
 }
 
@@ -490,23 +489,21 @@ func signString(signatureMethod SignatureMethod, strToSign []byte, secrets []str
 
 // genAuthHeader generates an OAuth1 Authorization header with a signature
 // calculated using an OAuth1 signature method.
-func genAuthHeader(signatureMethod SignatureMethod, method, u, b string, secrets []string) string {
+func genAuthHeader(signatureMethod SignatureMethod, method, u string, query url.Values, secrets []string) string {
 	var authHeader []string
-	params, _ := url.ParseQuery(b)
-
 	var keys []string
-	for k := range params {
+	for k := range query {
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
 
 	for _, k := range keys {
-		for _, v := range params[k] {
+		for _, v := range query[k] {
 			authHeader = append(authHeader, fmt.Sprintf("%s=%q", k, url.QueryEscape(v)))
 		}
 	}
 
-	strToSign := stringToSign(method, u, b)
+	strToSign := stringToSign(method, u, query)
 	signature := url.QueryEscape(signString(signatureMethod, strToSign, secrets))
 	authHeader = append(authHeader, fmt.Sprintf("oauth_signature=%q", signature))
 
--- a/openstack/identity/v3/extensions/oauth1/requests.go
+++ b/openstack/identity/v3/extensions/oauth1/requests.go
@@ -76,9 +76,18 @@ func (opts AuthOptions) ToTokenV3CreateHeaders(method string, url string) (map[s
 	}, nil
 }
 
-// ToTokenV3CreateMap builds a create request body from the AuthOptions.
+// ToTokenV3CreateMap builds a create request body.
 func (opts AuthOptions) ToTokenV3CreateMap(map[string]interface{}) (map[string]interface{}, error) {
-	return nil, nil
+	var req struct {
+		Auth struct {
+			Identity struct {
+				Methods []string `json:"methods"`
+				OAuth1  struct{} `json:"oauth1"`
+			} `json:"identity"`
+		} `json:"auth"`
+	}
+	req.Auth.Identity.Methods = []string{"oauth1"}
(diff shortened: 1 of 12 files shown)


kayrus:
@jtopjian done cleanup, ready for review.

openstack/identity/v3/extensions/oauth1/requests.go
```diff
@@ -0,0 +1,490 @@
...
+		query.Set("oauth_timestamp", strconv.FormatInt(timestamp.Unix(), 10))
+	} else {
+		// use current timestamp
+		query.Set("oauth_timestamp", strconv.FormatInt(time.Now().UTC().Unix(), 10))
+	}
+
+	if nonce == "" {
+		// when nonce is not set, generate a random one
+		query.Set("oauth_nonce", strconv.FormatInt(rand.Int63(), 10)+query.Get("oauth_timestamp"))
+	}
```
jtopjian, line 422: Should there be an `else` here?

openstack/identity/v3/extensions/oauth1/requests.go
```diff
@@ -0,0 +1,490 @@
...
+
+	// sorted by key
+	return query.Encode()
+}
+
+// stringToSign builds a string to be signed.
+func stringToSign(method string, u string, params string) []byte {
+	parsedURL, _ := url.Parse(u)
+	if strings.HasPrefix(parsedURL.Host, "127.0.0.1") {
+		// Workaround for unit tests with dynamic ports
```
jtopjian, line 437: I was able to get around this by doing the following:

```go
func TestAuthenticate(t *testing.T) {
  l, _ := net.Listen("tcp", "127.0.0.1:33199")
  th.Server = httptest.NewUnstartedServer(th.Mux)
  th.Server.Listener = l
  th.Server.Start()

  //th.SetupHTTP()
  defer th.TeardownHTTP()
  HandleAuthenticate(t)
...
```

However, the test is still failing, but I'm not sure why. 

I'd definitely like to see this conditional removed somehow, though.

openstack/identity/v3/extensions/oauth1/requests.go
kayrus: No, but you pointed me to the fact, that there is no need to pass "nonce" as a parameter, since `oauth_nonce` is already parsed by `gophercloud.BuildQueryString`.

openstack/identity/v3/extensions/oauth1/requests.go
kayrus: I created an extra testhelper function to listen the static port. Additionally I noticed that using "https://identity:443/" URL containing the default port is [not supported by oauth](https://oauth.net/core/1.0a/#rfc.section.9.1.2), therefore I modified the logic to normalize URLs with default port.

kayrus:
> In addition, it looks like a lot of the opts structs are missing builder interfaces - we should probably include those.

I don't think that build interface is necessary here. I don't expect that oauth method must be extended by additional parameters or logic.

> I hope to have more time tomorrow to continue looking at this - I'm sorry about that.

No worries.

jtopjian:
> I don't think that build interface is necessary here. I don't expect that oauth method must be extended by additional parameters or logic.

We should include the builder interfaces even if we can't think of a reason to use them. Any time in the past that I have assumed they wouldn't be used, someone usually reports how they're missing :)

kayrus:
> We should include the builder interfaces even if we can't think of a reason to use them. Any time in the past that I have assumed they wouldn't be used, someone usually reports how they're missing :)

I need to rewrite existing logic, since `opts.OAuthConsumerSecret`  and `opts.OAuthTokenSecret` are used inside the `Create` funcs. This requires type assertion and I'd like to avoid this.

jtopjian:
Right, I see. I'm not trying to cause large rewrites, but I think if the `Create` function is tightly coupled to the opts structure, it might be another good indication that an interface is needed. 

The interface methods can be pretty flexible with their return values, so that might be a way to link the Create function to the opts structure.

kayrus:
@jtopjian As you mentioned `Create` function is tightly coupled to the opts structure. Therefore I used type assertion, because I have to, otherwise I cannot get an access to secrets.

As for the rest `Create` functions, done.

jtopjian:
Could you create a new method to return the required information specific to this package?

kayrus:
@jtopjian Not sure if I understood you correctly, but see my latest commit.

jtopjian:
Yes, that should work.

kayrus:
@jtopjian while I took a look on these interfaces. I noticed that there is no need to pass scope map into the `ToTokenV3CreateMap(scope map[string]interface{}) (map[string]interface{}, error)` interface. We can call `ToTokenV3ScopeMap` method directly inside the `ToTokenV3CreateMap` method.
Thoughts?

kayrus:
recheck

jtopjian:
@kayrus Thank you for working on this. The general logic looks good and I appreciate the time you took investigating how to successfully generate oauth1-based requests.

I've done some refactoring to this which you can see here (https://github.com/jtopjian/gophercloud/commit/5aad4d75e74e82e28bdc2283c0bc05ebf82f33e5). Let me know what you think.

kayrus:
@jtopjian thanks for you time. I added my comments. Don't hesitate to commit directly to my branch to run the CI/CD pipeline next time. Github repo owners have this ability by default.

jtopjian:
> Don't hesitate to commit directly to my branch to run the CI/CD pipeline next time.

Thanks for the review - I've made changes per your comments and will commit to this branch. I didn't want to do that initially until you had a first look.

kayrus:
@jtopjian I moved json struct inside the `ToTokenV3CreateMap` to match other `ToTokenV3CreateMap` methods.

kayrus:
@jtopjian rebased to master

kayrus:
@jtopjian can you hold with this PR? Considering the #1949 I'd like to introduce a proper logic to extract/parse the `TokenResult`. Otherwise if `Extract()` is not called, the response body will remain opened.

kayrus:
recheck

kayrus:
@jtopjian PR is ready

PR #1937, opened 2020-04-15
Keystone V3: Fix time elements handling

### LINKED ISSUE #1938: Keystone V3: Change ApplicationCredentials CreateOpts.ExpiresAt from string to time
Proper time format for application credentials and trusts is `JSONRFC3339Milli`. AppCreds has string format. Trusts sets this format inside the `ToTrustCreateMap` instead of proper `MarshalJSON` extension.

kayrus:
Resolves #1938 
This PR fixes issues in time handling.
Additionally I noticed that there is no error check in baremetal packages.


jtopjian:
@kayrus Can you please create a corresponding issue that details the changes being made here?

I've been flexible with combining multiple features/fixes in a single PR, but we still need issues to describe why changes are being made.

kayrus:
@jtopjian updated

jtopjian:
Thank you!

openstack/identity/v3/applicationcredentials/requests.go
```diff
@@ -69,7 +72,25 @@ type CreateOpts struct {
 	// A list of access rules objects.
 	AccessRules []AccessRule `json:"access_rules,omitempty"`
 	// The expiration time of the application credential, if one was specified.
-	ExpiresAt string `json:"expires_at,omitempty"`
+	ExpiresAt *time.Time `json:"-"`
+}
+
+// MarshalJSON converts our struct request into JSON.
+func (opts CreateOpts) MarshalJSON() ([]byte, error) {
```
jtopjian, line 79: This can instead be implemented as:

```go
// ToApplicationCredentialCreateMap formats a CreateOpts into a create request.
func (opts CreateOpts) ToApplicationCredentialCreateMap() (map[string]interface{}, error) {
  b, err := gophercloud.BuildRequestBody(opts, "application_credential")
  if err != nil {
    return b, err
  }

  if opts.ExpiresAt != nil {
    appCred := b["application_credential"].(map[string]interface{})
    appCred["expires_at"] = opts.ExpiresAt.Format(gophercloud.RFC3339Milli)
    b["application_credential"] = appCred
  }

  return b, nil
}
```

This pattern is used in other areas of Gophercloud already.

In addition, we'll need to verify what time format Keystone is expecting. We're using Milli here, but MilliNoZ is being returned - that might cause some confusion and usage errors.

kayrus: My idea was to avoid body modification inside the `To***CreateMap`, but do this in `MarshalJSON`.

I clarified time format:
* Trusts accept and return RFC3339Milli, but response truncates milliseconds. Therefore for tests we need to truncate milliseconds.
* AppCreds accept and return RFC3339MilliNoZ. However for tests we need to truncate nanoseconds.

kayrus:
recheck

kayrus:
@jtopjian PR is ready to review. I hope that `MarshalJSON` is a better approach comparing to type assertion in `ToApplicationCredentialCreateMap`

kayrus:
@jtopjian from my perspective `To***Map` should be used, when `MarshalJSON` cannot be used, e.g. `Extra` field in `users` package is an excellent example: 
https://github.com/gophercloud/gophercloud/blob/b88f26317c347bc39d800e54663fd86c2a35d53e/openstack/identity/v3/users/requests.go#L142

Or in auth options
https://github.com/gophercloud/gophercloud/blob/b88f26317c347bc39d800e54663fd86c2a35d53e/auth_options.go#L140 when you need to create a complex body without altering the original `opts` structure. Let me know if you agree, and I'll add this info into a contributor guide.

kayrus:
Here are the places, where you can use JSONMarshal as well:

https://github.com/gophercloud/gophercloud/blob/b88f26317c347bc39d800e54663fd86c2a35d53e/openstack/keymanager/v1/orders/requests.go#L104

https://github.com/gophercloud/gophercloud/blob/b88f26317c347bc39d800e54663fd86c2a35d53e/openstack/keymanager/v1/secrets/requests.go#L234

https://github.com/gophercloud/gophercloud/blob/b88f26317c347bc39d800e54663fd86c2a35d53e/openstack/workflow/v2/crontriggers/requests.go#L54

jtopjian:
> Here are the places, where you can use JSONMarshal as well:

Two of those were written by me and I explicitly chose not to use `MarshalJSON` 🙂  

> from my perspective To***Map should be used, when MarshalJSON cannot be used,

I agree that it could be used, but I have to respectfully request that it isn't.

The `To***` method patterns are used to build a standard request as well as to modify a request beyond what simple marshalling can do. This is a standard pattern used throughout Gophercloud. By adding another option, `MarshalJSON`, you break that pattern and provide another choice that can already be solved by using the original pattern.

The opposite is not true: you can't use `MarshalJSON` for everything that the `To***` methods are used for.

In addition, `To***Map` provides easier interaction with the data. Compare the example I provided in this PR versus having to create a new type and build a new struct - it's much harder to comprehend.

There are a few occurrences of `MarshalJSON` in Gophercloud and I've often made notes to remove them, but it hasn't really been a pressing issue. Even this discussion we're having now isn't terribly pressing - it's not going to severely hurt the project if `MarshalJSON` is used, but given the amount of contributions you make to Gophercloud, I'd prefer us having a detailed conversation to ensure we're on the same page for future work.

I understand Gophercloud leverages `UnmarshalJSON` heavily and so it's understandable to think that if `UnmarshalJSON` is heavily used, then we should give equal weight to `MarshalJSON`. Perhaps Gophercloud could/should have a similar `ToResultExtractMap` method similar to the request methods for consistency - but that's a much larger topic to discuss.

For now, I'd prefer having the rules of:

1. Use the `To***` methods to build a request
2. Use the `UnmarshalJSON` method to build a response.

kayrus:
@jtopjian thanks for clarification. I'll use `To***` method.

kayrus:
@jtopjian applied fixes

PR #1944, opened 2020-04-19
Swift: Simplify json unmarshal using ",string" struct tag

### LINKED ISSUE #1943: Swift: Simplify json unmarshal using ",string" struct tag
https://golang.org/pkg/encoding/json/

> The "string" option signals that a field is stored as JSON inside a JSON-encoded string. It applies only to fields of string, floating point, integer, or boolean types. This extra level of encoding is sometimes used when communicating with JavaScript programs:

```ho
Int64String int64 `json:",string"`
```

This can be used in objectstorage packages.

kayrus:
Resolves #1943

### DIFF
--- a/openstack/objectstorage/v1/accounts/results.go
+++ b/openstack/objectstorage/v1/accounts/results.go
@@ -2,7 +2,6 @@ package accounts
 
 import (
 	"encoding/json"
-	"strconv"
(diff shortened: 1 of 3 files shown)


kayrus:
@jtopjian ready for review

Issue #1838, opened 2020-02-06
Connection is not reused


jonas0616:
In `provider_client.go`, it sets http request **Close** option to true to close the connection after replying to the request.
```go
// Set connection parameter to close the connection immediately when we've got the response
req.Close = true
```

It makes the connection cannot be reused, and every request requires to establish a new TCP connection. Not only does the latency increase, but also increase the overhead of both gophercloud clients and openstack services.

Furthermore, there are several requests do not close the response body after sending the request. It also makes the connections cannot be reused even after the **Close** option is set to false.

jtopjian:
@jonas0616 Thanks for opening this issue. If you have any suggestions or ideas on how to make connection management more flexible, please let me know or feel free to open a PR for review/discussion.

jonas0616:
@jtopjian In most cases, it should not set `req.Close` to true. If anyone needs to disable the connection keepalive, the `request` could be modified by `Transport` in `http.Client`.

Secondly, make sure all response body is closed. Most of the `DELETE` and `HEAD` requests do not close the body. Such as `ServiceClient.Delete()`, it does not close the body when `opts.JSONResponse` is not set. However, many of callers do not set `opts.JSONResponse`. e.g., `containers.Delete()`, `objects.Delete()`, etc.

```go
func (client *ServiceClient) Delete(url string, opts *RequestOpts) (*http.Response, error) {
    if opts == nil {
        opts = new(RequestOpts)
    }
    client.initReqOpts(url, nil, nil, opts)
    return client.Request("DELETE", url, opts)
}
```

kayrus:
@jonas0616 can you test the latest master branch and close an issue if it is solved?

jonas0616:
@kayrus It could reuse TCP connections after upgrade to v0.11.0. Thanks for your contributing.

PR #1952, opened 2020-04-21
Core: reuse HTTP connection when possible

### LINKED ISSUE #1949: Core: close response body, when body is not used
While working on #1942 I noticed that in certain cases there can be a resource leak. E.g. in delete operations:

https://github.com/gophercloud/gophercloud/blob/b88f26317c347bc39d800e54663fd86c2a35d53e/openstack/networking/v2/ports/requests.go#L185

When `JSONResponse` parameter is nil, the response body is not closed, therefore the existing connection cannot be reused.

See https://golang.org/pkg/net/http/#Client.Do

> If the returned error is nil, the Response will contain a non-nil Body which the user is expected to close. If the Body is not both read to EOF and closed, the Client's underlying RoundTripper (typically Transport) may not be able to re-use a persistent TCP connection to the server for a subsequent "keep-alive" request.

@jtopjian is it something that should be considered or do I dramatize things?

Relates to #1015

kayrus:
Resolves #1949 #1838

### DIFF
--- a/openstack/imageservice/v2/imagedata/doc.go
+++ b/openstack/imageservice/v2/imagedata/doc.go
@@ -40,6 +40,9 @@ Example to Download Image Data
 		panic(err)
 	}
 
+	// close the reader, when reading has finished
+	defer image.Close()
+
 	imageData, err := ioutil.ReadAll(image)
 	if err != nil {
 		panic(err)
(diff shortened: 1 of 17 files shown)


kayrus:
@jtopjian ready for review

PR #1962, opened 2020-05-11
Glance V2: Add create action response headers support

### LINKED ISSUE #1961: Glance V2: Add create action response headers support
> In a cloud in which interoperable image import is enabled, the Image Create response will include a **OpenStack-image-import-methods** header listing the types of import methods available in that cloud. Alternatively, these methods may be determined independently of creating an image by making the Import Method Discovery call.

> In a cloud in which multiple storage backends are available, the Image Create response will include a **OpenStack-image-store-ids header** listing the stores available in that cloud. Alternatively, these stores may be determined independently of creating an image by making the Stores Discovery call.

https://docs.openstack.org/api-ref/image/v2/#interoperable-image-import

https://github.com/openstack/glance/blob/stable/rocky/glance/api/v2/images.py#L961

kayrus:
resolves #1961

### DIFF
--- a/openstack/imageservice/v2/images/results.go
+++ b/openstack/imageservice/v2/images/results.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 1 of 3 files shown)


jtopjian:
That Travis error is unrelated, but is flagging an error with `tip`. This could be a temporary issue or a sign that we're going to have to dive into the most complicated part of Gophercloud...

jtopjian:
Per https://github.com/gophercloud/gophercloud/blob/master/results.go#L95-L102, this is probably related: https://github.com/golang/go/commit/11b2853e6f322306a55519d03671e256b966e8ca

kayrus:
recheck

kayrus:
recheck

kayrus:
@jtopjian ready for review

kayrus:
recheck

PR #1967, opened 2020-05-15
Cinder V3: Marshal an empty metadata object in backup import

### LINKED ISSUE #1966: Cinder V3: Marshal an empty metadata object in backup import
When backup is being imported with a nil metadata, cinder import process fails because it expects to have the metadata as a dict:
https://github.com/openstack/cinder/blob/6b5d1f45e9f1833ddf8c994c4203dfef64a51591/cinder/db/sqlalchemy/api.py#L5407

kayrus:
resolves #1966


kayrus:
@jtopjian ready for review

PR #1910, opened 2020-03-28
Support health_status and health_status_reason attributes

openstacker:
For #1909 

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:
https://review.opendev.org/#/c/710384/

Issue #1909, opened 2020-03-28
ContainerInfra v1: Support health_status and health_status_reason for cluster


openstacker:
This will allow user can check the health status and the health status reason of the cluster.

* [health_status](https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/cluster.py#L142)
* [health_status_reason](https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/cluster.py#L145)

Reference: https://review.opendev.org/#/c/570818/

jtopjian:
Implemented in #1910 and #1968

PR #1968, opened 2020-05-18
ContainerInfra v1: Fix cluster Get

openstacker:
This is a regression issue introduced by[1], which may impact the
cluster Get. The health_status_reason attribute should be a dict
instead of a string.

[1] https://github.com/gophercloud/gophercloud/pull/1910

For #1909


openstacker:
@jtopjian  I'm sorry this is a regression issue. Please review it as it's blocked the very basic cluster Get function. BTW, when the version 0.12 will be released? Thanks.

jtopjian:
> BTW, when the version 0.12 will be released? 

Probably not until next month.

openstack/containerinfra/v1/clusters/results.go
```diff
@@ -69,37 +69,37 @@ func (r ResizeResult) Extract() (string, error) {
...
+	Status             string                 `json:"status"`
+	StatusReason       string                 `json:"status_reason"`
+	UUID               string                 `json:"uuid"`
+	UpdatedAt          time.Time              `json:"updated_at"`
+	UserID             string                 `json:"user_id"`
+	FloatingIPEnabled  bool                   `json:"floating_ip_enabled"`
+	FixedNetwork       string                 `json:"fixed_network"`
+	FixedSubnet        string                 `json:"fixed_subnet"`
+	HealthStatus       string                 `json:"health_status"`
+	HealthStatusReason map[string]interface{} `json:"health_status_reason"`
```
jtopjian, line 102: https://github.com/openstack/magnum/blob/master/magnum/api/controllers/v1/cluster.py#L145

:+1:

jtopjian:
@openstacker We've had a few other small bug fixes come in. If you can confirm this works, I can release version 0.11.1 for you - if that would help?

openstacker:
> @openstacker We've had a few other small bug fixes come in. If you can confirm this works, I can release version 0.11.1 for you - if that would help?

Thank you very much. I think I can just use the latest commit in CPO. That's OK. Thanks again.

PR #1973, opened 2020-06-01
switch bootfromvolume from os-volumes_boot to servers

jichenjc:
according to
https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/routes.py#L726
os-volumes_boot is identical to servers routes and it will be deprecated later

https://docs.openstack.org/api-ref/compute/
os-volumes_boot is not listed at all

so switch from os-volumes_boot to servers

For #1974

### DIFF
--- a/openstack/compute/v2/extensions/bootfromvolume/urls.go
+++ b/openstack/compute/v2/extensions/bootfromvolume/urls.go
@@ -3,5 +3,5 @@ package bootfromvolume
 import "github.com/gophercloud/gophercloud"
 
 func createURL(c *gophercloud.ServiceClient) string {
-	return c.ServiceURL("os-volumes_boot")
+	return c.ServiceURL("servers")
 }


jtopjian:
recheck stable/newton

jtopjian:
recheck stable/ocata

jtopjian:
recheck stable/pike

jtopjian:
@jichenjc Thanks for submitting this. I'd like to make sure this change doesn't break with older releases of openstack, but it appears there's an issue with running these tests. I'll have to open an issue about this.

jichenjc:
thanks a lot for the help, yes, I saw the test is not running well
let's see the result first about older openstack release, thanks

jtopjian:
recheck stable/queens

jtopjian:
recheck stable/rocky

jtopjian:
recheck stable/stein

jtopjian:
recheck stable/stein

wangxiyuan:
recheck

wangxiyuan:
recheck stable/stein

wangxiyuan:
recheck stable/rocky

wangxiyuan:
recheck stable/queens

jtopjian:
@wangxiyuan It looks like tests for "stein" are correctly running. The test failure seems legitimate and something I can look into later. Everything before stein seems to not even start, though.

wangxiyuan:
recheck stable/pike

wangxiyuan:
recheck stable/ocata

wangxiyuan:
recheck stable/newton

wangxiyuan:
@jtopjian devstack failed setup before rocky release. It needs more time to debug. I guess it related to somthing with python3 or ubuntu version adoption.

jtopjian:
@wangxiyuan 

> devstack failed setup before rocky release. It needs more time to debug. I guess it related to somthing with python3 or ubuntu version adoption.

Thank you so much for looking into this. Please take all the time you need :)

jichenjc:
@wangxiyuan thanks, yes, I think py2->3 and ubuntu upgrade might be a challange

as some older releases has been unsupported by community itself https://releases.openstack.org/
（from Ocata ) ,do we really need to run such test? @jtopjian

jtopjian:
It looks like all jobs are being run on Ubuntu Bionic? If so, is it possible to have certain older jobs run on Ubuntu Xenial?

As for older jobs, I try to test with as many older releases as possible as that's what reflects real-world deployments. That's key for changes such as what this PR is doing. But if we have to drop support for some releases, then we can drop Ocata and Pike.

wangxiyuan:
openlab has Ubuntu 16.04 node. It's called `ubuntu-xenial`. You should update the `nodeset` in  `.zuul.yaml` to use it.

https://github.com/gophercloud/gophercloud/blob/master/.zuul.yaml#L23

Just override it in the job you want.

jtopjian:
@wangxiyuan Thanks! I'm testing this in https://github.com/gophercloud/gophercloud/pull/1975

jtopjian:
@jichenjc I've merged #1975 which should help fix the issue with the immediate test failures. 

Can you rebase this PR off of master to pull in the new `.zuul.yaml`?

It's OK if the entire test suite fails. I just want to make sure that changing this URL isn't going to break support in older releases. From looking through the git history, this URL alias has been in place for a number of years, so we should be safe. I just want to double check.

Thank you for your patience with this.

jichenjc:
@jtopjian thanks for the info and the fix for older releases

I rebased just now and hopefully it helps

jtopjian:
recheck

jtopjian:
recheck stable/ocata

jichenjc:
thanks a lot for the help :)

PR #1975, opened 2020-06-04
Acc Tests: Use Ubuntu Xenial for older releases

jtopjian:
For #818

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -36,6 +36,7 @@
     parent: gophercloud-acceptance-test
     description: |
       Run gophercloud acceptance test on rocky branch
+    nodeset: ubuntu-xenial
     vars:
       global_env:
         OS_BRANCH: stable/rocky
@@ -45,6 +46,7 @@
     parent: gophercloud-acceptance-test
     description: |
       Run gophercloud acceptance test on queens branch
+    nodeset: ubuntu-xenial
     vars:
       global_env:
         OS_BRANCH: stable/queens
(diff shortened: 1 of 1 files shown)


jtopjian:
recheck stable/ocata

wangxiyuan:
It looks that the failure relates to the test itself.

jtopjian:
@wangxiyuan Yes, looks like the core issue was having everything run on BIonic. I can begin looking into the individual test failures.

jtopjian:
@elfosardo The way I've viewed this is that just because it's going to be in an unmaintained state doesn't mean that everyone currently running Ocata in production will immediately upgrade.

Personally, I wanted to keep testing as far back as Mitaka since Ubuntu officially supports it until April 2021, but there were some issues running it in OpenLab and so we decided to remove it.

The OpenStack APIs have been designed with backwards compatibility in mind (well, most of the time). I see no harm in continuing to support and test older releases as long as we're able to do so. 🙂

jtopjian:
I'm going to merge this for now. This will help us move forward with #1973.

I've created #1976 to track fixing test failures with older releases.

elfosardo:
@jtopjian if we're talking about LTS support, and strictly talking about packaged Openstack, then Ocata should not be supported, but Mitaka should be instead.
Ubuntu is not supporting Ocata anymore.

jtopjian:
Right, I know. I was only highlighting that an older release such as Mitaka is still supported, but we're unable to use it in OpenLab.

To me, the status of support doesn't matter. There are numerous deployments of unsupported releases of OpenStack in production. I don't want to test as far back as, say Liberty or Icehouse or Diablo or all those nostalgic ones -- there's certainly a limit. 🙂

Issue #1977, opened 2020-06-05
goroutines leak in Reauthenticate


pamburus:
Following code in provider_client.go does not work as described:
```go
	// Report result to all other interested goroutines.
	//
	// This happens in a separate goroutine because another goroutine might have
	// acquired a copy of `client.reauthmut.ongoing` before we cleared it, but not
	// have come around to sending its request. By answering in a goroutine, we
	// can have that goroutine linger until all responseChannels have been sent.
	// When GC has collected all sendings ends of the channel, our receiving end
	// will be closed and the goroutine will end.
	go func() {
		for responseChannel := range messages {
			responseChannel <- err
		}
	}()
```

Garbage collector does not collect such goroutines and they leak.

Here is a simple example which proves that:
https://play.golang.org/p/99jyHd9R_AX

majewsky:
> Here is a simple example which proves that

This example doesn't prove anything. The goroutine "leak" in this example happens because you keep spawning goroutines on a long-lived channel. This is not the case in the reauth code, there we have short-lived channels and only one goroutine dealing with the channel teardown.

I'm not saying there is no problem with this code. The goroutine leak is probably true, but the presented evidence does not make that clear. I want an actual example code first, then we can verify that #1978 really fixes the issue.

majewsky:
I whipped up a quick real-life testcase to check if there is really a goroutine leak:

```go
package main

import (
	"fmt"
	"runtime"
	"time"

	"github.com/gophercloud/utils/openstack/clientconfig"
)

func main() {
	pc, err := clientconfig.AuthenticatedClient(nil)
	must(err)
	_ = pc

	for iteration := 0; iteration < 20; iteration++ {
		must(pc.Reauthenticate(pc.Token()))
		fmt.Printf("after % 3d reauths: %d goroutines\n", iteration, runtime.NumGoroutine())

		//don't put too much load on Keystone
		time.Sleep(1 * time.Second)
	}
}

func must(err error) {
	if err != nil {
		panic(err.Error())
	}
}
```

Turns out there isn't:

```
$ go run main.go
after   0 reauths: 2 goroutines
after   1 reauths: 2 goroutines
after   2 reauths: 2 goroutines
after   3 reauths: 2 goroutines
after   4 reauths: 2 goroutines
after   5 reauths: 2 goroutines
after   6 reauths: 2 goroutines
after   7 reauths: 2 goroutines
after   8 reauths: 2 goroutines
[...]

PR #1979, opened 2020-06-09
defense in depth against goroutine leak

majewsky:
For #1977. As explained over there (and also in the code comment in this PR's diff), I do not see any evidence for an actual problem, but given the general gnarliness of this code, I think it makes sense to be a bit more explicit here (esp. to reduce the chance of a later refactoring accidentally introducing a goroutine leak).


majewsky:
@pamburus @kayrus FYI

pamburus:
@majewsky 
Did you see this issue?
 https://github.com/golang/go/issues/19702

pamburus:
This pull request does not fix leaks.

majewsky:
You're right (as discussed in #1977).

Issue #1981, opened 2020-06-22
Lack `device hints` support for `software raid`


Hellcatlk:
[PhysicalDisks](https://github.com/gophercloud/gophercloud/blob/1f9e326ce889660d92ec175620ba7eb8c8dd7009/openstack/baremetal/v1/nodes/requests.go#L571)'s type is `[]string` unsupport `device hints` for  software raid.

Here is the description from [docs.openstack.org](https://docs.openstack.org/ironic/latest/admin/raid.html#backing-physical-disks):
> physical_disks - A list of physical disks to use as read by the RAID interface.
> For software RAID physical_disks is a list of device hints in the same format as used for Specifying the disk for deployment (root device hints). The number of provided hints must match the expected number of backing devices (repeat the same hint if necessary).

PR #1982, opened 2020-06-22
Modify PhysicalDisks's type to support physical disks hints

Hellcatlk:
For #1981 

code in this PR:
https://github.com/openstack/ironic/blob/master/ironic/conductor/manager.py#L2999
https://github.com/openstack/ironic/blob/master/ironic/common/raid.py#L63
https://github.com/openstack/ironic/blob/master/ironic/drivers/raid_config_schema.json#L59

### DIFF
--- a/openstack/baremetal/v1/nodes/requests.go
+++ b/openstack/baremetal/v1/nodes/requests.go
@@ -568,7 +568,7 @@ type LogicalDisk struct {
 	Controller string `json:"controller,omitempty"`
 
 	// A list of physical disks to use as read by the RAID interface.
-	PhysicalDisks []string `json:"physical_disks,omitempty"`
+	PhysicalDisks []interface{} `json:"physical_disks,omitempty"`
 }
 
 func (opts RAIDConfigOpts) ToRAIDConfigMap() (map[string]interface{}, error) {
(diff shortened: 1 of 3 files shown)


jtopjian:
@Hellcatlk Thanks for submitting this.

Per our [contributor's guide](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md), we also need to see the API service code that proves this change is valid. Let me know if you have any questions or need any help.

Hellcatlk:
@jtopjian Thanks, here is the link to `ironic` source code:
https://github.com/openstack/ironic/blob/master/ironic/conductor/manager.py#L2999
https://github.com/openstack/ironic/blob/master/ironic/common/raid.py#L63
https://github.com/openstack/ironic/blob/master/ironic/drivers/raid_config_schema.json#L59

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/gophercloud/gophercloud
 
+go 1.14
```
jtopjian, line 3: whoops - actually, I just noticed that Travis is flagging an error and I think it's due to this. Can you remove this? 

We can probably get rid of the Go 1.11 tests altogether, but I will do that in a separate PR when I get some time. Until then, removing this will be the fastest way.

Hellcatlk: Sorry , I forget delete it. This is auto added by IDE.

openstack/baremetal/v1/nodes/requests.go
```diff
@@ -568,7 +568,7 @@ type LogicalDisk struct {
 	Controller string `json:"controller,omitempty"`
 
 	// A list of physical disks to use as read by the RAID interface.
-	PhysicalDisks []string `json:"physical_disks,omitempty"`
+	PhysicalDisks interface{} `json:"physical_disks,omitempty"`
```
Hellcatlk, line 571: What do you think about `interface{}` and `[]interface{}`?
`interface {}` is easy to use, but `[] interface {}` can make the input must be a list.

jtopjian: `[]interface{}` would be more strict and that's more appealing than a generic `interface{}`. Sure, let's go with that.

Hellcatlk:
@jtopjian Thanks, done.

Issue #1634, opened 2019-06-27
Int overflow on 32 bits arches


eclipseo:
Golang 1.12.6 on i686 and armv7:

```
Testing    in: /builddir/build/BUILD/gophercloud-e0311c0920d47879e99ac9826142bc097eacd5b7/_build/src
         PATH: /builddir/build/BUILD/gophercloud-e0311c0920d47879e99ac9826142bc097eacd5b7/_build/bin:/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/sbin
       GOPATH: /builddir/build/BUILD/gophercloud-e0311c0920d47879e99ac9826142bc097eacd5b7/_build:/usr/share/gocode
  GO111MODULE: off
      command: go test -buildmode pie -compiler gc -ldflags "-X github.com/gophercloud/gophercloud/version.commit=e0311c0920d47879e99ac9826142bc097eacd5b7 -X github.com/gophercloud/gophercloud/version=0 -extldflags '-Wl,-z,relro -Wl,--as-needed  -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld '"
      testing: github.com/gophercloud/gophercloud
github.com/gophercloud/gophercloud/acceptance/openstack/loadbalancer/v2
PASS
ok  	github.com/gophercloud/gophercloud/acceptance/openstack/loadbalancer/v2	0.022s
github.com/gophercloud/gophercloud/internal/testing
PASS
ok  	github.com/gophercloud/gophercloud/internal/testing	0.013s
github.com/gophercloud/gophercloud/openstack/baremetal/v1/allocations/testing
PASS
ok  	github.com/gophercloud/gophercloud/openstack/baremetal/v1/allocations/testing	0.034s
github.com/gophercloud/gophercloud/openstack/baremetal/v1/drivers/testing
PASS
ok  	github.com/gophercloud/gophercloud/openstack/baremetal/v1/drivers/testing	0.036s
github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes/testing
PASS
ok  	github.com/gophercloud/gophercloud/openstack/baremetal/v1/nodes/testing	0.063s
github.com/gophercloud/gophercloud/openstack/baremetal/v1/ports/testing
PASS
ok  	github.com/gophercloud/gophercloud/openstack/baremetal/v1/ports/testing	0.038s
github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection/testing
[...]

PR #1988, opened 2020-07-06
make introspection.RootDiskType int64 rather than int (#1634)

mwhudson:
Fixes tests (and functionality I suspect!) on 32-bit systems.

### DIFF
--- a/openstack/baremetalintrospection/v1/introspection/results.go
+++ b/openstack/baremetalintrospection/v1/introspection/results.go
@@ -238,7 +238,7 @@ type RootDiskType struct {
 	ByPath             string `json:"by_path"`
 	Rotational         bool   `json:"rotational"`
 	Serial             string `json:"serial"`
-	Size               int    `json:"size"`
+	Size               int64  `json:"size"`
 	Vendor             string `json:"vendor"`
 	Wwn                string `json:"wwn"`
 	WwnVendorExtension string `json:"wwn_vendor_extension"`


go.mod
```diff
@@ -11,3 +11,5 @@ require (
 	gopkg.in/check.v1 v1.0.0-20190902080502-41f04d3bba15 // indirect
 	gopkg.in/yaml.v2 v2.2.7
 )
+
+go 1.13
```
jtopjian, line 15: Can this be removed? It's causing the Travis failure. I'll clean this up in another PR.

jtopjian:
@mwhudson Thank you for looking into this - it's really appreciated.

From your comments in #1634:

> I don't know if it would be too much of an abi break, but 

Breaking changes are OK as long as they are justified.


> github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection.RootDiskType being an "int" vs an "int64" definitely seems like a mistake. 

Yes, from the error message given in #1634, I agree.  I'd also be inclined to change other `int`'s to `int64` while we're at it.

mwhudson:
> @mwhudson Thank you for looking into this - it's really appreciated.

No worries, thanks for the quick review!

> From your comments in #1634:
> 
> > I don't know if it would be too much of an abi break, but
> 
> Breaking changes are OK as long as they are justified.

Fair enough.

> > github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection.RootDiskType being an "int" vs an "int64" definitely seems like a mistake.
> 
> Yes, from the error message given in #1634, I agree. I'd also be inclined to change other `int`'s to `int64` in this package while we're at it.

All of them? I can do that easily enough, of course. I did have a quick scan and didn't think any of the others would be problematic but it might be best to be consistent.

jtopjian:
> All of them? I can do that easily enough, of course. I did have a quick scan and didn't think any of the others would be problematic but it might be best to be consistent.

Sorry - I may have been too broad in my last message 🙂 

I should have said that we should look at changing them where it makes sense, and only in this introspection package. For example, the Memory fields _might_ work better as an int64. I know it's a long shot to have that much memory, but who knows 2-3 years from now. Things like that.

go.mod
mwhudson: Oh yes, I meant to remove that before pushing, cleaned up now.

mwhudson:
Well OK. Want to merge this, or fix it yourself how ever you like? :)

PR #1990, opened 2020-07-16
Octavia: handle pure boolean values in "insert_headers" object

### LINKED ISSUE #1989: Unmarshal error for LB listener "insert_headers"
The following error occurs when load balancer listeners are retrieved:

```
json: cannot unmarshal bool into Go struct field Listener.listeners.insert_headers of type string
```

https://github.com/gophercloud/gophercloud/pull/1514 added support for _insert_headers_ and with the Go struct it tries to unmarshal the retrived listener's  _insert_headers_ to `map[string]string`. In our case (Open Telekom Cloud) the OpenStack API returns bool values for the headers instead of strings:

```
"insert_headers": {
    "X-Forwarded-ELB-IP": false,
    "X-Forwarded-Host": true
  },
...
```

/cc lingxiankong

kayrus:
Resolves #1989

### DIFF
--- a/openstack/loadbalancer/v2/listeners/results.go
+++ b/openstack/loadbalancer/v2/listeners/results.go
@@ -1,6 +1,9 @@
 package listeners
 
 import (
+	"encoding/json"
+	"fmt"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/l7policies"
 	"github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/pools"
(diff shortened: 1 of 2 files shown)


kayrus:
@jtopjian ready for review

jtopjian:
@kayrus Is this PR still valid? Looking through the previous discussion, I was concerned that this was an OpenTelekom-specific issue.

markbgrdt:
I think it is OTC specific as networking was heavily modified on OTC compared to vanilla OpenStack. I think we should not merge but close it.

jtopjian:
@markbgrdt sounds good - thank you!

Issue #1991, opened 2020-07-21
listeners lack support of TERMINATED_HTTPS protocol 


hamzazafar:
Octavia supports creating listeners of type TERMINATED_HTTPS.

Please see the *protocol* option in [requests section of creating listeners](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=create-listener-detail#id35)

TERMINATED_HTTPS protocol should be added to listeners in gophercloud!

jtopjian:
Implemented in #1992

PR #1992, opened 2020-07-21
support TERMINATED_HTTPS protocol for listeners

hamzazafar:
For [1991](https://github.com/gophercloud/gophercloud/issues/1991)

code in this PR:

- [ListenerPOST](https://github.com/openstack/octavia/blob/f14ccb52ed004f1c30c49e87407f2ce569c045e9/octavia/api/v2/types/listener.py#L116)
- [supported-protocols octavia common](https://github.com/openstack/octavia/blob/77786595e5803488f46e9b4faae02b5bccdfe88d/octavia/common/constants.py#L152)
- [supported-protocol octavia lib common](https://github.com/openstack/octavia-lib/blob/ec7195d713d71370fa6a72d3839253b9789b3b4e/octavia_lib/common/constants.py#L141)

### DIFF
--- a/openstack/loadbalancer/v2/listeners/requests.go
+++ b/openstack/loadbalancer/v2/listeners/requests.go
@@ -10,11 +10,12 @@ type Protocol string
 
 // Supported attributes for create/update operations.
 const (
-	ProtocolTCP   Protocol = "TCP"
-	ProtocolUDP   Protocol = "UDP"
-	ProtocolPROXY Protocol = "PROXY"
-	ProtocolHTTP  Protocol = "HTTP"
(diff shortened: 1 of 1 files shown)


kayrus:
@hamzazafar fyi: you san simply convert any string to a `Protocol` type and use it in struct. See an example in terraform provider: https://github.com/terraform-providers/terraform-provider-openstack/blob/152286f9ae3d016cb38d53ae2c0d1920d967d368/openstack/lb_v2_shared.go#L74

Probably it does make sense to add all supported protocols, or simply convert the type to a `string`.

Issue #1993, opened 2020-07-24
[object-storage] Creating Temporary URL with slash (/) returns invalid URL


Lirt:
When creating swift temporary URL using function `CreateTempURL` and `objectName` contains slash (`/`), generated link is invalid.

To reproduce issue create bucket with object containing slash and run simple go snippet below:

```bash
# Create bucket with object containing slash
swift post my-container
echo "this is my grandma" > grandma.png
swift upload my-container grandma.png --object-name "pictures/grandma.png"
```

```go
package main

import (
    "fmt"
    "net/http"

    "github.com/gophercloud/gophercloud"
    "github.com/gophercloud/gophercloud/openstack"
    "github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects"
)

func main() () {
    authOpts, _ := openstack.AuthOptionsFromEnv()
    pc, _ := openstack.AuthenticatedClient(authOpts)
    client, _ := openstack.NewObjectStorageV1(pc, gophercloud.EndpointOpts{
		Region: "default",
    })

    bucket := "my-container"
    key := "pictures/grandma.png"

    objUrl, _ := objects.CreateTempURL(client, bucket, key, objects.CreateTempURLOpts{
		Method: http.MethodGet,
		TTL:    3600,
    })
    fmt.Printf("%v\n", objUrl)
}
```

Then `curl` the URL and you get `401 Unauthorized: Temp URL invalid`.

The issue lies in encoding/escaping the URL before generating SHA1 hash (https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/objects/requests.go#L482).

Swift middleware documentation says:

> Do not URL-encode the path when you generate the HMAC-SHA1 signature. However, when you make the actual HTTP request, you should properly URL-encode the URL. [1]

[1] https://docs.openstack.org/swift/latest/api/temporary_url_middleware.html

kayrus:
@Lirt thanks for the bug report, especially to the TempURL API note. Can you confirm that temp url generation works fine in 0.10.0?

Lirt:
Yes, 0.10 works.

I can see that you refactored all urls to encoded in e54703519a9c3bcf1e3c25450a02614beb583c6d.

But I don't understand what is the reason to encode container name and object name before generating hash `url := getURL(c, url.QueryEscape(containerName), url.QueryEscape(objectName))`.

Can I submit PR for this? I think we need to only remove this query escape. Can it break some functionality?

[...]

PR #1994, opened 2020-07-24
Fix/create temp url escaping

Lirt:
For #1993 

Fix issue with generating temporary URLs, which are invalid when conatiner or object name contains characters, which are subject to query escaping.

### DIFF
--- a/acceptance/openstack/objectstorage/v1/objects_test.go
+++ b/acceptance/openstack/objectstorage/v1/objects_test.go
@@ -96,11 +96,11 @@ func TestObjects(t *testing.T) {
 
 		resp, err := http.Get(objURLs[i])
 		th.AssertNoErr(t, err)
-		defer resp.Body.Close()
 
 		body, err := ioutil.ReadAll(resp.Body)
 		th.AssertNoErr(t, err)
 		th.AssertEquals(t, oContents[i], body)
+		resp.Body.Close()
 	}
 
 	// Copy the contents of one object to another.
--- a/acceptance/openstack/objectstorage/v1/objects_test.go
+++ b/acceptance/openstack/objectstorage/v1/objects_test.go
@@ -4,6 +4,8 @@ package v1
 
 import (
 	"bytes"
+	"io/ioutil"
+	"net/http"
 	"strings"
 	"testing"
 
@@ -81,6 +83,25 @@ func TestObjects(t *testing.T) {
 		t.Fatalf("Unable to extract object info: %v", err)
 	}
 	th.AssertEquals(t, len(ois), len(oNames))
+
+	// Create temporary URL, download its contents and compare with what was originally created.
+	// Downloading the URL validates it (this cannot be done in unit tests).
+	objURLs := make([]string, numObjects)
+	for i := 0; i < numObjects; i++ {
+		objURLs[i], err = objects.CreateTempURL(client, cName, oNames[i], objects.CreateTempURLOpts{
+			Method: http.MethodGet,
+			TTL:    180,
+		})
+		th.AssertNoErr(t, err)
+
+		resp, err := http.Get(objURLs[i])
+		th.AssertNoErr(t, err)
+		defer resp.Body.Close()
+
+		body, err := ioutil.ReadAll(resp.Body)
+		th.AssertNoErr(t, err)
+		th.AssertEquals(t, oContents[i], body)
+	}
 
 	// Copy the contents of one object to another.
 	copyOpts := objects.CopyOpts{
--- a/acceptance/openstack/objectstorage/v1/objects_test.go
+++ b/acceptance/openstack/objectstorage/v1/objects_test.go
@@ -99,7 +99,7 @@ func TestObjects(t *testing.T) {
 
 		body, err := ioutil.ReadAll(resp.Body)
 		th.AssertNoErr(t, err)
-		th.AssertEquals(t, oContents[i].Bytes(), body)
+		th.AssertDeepEquals(t, oContents[i].Bytes(), body)
 		resp.Body.Close()
 	}
 
--- a/acceptance/openstack/objectstorage/v1/objects_test.go
+++ b/acceptance/openstack/objectstorage/v1/objects_test.go
@@ -99,7 +99,7 @@ func TestObjects(t *testing.T) {
 
 		body, err := ioutil.ReadAll(resp.Body)
 		th.AssertNoErr(t, err)
-		th.AssertEquals(t, oContents[i], body)
+		th.AssertEquals(t, oContents[i].Bytes(), body)
 		resp.Body.Close()
 	}
 
--- a/openstack/objectstorage/v1/objects/requests.go
+++ b/openstack/objectstorage/v1/objects/requests.go
@@ -40,6 +40,10 @@ type ListOpts struct {
 	Path      string `q:"path"`
 }
 
+func init() {
+	ResetClockImplementation()
+}
+
 // ToObjectListParams formats a ListOpts into a query string and boolean
 // representing whether to list complete information for each object.
 func (opts ListOpts) ToObjectListParams() (bool, string, error) {
@@ -464,6 +468,14 @@ var NowFunc nowFuncT
 
 func timeNow() time.Time {
 	return NowFunc().UTC()
+}
+
+// ResetClockImplementation resets implementation of Time.
+// Used only for testing.
+func ResetClockImplementation() {
+	NowFunc = func() time.Time {
+		return time.Now()
+	}
 }
 
 // CreateTempURL is a function for creating a temporary URL for an object. It
--- a/openstack/objectstorage/v1/objects/requests.go
+++ b/openstack/objectstorage/v1/objects/requests.go
@@ -456,6 +456,16 @@ type CreateTempURLOpts struct {
 	Split string
 }
 
+type nowFuncT func() time.Time
+
+// NowFunc allows to modify function for getting current time.
+// This offers developers flexibility to stub time in tests.
+var NowFunc nowFuncT
+
+func timeNow() time.Time {
+	return NowFunc().UTC()
+}
+
 // CreateTempURL is a function for creating a temporary URL for an object. It
 // allows users to have "GET" or "POST" access to a particular tenant's object
 // for a limited amount of time.
@@ -464,7 +474,7 @@ func CreateTempURL(c *gophercloud.ServiceClient, containerName, objectName strin
 		opts.Split = "/v1/"
 	}
 	duration := time.Duration(opts.TTL) * time.Second
-	expiry := time.Now().Add(duration).Unix()
+	expiry := timeNow().Add(duration).Unix()
 	getHeader, err := containers.Get(c, url.QueryEscape(containerName), nil).Extract()
 	if err != nil {
 		return "", err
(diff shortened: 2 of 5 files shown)


Lirt:
For tests to work, time.Now() in function `CreateTempURL` must always return consistent value https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/objects/requests.go#L467.

openstack/objectstorage/v1/objects/testing/requests_test.go
```diff
@@ -323,3 +326,27 @@ func TestObjectCreateParamsWithSeek(t *testing.T) {
...
+	th.SetupHTTP()
+	defer th.TeardownHTTP()
+
+	// Handle fetching of secret key inside of CreateTempURL
+	containerTesting.HandleGetContainerSuccessfully(t)
+	accountTesting.HandleGetAccountSuccessfully(t)
+	client := fake.ServiceClient()
+
+	// Get dynamically generated port number so that it can be used to compare final URL
+	port := strings.Split(client.Endpoint, ":")[2]
```
Lirt, line 340: If we can set port manually to some consistent number across tests, this could be removed

kayrus: You can find an example here: https://github.com/gophercloud/gophercloud/blob/bd999d0da882fe8c5b0077b7af2dcc019c1ab458/openstack/identity/v3/extensions/oauth1/testing/requests_test.go#L248

kayrus:
> For tests to work, time.Now() in function CreateTempURL must always return consistent value https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/objects/requests.go#L467.

I'd split this func and use the major logic func in unit tests.

I wonder how terraform functional tests passed this, and noticed that object name doesn't contain the slash: https://github.com/terraform-providers/terraform-provider-openstack/blob/152286f9ae3d016cb38d53ae2c0d1920d967d368/openstack/resource_openstack_objectstorage_tempurl_v1_test.go#L14
In addition it would be great to have functional tests in gophercloud as well.

openstack/objectstorage/v1/objects/testing/requests_test.go
Lirt: Great, thank you. Fixed

Lirt:
> > For tests to work, time.Now() in function CreateTempURL must always return consistent value https://github.com/gophercloud/gophercloud/blob/master/openstack/objectstorage/v1/objects/requests.go#L467.
> 
> I'd split this func and use the major logic func in unit tests.

I'm sorry, but I don't see way how to split it without adding new argument for `time` to be able to stub it.

What I came up after some research is allowing to change function of time at package level (https://labs.yulrizka.com/en/stubbing-time-dot-now-in-golang/) - https://github.com/gophercloud/gophercloud/pull/1994/commits/f4b9e4bbabaa70b57cbdb5a2f89bea34c0505088

openstack/objectstorage/v1/objects/testing/requests_test.go
```diff
@@ -323,3 +326,33 @@ func TestObjectCreateParamsWithSeek(t *testing.T) {
 	_, ok = headers["ETag"]
 	th.AssertEquals(t, true, ok)
 }
+
+func TestCreateTempURL(t *testing.T) {
+	port := 33199
```
kayrus, line 331: if there is a case, when oauth1 unit tests run in parallel with swift, this may lead to a conflict.

kayrus:
recheck

openstack/objectstorage/v1/objects/requests.go
```diff
@@ -464,7 +474,7 @@ func CreateTempURL(c *gophercloud.ServiceClient, containerName, objectName strin
 		opts.Split = "/v1/"
 	}
 	duration := time.Duration(opts.TTL) * time.Second
-	expiry := time.Now().Add(duration).Unix()
+	expiry := timeNow().Add(duration).Unix()
```
kayrus, line 477: I wonder how did it work before without `UTC()`

Lirt: Good question, I didn't think about htat. I was reading docs, and it says it should be consistent in any way :arrow_down:.

> `Unix()` The result does not depend on the location associated with t. Unix-like operating systems
https://golang.org/pkg/time/#Time.Unix

So I can actually revert back to `Now()` if it looks better for you.

Lirt: https://play.golang.org/p/so0mPFXWY7m

openstack/objectstorage/v1/objects/testing/requests_test.go
Lirt: Changed to `33200`

openstack/objectstorage/v1/objects/requests.go
kayrus: Not sure what should be used. @jtopjian please advice.

openstack/objectstorage/v1/objects/testing/requests_test.go
```diff
@@ -323,3 +326,33 @@ func TestObjectCreateParamsWithSeek(t *testing.T) {
...
+	// Append v1/ to client endpoint URL to be compliant with tempURL generator
+	client.Endpoint = client.Endpoint + "v1/"
+	tempURL, err := objects.CreateTempURL(client, "testContainer", "testObject/testFile.txt", objects.CreateTempURLOpts{
+		Method: http.MethodGet,
+		TTL:    60,
+	})
+
+	sig := "89be454a9c7e2e9f3f50a8441815e0b5801cba5b"
+	expiry := "1593565980"
+	expectedURL := fmt.Sprintf("http://127.0.0.1:%v/v1/testContainer/testObject/testFile.txt?temp_url_sig=%v&temp_url_expires=%v", port, sig, expiry)
```
kayrus, line 354:
```suggestion
	expectedURL := fmt.Sprintf("http://127.0.0.1:%d/v1/testContainer/testObject/testFile.txt?temp_url_sig=%v&temp_url_expires=%v", port, sig, expiry)
```

Lirt:
Had to fix some issues with Time packaging, but it works now correctly. Question is if this way is OK for you maintainers. As I said I didn't find better solution yet, but I am open to suggestions.

I will also add 1 acceptance test, to cover this function better, but I think it will take some time, since it will be my first.

acceptance/openstack/objectstorage/v1/objects_test.go
```diff
@@ -82,6 +84,25 @@ func TestObjects(t *testing.T) {
...
+	for i := 0; i < numObjects; i++ {
+		objURLs[i], err = objects.CreateTempURL(client, cName, oNames[i], objects.CreateTempURLOpts{
+			Method: http.MethodGet,
+			TTL:    180,
+		})
+		th.AssertNoErr(t, err)
+
+		resp, err := http.Get(objURLs[i])
+		th.AssertNoErr(t, err)
+		defer resp.Body.Close()
```
kayrus, line 99: not critical, but better to call `resp.Body.Close()` at the end of the loop. otherwise go will close the connection only when the test function exits.

Lirt: Fixed in 22628a7ff541e2ff3bfab63a75f1d9ef6fd71fb6

openstack/objectstorage/v1/objects/requests.go
```diff
@@ -456,6 +460,24 @@ type CreateTempURLOpts struct {
 	Split string
 }
 
+type nowFuncT func() time.Time
```
kayrus, line 463: What if add the date into the `CreateTempURLOpts` struct? If it is nil, then a default `time.Now()` is used. If not, the provided value is used. This will avoid hacking around the `ResetClockImplementation`

kayrus: See here for example: https://github.com/gophercloud/gophercloud/blob/v0.12.0/openstack/identity/v3/extensions/ec2tokens/requests.go#L74
https://github.com/gophercloud/gophercloud/blob/v0.12.0/openstack/identity/v3/extensions/ec2tokens/requests.go#L268

Lirt: Sounds good to me

Lirt: Please check 9a500de9eaad35ab3e640fe323ccfe47bf88532d

Notice I didn't use pointer to time.Time because as documentation says:
> Programs using times should typically store and pass them as values, not pointers. That is, time variables and struct fields should be of type time.Time, not *time.Time. 

But I used `isZero()` to check for null value.

https://golang.org/pkg/time/#Time

Lirt: I looks much cleaner :-)

openstack/objectstorage/v1/objects/requests.go
Lirt: Can we resolve as discussed below? 
It was commited here https://github.com/gophercloud/gophercloud/commit/9a500de9eaad35ab3e640fe323ccfe47bf88532d

Lirt:
Acceptance test for the function was successful, but some other failed, doing recheck.

Lirt:
recheck

Lirt:
@kayrus Can you please look at the pipeline? I'm not sure if I should just run recheck, or there is something wrong. I think it is failing on something different than my code.

```bash
2020-08-03 10:06:35.176116 | ubuntu-bionic | --- SKIP: TestMessageDelete (0.19s)
2020-08-03 10:06:35.176174 | ubuntu-bionic |     messages_test.go:91: No messages were found
2020-08-03 10:06:35.176219 | ubuntu-bionic | PASS
2020-08-03 10:06:35.177855 | ubuntu-bionic | ok  	github.com/gophercloud/gophercloud/acceptance/openstack/sharedfilesystems/v2/messages	0.531s
2020-08-03 10:06:35.201255 | ubuntu-bionic | + ./script/acceptancetest:main:101         :   [[ 0 != 0 ]]
2020-08-03 10:06:35.204759 | ubuntu-bionic | + ./script/acceptancetest:main:107         :   [[ -n 1 ]]
2020-08-03 10:06:35.206948 | ubuntu-bionic | + ./script/acceptancetest:main:108         :   exit 1
2020-08-03 10:06:39.396148 | ubuntu-bionic | ERROR
2020-08-03 10:06:39.397095 | ubuntu-bionic | {
2020-08-03 10:06:39.397221 | ubuntu-bionic |   "delta": "0:45:12.849894",
2020-08-03 10:06:39.397348 | ubuntu-bionic |   "end": "2020-08-03 10:06:38.006579",
2020-08-03 10:06:39.397430 | ubuntu-bionic |   "msg": "non-zero return code",
2020-08-03 10:06:39.397508 | ubuntu-bionic |   "rc": 1,
2020-08-03 10:06:39.397628 | ubuntu-bionic |   "start": "2020-08-03 09:21:25.156685"
2020-08-03 10:06:39.397715 | ubuntu-bionic | }
```

Lirt:
recheck

jtopjian:
The acceptance test failure should be fixed when #2000 is merged.

Lirt:
@jtopjian Thank you.

I'm pretty much done with this MR. You can do review when you have time for it.

jtopjian:
@Lirt Thanks! This looks good to me.

@kayrus are you OK with this?

kayrus:
@jtopjian LGTM as well. @Lirt, thanks a lot for your contribution.

Lirt:
Thank you for review and help with code @jtopjian and @kayrus.

Issue #1997, opened 2020-08-03
[blockstorage]UnmarshalJSON error with RFC3339 timestamps in volumes.


zh-f:
When I tried to get volume details in our OpenStack environment, which version is queens 17.0.3,  using gophercloud blockstorage v3 like this :

```go
        volume, err := volumes.Get(client, "45fdb4fc-ad85-4512-85e6-980451edc57e").Extract()
        if err != nil {
                log.Errorf("err is %v", err)
        }
        log.Infof("Volume is %v", volume)
```
The volume detail is :
```
 cinder show el-volume
+------------------------------+-------------------------------------------------+
| Property                     | Value                                           |
+------------------------------+-------------------------------------------------+
| attached_servers             | []                                              |
| attachment_ids               | []                                              |
| availability_zone            | nova                                            |
| bootable                     | true                                            |
| consistencygroup_id          | None                                            |
| created_at                   | 2020-07-31T07:18:53Z                            |
| description                  | el-volume                                       |
| encrypted                    | False                                           |
| id                           | 45fdb4fc-ad85-4512-85e6-980451edc57e            |
| metadata                     |                                                 |
| multiattach                  | False                                           |
| name                         | el-volume                                       |
| os-vol-tenant-attr:tenant_id | bd6f9720483b4e259e840ead64a5b5d1                |
| replication_status           | None                                            |
| size                         | 20                                              |
| snapshot_id                  | None                                            |
| source_volid                 | None                                            |
| status                       | available                                       |
| updated_at                   | 2020-07-31T07:18:56Z                            |
| user_id                      | bfa8089a9eb542a7b43288979f697883                |
| volume_image_metadata        | checksum : 441af00e2dd98f1b2795b4b9a1ef84f0     |
|                              | container_format : bare                         |
|                              | ctcm_enabled : true                             |
|                              | disk_format : raw                               |
|                              | hw_qemu_guest_agent : yes                       |
|                              | hw_vif_multiqueue_enabled : true                |
|                              | image_id : f9de62d7-1896-4561-8571-b8e83b52696b |
|                              | image_name : newCentOS7.6                       |
|                              | image_version : v1.1                            |
|                              | min_disk : 0                                    |
|                              | min_ram : 0                                     |
|                              | os_distro : centos                              |
|                              | os_type : linux                                 |
|                              | os_version : 7.6                                |
|                              | size : 21474836480                              |
| volume_type                  | default_volume_type                             |
[...]

PR #1998, opened 2020-08-03
Support RFC3339 when unmarshal volume CreatedAt and UpdatedAt

zh-f:
Fix this error by trying to unmarshal volume CreatedAt and UpdatedAt via time.Time, to support RFC3339 timestamp.

I've tested in our OpenStack environment, it seems after tryting to unmarshal voluem.CreatedAt and volume.UpdatedAt using this fix, I can get volume details via Extract() correctly :

```
Volume is &{45fdb4fc-ad85-4512-85e6-980451edc57e available 20 nova 2020-07-31 07:18:53 +0000 UTC 2020-07-31 07:18:56 +0000 UTC [] el-volume el-volume default_volume_type   <nil> map[] bfa8089a9eb542a7b43288979f697883 true false   false map[checksum:441af00e2dd98f1b2795b4b9a1ef84f0 container_format:bare ctcm_enabled:true disk_format:raw hw_qemu_guest_agent:yes hw_vif_multiqueue_enabled:true image_id:f9de62d7-1896-4561-8571-b8e83b52696b image_name:newCentOS7.6 image_version:v1.1 min_disk:0 min_ram:0 os_distro:centos os_type:linux os_version:7.6 size:21474836480]} 
```

Looks fine.

For  #1997
Signed-off-by: Fan Zhang <zh.f@outlook.com>

### DIFF
--- a/openstack/blockstorage/v2/volumes/results.go
+++ b/openstack/blockstorage/v2/volumes/results.go
@@ -95,8 +95,8 @@ func (r *Volume) UnmarshalJSON(b []byte) error {
 
 	var s1 struct{
 		tmp
-		CreatedAt time.Time `json: "created_at"`
-		UpdatedAt time.Time `json: "updated_at"`
+		CreatedAt time.Time `json:"created_at"`
+		UpdatedAt time.Time `json:"updated_at"`
 	}
 	err = json.Unmarshal(b, &s1)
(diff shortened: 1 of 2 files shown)


zh-f:
Close this PR as it's not a bug in gophercloud.

Issue #2011, opened 2020-08-27
add support for cells v1 DifferentCells filter


cardoe:
The filter was added in https://github.com/openstack/nova/commit/596ee8796a028596e5ea1ad0529aa3f14b6f5934 to the OpenStack codebase. gophercloud supports it's corresponding filter TargetCell but doesn't support DifferentCell. While it's a cells v1 feature it's useful to folks that still have cells v1 deployments.

jtopjian:
Implemented in #2012

PR #2012, opened 2020-08-27
schedulerhints: support DifferentCell filter

cardoe:
The TargetCell filter from cells v1 is supported in the code base but
support for the DifferentCell filter was never added. DifferentCell
filter is part of cells v1 but for anyone still using cells v1 this is
useful.

For #2011

code in this PR:

https://github.com/openstack/nova/commit/596ee8796a028596e5ea1ad0529aa3f14b6f5934

### DIFF
--- a/openstack/compute/v2/extensions/schedulerhints/requests.go
+++ b/openstack/compute/v2/extensions/schedulerhints/requests.go
@@ -31,6 +31,9 @@ type SchedulerHints struct {
 	// TargetCell specifies a cell name where the instance will be placed.
 	TargetCell string `json:"target_cell,omitempty"`
 
+	// DifferentCell specifies cells names where an instance should not be placed.
+	DifferentCell []string `json:"different_cell,omitempty"`
+
 	// BuildNearHostIP specifies a subnet of compute nodes to host the instance.
 	BuildNearHostIP string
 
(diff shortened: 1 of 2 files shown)


cardoe:
The tests fail because of the identity/v3 module uses the internal module which isn't allowed. My change doesn't involve identity at all so the failures seemed out of scope to me.

jtopjian:
@cardoe Thank you for submitting this.

> The tests fail because of the identity/v3 module uses the internal module which isn't allowed.

I'm not sure what you mean. Can you elaborate?

I believe the unit tests are failing because of a syntax error: https://travis-ci.org/github/gophercloud/gophercloud/jobs/721697827#L286

In addition, it looks like `go fmt` will need to be run on these changed files. You can see how the indentation isn't matching up in the diff.

cardoe:
Ah I was looking at https://travis-ci.com/github/cardoe/gophercloud/jobs/378393317

jtopjian:
How strange - especially because the latest run didn't trigger that error.

It's looking like more `go fmt` errors. Did you run the actual `go fmt` tool or just change the indentation? If the former, then this is some wonky behavior with `go fmt` between releases and I'll add it to an ignore list we have.

cardoe:
@jtopjian I've just made the changes to make it happy. A real rough test with this in the terraform-provider-openstack is all I've done. I'm ultimately looking to make a PR there to make this work.

cardoe:
https://github.com/cardoe/terraform-provider-openstack/tree/different-cell-scheduler ultimately this is what I'd like to land.

Issue #2015, opened 2020-09-11
Support Messaging v2 _enable_encrypt_messages field


galuszkak:
Add support for `_enable_encrypt_messages` in `messaging/v2/queues`

jtopjian:
Implemented in #2016

PR #2016, opened 2020-09-11
feat: update Messaging v2 API to support _enable_encrypt_messages field

galuszkak:

For #2015 

code in this PR:
https://docs.openstack.org/api-ref/message/?expanded=show-queue-details-detail

### DIFF
--- a/openstack/messaging/v2/queues/requests.go
+++ b/openstack/messaging/v2/queues/requests.go
@@ -85,6 +85,9 @@ type CreateOpts struct {
 	// for any messages posted to the queue.
 	MaxMessagesPostSize int `json:"_max_messages_post_size,omitempty"`
 
+	// Does messages should be encrypted
+	EnableEncryptMessages *bool `json:"_enable_encrypt_messages,omitempty"`
+
 	// Extra is free-form extra key/value pairs to describe the queue.
 	Extra map[string]interface{} `json:"-"`
 }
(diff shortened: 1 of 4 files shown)


jtopjian:
@galuszkak Thank you for submitting this.

Per the contributor tutorial linked above, API doc references aren't enough to determine if this is a valid OpenStack feature. Please see [Step 3](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md) specifically and let me know if you have any questions.

galuszkak:
Hi @jtopjian ,

If I understand your ask correctly does below proofs are enough, or there is something still needed?
https://github.com/openstack/zaqar/blob/master/zaqar/transport/validation.py#L344-L347
https://github.com/openstack/zaqar/commit/e12c65a369825ea5469bdb31f5c7151268d7926b

Thanks for helping!

jtopjian:
@galuszkak Yes, that's perfect 🙂 

One extra thing to consider here is that you have a combination of `bool` and `omitempty`. This means that a value of false will _never_ be sent in the JSON request. According to the service-side code (https://github.com/openstack/zaqar/blob/master/zaqar/transport/validation.py#L344), the default value is false, so this combination should be fine. However, if you want the ability to explicitly send a value of false, this will need changed to `*bool`.

There's also some `ListOpts` fields you've added in this PR. We'll need to dig up the code references for those, too. Alternatively, you can leave them out of this PR and submit them in a new PR.

galuszkak:
@jtopjian thanks for feedback.

I've applied it. I will sent `name` and `with_count` in separate PR  as it makes more sense to seperate those changes.

jtopjian (approved):
LGTM - thank you!

Issue #2029, opened 2020-09-25
Missing 'Enabled' field in identity Endpoint struct


reenjii:
identity/v3 endpoints.Endpoint struct is lacking Enabled field.
https://github.com/gophercloud/gophercloud/blob/master/openstack/identity/v3/endpoints/results.go#L41

It is returned by OpenStack API.
https://docs.openstack.org/api-ref/identity/v3/?expanded=list-endpoints-detail#service-catalog-and-endpoints

> enabled | body | boolean | Indicates whether the endpoint appears in the service catalog: - false. The endpoint does not appear in the service catalog. - true. The endpoint appears in the service catalog.

jtopjian:
@reenjii Thank you for reporting this. Please note that we'll also need to see the area in the Keystone service code where the `Enabled` field is defined. More information is [here](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md) in our Contributor Tutorial. Let me know if you have any questions.

reenjii:
Hi @jtopjian 

Sure here are the related keystone code parts.

Keystone API get/list endpoints gets endpoints from catalog_api

https://github.com/openstack/keystone/blob/master/keystone/api/endpoints.py#L75
https://github.com/openstack/keystone/blob/master/keystone/api/endpoints.py#L82

Comes from SQL backend

https://github.com/openstack/keystone/blob/master/keystone/catalog/backends/sql.py#L240
https://github.com/openstack/keystone/blob/master/keystone/catalog/backends/sql.py#L245

Which returns an Endpoint object

https://github.com/openstack/keystone/blob/master/keystone/catalog/backends/sql.py#L63

Which has 'enabled' attribute filled here

https://github.com/openstack/keystone/blob/master/keystone/catalog/backends/sql.py#L79

- boolean
- NOT NULL
- default: true

jtopjian:
Implemented in #2030

PR #2030, opened 2020-09-25
Identity: add 'enabled' field on Endpoint

reenjii:
For #2029


jtopjian (approved):
@reenjii I'm assuming this PR was ready to be reviewed/merged? If so, it looks good to me.

Issue #2031, opened 2020-09-28
Support Magnum cluster upgrade API


jkmw:
Currently there is no support for the cluster `upgrade`-action.

API doc:
- https://docs.openstack.org/api-ref/container-infrastructure-management/?expanded=upgrade-a-cluster-detail#upgrade-a-cluster

source from cluster-actions controller:
- https://github.com/openstack/magnum/blob/e8467e94d88b6feb67ff7c167f5e167cd452c772/magnum/api/controllers/v1/cluster_actions.py#L58
- https://github.com/openstack/magnum/blob/e8467e94d88b6feb67ff7c167f5e167cd452c772/magnum/api/controllers/v1/cluster_actions.py#L132

jtopjian:
@jkmw Thank you for opening this. Please note that API documentation alone is not enough to support a feature being added. More information is [here](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md) in our Contributor Tutorial. Let me know if you have any questions.

jkmw:
@jtopjian well, I have linked the code location now, so that it's clear that my change is supported.

I hope this is enough?

jtopjian:
Thanks. 

In your PR, you are defining a new struct with a set of fields. Do you know where these fields are defined? In other words, how do we know that `max_batch_size` is a valid field to be sent in a JSON request?

jkmw:
I have now also linked the class, which contains all parameters that are used for the upgrade call. These are in the same file as the reference to calling the controller.

I assumed that the rest of the context is self-evident from the file :shrug:

jtopjian:
Implemented in #2032

PR #2032, opened 2020-09-28
Support Magnum upgrade API

jkmw:
Currently there is no support for the cluster upgrade-action. This PR implements the missing api-call.

For #2031

API doc:
- https://docs.openstack.org/api-ref/container-infrastructure-management/?expanded=upgrade-a-cluster-detail#upgrade-a-cluster

source from cluster-actions controller:
- https://github.com/openstack/magnum/blob/e8467e94d88b6feb67ff7c167f5e167cd452c772/magnum/api/controllers/v1/cluster_actions.py#L58
- https://github.com/openstack/magnum/blob/e8467e94d88b6feb67ff7c167f5e167cd452c772/magnum/api/controllers/v1/cluster_actions.py#L132

### DIFF
--- a/openstack/containerinfra/v1/clusters/doc.go
+++ b/openstack/containerinfra/v1/clusters/doc.go
@@ -89,6 +89,17 @@ Example to Update a Cluster
 	}
 	fmt.Printf("%s\n", clusterUUID)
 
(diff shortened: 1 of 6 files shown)


jtopjian (approved):
LGTM

PR #2033, opened 2020-09-30
openstack/networking: fix dropped test error

alrs:
For #2034 

This fixes a dropped test error in `openstack/networking/v2/extensions/layer3/routers/testing`.


jtopjian (approved):
LGTM - thank you!

Issue #2035, opened 2020-10-01
Add ip, ip6, user_id fields to servers.ListOpts


sfiera:
I’d like to submit a PR to add three fields to servers.ListOpts: `ip`, `ip6`, and `user_id`.

Documentation: https://docs.openstack.org/api-guide/compute/server_concepts.html#server-query
Source: https://github.com/openstack/nova/blob/a948a803b561606a892133940915caae610c080b/nova/api/openstack/compute/schemas/servers.py#L567

jtopjian:
@sfiera Sure, sounds good to me.

jtopjian:
Implemented in #2038

PR #2038, opened 2020-10-05
Add ip, ip6, and user_id to servers.ListOpts

sfiera:
Add ip, ip6, and user_id to servers.ListOpts

As listed in https://docs.openstack.org/api-guide/compute/server_concepts.html#server-query:
* ip
* ip6 (New in version 2.5)
* user_id (New in version 2.83)

And the CLI documentation:

    --ip <ip-address-regex>
                          Regular expression to match IP addresses
    --ip6 <ip-address-regex>
                          Regular expression to match IPv6 addresses. Note that
                          this option only applies for non-admin users when
                          using ``--os-compute-api-version`` 2.5 or greater.
    --user <user>         Search by user (admin only) (name or ID)

Note the difference between the user_id docs. My understanding is that
it’s the same as ip6: the option was initially admin-only, and became
generally available with API 2.83 (this is what I have documented).

For #2035

code in this PR:

https://github.com/openstack/nova/blob/a948a803b561606a892133940915caae610c080b/nova/api/openstack/compute/schemas/servers.py#L567

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -29,6 +29,14 @@ type ListOpts struct {
 	// Flavor is the name of the flavor in URL format.
 	Flavor string `q:"flavor"`
 
+	// IP is a regular expression to match the IPv4 address of the server.
+	IP string `q:"ip"`
+
+	// This requires the client to be set to microversion 2.5 or later, unless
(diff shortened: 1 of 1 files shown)


jtopjian (approved):
LGTM - thank you!

PR #2043, opened 2020-10-26
Networking / routers: Allow updating routers, without specifying the static routes

stefreak:
This is helpful when it is necessary to rename a router,
but it is not desired by the user to touch the static routes.

The OpenStack API just allows to omit the static routes (e.g.
verifyable with `openstack router set --name new_name --debug`).

This change makes sure that static routes parameter can be omitted with
gophercloud as well.

Also added a unit test.

For #2044

code in this PR:

(I could not find a specific line of code, but provided the openstack CLI command for quick verification that this works above)

### DIFF
--- a/openstack/networking/v2/extensions/layer3/routers/doc.go
+++ b/openstack/networking/v2/extensions/layer3/routers/doc.go
@@ -48,7 +48,20 @@ Example to Update a Router
 
 	updateOpts := routers.UpdateOpts{
 		Name:   "new_name",
-		Routes: routes,
+		Routes: &routes,
+	}
+
+	router, err := routers.Update(networkClient, routerID, updateOpts).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+Example to Update just the Router name, keeping everything else as-is
+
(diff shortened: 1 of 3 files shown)


jtopjian:
recheck

stefreak:
I also noticed that the documentation needed an update. 
Also, I simplified the unit test.

@jtopjian I am not sure how I can see what fails in the acceptance test – or is it just flaky? Thanks for sending the retest command ;)

jtopjian:
@stefreak The full suite usually takes 2-3 hours, so 1.5hrs may be pointing to an issue with OpenLab. It happens from time to time. Let's give it a day and if it's still failing, I'll open a ticket.

If I get some time, I'll run the tests in devstack to confirm if any of them need updating from this change.

stefreak:
recheck

jtopjian:
From the logs, it looks like the devstack installation is not completing successfully. I'll give it one more try and then open a ticket.

jtopjian:
recheck

jtopjian (approved):
LGTM - thank you!

Issue #2051, opened 2020-11-06
Add some missing Baremetal ProvisionState states


andfasano:
It looks like that not all the Ironic states are enumerated by the ProvisionState constants defined in https://github.com/gophercloud/gophercloud/blob/7495a1acc79a29a25751e8ad89c4944daa92eafe/openstack/baremetal/v1/nodes/requests.go#L18.

I'd like to open a new PR to add the following states:

* _rescue wait_
* _unrescuing_

Reference: https://docs.openstack.org/ironic/latest/contributor/states.html

jtopjian:
@andfasano Sure - thank you for looking into this :slightly_smiling_face:

jtopjian:
Implemented in #2052

PR #2052, opened 2020-11-09
Add RescueWait and Unrescuing ProvisionState states

andfasano:
For #2051 

code in this PR:

https://github.com/openstack/ironic/blob/f118307e2bc94f9d524e21b44b9b64357dd7a417/ironic/common/states.py#L203
https://github.com/openstack/ironic/blob/f118307e2bc94f9d524e21b44b9b64357dd7a417/ironic/common/states.py#L216


jtopjian (approved):
LGTM - thank you!

Issue #2055, opened 2020-11-13
Octavia pool member types are missing Backup, MonitorAddress, MonitorPort and Tags


mtneug:
#1824 added support for Backup, MonitorAddress and MonitorPort when creating members. It is still missing for other operations. Tags is an additional field currently not supported by Gophercloud.

jtopjian:
Implemented in #2056

PR #2056, opened 2020-11-13
Add missing options to member types

mtneug:
For #2055

This adds Backup, MonitorAddress, MonitorPort and Tags to member types.

**`CreateMemberOpts`**

[Documentation](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=create-member-detail#create-member)
[Code](https://github.com/openstack/octavia/blob/198980639cbe02820f90783fa04d5239717ed80d/octavia/api/v2/types/member.py#L87)

**`UpdateMemberOpts`**

[Documentation](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=update-a-member-detail#update-a-member)
[Code](https://github.com/openstack/octavia/blob/198980639cbe02820f90783fa04d5239717ed80d/octavia/api/v2/types/member.py#L100-L104)

**`BatchUpdateMemberOpts`**

[Documentation](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=batch-update-members-detail#batch-update-members)
[Code](https://github.com/openstack/octavia/blob/198980639cbe02820f90783fa04d5239717ed80d/octavia/api/v2/types/member.py#L79-L87)

### DIFF
--- a/openstack/loadbalancer/v2/pools/requests.go
+++ b/openstack/loadbalancer/v2/pools/requests.go
@@ -280,14 +280,20 @@ type CreateMemberOpts struct {
 	// or false (DOWN).
 	AdminStateUp *bool `json:"admin_state_up,omitempty"`
 
-	// Is the member a backup? Backup members only receive traffic when all non-backup members are down.
+	// Is the member a backup? Backup members only receive traffic when all
+	// non-backup members are down.
+	// Requires microversion 2.1 or later.
 	Backup *bool `json:"backup,omitempty"`
 
 	// An alternate IP address used for health monitoring a backend member.
-	MonitorAddress string `json:"monitor_address,omitempty"`
+	MonitorAddress *string `json:"monitor_address,omitempty"`
 
 	// An alternate protocol port used for health monitoring a backend member.
 	MonitorPort *int `json:"monitor_port,omitempty"`
+
+	// A list of simple strings assigned to the resource.
+	// Requires microversion 2.5 or later.
+	Tags []string `json:"tags,omitempty"`
 }
 
 // ToMemberCreateMap builds a request body from CreateMemberOpts.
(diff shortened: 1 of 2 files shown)


openstack/loadbalancer/v2/pools/requests.go
```diff
@@ -390,6 +411,21 @@ type BatchUpdateMemberOpts struct {
...
 	// or false (DOWN).
 	AdminStateUp *bool `json:"admin_state_up,omitempty"`
+
+	// Is the member a backup? Backup members only receive traffic when all
+	// non-backup members are down.
+	// Requires microversion 2.1 or later.
+	Backup *bool `json:"backup,omitempty"`
+
+	// An alternate IP address used for health monitoring a backend member.
+	MonitorAddress string `json:"monitor_address,omitempty"`
```
jtopjian, line 421: Is removing/unsetting a monitor address from a member possible? If so, is it done via an empty string (`""`) or with nil/null?

The combination of `string` and `omitempty` will cause values of empty strings to be removed from the request body. If an empty string is used to remove/unset the monitor address, this should be `*string`.

jtopjian (commented):
@alex Thank you for submitting this. I've left one question.

openstack/loadbalancer/v2/pools/requests.go
mtneug: Yes, it is possible. I tested it and for me

* setting it to `""`
* setting it to `null`
* omitting it

will remove/unset the monitor address. I'm unsure what the preferred way is.

jtopjian: I recommend setting it to:

```go
MonitorAddress *string `json:"monitor_address,omitempty"`
```

jtopjian (approved):
Thank you for working on this. I've noted one small change, but I can fix this post-merge.

edit: I have no idea why Github auto-corrected my at reply to "Alex" and not @mtneug.

Issue #2060, opened 2020-11-23
Networking v2: Add detailed quota reporting


EmilienM:
Neutron supports the ability to get not just the current quota for a project, but also the current allocation against the quota [1] through a "details.json" endpoint. Please enable this collection.

[1] https://docs.openstack.org/api-ref/network/v2/index.html?expanded=show-quota-details-for-a-tenant-detail#quotas-extension-quotas

This action would look similar to the existing quota fetch action, with a more complex response set.

jtopjian:
Implemented in #2061

PR #2061, opened 2020-11-23
Networking v2:  Add Get Detailed Quota for project + 'trunk' extension support

EmilienM:
* Networking v2: Add Get Detailed Quota for Tenant
* Add GetDetail function for the quotas extension in Gophercloud's OpenStack Networking v2 API
* Add 'trunk' resource in Networking v2 quotas

- Allows a caller to get the list of quota types with in_use, reserved, and limit values for each
- Effectively wraps OpenStack Networking v2 API function /v2.0/quotas/{project_id}/details.json
- Inspired from the Compute API implementation.
- When the 'trunk' extension is enabled, it's possible to set quotas for 
trunks resources, the same way as other networking resources.

For #2060 (detailed quota)
For #2064 (trunk support)
API reference: https://docs.openstack.org/api-ref/network/v2/index.html?expanded=show-quota-details-for-a-tenant-detail#quotas-extension-quotas

Code references:
* (extension) https://opendev.org/openstack/neutron/src/branch/master/neutron/extensions/quotasv2_detail.py
* (options) https://opendev.org/openstack/neutron/src/branch/master/neutron/conf/quota.py
* (trunk api) https://opendev.org/openstack/neutron-lib/src/branch/master/neutron_lib/api/definitions/trunk.py#L49
* It is registered as a quota resource in neutron here: https://opendev.org/openstack/neutron/src/branch/master/neutron/extensions/trunk.py

### DIFF
--- a/openstack/networking/v2/extensions/quotas/results.go
+++ b/openstack/networking/v2/extensions/quotas/results.go
@@ -74,6 +74,9 @@ type Quota struct {
 
 	// SubnetPool represents a number of subnet pools. A "-1" value means no limit.
 	SubnetPool int `json:"subnetpool"`
+
+	// Trunk represents a number of trunks. A "-1" value means no limit.
+	Trunk int `json:"trunk"`
 }
 
 // QuotaDetailSet represents details of both operational limits of Networking resources for a project
@@ -105,14 +108,17 @@ type QuotaDetailSet struct {
 
 	// SubnetPool represents a number of subnet pools. A "-1" value means no limit.
 	SubnetPool QuotaDetail `json:"subnetpool"`
+
+	// Trunk represents a number of trunks. A "-1" value means no limit.
+	Trunk QuotaDetail `json:"trunk"`
 }
 
 // QuotaDetail is a set of details about a single operational limit that allows
 // for control of networking usage.
 type QuotaDetail struct {
-	// InUse is the current number of provisioned/allocated resources of the
+	// Used is the current number of provisioned/allocated resources of the
 	// given type.
-	InUse int `json:"used"`
+	Used int `json:"used"`
 
 	// Reserved is a transitional state when a claim against quota has been made
 	// but the resource is not yet fully online.
--- a/openstack/networking/v2/extensions/quotas/results.go
+++ b/openstack/networking/v2/extensions/quotas/results.go
@@ -6,6 +6,10 @@ type commonResult struct {
 	gophercloud.Result
 }
 
+type detailResult struct {
+	gophercloud.Result
+}
+
 // Extract is a function that accepts a result and extracts a Quota resource.
 func (r commonResult) Extract() (*Quota, error) {
 	var s struct {
@@ -15,10 +19,25 @@ func (r commonResult) Extract() (*Quota, error) {
 	return s.Quota, err
 }
 
+// Extract is a function that accepts a result and extracts a QuotaDetailSet resource.
+func (r detailResult) Extract() (*QuotaDetailSet, error) {
+	var s struct {
+		Quota *QuotaDetailSet `json:"quota"`
+	}
+	err := r.ExtractInto(&s)
+	return s.Quota, err
+}
+
 // GetResult represents the result of a get operation. Call its Extract
 // method to interpret it as a Quota.
 type GetResult struct {
 	commonResult
+}
+
+// GetDetailResult represents the detailed result of a get operation. Call its Extract
+// method to interpret it as a Quota.
+type GetDetailResult struct {
+	detailResult
 }
 
 // UpdateResult represents the result of an update operation. Call its Extract
@@ -55,4 +74,51 @@ type Quota struct {
 
 	// SubnetPool represents a number of subnet pools. A "-1" value means no limit.
 	SubnetPool int `json:"subnetpool"`
+}
+
+// QuotaDetailSet represents details of both operational limits of Networking resources for a project
+// and the current usage of those resources.
+type QuotaDetailSet struct {
+	// FloatingIP represents a number of floating IPs. A "-1" value means no limit.
+	FloatingIP QuotaDetail `json:"floatingip"`
+
+	// Network represents a number of networks. A "-1" value means no limit.
+	Network QuotaDetail `json:"network"`
+
+	// Port represents a number of ports. A "-1" value means no limit.
+	Port QuotaDetail `json:"port"`
+
+	// RBACPolicy represents a number of RBAC policies. A "-1" value means no limit.
+	RBACPolicy QuotaDetail `json:"rbac_policy"`
+
+	// Router represents a number of routers. A "-1" value means no limit.
+	Router QuotaDetail `json:"router"`
+
+	// SecurityGroup represents a number of security groups. A "-1" value means no limit.
+	SecurityGroup QuotaDetail `json:"security_group"`
+
+	// SecurityGroupRule represents a number of security group rules. A "-1" value means no limit.
+	SecurityGroupRule QuotaDetail `json:"security_group_rule"`
+
+	// Subnet represents a number of subnets. A "-1" value means no limit.
+	Subnet QuotaDetail `json:"subnet"`
+
+	// SubnetPool represents a number of subnet pools. A "-1" value means no limit.
+	SubnetPool QuotaDetail `json:"subnetpool"`
+}
+
+// QuotaDetail is a set of details about a single operational limit that allows
+// for control of networking usage.
+type QuotaDetail struct {
+	// InUse is the current number of provisioned/allocated resources of the
+	// given type.
+	InUse int `json:"used"`
+
+	// Reserved is a transitional state when a claim against quota has been made
+	// but the resource is not yet fully online.
+	Reserved int `json:"reserved"`
+
+	// Limit is the maximum number of a given resource that can be
+	// allocated/provisioned.  This is what "quota" usually refers to.
+	Limit int `json:"limit"`
 }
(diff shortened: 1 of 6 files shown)


jtopjian:
@EmilienM Thank you for submitting this! 

For the code review, we also need to see the actual Neutron Service API code that defines this API call. Please see here for more details:

https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md

The most important part is providing a reference that the `QuotaDetailSet` you've created is accurate and all of the fields are valid.

EmilienM:
> @EmilienM Thank you for submitting this!

Hey, long time no see! ;-)

> For the code review, we also need to see the actual Neutron Service API code that defines this API call. Please see here for more details:
> 
> https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md
> 
> The most important part is providing a reference that the `QuotaDetailSet` you've created is accurate and all of the fields are valid.
> 
> Let me know if you have any questions or need any help.

Ack. I updated the PR body, let me know if it's enough.
Please note that the fields are the same as the QuotaSet, but for each field Neutron will figure out the resources usage/reservations.
Thanks!

jtopjian:
> Hey, long time no see!

Yes indeed. I hope you're doing well! 🙂 

> I updated the PR body, let me know if it's not enough.

I think it is - the extension definition should be good enough to trace everything else.

One thing I noticed immediately is that both the API docs and the code are using the word `used` and not `in_use`. Can you confirm which one you're seeing in your environment?

https://github.com/openstack/neutron/blob/b303f4409ea270d53866c7e296aa76d6b603ead8/neutron/db/quota/driver.py#L101-L104

EmilienM:
> One thing I noticed immediately is that both the API docs and the code are using the word `used` and not `in_use`. Can you confirm which one you're seeing in your environment?

You're right, it was an error on my side and in my testing I didn't have used resources so I didn't catch it.
Patch is updated. Thanks for the quick reviews!

openstack/networking/v2/extensions/quotas/results.go
```diff
@@ -56,3 +75,50 @@ type Quota struct {
 	// SubnetPool represents a number of subnet pools. A "-1" value means no limit.
 	SubnetPool int `json:"subnetpool"`
 }
+
+// QuotaDetailSet represents details of both operational limits of Networking resources for a project
+// and the current usage of those resources.
+type QuotaDetailSet struct {
```
mdbooth, line 81: This set is potentially problematic, because it's generated dynamically. It can certainly differ between clouds. E.g. my cloud outputs a 'trunk' resource:

```
{
  "quota": {
    "network": {
      "limit": -1,
      "used": 2,
      "reserved": 0
    },
    "subnet": {
      "limit": -1,
      "used": 2,
      "reserved": 0
    },
    "subnetpool": {
      "limit": -1,
      "used": 0,
      "reserved": 0
    },
    "port": {
      "limit": -1,
      "used": 14,
      "reserved": 0
    },
    "router": {
      "limit": -1,
      "used": 2,
      "reserved": 0
    },
    "floatingip": {
      "limit": -1,
      "used": 2,
      "reserved": 0
    },
    "rbac_policy": {
      "limit": -1,
      "used": 0,
      "reserved": 0
    },
    "security_group": {
      "limit": -1,
      "used": 4,
      "reserved": 0
    },
    "security_group_rule": {
      "limit": -1,
      "used": 46,
      "reserved": 0
    },
    "trunk": {
      "limit": -1,
      "used": 6,
      "reserved": 0
    }
  }
}
```

'trunk' originates here:

https://opendev.org/openstack/neutron-lib/src/branch/master/neutron_lib/api/definitions/trunk.py#L49

It is registered as a quota resource in neutron here:

https://opendev.org/openstack/neutron/src/branch/master/neutron/extensions/trunk.py

I did a quick audit of other resources which are registered in the same way and only found 'router', 'floatingip', and 'subnetpool', all of which you already have.

At a minimum you should add 'trunk'. It would again be down to the style goals of gophercloud whether this data should be returned as a struct or a map.

Something else to call out: if we continue to return a struct, any field in this struct which wasn't returned by the server will be filled with {InUse: 0, Reserved: 0, Limit: 0}. This is valid quota, so the caller wouldn't be able to distinguish this from the resource not being present at all. This might be ok, but we should probably make this decision deliberately.

mdbooth (changes requested):
QuotaDetail.InUse is inconsistent with the api field, which is called user. However, it is consistent with QuotaUsage in blockstorage, and QuotaDetail in compute.

Whether we should rename the field comes down to the goals of the gophercloud project: does it aim to 'fix' inconsistencies in the OpenStack API such as this, or faithfully reproduce them? I don't know the answer.

See inline discussion about dynamic generation of quota resources by neutron. 'Request changes' is specifically for the missing trunk field. Everything else is for discussion.

openstack/networking/v2/extensions/quotas/results.go
EmilienM: [...]
> At a minimum you should add 'trunk'. It would again be down to the style goals of gophercloud whether this data should be returned as a struct or a map.

I'll add `trunk` (since we'll need it for our needs), but I suspect there are more extensions that register a resource in the quotas, they can be added later.

mdbooth: Every field in this struct registered dynamically, btw, not just trunk. router, floatingip, and subnetpool are also registered in the same way. Also, the whole of neutron seems to be extensions. I wouldn't call out trunk as being different: it's probably just the most recent extension to be added. With an architecture like this you can almost guarantee that the api docs for quota detail will often/usually be out of date.

mdbooth:
Questions for a regular gophercloud maintainer:

1. Should QuotaDetail.InUse be renamed to QuotaDetail.Used to be consistent with the neutron api, or should it remain QuotaDetail.InUse to be consistent with similar uses in Cinder and Nova.

2. Should QuotaDetailSet remain a struct, or become a map?

Map advantages:
- Reflects how the neutron api generates this set (dynamically registered)
- Newly added extensions are added automatically without changes in gophercloud
- API user can differentiate between a zero resource and a resource which isn't present
Map disadvantages:
- It's considerably less ergonomic

openstack/networking/v2/extensions/quotas/results.go
pierreprinetti: What is the expected QuotaDetail value when the resource is not in the response?

If this question has a satisfying answer, then the struct would be much more usable than a map on the consumer side.

jtopjian:
> Should QuotaDetail.InUse be renamed to QuotaDetail.Used to be consistent with the neutron api, or should it remain QuotaDetail.InUse to be consistent with similar uses in Cinder and Nova.

We always stick to what the API is returning. If Gophercloud is inconsistent, it's because the API is inconsistent 🙂 

> Every field in this struct registered dynamically
> Should QuotaDetailSet remain a struct, or become a map?

Good question. I think this all depends on how often the resources in the returned result will differ. As an extreme example, if a cloud admin was able to create arbitrary network resources via the API, then yes, this absolutely should be a map. But on the other side of the spectrum, if Neutron (as a collection of extensions) ultimately only has a finite set of resources, then a struct probably works better. 

If it's common for very large OpenStack environments or forks of OpenStack to add additional Neutron extensions that register in-house/site-specific resources, then that shouldn't be taken into consideration here. Gophercloud only ensures compatibility with "vanilla" OpenStack. For this situation, the in-house environment would need to create a Go package that redefines the `QuotaDetailSet` struct.

This is not set in stone - just the general guidelines used in Gophercloud. Happy to discuss more.

mdbooth:
> > Should QuotaDetail.InUse be renamed to QuotaDetail.Used to be consistent with the neutron api, or should it remain QuotaDetail.InUse to be consistent with similar uses in Cinder and Nova.
> 
> We always stick to what the API is returning. If Gophercloud is inconsistent, it's because the APIs are inconsistent 🙂
> 
> > Every field in this struct registered dynamically
> > Should QuotaDetailSet remain a struct, or become a map?
> 
> Good question. I think this all depends on how often the resources in the returned result will differ. As an extreme example, if a cloud admin was able to create arbitrary network resources via the API, then yes, this absolutely should be a map. But on the other side of the spectrum, if Neutron (as a collection of extensions) ultimately only has a finite set of resources, then a struct probably works better.
> 
> If it's common for very large OpenStack environments or forks of OpenStack to add additional Neutron extensions that register in-house/site-specific resources, then that shouldn't be taken into consideration here. Gophercloud only ensures compatibility with "vanilla" OpenStack. For this situation, the in-house environment would need to create a Go package that redefines the `QuotaDetailSet` struct.
> 
> This is not set in stone - just the general guidelines used in Gophercloud. Happy to discuss more.

Both of those are the answers I was expecting :) I'm happy with this patch if InUse is renamed to Used and a Trunk field is added to the struct.

EmilienM:
> I'm happy with this patch if InUse is renamed to Used and a Trunk field is added to the struct.

Done.

Thanks for the reviews everyone!

openstack/networking/v2/extensions/quotas/urls.go
```diff
@@ -3,15 +3,24 @@ package quotas
...
+	return c.ServiceURL(resourcePath, projectID, resourcePathDetail)
+}
+
 func getURL(c *gophercloud.ServiceClient, projectID string) string {
 	return resourceURL(c, projectID)
 }
 
+func getDetailURL(c *gophercloud.ServiceClient, projectID string) string {
+	return resourceDetailURL(c, projectID)
+}
```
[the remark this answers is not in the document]
jtopjian: Yeah, I agree that this is an odd duplication. Looking at the Compute quota code, I can see the direction the author was going in. But yeah, let's just keep this as-is.

jtopjian (approved):
LGTM - I confirmed this works in a live environment.

@EmilienM Thank you for submitting this.

And thank you to everyone else for your input and reviews.

Issue #2062, opened 2020-11-24
Objectstorage v1: metadata cannot be removed from accounts and objects


reenjii:
Related to [#1511](https://github.com/gophercloud/gophercloud/issues/1511)

When updating an objectstorage account or object, it is not possible to remove a metadata. If you set a key to an empty string [the empty header is stripped from the request when the POST is made](https://github.com/gophercloud/gophercloud/blob/master/provider_client.go#L307).

> // MoreHeaders specifies additional HTTP headers to be provide on the request. If a header is
> // provided with a blank value (""), that header will be *omitted* instead: use this to suppress
> // the default Accept header or an inferred Content-Type, for example.
> MoreHeaders map[string]string

Both calls should have a `RemoveMetadata` slice to allow sending `X-Remove-xxx-Meta-xxx` headers.

jtopjian:
Implemented in #2063

PR #2063, opened 2020-11-24
Objectstore v1: allow to remove account and object metadata

reenjii:
For #2062

Documentation about metadata handling in middlewares.
https://github.com/openstack/swift/blob/master/doc/source/development_middleware.rst#user-metadata

Actual code that handles `X-Remove-xxx-Meta-name` haders.
https://github.com/openstack/swift/blob/master/swift/proxy/controllers/base.py#L1647

Documentation update account.
https://docs.openstack.org/api-ref/object-store/#create-update-or-delete-account-metadata

> X-Remove-Account-name (Optional) | header | string | Removes the metadata item named name. For example, X-Remove-Account-Meta-Blue removes custom metadata.

### DIFF
--- a/openstack/objectstorage/v1/accounts/requests.go
+++ b/openstack/objectstorage/v1/accounts/requests.go
@@ -53,6 +53,7 @@ type UpdateOptsBuilder interface {
 // deleting an account's metadata.
 type UpdateOpts struct {
 	Metadata          map[string]string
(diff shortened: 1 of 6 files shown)


jtopjian (approved):
I apologize for the delay with reviewing this.

LGTM - thank you!

Issue #1719, opened 2019-09-29
Identity V3: Get available catalog/project/domain/system scopes  


zhsj:
The following APIs are missing,

1. [GET /v3/auth/catalog](https://docs.openstack.org/api-ref/identity/v3/#get-service-catalog)
1. [GET /v3/auth/projects](https://docs.openstack.org/api-ref/identity/v3/#get-available-project-scopes)
1. [GET /v3/auth/domains](https://docs.openstack.org/api-ref/identity/v3/#get-available-domain-scopes)
1. [GET /v3/auth/system](https://docs.openstack.org/api-ref/identity/v3/#get-available-system-scopes)

PR #2067, opened 2020-12-04
[Identity v3] Add service catalog listing

outcatcher:
For #1719 (one of four requested resources)

Add support of service catalog list

API reference:
https://docs.openstack.org/api-ref/identity/v3/?expanded=get-service-catalog-detail#get-service-catalog

Implementation:
https://github.com/openstack/keystone/blob/stable/victoria/keystone/catalog/core.py#L237

### DIFF
--- a/acceptance/openstack/identity/v3/catalog_test.go
+++ b/acceptance/openstack/identity/v3/catalog_test.go
@@ -1,0 +1,25 @@
+package v3
+
+import (
+	"testing"
+
+	"github.com/gophercloud/gophercloud/acceptance/clients"
+	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/identity/v3/catalog"
+	th "github.com/gophercloud/gophercloud/testhelper"
+)
+
+func TestCatalogList(t *testing.T) {
+	client, err := clients.NewIdentityV3Client()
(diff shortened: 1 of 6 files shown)


jtopjian:
@outcatcher Thanks for working on this. Please let me know if this is ready for review.

outcatcher:
@jtopjian It's already ready to be reviewed.
I wanted to minimize the scope of my first contribution, so that's all I wanted to submit.

jtopjian (approved):
LGTM - thank you!

A few nits for the next PR:

1. The following area of code better defines this API call: https://github.com/openstack/keystone/blob/master/keystone/api/auth.py#L230. You can see how this code calls the code that you provided. It also defines the JSON response structure.

2. When possible, please try not to use the JSON responses from the API documentation for unit tests. Please try to obtain an actual JSON response from a live environment. 

As an example, search this link (https://logs.openlabtesting.org/logs/67/2067/219f98f8514e6f150450b4d1bda325f73f7b6c5d/check/gophercloud-acceptance-test/ce06b4b/job-output.txt.gz) for `TestCatalogList`. You'll see the JSON requests and responses included in this output file. The JSON response is slightly different than the API documentation in that it includes more than one service as well as includes a `region_id` field. 

The fixture you used is good enough to keep, so just a nit.

All in all, really nice work - thank you for submitting this!

PR #2050, opened 2020-11-02
Fwaasv2 firewall groups acc tests (depends on #2049)

codemanufaktur:
For #514

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:

https://github.com/openstack/neutron-fwaas/blob/stable/ussuri/neutron_fwaas/db/firewall/v2/firewall_db_v2.py#L85-L103
https://github.com/openstack/neutron-fwaas/blob/stable/ussuri/neutron_fwaas/db/firewall/v2/firewall_db_v2.py#L307:L322
https://github.com/openstack/neutron-fwaas/blob/stable/ussuri/neutron_fwaas/services/firewall/fwaas_plugin_v2.py#L313-L369

**this PR:
Acceptance test for fwaasv2 firewall groups**

kind regards, 
Georg

<sub>Georg Schreiber georg.g.schreiber@daimler.com, Daimler TSS GmbH, [legal info/Impressum](https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md)</sub>

PR #2073, opened 2020-12-16
openstack/networking/v2/extensions/fwaas_v2/groups/requests.go 

chrischdi:
For #2099

make UpdateOpts.Ports to pointer to allow unsetting it explicitly.

By this change it will be possible to do the same as `openstack firewall group unset --all-ports <id>`.

As long as the Ports field is only a `[]string`:
* whenever I want to set the empty array, the attribute will get dropped due to `omitempty` resulting in not unsetting the value.
* however removing the `omitempty` will result in always having to set the value (and implicitly do a unset, if not setting the attribute, when only updating other values is wanted).

The struct was introduced in https://github.com/gophercloud/gophercloud/pull/2050

We need to be able to unset the value in our openstack, because on our openstack the fwaas group will only be able to get deleted when we get a status of `DOWN`. So we have to delete by:
* setting the firewall group to `disable`
* setting the ports of the firewall group to empty (to trigger reconciling the status in the fwaas driver)

---
<sup>
Christian Schlotter <christian.schlotter@daimler.com>, Daimler TSS GmbH, 
<a href="https://github.com/Daimler/daimler-foss/blob/master/LEGAL_IMPRINT.md">Imprint</a>
</sup>

### DIFF
--- a/openstack/networking/v2/extensions/fwaas_v2/groups/requests.go
+++ b/openstack/networking/v2/extensions/fwaas_v2/groups/requests.go
@@ -119,13 +119,13 @@ type UpdateOptsBuilder interface {
 
 // UpdateOpts contains the values used when updating a firewall group.
 type UpdateOpts struct {
-	Name                    *string  `json:"name,omitempty"`
-	Description             *string  `json:"description,omitempty"`
-	IngressFirewallPolicyID *string  `json:"ingress_firewall_policy_id,omitempty"`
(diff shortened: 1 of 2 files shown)


chrischdi:
recheck

chrischdi:
@jtopjian is there a chance getting the openlab check running?

chrischdi:
> @jtopjian is there a chance getting the openlab check running?

Sorry for disturbing you :-) looks like the recheck did the trick :+1:

jtopjian (approved):
LGTM - thank you!

Issue #2066, opened 2020-12-02
Support of update snapshot operation


TenSt:
Hi Gophercloud team.
I was going through crud operations of snapshots and I see that update is not there. There is update metadata, however no update for snapshot. There is an API call for this operation in OpenStack:
https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=#update-a-snapshot

I'm willing to work on contribution of it, but before starting wanted to know if there was some reason behind not covering it from the start? Is there some issue with it or it was just missed and I can work on contributing it?

Thanks,
Stepan

jtopjian:
@TenSt Thanks for opening this issue. It looks like the update call was just overlooked and never added.

TenSt:
@jtopjian thanks. Then I will start working on contributing it. It may take some time as it would be my first contribution to gophercloud and I will be configuring local environment for the first time.

jtopjian:
@TenSt No problem. Let me know if you need any help or have any questions.

TenSt:
@jtopjian 
Sorry for the delay, I was really pinned down at work. Here is the PR, I hope I've covered everything: #2081

TenSt:
Closing, as the code was successfully merged and released in v0.15.0

PR #2081, opened 2020-12-20
Block storage: adding support of update snapshot api call

TenSt:
For #2066 

- [x] code
- [x] unit tests
- [x] acceptance tests
- [x] documentation

API:
https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=update-a-snapshot-detail#update-a-snapshot

Schema:
https://github.com/openstack/cinder/blob/482e6a3cc5cca697b54ee1d853a4eca6e6f3cfc7/cinder/api/schemas/snapshots.py#L45

It seems that v3 uses the same method as v2 under the hood:
https://github.com/openstack/cinder/blob/482e6a3cc5cca697b54ee1d853a4eca6e6f3cfc7/cinder/api/v3/snapshots.py#L26

Update api:
https://github.com/openstack/cinder/blob/482e6a3cc5cca697b54ee1d853a4eca6e6f3cfc7/cinder/api/v2/snapshots.py#L148

Additional information:
Acceptance tests were ran against live OpenStack API(Ussuri). Here are logs from its run:
[AcceptanceTestsRun_logs.txt](https://github.com/gophercloud/gophercloud/files/5721502/AcceptanceTestsRun_logs.txt)

### DIFF
--- a/acceptance/openstack/blockstorage/v3/snapshots_test.go
+++ b/acceptance/openstack/blockstorage/v3/snapshots_test.go
@@ -28,8 +28,8 @@ func TestSnapshots(t *testing.T) {
 	defer DeleteSnapshot(t, client, snapshot1)
 
 	// Update snapshot
-	updatedSnapshotName := "ACPTTEST-002"
-	updatedSnapshotDescription := "ACPTTEST-002"
+	updatedSnapshotName := tools.RandomString("ACPTTEST", 16)
+	updatedSnapshotDescription := tools.RandomString("ACPTTEST", 16)
 	updateOpts := snapshots.UpdateOpts{
 		Name:        &updatedSnapshotName,
 		Description: &updatedSnapshotDescription,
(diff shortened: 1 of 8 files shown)


acceptance/openstack/blockstorage/v3/snapshots_test.go
```diff
@@ -26,6 +27,21 @@ func TestSnapshots(t *testing.T) {
 	th.AssertNoErr(t, err)
 	defer DeleteSnapshot(t, client, snapshot1)
 
+	// Update snapshot
+	updatedSnapshotName := "ACPTTEST-002"
+	updatedSnapshotDescription := "ACPTTEST-002"
```
jtopjian, line 32: nit: It would be better to use values that are randomly generated instead of hard-coded. You can randomly generate a name and description using this method: https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/blockstorage/v3/blockstorage.go#L20-L21

jtopjian (changes requested):
@TenSt Thanks! This all looks really good - really nice work. Just one small nit with the acceptance test.

You can ignore the OpenLab error. The failure is coming from an unrelated test. 

TenSt:
@jtopjian Thanks! I've done the change you requested.

jtopjian:
recheck

jtopjian (approved):
LGTM - thanks!

Issue #1702, opened 2019-09-16
Blockstorage: Add limits extension


thcdrt:
Hello,

I opened recently an issue about nova limits call (https://github.com/gophercloud/gophercloud/issues/1697).
Then I found out that the missing information are in another limits call, but in the volume API.

The code for nova limits is at this path: openstack/compute/v2/extensions/limits
But I can't found the equivalence on volume side in https://github.com/gophercloud/gophercloud/tree/master/openstack/blockstorage/extensions.

Are you ok to allow me to add it?

API Documentation: 
[https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=show-absolute-limits-for-project-detail#limits-limits](https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=show-absolute-limits-for-project-detail#limits-limits).

V2 implementation:
[https://github.com/openstack/cinder/blob/master/cinder/api/v2/limits.py](https://github.com/openstack/cinder/blob/master/cinder/api/v2/limits.py)

V3 implementation:
[https://github.com/openstack/cinder/blob/master/cinder/api/v3/limits.py](https://github.com/openstack/cinder/blob/master/cinder/api/v3/limits.py)

Thanks,
Thomas

jtopjian:
@thcdrt Yes, thank you for offering 🙂 

We have a [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial) that should help get you going. Note that you'll need to reference more than just the API docs for this.

Please let me know if you need any help or have any questions.

thcdrt:
Hello @jtopjian, sorry but I don't understand what you want me to add.
In the [create an issue tutorial ](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-02-issues.md) in the [add a missing call section](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-02-issues.md#adding-a-missing-api-call), the [example](https://github.com/gophercloud/gophercloud/issues/715) just add the link of the doc.

Thomas

jtopjian:
@thcdrt Yes, true. 

Most people contributing PRs skip Part 3 and don't mention the actual API service code either in the issue or PR (https://github.com/gophercloud/gophercloud/pull/716). Either is fine, but it's just become habit to always make mention of it.

thcdrt:
@jtopjian  I updated it, is it better ?

[...]

PR #2084, opened 2020-12-28
Blockstorage extensions limits

TenSt:
For #1702

All links are already included in the issue.

Notes on Limits struct:
The original code contains fields only for max values, but API docs states that there are also total values. I've tested out on OpenStack Ussuri and I see that API docs are valid. I'm not very good at python, so might've missed where exactly total fields are added to the response.
Original code: https://github.com/openstack/cinder/blob/master/cinder/api/views/limits.py#L42-L48
API docs: https://docs.openstack.org/api-ref/block-storage/v3/index.html?#show-absolute-limits-for-project

Also, I've left "rate" as interface because it is not clear from the code what will return there and docs example is just an empty array (as well as response from real OpenStack).

Note on acceptance test: the hard-coded values are default ones, so should work for any OpenStack Project with default values.


jtopjian:
@TenSt Thank you for working on this.

The OpenLab failure is from the acceptance test you've added. You can see the log here: https://logs.openlabtesting.org/logs/84/2084/1f255e0aca45b3545996a6691c5c93121b8e213e/check/gophercloud-acceptance-test/6eee85a/job-output.txt.gz and search for the string `--- FAIL`. 

The cause of the error looks to be from the backup created in the previous test not deleting in time for the new test to run. I recommend changing the assertion to compare against snapshots, maybe.

> Also, I've left "rate" as interface because it is not clear from the code what will return there and docs example is just an empty array (as well as response from real OpenStack).

The code that builds the rate can be found here: https://github.com/openstack/cinder/blob/master/cinder/api/views/limits.py#L56-L91

It looks like there's a unit test within Cinder that details how the data structure should look: https://github.com/openstack/cinder/blob/84a084d1bdc88c649c1f3b728beb36dd0707dc22/cinder/tests/unit/api/v2/test_limits.py#L119-L156

You can use this code for your own test fixture, which should allow you to create a detailed struct for the `rate`. 

TenSt:
@jtopjian Thank you for the review and pointing out where to find rate object.

Rate part of Limits:
Added rate struct and updated unit tests accordingly.

Acceptance tests:
Updated acceptance tests to verify that current values greater or equal than 0 and lesser or equal than the max value. This way the test is more flexible to dynamic values of current usage of blockstorage in the project.

TenSt:
@jtopjian I've fixed acceptance tests, so PR is ready for the next review round.

jtopjian (approved):
LGTM - thank you!

Issue #2083, opened 2020-12-24
Backups list detail


rendaw:
`openstack volume backup list --volume X` uses `/v3/.../backups/detail?volume_id=...` whereas gophercloud uses `/v3/.../backups?volume_id=...`.  `detail` adds the backup status which can be useful for filtering out failed backups (significantly reduces number of requests if there are a lot).

I'd be glad to make a PR adding this if someone can give me an overview of the form it should take.  New `ListDetail` function and `listDetailUrl` which takes the same query struct?

jtopjian:
@rendaw Thank you for opening this issue.

I think it's best to implement a whole new set of code, duplicating the existing [regular list call](https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/extensions/backups/requests.go#L86-L146), but for `ListDetail`. 

Perhaps the detailed list call is backwards compatible with the regular list call, and perhaps could have superseded the regular list call, but it's probably better and more flexible to users to have both as separate calls.

Let me know if this makes sense or if you have any questions.

rendaw:
Perfect, thanks!

jtopjian:
Implemented in #2085

PR #2085, opened 2020-12-31
Backups list detail #2083

rendaw:
For #2083 

Here's the API docs:

https://docs.openstack.org/api-ref/block-storage/v3/?expanded=list-backups-with-detail-detail#list-backups-with-detail

Sorry, I'm not sure where to find this in the openstack source.

Also I modified the List tests to not include additional information like `status` - this is what I observed against OVH, and appears to match the docs here: https://docs.openstack.org/api-ref/block-storage/v3/?expanded=list-backups-for-project-detail#list-backups-for-project

### DIFF
--- a/openstack/blockstorage/extensions/backups/requests.go
+++ b/openstack/blockstorage/extensions/backups/requests.go
@@ -145,6 +145,55 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 	})
 }
 
+// ListDetailOptsBuilder allows extensions to add additional parameters to the ListDetail
+// request.
+type ListDetailOptsBuilder interface {
+	ToBackupListDetailQuery() (string, error)
+}
+
+type ListDetailOpts struct {
+	// AllTenants will retrieve backups of all tenants/projects.
+	AllTenants bool `q:"all_tenants"`
+
+	// Comma-separated list of sort keys and optional sort directions in the
+	// form of <key>[:<direction>].
+	Sort string `q:"sort"`
+
+	// Requests a page size of items.
+	Limit int `q:"limit"`
+
+	// Used in conjunction with limit to return a slice of items.
+	Offset int `q:"offset"`
+
+	// The ID of the last-seen item.
+	Marker string `q:"marker"`
+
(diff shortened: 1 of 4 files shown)


rendaw:
Sorry, I haven't tested the code as a whole live, I was going to hook it into the project I'm on but go complains about the module name not matching...

jtopjian:
> Sorry, I'm not sure where to find this in the openstack source.

The source code for this API action is here:

* https://github.com/openstack/cinder/blob/acfc148d7cec08362873115abb4551d2ba531d06/cinder/api/views/backups.py#L32-L35
* https://github.com/openstack/cinder/blob/acfc148d7cec08362873115abb4551d2ba531d06/cinder/api/views/backups.py#L84-L100
* https://github.com/openstack/cinder/blob/acfc148d7cec08362873115abb4551d2ba531d06/cinder/api/views/backups.py#L58-L80

jtopjian:
I don't have access to an OpenStack environment that supports volume backups to test this out.

> Sorry, I haven't tested the code as a whole live, I was going to hook it into the project I'm on but go complains about the module name not matching...

You can run the acceptance test for this against your cloud by doing the following:

1. Edit https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/blockstorage/extensions/backups_test.go#L29 and change `List` to `ListDetail`.
2. Source an `openrc` file
3. Run:

```
$ cd gophercloud
$ go test -v -tags "fixtures acceptance" -run "TestBackupsCRUD" github.com/gophercloud/gophercloud/acceptance/openstack/blockstorage/extensions
```

Let me know if that works or if you run into any problems.

rendaw:
Hey, sorry for the delay!  I tried the acceptance tests but the openrc I have for this provider (OVH) doesn't work with the standard env auth parser (we use clientconfig), and I was getting errors in lines that I had modified locally to try to get it to work; I wasn't confident that the acceptance test was running the up to date acceptance test code.  Also I was hesitant to create and delete backups on the region :sweat_smile: 

I figured out `replace` though and tested just ListDetail, it works with no issues (returns all backups with full details).

jtopjian (approved):
@rendaw LGTM - thank you for working on this and for testing it out.

Issue #2088, opened 2021-01-04
Clustering: Return ActionResult for node remove API


dkt26111:
The Node Remove API in Senlin Clustering returns a actionID same as Node Add API:

https://docs.openstack.org/api-ref/clustering/?expanded=remove-nodes-from-a-cluster-detail#remove-nodes-from-a-cluster

Therefore, the Node Remove API should return an ActionResult same as Node Add API.

jtopjian:
Implemented in #2089

PR #2089, opened 2021-01-04
Return actionID for Node Remove API

dkt26111:
For #2088 

code in this PR:

https://github.com/openstack/senlin/blob/master/senlin/conductor/service.py#L970
https://docs.openstack.org/api-ref/clustering/?expanded=remove-nodes-from-a-cluster-detail#remove-nodes-from-a-cluster


jtopjian (approved):
LGTM - thank you!

Issue #2091, opened 2021-01-05
Remove imports of golang.org/x/crypto older than v0.0.0-20201203163018-be400aefbc4c


pierreprinetti:
Even though Gophercloud might now be directly affectred by
CVE-2020-29652, removing the vulnerable dependency will:

* endure that such a vulnerability is not introduced in Gophercloud in
the future;
* simplify threat analysis in dependent packages.

jtopjian:
Implemented in #2092

PR #2092, opened 2021-01-05
Bump golang.org/x/crypto

pierreprinetti:
Even though Gophercloud might now be directly affectred by
CVE-2020-29652, removing the vulnerable dependency will:

* endure that such a vulnerability is not introduced in Gophercloud in
the future;
* simplify threat analysis in dependent packages.

After this patch, only one occurrence of the vulnerable library will
remain; it is imported indirectly by the new, fixed version of the same
golang.org/x/crypto.

ref.: https://go-review.googlesource.com/c/crypto/+/281632

For #2091


jtopjian (approved):
LGTM - Thank you!

pierreprinetti:
Below the (hopefully) typo-free version of my PR description:

---

Even though Gophercloud might not be directly affected by
CVE-2020-29652, removing the vulnerable dependency will:

* ensure that such a vulnerability is not introduced in Gophercloud in
    the future;
* simplify threat analysis in dependent packages.

After this patch, only one occurrence of the vulnerable library will
remain; it is imported indirectly by the new, fixed version of the same
golang.org/x/crypto.

jtopjian:
@pierreprinetti It's all good - I knew what you meant :)

Issue #2096, opened 2021-01-15
client: gophercloud should respect "Retry-After" response header


kayrus:
When OpenStack service throttles and responses with the 429 response code and "Retry-After" response header, provider client should backoff and wait for "Retry-After" before the next try.

The default backoff function should be called unless the `RetryBackoffDisabled` provider struct member is true. This will allow to enable the backoff logic by default and simplify the integration in dependent projects.

In order to give more control, the `RetryBackoffFunc` provider struct member can be defined (by analogy with `ReauthFunc`).

The default `MaxBackoffRetries` provider struct member should be set to `DefaultMaxBackoffRetries` const. When an amount of backoff retries is greater than `MaxBackoffRetries`, provider client should return the 429 error.

Reference:
* https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.37
* https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

see also #1918

jtopjian:
Implemented in #2097

PR #2097, opened 2021-01-15
Client: introduce the retry backoff

kayrus:
For #2096

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -13,7 +13,10 @@ import (
 )
 
 // DefaultUserAgent is the default User-Agent string set in the request header.
-const DefaultUserAgent = "gophercloud/2.0.0"
+const (
+	DefaultUserAgent         = "gophercloud/2.0.0"
+	DefaultMaxBackoffRetries = 60
+)
 
 // UserAgent represents a User-Agent header.
 type UserAgent struct {
@@ -79,6 +82,12 @@ type ProviderClient struct {
 
 	// Context is the context passed to the HTTP request.
 	Context context.Context
+
+	// Retry backoff func
+	RetryBackoffFunc func(context.Context, *ErrUnexpectedResponseCode, error, uint) error
+
+	// MaxBackoffRetries set the maximum number of backoffs. When not set, defaults to DefaultMaxBackoffRetries
+	MaxBackoffRetries uint
 
 	// mut is a mutex for the client. It protects read and write access to client attributes such as getting
 	// and setting the TokenID.
@@ -323,6 +332,8 @@ type requestState struct {
 	// reauthenticate, but keep getting 401 responses with the fresh token, reauthenticating some more
 	// will just get us into an infinite loop.
 	hasReauthenticated bool
+	// Retry-After backoff counter, increments during each backoff call
+	retries uint
 }
 
 var applicationJSON = "application/json"
@@ -502,6 +513,24 @@ func (client *ProviderClient) doRequest(method, url string, options *RequestOpts
 			err = ErrDefault429{respErr}
 			if error429er, ok := errType.(Err429er); ok {
 				err = error429er.Error429(respErr)
+			}
+
+			maxTries := client.MaxBackoffRetries
+			if maxTries == 0 {
+				maxTries = DefaultMaxBackoffRetries
+			}
+
+			if f := client.RetryBackoffFunc; f != nil && state.retries < maxTries {
+				var e error
+
+				state.retries = state.retries + 1
+				e = f(client.Context, &respErr, err, state.retries)
+
+				if e != nil {
+					return resp, e
+				}
+
+				return client.doRequest(method, url, options, state)
 			}
 		case http.StatusInternalServerError:
 			err = ErrDefault500{respErr}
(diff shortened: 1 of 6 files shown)


provider_client.go
```diff
@@ -335,6 +351,29 @@ func (client *ProviderClient) Request(method, url string, options *RequestOpts)
 	})
 }
 
+func (client *ProviderClient) retryBackoff(respErr *ErrUnexpectedResponseCode, e error, retries uint) error {
+	var sleep time.Duration
+
+	retryAfter := respErr.ResponseHeader.Get("Retry-After")
+	if retryAfter == "" {
+		return e
+	}
```
jtopjian, line 360: I understand that a 429 should always include a `Retry-After` header, but in the event that one is missing, should the code silently continue (return nil) instead of fail?

I feel that if the code returns `nil` here, it more closely matches the current behaviour of not supporting a backoff at all. Therefore, we're not introducing a large change in behaviour to the client.

provider_client.go
```diff
@@ -335,6 +351,29 @@ func (client *ProviderClient) Request(method, url string, options *RequestOpts)
...
+
+	if v, err := strconv.ParseUint(retryAfter, 10, 32); err == nil {
+		sleep = time.Duration(v) * time.Second
+	}
+
+	if v, err := time.Parse(http.TimeFormat, retryAfter); err != nil {
```
jtopjian, line 366: Should this code only execute if the above `ParseUint` was unsuccessful?

jtopjian (commented):
@kayrus Thanks for suggesting this. I agree that that it makes sense to add this functionality to the core client.

I left two general comments.

provider_client.go
kayrus: no, it should not return nil. `e` is the error, which comes from the `doRequest` method. So when there is no `Retry-After` header or it's wrong, a provider client will behave like the `RetryBackoffDisabled` is true.

provider_client.go
kayrus: thanks, fixed

kayrus:
@jtopjian I think that there should be an indication, that the client received `Retry-After` code. Otherwise it will sleep silently and this can cause confusion from developer side. Thoughts?

provider_client.go
jtopjian: Yes, you're right - I misread the code.

jtopjian:
Yeah, I thought about this as well.

I'm thinking the best option is to not have a default func for this. Instead, document it in the FAQ, similar to how the round tripper is documented. This way, all implementations of Gophercloud that want to do retries have to implement their own function which can include their own logger, signals, and such.

kayrus:
@jtopjian ready for review

jtopjian (approved):
@kayrus Nice work. This looks good to me - we'll give it a shot and see how it goes :)

majewsky:
This should also be done for 498. Swift uses 498 instead of 429 for historical reasons.

kayrus:
@majewsky indeed, it uses [498 response code](https://github.com/openstack/swift/blob/833d7f1d1257de852c21f38b226b108739628ef6/swift/common/middleware/ratelimit.py#L293)

provider_client.go
mboruta1: @jtopjian Hello, I don't know if this is the correct place to post this but I believe your statement regarding `Retry-After` needing to be included in 429 responses is incorrect. If you have a look at the relevant section of the [RFC](https://datatracker.ietf.org/doc/html/rfc6585#section-4) it states:

```
   The response representations SHOULD include details explaining the
   condition, and MAY include a Retry-After header indicating how long
   to wait before making a new request.
```

Given this I believe it makes sense to, when the header is missing, sleep for a user-modifiable amount of time instead of simply treating the 429 request as an error.

jtopjian: @mboruta1 This is a closed pull request, so it's probably not the best place to discuss code. Are you running into a problem that this PR caused?

mboruta1: The Openstack cluster I run terraform against does not specify `Retry-After` in its 429 error response. I can change it to do so and thus enable this PR's retry-on-429 feature, but other users of Gopher cloud may not have that option (e.g. users of a public Openstack cloud). Thus I think it makes sense to not pre-condition this feature on the presence of a header that does not need to be present.

jtopjian: @mboruta1 I think you might be looking at outdated code. The code that was merged (https://github.com/gophercloud/gophercloud/pull/2097/files) doesn't explicitly look for a `Retry-After` header. It provides the ability for a developer to define their own `RetryBackoffFunc` function. This PR included documentation and unit tests that define example `RetryBackoffFunc` functions, but by default no function is defined and therefore no logic takes place.

mboruta1: Indeed, I suppose I should have commented on [this](https://github.com/gophercloud/utils/pull/144) instead which defines what appears to be the default `RetryBackoffFunc` function which, correct me if I am wrong, _does_ look for the `Retry-After` header. This is the same understanding that Terraform's Openstack Provider plugin has (see [this](https://registry.terraform.io/providers/terraform-provider-openstack/openstack/latest/docs#max_retries)). 

Instead of asking users to override the default `RetryBackoffFunc` when dealing with providers that do not provide the `Retry-After` header, this case could be handled by the default retry function itself.

jtopjian: @mboruta1 Gotcha. 

In my opinion, if there is at least one OpenStack service that does not respond with a `Retry-After` header along with a 429 response, the default `RetryBackoffFunc` function bundled in `gophercloud/utils` (and being used by Terraform) should account for that. Opening an Issue/PR at https://github.com/gophercloud/utils to track this would be the best way forward.

PR #2098, opened 2021-01-17
nova: support query hosts with availability-zone

TonyZhang1989:
For #2188 

nova: add support query hosts with availability-zone parameter

Filter the server list result by server availability zone.

This parameter is restricted to administrators until microversion 2.83. If non-admin users specify this parameter on a microversion less than 2.83, it will be ignored.

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -84,8 +84,8 @@ type ListOpts struct {
 	// NotTagsAny filters on specific server tags. At least one of the tags must be absent for the server.
 	NotTagsAny string `q:"not-tags-any"`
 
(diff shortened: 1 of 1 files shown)


EmilienM:
please squash the commits or rebase, so we have one single commit.

jtopjian:
@EmilienM Squashing can be done via GitHub. Multiple commits within PRs are actually preferred here since it's easier to see revisions.

However, per the [contributor's tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial), the following items need addressed in this PR:

1. A corresponding issue needs to be opened that describes the issue or feature.
2. The Nova code that implements this feature on the server-side.

Issue #2107, opened 2021-02-01
Object Store v1: DownloadOpts lost IfModifiedSince and IfUnmodifiedSince


toliu:
The `time.Time` carried by DownloadOpts is not added to the request header. I think this is needed when caching objects

jtopjian:
Fixed in #2108

PR #2108, opened 2021-02-01
swift: support modified-since for cache

toliu:
For https://github.com/gophercloud/gophercloud/issues/2107

API doc: 
https://docs.openstack.org/api-ref/object-store/?expanded=get-object-content-and-metadata-detail,show-container-details-and-list-objects-detail#get-object-content-and-metadata

Source code: https://github.com/openstack/swift/blob/ed990999170e7789dc6e32f299c0a3818636d9ea/swift/common/swob.py#L903-L904


jtopjian (approved):
LGTM - thank you!

Issue #2072, opened 2020-12-14
Add support for volume_type blockstorage quota


nikParasyr:
Before starting a PR, please read our [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial).

Block Storage api allows to set quota per volume type as shown [here](https://docs.openstack.org/api-ref/block-storage/v3/?expanded=update-quotas-for-a-project-detail#quota-sets-extension-os-quota-sets).

I'd like to work on this but i'm a bit unsure given the nature that the blockstorage api approaches this with `volumes_<volume_type>` etc and how to updateOpts are [set](https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/extensions/quotasets/requests.go#L62). Any suggestions or direction would be welcome. :)

jtopjian:
@nikParasyr I apologize for the late reply.

Yes, this looks like a tricky feature to implement. 

There are a few areas in Gophercloud where we use a utility function called [`RemainingKeys`](https://github.com/gophercloud/gophercloud/search?q=RemainingKeys). This function will bundle all JSON response fields that don't match the corresponding struct into a map. In the case of per-volume-type quotas, all returned fields such as `volumes_foo` could be bundled into a map. 

For `UpdateOpts`, we might have to add a field called `Extra` which will allow arbitrary fields to be added to the JSON body, similar to what's done [here](https://github.com/gophercloud/gophercloud/blob/f0b37a3eb4b4251a1041975345294c22fa083822/openstack/identity/v3/groups/requests.go#L147-L163).

I hope that helps. Please let me know if you have any questions.

jtopjian:
Implemented in #2109

PR #2109, opened 2021-02-01
Add support for blockstorage volume_type quota

nikParasyr:
For #2072 

Api docs:
- https://docs.openstack.org/api-ref/block-storage/v3/?expanded=update-quotas-for-a-project-detail#quota-sets-extension-os-quota-sets

code in this PR:
- https://github.com/openstack/cinder/blob/0fe910f130161edf5025d59020093ae163dfcba5/cinder/quota.py#L1157
- https://github.com/openstack/cinder/blob/0fe910f130161edf5025d59020093ae163dfcba5/cinder/quota.py#L1242
- https://github.com/openstack/cinder/blob/0fe910f130161edf5025d59020093ae163dfcba5/cinder/api/contrib/quotas.py#L31
- https://github.com/openstack/cinder/blob/0fe910f130161edf5025d59020093ae163dfcba5/cinder/api/contrib/quotas.py#L192-L275

### DIFF
--- a/openstack/blockstorage/extensions/quotasets/results.go
+++ b/openstack/blockstorage/extensions/quotasets/results.go
@@ -1,6 +1,8 @@
 package quotasets
 
 import (
+	"encoding/json"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
@@ -34,6 +36,36 @@ type QuotaSet struct {
 
 	// Groups is the number of groups that are allowed for each project.
 	Groups int `json:"groups,omitempty"`
+
+	// Extra is a collection of miscellaneous key/values used to set
+	// quota per volume_type
+	Extra map[string]interface{} `json:"-"`
+}
+
+// UnmarshalJSON is used on QuotaSet to unmarshal extra keys that are
+// used for volume_type quota
+func (r *QuotaSet) UnmarshalJSON(b []byte) error {
+	type tmp QuotaSet
+	var s struct {
+		tmp
+		Extra map[string]interface{} `json:"extra"`
+	}
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*r = QuotaSet(s.tmp)
+
+	var result interface{}
+	err = json.Unmarshal(b, &result)
+	if err != nil {
+		return err
+	}
+	if resultMap, ok := result.(map[string]interface{}); ok {
+		r.Extra = gophercloud.RemainingKeys(QuotaSet{}, resultMap)
+	}
+
+	return err
 }
 
 // QuotaUsageSet represents details of both operational limits of block
--- a/acceptance/openstack/blockstorage/v3/quotaset_test.go
+++ b/acceptance/openstack/blockstorage/v3/quotaset_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/extensions/quotasets"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumetypes"
 	th "github.com/gophercloud/gophercloud/testhelper"
 )
 
@@ -57,6 +58,9 @@ var UpdateQuotaOpts = quotasets.UpdateOpts{
 	Backups:            gophercloud.IntToPointer(2),
 	BackupGigabytes:    gophercloud.IntToPointer(300),
 	Groups:             gophercloud.IntToPointer(350),
+	Extra: map[string]interface{}{
+		"volumes_foo": gophercloud.IntToPointer(100),
+	},
 }
 
 var UpdatedQuotas = quotasets.QuotaSet{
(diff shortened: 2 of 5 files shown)


jtopjian:
recheck

jtopjian:
@nikParasyr Thanks for working on this. I haven't looked it over since it's still a draft, but wanted to mention that the acceptance test failure is an actual failure. You can see details about the failure by clicking on the `gophercloud-acceptance-test ` link above, and then going to the `job-output.txt.gz` file and searching for:

```
--- FAIL: TestQuotasetUpdate
```

Let me know if you have any questions or need any help. One common issue seen is that people will test on their own local OpenStack deployment which might act differently than a "vanilla" devstack environment. If this is your situation, it might be helpful to launch a vm, install devstack and go and work on the test from within that environment.

nikParasyr:
Hello @jtopjian. Thank you for your input. I'd like some more input and help if possible. The acceptance test fails for 2 reasons. I was aware of both of them from running it locally but i wanted to see the openlab result as well as get an opinion.
- Currently I have updated `TestQuotasetUpdate` to create a volume type foo in order to test that the patch works as intended. That isn't rly workable though because `Extra` is populated by volumes_/snapshots_/gigabytes_<volume_type_name> for every volumeType that exists on the testing cloud. So it doesn't only contain for the foo volume type but also for all existing volumeTypes. The volumeTypes differ from cloud to cloud. So hardcoding the expected Extra map won't work. Therefore to be able to test this properly we probably need to create `Extra` dynamically based on the existing volumeTypes found. The will sort of increase the complexity of the acceptance test but it's the only way i can think of to make it run properly on any testing cloud. Let me know if i should work towards that direction.
- Secondly, even when i hardcoded the Extra based on my testing cloud volume types I still get the following error which I'm a bit unsure how to handle. Same error occurs on openlab. It's a type error. Even more "weird" is that between runs I get different results. One run only gigabytes_foo is complaining, then all of them, then only volumes_foo etc. I'm probably missing something very basic here...
```
	=== RUN   TestQuotasetUpdate
	TestQuotasetUpdate: convenience.go:236: Failure in quotaset_test.go, line 122:at .Extra[gigabytes_foo] expected 100, but got 100
	TestQuotasetUpdate: convenience.go:236: Failure in quotaset_test.go, line 122:at .Extra[volumes_foo] expected 10, but got 10
	TestQuotasetUpdate: convenience.go:236: Failure in quotaset_test.go, line 122:at .Extra[snapshots_foo] expected 10, but got 10
	TestQuotasetUpdate: convenience.go:236: Failure in quotaset_test.go, line 122:at .Extra expected <nil>, but got map[string]interface {}{"gigabytes_foo":100, "snapshots_foo":10, "volumes_foo":10}
	TestQuotasetUpdate: convenience.go:35: Failure in quotaset_test.go, line 122: The structures were different.
	--- FAIL: TestQuotasetUpdate (0.88s)
	FAIL
```

jtopjian:
> Currently I have updated TestQuotasetUpdate to create a volume type foo in order to test that the patch works as intended. That isn't rly workable though because Extra is populated by volumes_/snapshots_/gigabytes_<volume_type_name> for every volumeType that exists on the testing cloud.

This is OK. Take a look at the other acceptance tests for different examples. A lot of them contain a pattern that defines a variable called `found` and then only checks for a specific resource, ignoring other resources. This is to account for existing resources in the cloud.

> Secondly, even when i hardcoded the Extra based on my testing cloud volume types I still get the following error which I'm a bit unsure how to handle. 

At first glance, this looks like some kind of type assertion issue. I would double-check the type that is being marshaled into "Extra" and make sure it's being compared to a type of the same kind.

I hope this helps, but let me know if you need more details.

nikParasyr:
@jtopjian Thanks for all the help. This is ready for a first review.
Tried to follow the contributing guide as much as possible but I might have missed some spots.

nikParasyr:
recheck

nikParasyr:
recheck

nikParasyr:
recheck

jtopjian:
@nikParasyr There might be an issue with OpenLab. I recommend trying again in a few hours. I'll try to get some time to review this either today or tomorrow.

nikParasyr:
recheck

openstack/blockstorage/extensions/quotasets/results.go
```diff
@@ -34,6 +36,42 @@ type QuotaSet struct {
...
+	}
+	err := json.Unmarshal(b, &s)
+	if err != nil {
+		return err
+	}
+	*r = QuotaSet(s.tmp)
+
+	// Collect other fields and bundle them into Extra
+	// but only if a field titled "extra" wasn't sent.
+	if s.Extra != nil {
```
jtopjian, line 61: This is specific to the Keystone/Identity API and can be removed here. The body of the else clause should only be used.

jtopjian:
recheck

openstack/blockstorage/extensions/quotasets/results.go
nikParasyr: done

nikParasyr:
hmmmm openlab is consistently failing. maybe [this](https://github.com/theopenlab/openlab/issues/674) is related? not the occm stuff but openlab upgrading to focal? ill stop rechecking for the time being

jtopjian:
Yes, it's possible. I've subscribed to the issue and will keep an eye on it.

In the meantime, I have run the tests in my own devstack environment and everything looks good to me. If you're finished with this, I can go ahead and merge it.

nikParasyr:
Yes. I've finished with this one. Thank you for all the help :)

jtopjian (approved):
LGTM - thank you!

jtopjian:
@nikParasyr Thank you for submitting this - this was really good work.

Issue #2110, opened 2021-02-02
Blockstorage: Add support for "retype" volume action


nikParasyr:
Blockstorage api allows changing the volume_type of a volume via:

https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=retype-a-volume-detail#volume-actions-volumes-action

I'll work on this later this (next) week

jtopjian:
Implemented in #2113

PR #2113, opened 2021-02-04
Add support for os-retype volumeaction

nikParasyr:
For #2110 

Add support for the "os-retype" volumeaction that changes the volume_type of a volume.

The lowe-level implementation of the os-retype is driver specific on cinder. The links below are blobs on the api level.
- [volumeaction](https://github.com/openstack/cinder/blob/876ac4e79fe14887ee3902ef833be5e2273d800b/cinder/api/contrib/volume_actions.py#L293-L303)
- [schema](https://github.com/openstack/cinder/blob/876ac4e79fe14887ee3902ef833be5e2273d800b/cinder/api/schemas/volume_actions.py#L87-L104)
- [volume-api](https://github.com/openstack/cinder/blob/876ac4e79fe14887ee3902ef833be5e2273d800b/cinder/volume/api.py#L1633)

### DIFF
--- a/acceptance/openstack/blockstorage/v3/blockstorage.go
+++ b/acceptance/openstack/blockstorage/v3/blockstorage.go
@@ -79,6 +79,42 @@ func CreateVolume(t *testing.T, client *gophercloud.ServiceClient) (*volumes.Vol
 	return volume, nil
 }
 
+// CreateVolumeWithType will create a volume of the given volume type
+// with a random name and size of 1GB. An error will be returned if
+// the volume was unable to be created.
+func CreateVolumeWithType(t *testing.T, client *gophercloud.ServiceClient, vt *volumetypes.VolumeType) (*volumes.Volume, error) {
+	volumeName := tools.RandomString("ACPTTEST", 16)
+	volumeDescription := tools.RandomString("ACPTTEST-DESC", 16)
+	t.Logf("Attempting to create volume: %s", volumeName)
+
+	createOpts := volumes.CreateOpts{
+		Size:        1,
+		Name:        volumeName,
+		Description: volumeDescription,
+		VolumeType:  vt.Name,
+	}
+
+	volume, err := volumes.Create(client, createOpts).Extract()
+	if err != nil {
+		return volume, err
+	}
+
+	err = volumes.WaitForStatus(client, volume.ID, "available", 60)
+	if err != nil {
+		return volume, err
+	}
+
+	tools.PrintResource(t, volume)
+	th.AssertEquals(t, volume.Name, volumeName)
+	th.AssertEquals(t, volume.Description, volumeDescription)
+	th.AssertEquals(t, volume.Size, 1)
+	th.AssertEquals(t, volume.VolumeType, vt.Name)
+
+	t.Logf("Successfully created volume: %s", volume.ID)
+
+	return volume, nil
+}
+
 // CreateVolumeType will create a volume type with a random name. An
 // error will be returned if the volume was unable to be created.
 func CreateVolumeType(t *testing.T, client *gophercloud.ServiceClient) (*volumetypes.VolumeType, error) {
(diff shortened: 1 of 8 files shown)


acceptance/openstack/blockstorage/v3/blockstorage.go
```diff
@@ -79,6 +79,42 @@ func CreateVolume(t *testing.T, client *gophercloud.ServiceClient) (*volumes.Vol
 	return volume, nil
 }
 
+// CreateVolumeWithType will create a volume of the given volume type
+// with a random name and size of 1GB. An error will be returned if
+// the volume was unable to be created.
+func CreateVolumeWithType(t *testing.T, client *gophercloud.ServiceClient, vt *volumetypes.VolumeType) (*volumes.Volume, error) {
```
nikParasyr, line 85: This could be incorporated into `CreateVolume` but i opted into creating a new func for this. Let me know if its better to incorporate it instead.

acceptance/openstack/blockstorage/v3/blockstorage.go
```diff
@@ -110,6 +146,36 @@ func CreateVolumeType(t *testing.T, client *gophercloud.ServiceClient) (*volumet
 	return vt, nil
 }
 
+// CreateVolumeTypeNoExtraSpecs will create a volume type with a random name and
+// no extra specs. This is required to bypass cinder-scheduler filters and be able
+// to create a volume with this volumeType. An error will be returned if the volume
+// type was unable to be created.
+func CreateVolumeTypeNoExtraSpecs(t *testing.T, client *gophercloud.ServiceClient) (*volumetypes.VolumeType, error) {
```
nikParasyr, line 153: VolumeType needs to have empty extra specs in order to be able to create a volume with this type. Otherwise the cinder-scheduler and filterOpts get in the middle and the volume cannot be created. Currently on `CreateVolumeType` the extra specs are not tested due to some limitations, so it could be incorporated but i'm not sure if its the right move.

acceptance/openstack/blockstorage/v3/blockstorage.go
```diff
@@ -145,6 +211,20 @@ func DeleteVolume(t *testing.T, client *gophercloud.ServiceClient, volume *volum
 		t.Fatalf("Unable to delete volume %s: %v", volume.ID, err)
 	}
 
+	// VolumeTypes can't be deleted until their volumes have been,
```
nikParasyr, line 214: This change affects all tests that use DeleteVolume, while its only sort of needed when extra resources depend on the volume such as a volumeType. Let me know if a seperate DeleteVolumeAndWait func is better

nikParasyr (commented):
@jtopjian I've added some comments/questions I have. Also the acceptance tests cover only blockstorage V3. The action is valid for V2 as well. let me know if you want me to add acceptance tests for V2 as well.

openstack/blockstorage/extensions/volumeactions/requests.go
```diff
@@ -343,3 +343,42 @@ func SetBootable(client *gophercloud.ServiceClient, id string, opts BootableOpts
...
+// ChangeTypeOpts contains options for changing the type of an existing Volume.
+// This object is passed to the volumes.ChangeType function.
+type ChangeTypeOpts struct {
+	// NewType is the name of the new volume type of the volume.
+	NewType string `json:"new_type" required:"true"`
+
+	// MigrationPolicy specifies if the volume should be migrated when it is
+	// re-typed. Possible values are "on-demand" or "never". If not specified,
+	// the default is "never".
+	MigrationPolicy string `json:"migration_policy,omitempty"`
```
jtopjian, line 362: It looks like there are only two possible values for `MigrationPolicy`: https://github.com/openstack/cinder/blob/876ac4e79fe14887ee3902ef833be5e2273d800b/cinder/api/schemas/volume_actions.py#L94-L96

Using a generic `string` is fine, but it's also possible to create a specific type for this like how it's done here:

* https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/lbaas_v2/listeners/requests.go#L8-L17
* https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/lbaas_v2/listeners/requests.go#L85

If the migration policy choices rarely change, creating a type is usually a good method for more specific data modeling.

acceptance/openstack/blockstorage/v3/blockstorage.go
jtopjian: A new function is a good idea. This keeps the testing functions specific to only one type of resource.

acceptance/openstack/blockstorage/v3/blockstorage.go
jtopjian: I think the limitations have to do with the OpenLab testing environment? If so it's OK to pass on doing that kind of test. As long as functionality was tested some other way - such as a local devstack environment.

acceptance/openstack/blockstorage/v3/blockstorage.go
jtopjian: This is a good addition. Thank you!

acceptance/openstack/blockstorage/v3/blockstorage.go
nikParasyr: In my understanding, if you set extra_specs then the cinder-scheduler tries to map the volumeType to a back-end that matches these extra_specs. That behavior can be different per cloud depending on how the cinder-scheduler has been configured (which [filters](https://opendev.org/openstack/cinder/src/branch/master/cinder/scheduler/filters) to use etc). I'd still say that the acceptance test is useful and i expect without extra_specs to avoid all the cloud-specific limitations of how the backens and cinder-scheduler is set-up.

I tested this on our test cluster and the acceptance passed (volume created properly, and type changed without an issue etc). I've also run a retype, basically the example on docs.go, between 2 valid volume_types that also went fine. I'll do the same with an in-use volume which should also be working

openstack/blockstorage/extensions/volumeactions/requests.go
nikParasyr: Done :)

nikParasyr:
recheck

jtopjian:
@nikParasyr Is this ready for review/merge?

nikParasyr:
Yes

jtopjian (approved):
LGTM - Thank you!

PR #2116, opened 2021-02-05
Change Microversion to 2.53

lrainsun:
When list nova services, we want to get uuid instead of id. It is
supported > 2.53
The id of the service as a uuid.
New in version 2.53

For #[PUT ISSUE NUMBER HERE]

code in this PR:

[PUT URLS HERE]

### DIFF
--- a/openstack/compute/v2/extensions/services/requests.go
+++ b/openstack/compute/v2/extensions/services/requests.go
@@ -25,6 +25,7 @@ func (opts ListOpts) ToServicesListQuery() (string, error) {
 
 // List makes a request against the API to list services.
 func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pager {
+	client.Microversion = "2.53"
 	url := listURL(client)
 	if opts != nil {
 		query, err := opts.ToServicesListQuery()


kayrus:
@lrainsun it's better to extend the json unmarshaller, otherwise your fix will break the logic for older openstack API revisions.

https://github.com/gophercloud/gophercloud/blob/633d73521055b7b9dfa92da51e6982c398683d58/openstack/compute/v2/extensions/services/results.go#L44-L73

lrainsun:
> @lrainsun it's better to extend the json unmarshaller, otherwise your fix will break the logic for older openstack API revisions.
> 
> https://github.com/gophercloud/gophercloud/blob/633d73521055b7b9dfa92da51e6982c398683d58/openstack/compute/v2/extensions/services/results.go#L44-L73

Hi kayrus, 
Thanks for reply
I'm not quite familiar with the code
What I want is using microversion >= 2.53 to do nova list serivces
Cause in our situation we want to get the service uuid ot service id
So, where is the proper place that I can config the microversion when listing nova services?

kayrus:
> What I want is using microversion >= 2.53 to do nova list serivces

You can define a custom microversion in your application. See https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md

lrainsun:
> > What I want is using microversion >= 2.53 to do nova list serivces
> 
> You can define a custom microversion in your application. See https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md

Got u, thanks

jtopjian:
@lrainsun I'm going to close this PR but let us know if you have any questions.

Issue #2119, opened 2021-02-16
Baremetal: support deploy_steps for the provisioning API (microversion 1.69)


dtantsur:
A new addition in the Wallaby release: `/v1/node/<node>/states/provision` now supports a `deploy_steps` argument.

Deploy steps explained: https://docs.openstack.org/ironic/latest/admin/node-deployment.html#deploy-steps
New API addition: https://docs.openstack.org/ironic/latest/contributor/webapi-version-history.html#wallaby-16-2

jtopjian:
Implemented in #2120

PR #2120, opened 2021-02-16
Baremetal: support deploy_steps for the provisioning API

dtantsur:
For: #2119

### DIFF
--- a/openstack/baremetal/v1/nodes/requests.go
+++ b/openstack/baremetal/v1/nodes/requests.go
@@ -394,13 +394,34 @@ func GetSupportedBootDevices(client *gophercloud.ServiceClient, id string) (r Su
 	return
 }
 
+// An interface type for a deploy (or clean) step.
+type StepInterface string
+
+const (
+	InterfaceBIOS       StepInterface = "bios"
+	InterfaceDeploy     StepInterface = "deploy"
+	InterfaceManagement StepInterface = "management"
+	InterfacePower      StepInterface = "power"
+	InterfaceRAID       StepInterface = "raid"
+)
+
 // A cleaning step has required keys ‘interface’ and ‘step’, and optional key ‘args’. If specified,
 // the value for ‘args’ is a keyword variable argument dictionary that is passed to the cleaning step
 // method.
 type CleanStep struct {
-	Interface string                 `json:"interface" required:"true"`
+	Interface StepInterface          `json:"interface" required:"true"`
 	Step      string                 `json:"step" required:"true"`
 	Args      map[string]interface{} `json:"args,omitempty"`
+}
+
+// A deploy step has required keys ‘interface’, ‘step’, ’args’ and ’priority’.
+// The value for ‘args’ is a keyword variable argument dictionary that is passed to the deploy step
+// method. Priority is a numeric priority at which the step is running.
+type DeployStep struct {
+	Interface StepInterface          `json:"interface" required:"true"`
+	Step      string                 `json:"step" required:"true"`
+	Args      map[string]interface{} `json:"args" required:"true"`
+	Priority  int                    `json:"priority" required:"true"`
 }
 
 // ProvisionStateOptsBuilder allows extensions to add additional parameters to the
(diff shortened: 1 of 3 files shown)


elfosardo (approved):

jtopjian:
@dtantsur I took a stab at doing the [code hunting](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md) requirement and came up with:

* https://github.com/openstack/ironic/blob/766d8f11b4101570829788ed11b2ab3fef85e3f8/ironic/conductor/steps.py#L234
* https://github.com/openstack/ironic/blob/766d8f11b4101570829788ed11b2ab3fef85e3f8/ironic/conductor/steps.py#L264
* https://github.com/openstack/ironic/blob/766d8f11b4101570829788ed11b2ab3fef85e3f8/ironic/conductor/steps.py#L600
* https://github.com/openstack/ironic/blob/766d8f11b4101570829788ed11b2ab3fef85e3f8/ironic/conductor/steps.py#L538-L554

If this is correct, please let me know if this is ready for review.

dtantsur:
HI @jtopjian! Yeah, you're on the right track. This feature was added in https://opendev.org/openstack/ironic/commit/3138acc836c933aa179b9a92c14a226a8b903fe5, its API part starts in https://opendev.org/openstack/ironic/src/commit/6ae5bc4642c4619946a48512501b4ae37246f7ff/ironic/api/controllers/v1/node.py#L850 and progresses further ending up in the code you've found. Note that there is a related concept of deploy templates which makes code hunting somewhere more complicated. This change does not implement deploy templates, only the smaller API addition from the above mentioned commit.

dtantsur:
Tested with the following code:
```go
    client.Microversion = "1.69"

    updates := nodes.UpdateOpts {
        nodes.UpdateOperation {
            Op: nodes.AddOp,
            Path: "/instance_info/image_source",
            Value: "file:///httpboot/centos8.qcow2",
        },
    }

    _, err = nodes.Update(client, "testvm1", updates).Extract()
    if err != nil {
        panic(fmt.Sprintf("Cannot update: %s", err))
    }

    files := []struct {
        Path string `json:"path"`
        Content string `json:"content"`
    } {
        {
            Path: "/etc/motd",
            Content: "SGVsbG8gSXJvbmljIFdvcmxkIQo=",
        },
    }

    opts := nodes.ProvisionStateOpts {
        Target: nodes.TargetActive,
        DeploySteps: []nodes.DeployStep {
            nodes.DeployStep {
                Step: "inject_files",
                Interface: "deploy",
                Priority: 50,
                Args: map[string]interface{}{
                    "files": files,
                },
            },
        },
    }

    err = nodes.ChangeProvisionState(client, "testvm1", opts).ExtractErr()
    if err != nil {
        panic(fmt.Sprintf("Cannot provision: %s", err))
    }
```

(Note that I'm using a deploy step that [has not merged yet](https://review.opendev.org/c/openstack/ironic-python-agent/+/775428))

jtopjian:
@dtantsur Thanks! Those links helped. Notably:

https://opendev.org/openstack/ironic/src/commit/3138acc836c933aa179b9a92c14a226a8b903fe5/ironic/api/controllers/v1/utils.py#L125-L141

which defines the data model being added in this PR. I just needed to see something that showed the fields of the `DeployStep` struct are defined somewhere and `DeployStep` shouldn't be a generic `map[string]interface`.

It looks like the `Interface` field is an [enum](https://opendev.org/openstack/ironic/src/commit/3138acc836c933aa179b9a92c14a226a8b903fe5/ironic/api/controllers/v1/utils.py#L129-L132), with values defined [here](https://github.com/openstack/ironic/blob/766d8f11b4101570829788ed11b2ab3fef85e3f8/ironic/conductor/steps.py#L39-L51) (sorry for linking to GitHub, I wasn't getting any results when using the "search" function in opendev). 

If you wanted, you could create a custom string type called `DeployStepInterfaceType` (or something similar). It would then look like:

```go

type DeployStepInterfaceType string

const (
  InterfaceDeploy DeployStepInterfaceType = "deploy"
  ...
)

    ....

    opts := nodes.ProvisionStateOpts {
        Target: nodes.TargetActive,
        DeploySteps: []nodes.DeployStep {
            nodes.DeployStep {
                Step: "inject_files",
                Interface: nodes.InterfaceDeploy,
                Priority: 50,
                Args: map[string]interface{}{
                    "files": files,
                },
            },
        },
    }
```

This is up to you, though. There are areas in Gophercloud that implement this and other areas that use a generic "string". If you prefer to use a `string`, let me know and I'll go ahead and merge this PR as-is.

dtantsur:
> It looks like the Interface field is an enum, with values defined here

Good call, I completely forgot that we limit the interfaces supported for deploy steps (I also forgot *why* we do that, but that's a different conversation). Will change.

dtantsur:
@jtopjian updated and applied the same change to clean steps.

EmilienM (approved):
:+1 LGTM

jtopjian (approved):
LGTM - thank you!

Issue #2121, opened 2021-02-22
Baremetal: Add automated_clean field to Ironic node


fmuyassarov:
I would like to add Ironic `Node.automated_clean` field which is currently missing in https://github.com/gophercloud/gophercloud/tree/master/openstack/baremetal/v1/nodes

Reference to the field in Ironic: [automated_clean](https://github.com/openstack/ironic/blob/8604f84fd7bda4e30d3f07005c4901f3662303a7/ironic/objects/node.py#L145)

fmuyassarov:
Seems the issue didn't get closed automatically. Shall we close it now @jtopjian ?

jtopjian:
Implemented in #2122

jtopjian:
@fmuyassarov Yes, thank you 🙂

PR #2122, opened 2021-02-22
Baremetal: add nodes.automated_clean

fmuyassarov:
For https://github.com/gophercloud/gophercloud/issues/2121

### DIFF
--- a/openstack/baremetal/v1/nodes/requests.go
+++ b/openstack/baremetal/v1/nodes/requests.go
@@ -181,6 +181,10 @@ type CreateOptsBuilder interface {
 
 // CreateOpts specifies node creation parameters.
 type CreateOpts struct {
+	// The interface to configure automated cleaning for a Node.
+	// Requires microversion 1.47 or later.
+	AutomatedClean *bool `json:"automated_clean,omitempty"`
+
 	// The boot interface for a Node, e.g. “pxe”.
 	BootInterface string `json:"boot_interface,omitempty"`
 
--- a/openstack/baremetal/v1/nodes/results.go
+++ b/openstack/baremetal/v1/nodes/results.go
@@ -50,6 +50,10 @@ func ExtractNodesInto(r pagination.Page, v interface{}) error {
 
 // Node represents a node in the OpenStack Bare Metal API.
 type Node struct {
+	// Whether automated cleaning is enabled or disabled on this node.
+	// Requires microversion 1.47 or later.
+	AutomatedClean *bool `json:"automated_clean"`
+
 	// UUID for the resource.
 	UUID string `json:"uuid"`
 
--- a/openstack/baremetal/v1/nodes/testing/fixtures.go
+++ b/openstack/baremetal/v1/nodes/testing/fixtures.go
@@ -74,6 +74,7 @@ const NodeListDetailBody = `
  {
   "nodes": [
     {
+      "automated_clean": null,
       "bios_interface": "no-bios",
       "boot_interface": "pxe",
       "chassis_uuid": null,
@@ -178,6 +179,7 @@ const NodeListDetailBody = `
       ]
     },
     {
+      "automated_clean": null,
       "bios_interface": "no-bios",
       "boot_interface": "pxe",
       "chassis_uuid": null,
@@ -274,6 +276,7 @@ const NodeListDetailBody = `
       ]
     },
     {
+      "automated_clean": null,
       "bios_interface": "no-bios",
       "boot_interface": "pxe",
       "chassis_uuid": null,
@@ -376,6 +379,7 @@ const NodeListDetailBody = `
 // SingleNodeBody is the canned body of a Get request on an existing node.
 const SingleNodeBody = `
 {
+  "automated_clean": null,
   "bios_interface": "no-bios",
   "boot_interface": "pxe",
   "chassis_uuid": null,


openstack/baremetal/v1/nodes/results.go
```diff
@@ -50,6 +50,9 @@ func ExtractNodesInto(r pagination.Page, v interface{}) error {
 
 // Node represents a node in the OpenStack Bare Metal API.
 type Node struct {
+	// Whether automated cleaning is enabled or disabled on this node.
+	AutomatedClean bool `json:"automated_clean"`
```
dtantsur, line 54: The field will be missing on old microversions, you need to avoid defaulting to `false`.

openstack/baremetal/v1/nodes/requests.go
```diff
@@ -181,6 +181,10 @@ type CreateOptsBuilder interface {
 
 // CreateOpts specifies node creation parameters.
 type CreateOpts struct {
+
+	// The interface to configure automated cleaning for a Node
+	AutomatedClean bool `json:"automated_clean,omitempty"`
```
dtantsur, line 186: This will default to `false` I assume? You need to allow explicit `true` and `false`, skipping completely if no value is provided (so `*bool` I guess).

dtantsur (changes requested):

openstack/baremetal/v1/nodes/results.go
fmuyassarov: @dtantsur 
I'm not sure how to enforce the default value here, but maybe we could have
```
AutomatedClean *bool `json:"automated_clean"
```
and set it to `true` if we see nil value? I assume this would need to be done in Ironic code :(

dtantsur: If you use `*bool` as the type then `omitempty` should work (`nil` = empty if I understand Go right).

dtantsur: Oh, wait, my comment was about a request. I'm not sure, I guess we can live with `nil` if ironic doesn't return this field? You need to ask someone who knows more about Go, to be honest.

jtopjian: If `AutomatedClean` is a field that was added from a microversion, then please use `*bool`. This way the app/developer can check if it is `nil` or not. It's a little hacky, but the most sane solution I could come up with.

Also, please include a docstring that details what microversion the field was added in.

And let me know if anyone has any questions :) 

https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md#new-response-fields

openstack/baremetal/v1/nodes/requests.go
jtopjian: `*bool` and `omitempty` should do it. And please also add which microversion this was added as part of the docstring/comment.

https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md#new-response-fields

openstack/baremetal/v1/nodes/results.go
fmuyassarov: Thank you @dtantsur and @jtopjian . Fixed.

openstack/baremetal/v1/nodes/requests.go
fmuyassarov: Fixed, thanks.

fmuyassarov:
@jtopjian , @dtantsur PTAL, addressed the comments above. Thanks.

openstack/baremetal/v1/nodes/requests.go
```diff
@@ -181,6 +181,10 @@ type CreateOptsBuilder interface {
 
 // CreateOpts specifies node creation parameters.
 type CreateOpts struct {
+
+	// The interface to configure automated cleaning for a Node
+	AutomatedClean *bool `json:"automated_clean,omitempty"`
```
jtopjian, line 186: Format nit: Please remove the blank line on line 184.

Can you also add a comment to note which microversion introduced this field? As an exmaple:

https://github.com/gophercloud/gophercloud/blob/917735ee91e24fe1493e57869c3b42ee89bc95d8/openstack/compute/v2/servers/requests.go#L215-L217

openstack/baremetal/v1/nodes/results.go
```diff
@@ -50,6 +50,9 @@ func ExtractNodesInto(r pagination.Page, v interface{}) error {
 
 // Node represents a node in the OpenStack Bare Metal API.
 type Node struct {
+	// Whether automated cleaning is enabled or disabled on this node.
+	AutomatedClean *bool `json:"automated_clean"`
```
jtopjian, line 54: Similar here, can you add a comment to note which microversion this field was added in? As an example:

https://github.com/gophercloud/gophercloud/blob/ed4deec00ff1d4d4c8a762af0c6360d4184a4bf4/openstack/compute/v2/servers/results.go#L221-L223

jtopjian (changes requested):
@fmuyassarov I've noted three small changes. Please let me know if you have any questions.

openstack/baremetal/v1/nodes/requests.go
fmuyassarov: @jtopjian thanks, fixed it.

openstack/baremetal/v1/nodes/results.go
fmuyassarov: Fixed, thank you.

openstack/baremetal/v1/nodes/requests.go
dtantsur: Honestly, half of these fields were introduced in a microversion. Somebody should go through https://docs.openstack.org/ironic/latest/contributor/webapi-version-history.html and add these eventually.

/me uses this comment to (once again) express his disagreement with how gophercloud handles microversions

dtantsur (approved):

openstack/baremetal/v1/nodes/requests.go
fmuyassarov: > Honestly, half of these fields were introduced in a microversion. Somebody should go through https://docs.openstack.org/ironic/latest/contributor/webapi-version-history.html and add these eventually.
>

I can try to work on this, but can't tell the exact date right now.

dtantsur: Only if you have spare cycles to spend. Otherwise I'll probably get to it (no ETA as well).

fmuyassarov: Sure, will let you know let's say somewhere during next week.

jtopjian: > I'd like to use this comment to (once again) express his disagreement with how gophercloud handles microversions. They have to be set automatically, not like this:

I would be _more_ than happy to discuss a change in implementation. In fact, I'd appreciate the help.

jtopjian (approved):
LGTM - thank you!

openstack/baremetal/v1/nodes/requests.go
dtantsur: Not sure which venue we should use, but the vision of microversions by the OpenStack API SIG is that the microversions should not be exposed to consumers of all but the most basic and low-level SDKs. Back at the times I wrote https://specs.openstack.org/openstack/api-sig/guidelines/sdk-exposing-microversions.html.

Let's take this patch as an example. If `client.Microversion` is empty, we provide the minimum microversion we support. But if `automated_clean` is not `nil`, we understand that this request can only be executed with API version `1.47` and automatically use it (consumers can still override with by setting `client.Microversion` explicitly).

When retrieving anything, the situation is a bit more interesting. You need to figure out the maximum version that both the server and gophercloud understand. In openstacksdk, for example, we set the latter per resource: https://github.com/openstack/openstacksdk/blob/master/openstack/baremetal/v1/node.py#L95. This procedure allows exposing maximum fields without the users guessing the right microversion. Instead they can rely on the fact that e.g. `automated_clean` is not `nil`.

I hope it all makes sense, I'm open to a more detailed/broad conversation since microversions happened to be more specialization (even though I personally dislike them :) )

jtopjian: Thank you for the reply - it's very helpful.

> Not sure which venue we should use

If there's substantial discussion that won't be resolved in a few comments, then a new Issue would be best.

It sounds like there isn't a problem with the technical details of getting microversions to work. To be honest, that's my biggest priority/concern.

I understand the implementation you've described and I agree with it. 

Gophercloud attempts to be, basically, one step above raw HTTP requests. It doesn't try to do anything intelligent unless it's absolutely needed. There isn't a technical reason for this or even an ideological reason. It's really just an issue of available hands and eyes to work on Gophercloud. The amount of volunteer effort available right now is just enough for "one step above raw HTTP requests" and an occasional feature at https://github.com/gophercloud/utils. It'd be awesome if Gophercloud had feature parity with openstacksdk, but it's not possible unless more people work on it regularly. 

Again, I do agree with the implementation you described, but it would take a good amount of work. It would take more effort to also make sure Gophercloud fully supported a microversion to the point where I'd feel comfortable tagging a certain Gophercloud release with an officially supported microversion (the concern largely comes from making sure all OpenStack projects/services within Gophercloud have this level of support and one service doesn't get preferential treatment).

This is why anything "two steps" above a raw HTTP requests is punted up the stack. I don't like doing it, but it's better than 2-3 years ago when microversioned features were blocked entirely while trying to figure out a clean solution. If someone would argue that not doing it the ideal way is worse than no support at all, I wouldn't disagree with them, but I'd ask them to field the feature requests that come in :)

Yeah, I realize I'm giving the "not enough time" excuse. But if people began submitting patches to change how Gophercloud works with microversions, I would absolutely be open to merging them.

dtantsur: > Gophercloud had feature parity with openstacksdk, but it's not possible unless more people work on it regularly

Sigh, don't get me started on how many people work full time on openstacksdk (zero)..

> I realize I'm giving the "not enough time" excuse

And I understand you very well! Where is my human cloning machine?

> Gophercloud attempts to be, basically, one step above raw HTTP requests

There is one problem though: microversions are supposed to be set per request, not per client. It seems that Gophercloud does the latter though. A low-level SDK should do the former. Not sure how ugly it will make the API though. In rust-osauth I ended up with an explicit argument: https://docs.rs/osauth/0.3.4/osauth/struct.Session.html#method.request (but rust-osauth is actually one stop closer to the wire).

jtopjian: >  microversions are supposed to be set per request, not per client. It seems that Gophercloud does the latter though

I can't speak for everything that has used Gophercloud+microversions, so it's best to assume that there are cases where this is being done.

However, it's possible to do it on a per-request basis above Gophercloud right now. This is how it's been implemented in Terraform. Some examples are:

* https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/53f886ad12a1abcce778d535b101a91d06911a7c/openstack/resource_openstack_sharedfilesystem_share_access_v2.go#L88-L94
* https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/53f886ad12a1abcce778d535b101a91d06911a7c/openstack/resource_openstack_sharedfilesystem_share_access_v2.go#L153-L170
* https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/b1f4529d38c42062b39a43e006d728d7a890945e/openstack/resource_openstack_lb_quota_v2.go#L114-L124

These examples could be tightened up to where the microversion is reverted after the individual request, but Terraform's Create, Read, Update, and Delete functions can be thought of as a single request to perform one action.

I do understand that the logic being done in Terraform _should_ be done in Gophercloud. Again, if someone proposed a way to redesign and enable this, I'm open to merging it.

All API calls in Gophercloud have the ability add additional headers, so it should be possible to tuck all of this logic away from the user. For example: https://github.com/gophercloud/gophercloud/blob/841705fad1df2f50b23c31ee60f51702063be7c6/openstack/imageservice/v2/imagedata/requests.go#L12

PR #2126, opened 2021-03-11
Neutron quota: int-as-string workaround

### LINKED ISSUE #2125: Tolerate a Neutron bug where quota values may be returned as strings
When its value is greater than zero, Neutron is likely to return the numerical "reserved" property as a string instead of an integer.

In Gophercloud, this event produces a fatal error:

```
json: cannot unmarshal string into Go struct field QuotaDetail.quota.network.reserved of type int
```

cf https://bugs.launchpad.net/neutron/+bug/1918565

pierreprinetti:
In some conditions, Neutron will JSON-encode quota values as strings
instead of integers.

With this patch, the unmarshaling method is overridden to accept
integers as strings.

See https://bugs.launchpad.net/neutron/+bug/1918565

Fixes #2125

### DIFF
--- a/openstack/networking/v2/extensions/quotas/results.go
+++ b/openstack/networking/v2/extensions/quotas/results.go
@@ -1,6 +1,12 @@
 package quotas
 
-import "github.com/gophercloud/gophercloud"
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/gophercloud/gophercloud"
+)
 
 type commonResult struct {
 	gophercloud.Result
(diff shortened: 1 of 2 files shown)


openstack/networking/v2/extensions/quotas/results.go
```diff
@@ -128,3 +133,38 @@ type QuotaDetail struct {
...
+// Due to a bug in Neutron, under some conditions Reserved is returned as a
+// string.
+//
+// This method is left for compatibility with unpatched versions of Neutron.
+//
+// cf. https://bugs.launchpad.net/neutron/+bug/1918565
+func (q *QuotaDetail) UnmarshalJSON(b []byte) error {
+	var quotaDetail struct {
+		Used     int             `json:"used"`
+		Reserved json.RawMessage `json:"reserved"`
```
kayrus, line 149: it's better to use interface to align with the rest gophercloud style:
see a https://github.com/gophercloud/gophercloud/blob/cd9c207e93f4f76af2c0a06c6d449ab342bfbe56/openstack/compute/v2/extensions/hypervisors/results.go#L39 for example

EmilienM:
great, I'll go with that until this is fixed in Neutron and backported. I suspect we'll carry this workaround for some time...
Thanks!

jtopjian:
@pierreprinetti This looks good to me. To confirm: this is ready to be merged?

pierreprinetti:
> @pierreprinetti This looks good to me. To confirm: this is ready to be merged?

As far as I can tell, the patch is ready.

EmilienM:
@jtopjian LGTM, this is ready to go I think.

Thanks Pierre for the quick workaround!

jtopjian:
Thanks, all! I just like to make sure before I accidentally hit the merge button too early 🙂

openstack/networking/v2/extensions/quotas/results.go
jtopjian: @kayrus Thanks!

jtopjian (approved):
LGTM - thank you!

EmilienM:
thanks @jtopjian for the ninja review!

pierreprinetti:
Thank you @jtopjian!

PR #2131, opened 2021-03-18
Compute: Handle Hypervisor request parameters

ntk148v:
code in this PR:

https://github.com/openstack/nova/blob/stable/victoria/nova/api/openstack/compute/hypervisors.py#L89

This field has become mandatory in microversion 2.75. If no servers is on hypervisor then empty list is returned.

Signed-off-by: Kien Nguyen <kiennt2609@gmail.com>

### DIFF
--- a/openstack/compute/v2/extensions/hypervisors/requests.go
+++ b/openstack/compute/v2/extensions/hypervisors/requests.go
@@ -5,9 +5,52 @@ import (
 	"github.com/gophercloud/gophercloud/pagination"
 )
 
+// ListOptsBuilder allows extensions to add additional parameters to the
+// List request.
+type ListOptsBuilder interface {
+	ToHypervisorListQuery() (string, error)
+}
+
+// ListOpts allows the filtering and sorting of paginated collections through
+// the API. Filtering is achieved by passing in struct field values that map to
+// the server attributes you want to see returned. Marker and Limit are used
+// for pagination.
+type ListOpts struct {
+	// Limit is an integer value for the limit of values to return.
+	// This requires microversion 2.33 or later.
+	Limit *int `q:"limit"`
+
+	// Marker is the ID of the last-seen item as a UUID.
+	// This requires microversion 2.53 or later.
+	Marker *string `q:"marker"`
+
+	// HypervisorHostnamePattern is the hypervisor hostname or a portion of it.
+	// This requires microversion 2.53 or later
+	HypervisorHostnamePattern *string `q:"hypervisor_hostname_pattern"`
+
+	// WithServers is a bool to include all servers which belong to each hypervisor
+	// This requires microversion 2.53 or later
+	WithServers *bool `q:"with_servers"`
+}
+
+// ToHypervisorListQuery formats a ListOpts into a query string.
+func (opts ListOpts) ToHypervisorListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
 // List makes a request against the API to list hypervisors.
-func List(client *gophercloud.ServiceClient) pagination.Pager {
-	return pagination.NewPager(client, hypervisorsListDetailURL(client), func(r pagination.PageResult) pagination.Page {
+func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pager {
+	url := hypervisorsListDetailURL(client)
+	if opts != nil {
+		query, err := opts.ToHypervisorListQuery()
+		if err != nil {
+			return pagination.Pager{Err: err}
+		}
+		url += query
+	}
+
+	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
 		return HypervisorPage{pagination.SinglePageBase(r)}
 	})
 }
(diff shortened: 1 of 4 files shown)


jtopjian:
@ntk148v Thank you for submitting this. When you have a moment, please read over our [contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial) which outlines additional information needed for submitting a Pull Request. Please let me know if you have any questions.

ntk148v:
@jtopjian I've updated the PR description.

ntk148v:
- An issue: https://github.com/gophercloud/gophercloud/issues/2132
- According mandatory, actually this is a word from OpenStack Compute API documentation.
https://docs.openstack.org/api-ref/compute/?expanded=list-hypervisors-details-detail#hypervisors-os-hypervisors

openstack/compute/v2/extensions/hypervisors/results.go
```diff
@@ -123,6 +144,12 @@ type Hypervisor struct {
 	// Service is the service this hypervisor represents.
 	Service Service `json:"service"`
 
+	// Servers is a list of Server object.
+	// The requires microversion 2.53 or later.
+	// This field has become mandatory in microversion 2.75.
```
jtopjian, line 149: I recommend we remove this line. "mandatory" doesn't make sense here, even if it's in the OpenStack documentation. This is a field that is being returned from OpenStack, so there's no action that a developer or user can do that is "mandatory".

openstack/compute/v2/extensions/hypervisors/results.go
```diff
@@ -123,6 +144,12 @@ type Hypervisor struct {
...
+	// If no servers is on hypervisor then empty list is returned.
+	Servers []Server `json:"servers"`
```
jtopjian, line 151: Please change this to `*[]Server`. The reason is described in our [microversions](https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md#new-response-fields) document, but let me know if you have any questions.

A similar example of using `*[]` is here: https://github.com/gophercloud/gophercloud/blob/841705fad1df2f50b23c31ee60f51702063be7c6/openstack/compute/v2/extensions/instanceactions/results.go#L141-L143

openstack/compute/v2/extensions/hypervisors/results.go
```diff
@@ -62,6 +62,27 @@ func (r *Service) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// Server represents an instance running on the hypervisor
+type Server struct {
+	Name string `json:"name"`
+	UUID string `json:"uuid"`
+}
+
+func (r *Server) UnmarshalJSON(b []byte) error {
```
jtopjian, line 71: This method isn't needed since it's not doing any custom logic within the body. You should be able to remove this whole method and not see any errors.

openstack/compute/v2/extensions/hypervisors/requests.go
```diff
@@ -5,9 +5,48 @@ import (
...
+// List request.
+type ListOptsBuilder interface {
+	ToHypervisorListQuery() (string, error)
+}
+
+// ListOpts allows the filtering and sorting of paginated collections through
+// the API. Filtering is achieved by passing in struct field values that map to
+// the server attributes you want to see returned. Marker and Limit are used
+// for pagination.
+type ListOpts struct {
```
jtopjian, line 18: All 4 of the options in this struct require a specific microversion to be set. Please see the "New Request Fields" section of our [microversions](https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md#new-request-fields) document for how to handle this.

Basically, please add a doc comment stating which microversion is required and change the type of each parameter to a pointer (`*string`, `*int`, etc). For example:

```go
// Marker is a ID of the last-seen item as a UUID.
// This requires microversion 2.33
Marker *string `q:"marker"`
```

(note: `q` tags don't support `omitempty`, so you don't need to add them here)

jtopjian (changes requested):
@ntk148v Thank you for your patience with this. Please see the review comments and let me know if you have any questions.

openstack/compute/v2/extensions/hypervisors/requests.go
ntk148v: I've checked that `Limit` and `Marker` fields don't need to be a pointer:

https://github.com/gophercloud/gophercloud/blob/fa957f3e1f14997f31f301788af11e247d1facc6/openstack/loadbalancer/v2/listeners/requests.go#L44

I changed the other fields to pointer and kept these two.

jtopjian: `Limit` and `Marker` are part of the standard API set for the LoadBalancer service. However, they were added via a microversion for the Compute Hypervisor API. You can see this in the following code, where there are two blocks of code that define these two parameters for microversions 2.33 to 2.52 and then 2.53 and up:

https://github.com/openstack/nova/blob/stable/victoria/nova/api/openstack/compute/hypervisors.py#L196-L215

It's also in the API documentation: https://docs.openstack.org/api-ref/compute/?expanded=list-hypervisors-details-detail#hypervisors-os-hypervisors

Please change these to pointers.

wangxiyuan:
recheck

jtopjian:
This has been merged as part of #2187

Issue #2124, opened 2021-03-02
Support include_names option in role_assignments endpoint


objque:
Hello, after solving issue https://github.com/gophercloud/gophercloud/issues/550, support for role-assignments was added. 
The Identity API v3.6 introduce ability to load names with query arguments `include_names=True`

It will be great if someone add support of it

doc https://docs.openstack.org/api-ref/identity/v3/index.html?expanded=id627-detail#id627

objque:
@jtopjian

jtopjian:
I agree that it would be great if someone added support for this.

objque:
Okay, I'll take it

objque:
Hey! delayed a little due to tasks at work. I'm working on a patch right now.

jtopjian:
Implemented in #2133

PR #2133, opened 2021-03-25
Identity: Support include_names query arg in /role_assignments

objque:
Keystone may include names into RoleAssignment entity, if provide include_names=true as query arguemnt

For #2124

code in this PR:

https://docs.openstack.org/api-ref/identity/v3/index.html?expanded=id627-detail#id627

### DIFF
--- a/openstack/identity/v3/roles/requests.go
+++ b/openstack/identity/v3/roles/requests.go
@@ -204,6 +204,9 @@ type ListAssignmentsOpts struct {
 	// Effective lists effective assignments at the user, project, and domain
 	// level, allowing for the effects of group membership.
 	Effective *bool `q:"effective"`
+
+	// IncludeNames indicates whether to include names of any returned entities.
+	IncludeNames *bool `q:"include_names"`
 }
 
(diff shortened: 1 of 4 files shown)


jtopjian:
@objque Thanks. Just for reference, API documentation isn't used as it may be outdated. Instead, we need to see the actual Keystone code that defines this feature. This is explained [here](https://github.com/gophercloud/gophercloud/blob/master/docs/contributor-tutorial/step-03-code-hunting.md).

For this PR, the following links are relevant:

* https://github.com/openstack/keystone/blob/a98f006f854be02e5682390012d8bb917f4f3940/keystone/api/role_assignments.py#L120
* https://github.com/openstack/keystone/blob/a98f006f854be02e5682390012d8bb917f4f3940/keystone/api/role_assignments.py#L267-L335

Please let me know when this PR is ready for review.

objque:
Yes, you definitely right. In the next PR's I will reference on the actual Keystone code.
This PR is ready for review. No `wip/draft` tag here :)

jtopjian (approved):
LGTM - thank you!

Issue #2139, opened 2021-04-01
Blockstorage_v3: Add support for QoS specifications


nikParasyr:
Blockstorage v3 supports multiple different api calls for quality-of-service specifications. The full list to be added:
- [x] List QoS Specifications
- [x] Create a QoS specification
- [x] Delete a QoS specification
- [x] Show a QoS specification details
- [x] Set keys in a QoS specification
- [x] Unset keys in a QoS specification
- [ ] Associate QoS specification with a volume type
- [ ] Disassociate QoS specification from a volume type
- [ ] Disassociate a QoS specification from all associations
- [ ] Get all associations for a QoS specification 

Docs for the api: https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=#quality-of-service-qos-specifications-qos-specs
i'll be gradually working on this

PR #2140, opened 2021-04-04
blockstorage/v3: Add support for create/delete qos

nikParasyr:
For #2139 

Create:
Return code on doc is wrong, and parameters other than `name` are missing. Code showcases more how consumer and specs are used.
[docs](https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=create-a-qos-specification-detail#quality-of-service-qos-specifications-qos-specs)
[code_schema](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/api/schemas/qos_specs.py#L19) [code_api](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/api/contrib/qos_specs_manage.py#L79) [code](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/volume/qos_specs.py#L36)

Delete:
[docs](https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=delete-a-qos-specification-detail#quality-of-service-qos-specifications-qos-specs)
[code_api](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/api/contrib/qos_specs_manage.py#L161) [code](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/volume/qos_specs.py#L107)

### DIFF
--- a/acceptance/openstack/blockstorage/v3/blockstorage.go
+++ b/acceptance/openstack/blockstorage/v3/blockstorage.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
+	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/qos"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/snapshots"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumetypes"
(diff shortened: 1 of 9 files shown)


nikParasyr:
@jtopjian this is ready from my side.

jtopjian:
@nikParasyr Thanks!

This set of API calls is a bit strange. The schema only defines a single field called `name` and then allows any other arbitrary field to be specified. I'm sure this is because of the many variations of QoS parameters that are possible. However, in theory, `foobar` is also a valid field.

With that in mind, from reading the other Python code, I do agree with your assessment for using `Name`, `Consumer`, and `Specs` as fields. I also agree with how the `Specs` are modified to be inline with the top-level of the request body.

One optional change you can make is to define the different types of consumers as types, since there is only a finite set of acceptable values for consumer: https://github.com/openstack/cinder/blob/acfc148d7cec08362873115abb4551d2ba531d06/cinder/objects/fields.py#L139-L142

You can see an example of this here: 

* https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/security/rules/requests.go#L50-L77
* https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/security/rules/requests.go#L90

Note that this type would only be used in the functions in `requests.go`. For the resulting struct in `results.go`, the field would still be a string.

Again, this is optional, but encouraged as it helps model the data more accurately.

nikParasyr:
recheck

nikParasyr:
recheck

nikParasyr:
@jtopjian thanks for the help :) The changes are done. Let me know if anything is missing

jtopjian (approved):
LGTM - nice work. Thank you!

PR #2143, opened 2021-04-21
networking/v2: fix documentation for extensions/networkipavailabilities

psvmcc:
This is a documentation-only patch.

For #2144


jtopjian (approved):
LGTM - thank you!

The test failure is ignorable here.

Issue #2153, opened 2021-05-07
Add support for `network_data` field in baremetal Node


zaneb:
[In the Victoria release](https://docs.openstack.org/releasenotes/ironic/victoria.html#relnotes-15-2-0-stable-victoria-new-features), Ironic added a new `network_data` field to the Node API. It can be set on [create](https://docs.openstack.org/api-ref/baremetal/?expanded=create-node-detail#create-node) and retrieved in [show node details](https://docs.openstack.org/api-ref/baremetal/?expanded=show-node-details-detail#show-node-details). The minimum microversion required is 1.66.

jtopjian:
Implemented in #2154

PR #2154, opened 2021-05-07
baremetal: add network_data to Node interface

zaneb:
This field was added in microversion 1.66 and appears in the Victoria release.

For #2153

code in this PR:

https://opendev.org/openstack/ironic/src/branch/master/ironic/api/controllers/v1/node.py#L161-L165
https://review.opendev.org/c/openstack/ironic/+/687910/

### DIFF
--- a/openstack/baremetal/v1/nodes/requests.go
+++ b/openstack/baremetal/v1/nodes/requests.go
@@ -247,6 +247,9 @@ type CreateOpts struct {
 
 	// A string or UUID of the tenant who owns the baremetal node.
 	Owner string `json:"owner,omitempty"`
+
+	// Static network configuration to use during deployment and cleaning.
+	NetworkData map[string]interface{} `json:"network_data,omitempty"`
 }
 
 // ToNodeCreateMap assembles a request body based on the contents of a CreateOpts.
--- a/openstack/baremetal/v1/nodes/results.go
+++ b/openstack/baremetal/v1/nodes/results.go
@@ -192,6 +192,9 @@ type Node struct {
 
 	// A string or UUID of the tenant who owns the baremetal node.
 	Owner string `json:"owner"`
+
+	// Static network configuration to use during deployment and cleaning.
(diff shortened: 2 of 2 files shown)


zaneb:
recheck

zaneb:
Opened #2155 to report the (pre-existing) test failure.

openstack/baremetal/v1/nodes/requests.go
```diff
@@ -247,6 +247,9 @@ type CreateOpts struct {
 
 	// A string or UUID of the tenant who owns the baremetal node.
 	Owner string `json:"owner,omitempty"`
+
+	// Static network configuration to use during deployment and cleaning.
+	NetworkData map[string]interface{} `json:"network_data,omitempty"`
```
jtopjian, line 252: To confirm: this is just an arbitrary/generic map/dict/hash that will accept any data?

https://github.com/openstack/ironic/blob/462969c68f2978b4b250ba99ef491ef6858437fe/ironic/objects/node.py#L168

Normally if a field was added from a microversion, it should be a pointer so that the field is omitted from the boddy if it isn't set. But it looks like Ironic is accounting for having this field set when the microversion is less than 1.35: https://github.com/openstack/ironic/blob/462969c68f2978b4b250ba99ef491ef6858437fe/ironic/objects/node.py#L554

Because of this, I'm okay with leaving it as `map[string]interface{}`, but if you wanted to change it to `*map[string]interface{}`, go for it.

jtopjian (commented):
@zaneb Thanks for submitting this. Just two comments that may not require any changes.

openstack/baremetal/v1/nodes/requests.go
jtopjian: > If you're OK with it I think I'd prefer to leave it, because there's no useful distinction between the field being empty vs. not set.

In this case, I agree. Usually it's because the API will return an error for an invalid field when the microversion isn't set correctly.

PR #2160, opened 2021-05-12
objects: partially revert e54703519

### LINKED ISSUE #2159: objects: unable to create files with / in the name as a pseudo folder
Since https://github.com/gophercloud/gophercloud/commit/e54703519a9c3bcf1e3c25450a02614beb583c6d you are unable to create a file with a / in the name as a pseudo folder. This is used by kOps and documented as broken in kubernetes/kops#9933.

cardoe:
Since this change you are no longer able to create objects with a / in
them to create pseudo-folders. This behavior is relied upon by kOps
documented as broken in kubernetes/kops#9933. fixes #2159.


jtopjian:
It looks like the failure is with the `TestObjectsBulkDelete` test:

```
2021-05-13 00:17:02.588362 | ubuntu-bionic |     convenience.go:35: [1;31mFailure in objects_test.go, line 316: unexpected error [0m[1;33m"Expected HTTP response code [] when accessing [DELETE http://192.168.0.18:8080/v1/AUTH_6b79870e7d4b4022a9a51dc246cf1174/test%26happy%3F-eX1Hu1AF], but got 409 instead\n<html><h1>Conflict</h1><p>There was a conflict when trying to complete your request.</p></html>"[0m[1;31m[0m
2021-05-13 00:17:02.640192 | ubuntu-bionic | --- FAIL: TestObjectsBulkDelete (3.14s)
```

I'm thinking that `url. QueryEscape` should be removed from all occurrences of being used with `objectName`. It feels inconsistent to only remove it for the `Create` API call.

cardoe:
Rebased and updated based on your suggestion @jtopjian

cardoe:
@jtopjian it seems it couldn't find a host to test against?

jtopjian:
@cardoe Yes, there's been some hiccups with OpenLab since this PR first opened. If you can, please do a `git pull` from the master branch and then rebase this branch off of the updated master. Once you do a force-push, testing _should_ work again. Let me know if you have any problems, though. I apologize about the hassle.

cardoe:
The failure happened when I was rebased atop of 9abac83404d86 which is the current master commit.

jtopjian:
ah - well that's interesting! Let's hope it was a one-off issue. I'll dig into it if it fails again. Additionally, I'll look at testing this in a different environment and try to get this reviewed/merged quickly for you.

jtopjian:
The errors indicate that Devstack is still being created with Ubuntu Bionic, thought here's no more mention of Bionic in the repo. I wonder if OpenLab/Zuul is not picking up the rebase somehow... 

I'll take a look at this shortly and try to run the tests in a separate environment.

cardoe:
Thank you for the help with this.

jtopjian:
@cardoe I've added a commit with additional changes. 

I tested this on Vexxhost and ran into errors with the QueryEscape calls in general - I'm not sure how this is/was passing in OpenLab. I decided to just remove all QueryEscape calls from within Gophercloud. A dev/user is still able to escape the paths if they like by performing a `url.QueryEscape` on the name before calling the various Object Storage functions - there's no reason for Gophercloud to mandate it.

Let me know your thoughts. I'm happy to merge this.

kayrus:
> A dev/user is still able to escape the paths if they like by performing a url.QueryEscape on the name before calling the various Object Storage functions - there's no reason for Gophercloud to mandate it.

Let's hope other dependent apps won't fail :\

jtopjian:
There's a good possibility this will affect some applications - this is a breaking change. However, leaving this in is breaking functionality in other areas. From the behaviour I observed on Vexxhost, it's best to leave the URL encoding to the application so it can handle these situations in a more controlled and fine-grained manner.

cardoe:
So I definitely feel that encoding the object name is wrong and bad. Playing with this further shows a number of valid objects that cannot be properly manipulated with GopherCloud. I'm less opinionated about the container name being escaped.

I see with your changes I missed a spot in the bulk delete with the object name when it was in a loop which would have caused test failures with my change.

jtopjian:
OK, I'm going to merge this.

I do understand that this is a breaking change. For anyone who reads this - I apologize. And if this behaviour is still not working correctly, please let me know and I'll do what I can to fix it.

For anyone who was relying on URL encoded container or object names, you can still do this by encoding the name before passing the name to the object storage functions.

PR #2162, opened 2021-05-13
Add resource provider allocations API for placement

abnvanand:
For #526

code in this PR:

API doc:
https://docs.openstack.org/api-ref/placement/?expanded=list-resource-provider-usages-detail#list-resource-provider-allocations

API code:
https://github.com/openstack/placement/blob/2150dbf6880c379b52035b604c51048743c446a5/placement/handlers/allocation.py#L271

### DIFF
--- a/acceptance/openstack/placement/v1/resourceproviders_test.go
+++ b/acceptance/openstack/placement/v1/resourceproviders_test.go
@@ -142,3 +142,33 @@ func TestResourceProviderTraits(t *testing.T) {
 
 	tools.PrintResource(t, usage)
 }
+
+func TestResourceProviderAllocations(t *testing.T) {
+	clients.SkipRelease(t, "stable/mitaka")
+	clients.SkipRelease(t, "stable/newton")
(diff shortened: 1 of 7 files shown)


abnvanand:
Hi @jtopjian. This PR adds API to get allocations for a resource provider.
Can you please review? 

Thanks

jtopjian:
I think this is the best representation of the schema: https://github.com/openstack/placement/blob/6f00ba5f685183539d0ebf62a4741f2f6930e051/placement/schemas/allocation.py#L75-L126

It also details a `project_id` and `user_id`, however, from looking at the acceptance test output, those are not used in the final response.

jtopjian:
@abnvanand Looks like this needs `gofmt` run on it. Other than that, this looks good to me and I can merge once that's fixed.

abnvanand:
Fixed formatting.

jtopjian (approved):
LGTM - Thank you!

jtopjian:
@abnvanand Thanks so much for working on this.

PR #2164, opened 2021-05-19
Add missing bios interface fields to Ironic API

bfournie:
A few bios interface fields are in the Ironic API but were missing from
the Ironic API documentation and not included in gophercloud. This
change adds in the bios interface to the node create, node list,
and node driver validation APIs.

For #[2163]

code in this PR:
https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/node.py#L143
https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/node.py#L1225
https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/utils.py#L1262

### DIFF
--- a/openstack/baremetal/v1/nodes/requests.go
+++ b/openstack/baremetal/v1/nodes/requests.go
@@ -185,6 +185,9 @@ type CreateOpts struct {
 	// Requires microversion 1.47 or later.
 	AutomatedClean *bool `json:"automated_clean,omitempty"`
 
+	// The BIOS interface for a Node, e.g. “redfish”.
+	BIOSInterface string `json:"bios_interface,omitempty"`
+
 	// The boot interface for a Node, e.g. “pxe”.
 	BootInterface string `json:"boot_interface,omitempty"`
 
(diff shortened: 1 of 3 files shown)


jtopjian:
@bfournie Thank you for submitting this. The OpenLab failure can be ignored - it looks like an issue with OpenLab itself.

It looks like you need to run `gofmt` on some of these files. Once that's done, this should be good to merge. Please let me know if you have any questions.

bfournie:
> @bfournie Thank you for submitting this. The OpenLab failure can be ignored - it looks like an issue with OpenLab itself.
> 
> It looks like you need to run `gofmt` on some of these files. Once that's done, this should be good to merge. Please let me know if you have any questions.

Thank you. Yes, I forgot to run `gofmt`, I've updated the files now.

jtopjian (approved):
LGTM - thank you!

jtopjian:
I think this is safe to merge without the OpenLab test passing - I don't think this field would end up causing an error. However, if it does, we'll get it fixed quickly.

Thanks again for working on this.

bfournie:
> I think this is safe to merge without the OpenLab test passing - I don't think this field would end up causing an error. However, if it does, we'll get it fixed quickly.
> 
> Thanks again for working on this.

Thank you

Issue #2165, opened 2021-05-19
Add support for Ironic bios API


bfournie:
The Ironic bios API - `/v1/nodes/{node_ident}/bios`, has been available since version 1.40 of the Ironic API, see
https://docs.openstack.org/api-ref/baremetal/?expanded=#node-bios-nodes.  It provides the ability to get all or individual BIOS settings on a node. 

Support should be added to Gophercloud for the bios API.

mandre:
Implemented via https://github.com/gophercloud/gophercloud/pull/2171.

PR #2171, opened 2021-05-21
Add support for baremetal bios endpoint

bfournie:
For #2165.

The baremetal bios endpoint has been available in the Ironic API since
version 1.40 and is documented in the API reference here -
https://docs.openstack.org/api-ref/baremetal/?expanded=#node-bios-nodes.
It provides a list of all bios settings by node or a particular setting on a node.

The ironic API code is here:
https://github.com/openstack/ironic/blob/master/ironic/api/controllers/v1/bios.py#L57

### DIFF
--- a/openstack/baremetal/v1/nodes/requests.go
+++ b/openstack/baremetal/v1/nodes/requests.go
@@ -636,3 +636,21 @@ func SetRAIDConfig(client *gophercloud.ServiceClient, id string, raidConfigOptsB
 	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
 	return
 }
+
+// Get the current BIOS Settings for the given Node.
+func ListBIOSSettings(client *gophercloud.ServiceClient, id string) (r ListBIOSSettingsResult) {
+	resp, err := client.Get(biosListSettingsURL(client, id), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
+	return
+}
+
+// Get one BIOS Setting for the given Node.
+func GetBIOSSetting(client *gophercloud.ServiceClient, id string, setting string) (r GetBIOSSettingResult) {
+	resp, err := client.Get(biosGetSettingURL(client, id, setting), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
+	return
+}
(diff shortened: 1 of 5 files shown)


zaneb:
Looks reasonable to me :)

openstack/baremetal/v1/nodes/results.go
```diff
@@ -48,6 +48,20 @@ func ExtractNodesInto(r pagination.Page, v interface{}) error {
 	return r.(NodePage).Result.ExtractIntoSlicePtr(v, "nodes")
 }
 
+// Extract interprets a BIOSSettingsResult as an array of BIOSSetting structs, if possible.
+func (r BIOSSettingsResult) Extract() (*[]BIOSSetting, error) {
+	var s BIOSSettings
+	err := r.ExtractInto(&s)
+	return &s.Settings, err
+}
```
jtopjian, line 56: What you have works, but I think the following version more closely matches other functions within Gophercloud:

```go
// Extract interprets a BIOSSettingsResult as an array of BIOSSetting structs, if possible.
func (r BIOSSettingsResult) Extract() ([]BIOSSetting, error) {
  var s struct {
    Settings []BIOSSetting `json:"bios"`
  }
  err := r.ExtractInto(&s)
  return s.Settings, err
}
```

You can then remove the `BIOSSettings` type as it's no longer used.

This change will require an update to the unit tests, but it's a small update.

jtopjian (changes requested):
@bfournie Thanks for submitting this. I've left two comments for review. Please let me know if you have any questions.

openstack/baremetal/v1/nodes/requests.go
```diff
@@ -636,3 +636,21 @@ func SetRAIDConfig(client *gophercloud.ServiceClient, id string, raidConfigOptsB
...
+func ListBIOSSettings(client *gophercloud.ServiceClient, id string) (r BIOSSettingsResult) {
+	resp, err := client.Get(biosListSettingsURL(client, id), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
+	return
+}
+
+// Get one BIOS Setting for the given Node.
+func GetSingleBIOSSetting(client *gophercloud.ServiceClient, id string, setting string) (r SingleBIOSSettingResult) {
```
jtopjian, line 650: Total nit pick: How about `GetBIOSSetting`? The use of "Get" (instead of "List") and the singular use of "Setting" implies that a single item will be returned. This is not a required change, though.

openstack/baremetal/v1/nodes/results.go
bfournie: Thanks, that's a good change.  Updated.

openstack/baremetal/v1/nodes/requests.go
bfournie: Yes, that naming is more clear. Fixed.

jtopjian (approved):
LGTM - thank you!

Issue #2166, opened 2021-05-19
Add support for additional bios API fields from Redfish bios registry


bfournie:
As described in the Ironic spec - https://review.opendev.org/c/openstack/ironic-specs/+/774681, additional fields are being added to the Ironic bios API in to include fields retrieved from the Redfish BIOS registry. This is currently up for review in 
https://review.opendev.org/c/openstack/ironic/+/786707 using the proposed micro-version 1.74.

This is a follow-on to the bios API support https://github.com/gophercloud/gophercloud/issues/2165 that should be implemented once the Ironic patch above merges.

jtopjian:
Implemented in #2174

PR #2169, opened 2021-05-20
Acc Tests: Moving to Ubuntu Focal

jtopjian:
For #2168

PR #2174, opened 2021-06-04
Add support for additional bios settings fields in Ironic

bfournie:
For #2166

A recent change to Ironic
(https://review.opendev.org/c/openstack/ironic/+/786707) has added
support for bios registry fields in the Ironic API for microversion 1.74. 
These can be retrieved using the `?detail=True` option when listing
all settings. The fields are also included when getting individual
settings without any options.

### DIFF
--- a/openstack/baremetal/v1/nodes/testing/fixtures.go
+++ b/openstack/baremetal/v1/nodes/testing/fixtures.go
@@ -622,6 +622,92 @@ const NodeBIOSSettingsBody = `
    ]
 }
 `
+
+const NodeDetailBIOSSettingsBody = `
+{
+  "bios": [
+   {
+      "created_at": "2021-05-11T21:33:44+00:00",
+      "updated_at": null,
+      "name": "Proc1L2Cache",
+      "value": "10x256 KB",
+      "attribute_type": "String",
+      "allowable_values": [],
+      "lower_bound": null,
+      "max_length": 16,
+      "min_length": 0,
+      "read_only": true,
+      "reset_required": null,
+      "unique": null,
+      "upper_bound": null,
+      "links": [
+        {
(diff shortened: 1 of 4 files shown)


elfosardo:
this probably depends on the focal update? https://github.com/gophercloud/gophercloud/pull/2169

elfosardo:
recheck

jtopjian:
This PR will need to be rebased off of master in order to take advantage of the recently fixed tests. However, since there aren't any acceptance tests in this PR, nothing will really be tested, so it's kind of moot.

openstack/baremetal/v1/nodes/testing/fixtures.go
```diff
@@ -631,6 +717,14 @@ const NodeSingleBIOSSettingBody = `
 }
 `
 
+func getIntPointer(x int) *int {
+	return &x
+}
+
+func getBoolPointer(x bool) *bool {
+	return &x
+}
```
jtopjian, line 726: There are a few existing ways of handling this:

1. There's some existing helper code here: openstack/identity/v3/users/testing/requests_test.go
2. Just declare the variables each time: https://github.com/gophercloud/gophercloud/blob/513734676e6495f6fec60e7aaf1f86f1ce807428/openstack/networking/v2/extensions/mtu/testing/results_test.go#L90-L94

I can't recall why there aren't central helper functions in the `testhelper` package. I have a feeling that, for whatever reason, it was discouraged to do this in Golang in general and to instead manually declare every occurrence of pointers, but it's been too long for me to remember at this point.

jtopjian (commented):
@bfournie Thanks for submitting this. This looks good to me. I've left one non-blocking comment. Let me know if you want to change things or leave them as-is.

openstack/baremetal/v1/nodes/testing/fixtures.go
bfournie: Thank you. I changed to use the variables.

jtopjian (approved):
LGTM - thank you!

Issue #2170, opened 2021-05-21
No access to X-Timestamp attribute in objectstorage container header


NeilW:
The object storage GetHeader structure within objectstorage/v1/containers lacks access to the 'X-Timestamp' attribute showing when the container was originally created. 

https://docs.openstack.org/api-ref/object-store/?expanded=show-container-metadata-detail#show-container-details-and-list-objects

jtopjian:
Implemented in #2185

PR #2185, opened 2021-06-30
Object Storage v1: Adding Timestamp to GetHeader results

jtopjian:
For #2170 

https://github.com/openstack/swift/blob/b79ca57fa98323b17f896af86c617d465066d544/swift/common/swob.py#L1014-L1039

### DIFF
--- a/openstack/objectstorage/v1/containers/results.go
+++ b/openstack/objectstorage/v1/containers/results.go
@@ -3,6 +3,7 @@ package containers
 import (
 	"encoding/json"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


openstack/objectstorage/v1/containers/results.go
```diff
@@ -104,16 +105,19 @@ type GetHeader struct {
 	StoragePolicy    string    `json:"X-Storage-Policy"`
 	TempURLKey       string    `json:"X-Container-Meta-Temp-URL-Key"`
 	TempURLKey2      string    `json:"X-Container-Meta-Temp-URL-Key-2"`
+	Timestamp        float64   `json:"-"`
```
kayrus, line 108: @jtopjian you can use `ContentLength  int64     `json:"Content-Length,string"``, see for example https://github.com/gophercloud/gophercloud/blob/c9471fccb2bb90d41832195b1134ac1c6f50f2bb/openstack/objectstorage/v1/objects/results.go#L136

jtopjian: Oh nice - thanks!

jtopjian:
recheck

jtopjian:
The Container tests pass on Vexxhost. I'm thinking the OpenLab failure was a timing issue.

PR #2146, opened 2021-04-25
Add the user id to the get create and delete keypair

randomswdev:
For #2145

The capability to create of get the keypair of another user is documented in the Nova Rest API reference:
https://docs.openstack.org/api-ref/compute/?expanded=list-keypairs-detail,show-keypair-details-detail,create-or-import-keypair-detail#create-or-import-keypair

This pull request adds the capability to create or get the keypairs also for another user.

* https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/schemas/keypairs.py#L64-L84
* https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/keypairs.py#L148-L156
* https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/keypairs.py#L175-L182

Issue #2145, opened 2021-04-25
Compute v2: add the capability to create keypairs for other users


randomswdev:
The nova API v 2.10 added the ability to create a keypair for another user:

https://docs.openstack.org/api-ref/compute/?expanded=list-keypairs-detail,show-keypair-details-detail#create-or-import-keypair

> The user_id for a keypair. This allows administrative users to upload keys for other users than themselves.

The gophercloud library currently does not support this functionality. I would like to contribute a pull request that adds this support.

jtopjian:
Implemented in #2186

PR #2186, opened 2021-06-30
Compute v2: Add the user id to the get create and delete keypair

jtopjian:
For #2145 
Supersedes #2146


jtopjian:
@kayrus Let me know your thoughts on this. Generally, I'm not opposed to adding an `opts` field to existing functions, but since Key Pairs are part of the earliest and foundational resources in Gophercloud, I want to be careful and retain backwards compatibility.

openstack/compute/v2/extensions/keypairs/requests.go
```diff
@@ -81,9 +85,85 @@ func Get(client *gophercloud.ServiceClient, name string) (r GetResult) {
...
+// ToKeyPairGetQuery formats a GetOpts into a query string.
+func (opts GetOpts) ToKeyPairGetQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
+// GetWithOpts is similar to Get, but can take a GetOpts struct.
+// The options within GetOpts require the client to have a microversion set.
+// The original Get function is retained for non-microverisoned requests.
+func GetWithOpts(client *gophercloud.ServiceClient, name string, opts GetOptsBuilder) (r GetResult) {
```
kayrus, line 110: @jtopjian `GetWithOpts` func looks hacky. I'd extend the `Get` function, even if it breaks the previous API.

kayrus:
@jtopjian can you also add a `UserID` parameter to the `List` function?

kayrus:
@jtopjian thanks a lot. LGTM.

Issue #2132, opened 2021-03-24
Compute: Handle Hypervisor request parameters


ntk148v:
Start from API version 2.53, Nova returns a list of server objects.

https://docs.openstack.org/api-ref/compute/?expanded=list-hypervisors-details-detail#hypervisors-os-hypervisors

https://github.com/openstack/nova/blob/stable/victoria/nova/api/openstack/compute/hypervisors.py#L89

jtopjian:
Implemented in #2187

PR #2187, opened 2021-07-01
Compute v2: Implement Hypervisor request options 

jtopjian:
For #2132 
Supersedes #2131

### DIFF
--- a/openstack/compute/v2/extensions/hypervisors/doc.go
+++ b/openstack/compute/v2/extensions/hypervisors/doc.go
@@ -24,7 +24,28 @@ Example of Show Hypervisor Details with Compute API microversion greater than 2.
 
 Example of Retrieving Details of All Hypervisors
 
-	allPages, err := hypervisors.List(computeClient).AllPages()
+	allPages, err := hypervisors.List(computeClient, nil).AllPages()
+	if err != nil {
+		panic(err)
+	}
+
+	allHypervisors, err := hypervisors.ExtractHypervisors(allPages)
+	if err != nil {
+		panic(err)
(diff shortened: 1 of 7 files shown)


jtopjian:
recheck

jtopjian:
@kayrus Do you mind reviewing this when you have time?

openstack/compute/v2/extensions/hypervisors/doc.go
```diff
@@ -24,7 +24,28 @@ Example of Show Hypervisor Details with Compute API microversion greater than 2.
...
+	allHypervisors, err := hypervisors.ExtractHypervisors(allPages)
+	if err != nil {
+		panic(err)
+	}
+
+	for _, hypervisor := range allHypervisors {
+		fmt.Printf("%+v\n", hypervisor)
+	}
+
+Example of Retrieving Details of All Hypervisors with Compute API microversion greater than 2.33
```
kayrus, line 41: Not clear which microversion end user needs to use. 2.33 or 2.53 (from the acceptance tests)

kayrus (commented):
I don't have a use case for this API call, I can only try to run acceptance tests.
The only thing that I have concerns on is an API version in docs.

openstack/compute/v2/extensions/hypervisors/doc.go
jtopjian: It is confusing, yes. The newly added `ListOpts` includes options that are valid in either 2.33 or 2.53. The title of this example is supposed to denote that microversion 2.33+ is required.

PR #2193, opened 2021-07-12
add `tag` attribute to servers.Network

### LINKED ISSUE #2192: servers: Tag is missing for servers.Network
This type:

https://github.com/gophercloud/gophercloud/blob/5a7ac6556e891d9dc387da1ed15311b1ab60023e/openstack/compute/v2/servers/requests.go#L118-L131

should have an additional field `Tag string` for device-role tags, as mentioned in the [API docs](https://docs.openstack.org/api-ref/compute/?expanded=create-server-detail#create-server). I tried finding a single line in the codebase that demonstrates this field being read, and was once again overwhelmed by OpenStack's signature dish, spaghetti code. As for evidence, let's go with [this commit](https://github.com/openstack/nova/commit/e80e2511cf825671a479053cc8d41463aab1caaa) that last touched the definition of the field in the API reference. It contains several API samples both for docs and tests that demonstrate the tag attribute on networks in create-server requests.

majewsky:
Closes #2192.

code in this PR: Links are given in the issue.

---

I was going to extend the tests to handle this attribute, but there are no tests at all that actually cover the `networks` attribute of server creation requests, so I hope this change is okay as-is.

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -128,6 +128,13 @@ type Network struct {
 
 	// FixedIP specifies a fixed IPv4 address to be used on this network.
 	FixedIP string
+
+	// Tag may contain an optional device role tag for the server's virtual
+	// network interface. This can be used to identify network interfaces when
+	// multiple networks are connected to one server.
+	//
+	// Requires microversion 2.32 through 2.36 or 2.42 or later.
+	Tag string
 }
 
 // Personality is an array of files that are injected into the server at launch.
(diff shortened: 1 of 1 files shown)


jtopjian:
recheck

jtopjian:
Interesting. The same failure as last time. Other PRs are passing, though. I'm going to try one more time.

jtopjian:
recheck

jtopjian:
Different test failure, but it was with submitting the test results. Everything looks like it passed.

@majewsky This looks good to me. Let me know if you are ready for this to be merged.

majewsky:
Yes, please, by all means.

jtopjian (approved):
LGTM - thank you!

Issue #2183, opened 2021-06-30
Automatic retries


andrewbaxter:
I'd like to request that Gophercloud support (an optional) automatic retry mechanism.  I've seen this mentioned and rejected a couple times, but I'd like to make additional arguments for why it would be good if this was supported.

There are status codes and errors that we always want to retry - timeouts, internal service error (500) from unreliable upstream providers, maybe reauthentication when creds expire, other random network glitches, etc.  In order for us to retry these on every call to every openstack service we use, even assuming we only call each once, it's still 50+ method calls that need to be wrapped.  We could use go's reflection to make a wrapper method but at the same time lose type safety and we'd still need to go in and touch each call.

Gophercloud's code is in a much better position to do this: all requests go through the doRequest method in the client object which deals with generic requests (and retrying already exists for one error code).  This isn't something that can be trivially worked around downstream - `ProviderClient` is a concrete type so a custom implementation can't be provided and there's no way to inject custom request processing code (AFAICT) since each of the request methods call `client.Request` directly.

FWIW other iaas api libraries (AWS, GCP, Azure) etc all do 3-5 retries by default to my knowledge.

I assume there were likely concerns about duplicate resources being created or requests failing because they caused a state change before being retried, etc. For the most part we're creating resources, listing resources, and deleting resources (the latter of which we mostly log and ignore the result) where this doesn't matter or else would be an issue regardless of retries.  For other apis (attaching volumes, maybe) which involve state changes, handling an error from a duplicate execution wouldn't be harder than manually retrying in the case of a temporary error.  Regardless though I think making it optional would be reasonable.

jtopjian:
@andrewbaxter Thank you for sharing your thoughts.

Gophercloud does support retry mechanisms, specifically for reauthentication and for 429 responses. If prior requests for general "retrying" have been rejected, it was due to the implementation being too specific and covering only a single use-case. There's always concern about adding additional use-cases for the foreseeable future. Instead, the implementation should be generic enough for others to extend and customize. Leveraging hooks with function signatures seems to be the best method.

With that in mind, I would be happy to review and merge a PR which either implemented a general retry mechanism or added individual mechanisms for different HTTP status codes.

jtopjian:
Implemented in #2194

PR #2194, opened 2021-07-19
Add general request retry mechanism

andrewbaxter:

For #2183 

code in this PR:

None - this is purely client-side.

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -27,6 +27,12 @@ type UserAgent struct {
 
 type RetryFunc func(context.Context, *ErrUnexpectedResponseCode, error, uint) error
 
+// GeneralRetryFunc is a catch-all function for retrying failed API requests.
+// If it returns nil, the request will be retried.  If it returns an error,
+// the request method will exit with that error.  failCount is the number of
+// times the request has failed (starting at 1).
+type GeneralRetryFunc func(context context.Context, method, url string, options *RequestOpts, err error, failCount uint) error
+
 // Prepend prepends a user-defined string to the default User-Agent string. Users
 // may pass in one or more strings to prepend.
 func (ua *UserAgent) Prepend(s ...string) {
@@ -85,11 +91,15 @@ type ProviderClient struct {
 	// Context is the context passed to the HTTP request.
 	Context context.Context
 
-	// Retry backoff func
+	// Retry backoff func is called when rate limited.
 	RetryBackoffFunc RetryFunc
 
 	// MaxBackoffRetries set the maximum number of backoffs. When not set, defaults to DefaultMaxBackoffRetries
 	MaxBackoffRetries uint
+
+	// A general failed request handler method - this is always called in the end if a request failed. Leave as nil
+	// to abort when an error is encountered.
+	GeneralRetryFunc GeneralRetryFunc
 
 	// mut is a mutex for the client. It protects read and write access to client attributes such as getting
 	// and setting the TokenID.
@@ -348,7 +358,7 @@ func (client *ProviderClient) Request(method, url string, options *RequestOpts) 
 	})
 }
 
-func (client *ProviderClient) doRequest(method, url string, options *RequestOpts, state *requestState) (*http.Response, error) {
+func (client *ProviderClient) doRequestInner(method, url string, options *RequestOpts, state *requestState) (*http.Response, error) {
 	var body io.Reader
 	var contentType *string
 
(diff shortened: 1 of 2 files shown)


andrewbaxter:
This is passing tests (at least the ones I tried locally) but I haven't played around with it much yet.  I wanted initial feedback if you thought this needed to be redesigned.

Notes on the implementation:

- I left RetryBackoffFunc and the reauth stuff because I didn't want to make this backwards incompatible.  I think it should be possible to make a default `GeneralRetryFunc` that handles both reauth and the 429 stuff if so desired.
- There's no MaxRetry count field (and the Backoff one isn't used) since this can be handled by the `GeneralRetryFunc` itself.
- I added two simple tests, but I can add more if you think it would be useful.  I didn't want to go crazy in case you wanted changes to the implementation.
- I didn't include any premade `GeneralRetryFunc` - I think one that retries GETs and maybe PATCHs would be reasonable, but I haven't had time to dig into this yet and I'm sure some tuning would be required to make something that works well out of the box.

andrewbaxter:
I've tested this myself with `tc` to drop packets, but that's pretty limited and I can't exactly cause our openstack provider to go down ;) ... for what it's worth.  I was able to confirm that operation in normal circumstances didn't break though I believe.

jtopjian:
@andrewbaxter Thanks for working on this. This is a really good start.

The first area that stuck out was with the RetryBackoff implementation. To make things better named, I think it's best to rename the existing `RetryFunc` to `RetryBackoffFunc`. This change will also need done in the unit tests. This will have a slight backwards incompatibility, but I think it's worth it in the long run. 

After that, we can rename `GeneralRetryFunc` to simply `RetryFunc`.

Looking at the implementation, instead of creating a `doRequest` that wraps the renamed `doInnerRequest`, isn't it possible to keep a single `doRequest` and implement the retry mechanism like so:

```go
...
// existing code
    case http.StatusServiceUnavailable:
      err = ErrDefault503{respErr}
      if error503er, ok := errType.(Err503er); ok {
        err = error503er.Error503(respErr)
      }
    }

    if err == nil {
      err = respErr
    }

// add retry mechanism here
    if err != nil && client.RetryFunc != nil {
      var e error
      state.retries = state.retries + 1
      e = client.RetryFunc(client.Context, method, url, options, err, state.retries)
      if e != nil {
        return resp, e
      }

      return client.doRequest(method, url, options, state)
    }

// resume existing code
    return resp, err
  }
...
```

I tested this out and the unit tests pass, but maybe I am missing something.

Let me know your thoughts on the above. Once we discuss that, we can talk about extra tests, default functions, etc.

I'm not concerned about the failing acceptance test _yet_. OpenLab can be quirky sometimes and exhibit transient behaviour. However, if they continue to fail, it might be this RetryFunc somehow causing request issues.

andrewbaxter:
Thanks! I'll get on those changes.

For the single `doRequest` - it wasn't clear to me if go would do tail call optimization here and a loop would keep the stack short. To make sure I understand, you mean recurse `doRequest` rather than have a loop?  If that's the preference I'm happy to make that change.

Okay yeah, I didn't look at the acceptance tests quite yet (wasn't sure where to find the failing output exactly) :sweat_smile: I hope they do pass but I'll dig deeper after resolving the above.

Thanks for the review!

jtopjian:
> To make sure I understand, you mean recurse doRequest rather than have a loop?

Correct. 

We're already doing reauth retries and backoff retries in the single `doRequest`, so we might as well stick with consistency (unless we discover an issue with general retrying). As for optimization, again, since we're already doing other types of retries, I think we'd be looking at a major refactor.

> Okay yeah, I didn't look at the acceptance tests quite yet (wasn't sure where to find the failing output exactly)

If you click on the link for the failed test, you can access `job-output.txt.gz` and sift through 1.4mb of text 🙂 Searching for `--- FAIL` will take you to the failed tests. From what I can tell, there were one or two failed requests that were the cause of the failures.

jtopjian:
recheck

andrewbaxter:
I made the above changes, but there are at least 4 separate error return statements in `doRequest` which could result from temporary issues (upstream, network connectivity, etc).  I ended up copying the RetryFunc call snippet into each of them, but I'm not sure if there's a better way to handle this.

jtopjian:
@andrewbaxter Thanks for continuing to work on this.

I think the first two additions of calling RetryFunc are all that's needed (Lines 428 and 573).

On Line 593, the error would be an IO-based error, where the content is being discarded in order to trigger an EOF so that the connection can safely close. I don't think this is something that should be retried if it fails since it's not part of the HTTP/network request (other than ensuring a connection is closed, and if we retry, we might end up accumulating open connections).

And on Line 606, the error would be from unmarshalling a JSON response. At this point, the HTTP connection was successful, data was successfully retrieved, the HTTP connection was closed, but there was some kind of problem unmarshalling the data. IMO, this error should be handled within the application.

Thoughts?

andrewbaxter:
Okay, makes sense for 593.  Retrying would be a strange response to some internal connection management issue here, and it looks like the response _is_ returned (just along with the error).

For 606, I've seen servers partially write responses (with matching content lengths) randomly, and that's a situation where I'd want to retry - similarly to other temporary upstream issues.

jtopjian:
> For 606, I've seen servers partially write responses (with matching content lengths) randomly, and that's a situation where I'd want to retry - similarly to other temporary upstream issues.

That's fair and sounds good to me.

From what I can tell, this is looking good. 

@kayrus @ozerovandrei do either of you have any thoughts on this one?

jtopjian (approved):
LGTM - thank you!

jtopjian:
@andrewbaxter Thank you for working on this!

andrewbaxter:
Awesome, thanks for the reviews and merge!

kayrus:
This change must be reflected in gophercloud/utils as well:

```
vendor/github.com/gophercloud/utils/client/client.go:368:9: cannot use func literal (type func(context.Context, *gophercloud.ErrUnexpectedResponseCode, error, uint) error) as type gophercloud.RetryFunc in return argument
```

PR #2201, opened 2021-08-05
Vendor Passthru support for Subscriptions

iurygregory:
For #1429 
This PR adds support for the following vendor passthru methods:
- create_subscription
- delete_subscription
- get_subscription
- get_all_subscriptions

This methods are only supported by redfish implementations in ironic
that are using the vendor passthru.

Introduced in Ironic in  [change 801064](https://review.opendev.org/c/openstack/ironic/+/801064)

### DIFF
--- a/openstack/baremetal/v1/nodes/doc.go
+++ b/openstack/baremetal/v1/nodes/doc.go
@@ -126,5 +126,67 @@ Example to get boot device for a node
 	if err != nil {
 		panic(err)
 	}
+
+Example to list all vendor passthru methods
+
+	methods, err := nodes.GetVendorPassthruMethods(client, "a62b8495-52e2-407b-b3cb-62775d04c2b8").Extract()
+	if err != nil {
+		panic(err)
+	}
+
+Example to list all subscriptions
+
+	method := nodes.CallVendorPassthruOpts{
+		Method: "get_all_subscriptions",
+	}
+	allSubscriptions, err := nodes.GetAllSubscriptions(client, "a62b8495-52e2-407b-b3cb-62775d04c2b8", method).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+Example to get a subscription
+
+	method := nodes.CallVendorPassthruOpts{
+		Method: "get_subscription",
+	}
+	subscriptionOpt := nodes.GetSubscriptionOpts{
+		Id:     "subscription id",
+	}
+
+	subscription, err := nodes.GetSubscription(client, "a62b8495-52e2-407b-b3cb-62775d04c2b8", method, subscriptionOpt).Extract()
+	if err != nil {
+		panic(err)
+	}
+
+Example to delete a subscription
+
+	method := nodes.CallVendorPassthruOpts{
+		Method: "delete_subscription",
+	}
+	subscriptionDeleteOpt := nodes.DeleteSubscriptionOpts{
+		Id: "subscription id",
+	}
+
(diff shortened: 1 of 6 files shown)


iurygregory:
openlab-ci seems to be broken =(

jtopjian:
Indeed. I've been meaning to open an Issue about this. I'll try to get to that today.

jtopjian:
@iurygregory Thanks for submitting this. 

At first review, I think the pattern being used here is too "high level" for Gophercloud and I couldn't find an existing similar implementation. 

Gophercloud should have relatively simple and low level API calls with little logic used to build the request. In this case, I suggest having each function require a `CallVendorPassthruOpts` in addition to the per-method (create, delete, etc) opts. Once these are in place, helper functions that simplify the interaction with these low-level functions could be added to `gophercloud/utils` to help make this interaction easier. 

Another possible option would be to combine `CallVendorPassthruOpts` and the per-function opts like this:

```go
type CreateSubscriptionOpts struct {
        Method      string   `q:"method"`
	Destination string   `json:"Destination"`
	EventTypes  []string `json:"EventTypes,omitempty"`
	Context     string   `json:"Context,omitempty"`
	Protocol    string   `json:"Protocol,omitempty"`
}
```

You can then call two methods on the struct: one to create the query string and one to create the request body. It's still not a clean, intuitive solution, though.

I understand that this makes calling the functions rather awkward, but the way these routes are designed seem to be a little different than most. 

Let me know if you have any questions or would like to discuss further.

iurygregory:
Hi @jtopjian , thanks for the feedback!
I will give a try to the approach you mentioned of having a single struct including the method.

iurygregory:
Having the Method inside the same struct was causing gophercloud to send as part of the request, I'm trying now to add CallVendorPassthruOpts as parameter for each function.

jtopjian:
@iurygregory Thanks for making those changes. This code looks good to me, though I'm not familiar with Ironic. Have you been able to test this on a running Ironic service? If it works the way you're expecting, I can go ahead and merge this.

iurygregory:
Hi @jtopjian 

I've tested and I've found one small issue in the delete - ironic returns 200, so I've added 
`OkCodes:  []int{200, 202, 204},` 
to it just to avoid the message
```
Expected HTTP response code [202 204] when accessing [DELETE http://192.168.122.1:6385/nodes/c5a2bcdb-0ead-4162-a5ce-ea3f23929a87/vendor_passthru?method=delete_subscription], but got 200 instead
```
All other actions are working as expected, I'm pushing a new commit to address the problem with delete and it should be fine.

Thanks!

iurygregory:
@jtopjian let me know if there is something else I should do, thanks! =)

jtopjian:
@iurygregory Everything looks good. Sorry for the delay!

jtopjian (approved):
LGTM - thank you!

iurygregory:
@jtopjian Thanks! If possible can you publish a tag with this feature? It would make things easier to do the changes in metal3.

jtopjian:
@iurygregory Sure, I don't mind making a release for this. Do you have any additional PRs planned? If so, I can wait. But if not, I will create a release now.

iurygregory:
@jtopjian I've checked with some folks today, for now this would be the only PR we need to start working on the feature =)

jtopjian:
@iurygregory Sounds good. v0.20.0 has been created.

PR #2212, opened 2021-09-10
Add block volume extension for os-vol-host-attr

### LINKED ISSUE #2211: Value for os-vol-host-attr:host not returned for block volumes
Although OpenStack API returns the value for `os-vol-host-attr:host` when getting volumes, denoting the host that the volume was created on, the library ignores this value.

Even the related [test fixtures](https://github.com/kristian-lesko/gophercloud/blob/master/openstack/blockstorage/v3/volumes/testing/fixtures.go) contain this key in the respective outputs, but with `null` as its value.

kristian-lesko:
The os-vol-host-attr is returned by OpenStack API;
let's add an extension to allow using this value.

Fixes #2211.

### DIFF
--- a/openstack/blockstorage/extensions/volumehost/doc.go
+++ b/openstack/blockstorage/extensions/volumehost/doc.go
@@ -1,0 +1,26 @@
+/*
+Package volumehost provides the ability to extend a volume result with
+information about the Openstack host holding the volume. Example:
+
+	type VolumeWithHost struct {
+		volumes.Volume
+		volumehost.VolumeHostExt
+	}
+
+	var allVolumes []VolumeWithHost
+
+	allPages, err := volumes.List(client, nil).AllPages()
+	if err != nil {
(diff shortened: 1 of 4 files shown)


jtopjian:
@kristian-lesko Thank you for submitting this. 

This might seem overkill for a single field, but technically `os-vol-host-attr:host` is an API extension as defined here: https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/api/contrib/volume_host_attribute.py

Instead of adding this field to the main `Volume` struct, create a dedicated package for it like this one: https://github.com/gophercloud/gophercloud/tree/513734676e6495f6fec60e7aaf1f86f1ce807428/openstack/blockstorage/extensions/volumetenants

kristian-lesko:
Thanks for your comment @jtopjian; PR updated per your suggestion.

jtopjian (approved):
LGTM - thank you very much!

Issue #2216, opened 2021-09-16
Compute's Server type should carry information about Server groups


pierreprinetti:
Starting with microversion 1.71, Nova exposes the Server group IDs a Server was created in, [with the property `server_groups`](https://docs.openstack.org/api-ref/compute/?expanded=show-server-details-detail#show-server-details).

pierreprinetti:
Fixed by #2217

PR #2217, opened 2021-09-16
compute: Add ServerGroup property to Server

pierreprinetti:
Add to the Server type the list of server groups the server is part of,
as exposed in microversions 1.71 and above as "server_groups".

For #2216

Here are the docs describing the property of the API object "server": https://docs.openstack.org/api-ref/compute/?expanded=show-server-details-detail#show-server-details

* https://github.com/openstack/nova/blob/d2a5fe5621d6ff1ae8ba5087049e0c4347592cf6/nova/api/openstack/compute/servers.py#L454-L455
* https://github.com/openstack/nova/blob/52cae8801de9229d6cfb7871c2073b83b3e41b81/nova/api/openstack/compute/views/servers.py#L420-L423

### DIFF
--- a/openstack/compute/v2/servers/results.go
+++ b/openstack/compute/v2/servers/results.go
@@ -221,6 +221,12 @@ type Server struct {
 	// Tags is a slice/list of string tags in a server.
 	// The requires microversion 2.26 or later.
 	Tags *[]string `json:"tags"`
+
+	// ServerGroups is a slice of strings containing the UUIDs of the
+	// server groups to which the server belongs. Currently this can
+	// contain at most one entry.
+	// New in microversion 2.71
+	ServerGroups *[]string `json:"server_groups"`
 }
 
 type AttachedVolume struct {


pierreprinetti:
Tested against OpenStack 16:

```Go
	computeClient.Microversion = "2.71"

	s, err := servers.Get(computeClient, serverUUID).Extract()
	if err != nil {
		panic(err)
	}

	fmt.Println(s.ServerGroups)
```

openstack/compute/v2/servers/results.go
```diff
@@ -221,6 +221,12 @@ type Server struct {
 	// Tags is a slice/list of string tags in a server.
 	// The requires microversion 2.26 or later.
 	Tags *[]string `json:"tags"`
+
+	// ServerGroups is a slice of strings containing the UUIDs of the
+	// server groups to which the server belongs. Currently this can
+	// contain at most one entry.
+	// New in microversion 2.71
+	ServerGroups []string `json:"server_groups"`
```
jtopjian, line 229: This should be `*[]string` per: https://github.com/gophercloud/gophercloud/blob/master/docs/MICROVERSIONS.md#new-response-fields

jtopjian (changes requested):
@pierreprinetti Thanks for submitting this. Just one change - let me know if you have any questions.

In addition, for updates like this, we need to see the upstream Nova code that implements this feature. For this case, I believe this is the following:

* https://github.com/openstack/nova/blob/d2a5fe5621d6ff1ae8ba5087049e0c4347592cf6/nova/api/openstack/compute/servers.py#L454-L455
* https://github.com/openstack/nova/blob/52cae8801de9229d6cfb7871c2073b83b3e41b81/nova/api/openstack/compute/views/servers.py#L420-L423

Does that look correct to you?

openstack/compute/v2/servers/results.go
pierreprinetti: I am not a fan of indirection, but I am happy to do the change in the name of consistency.

However I wonder what case does this "pointer rule" address? When setting the microversion to a high enough value, Nova becomes pretty loud:

```plaintext
$ go build ./cmd/test && ./test
panic: Expected HTTP response code [200 203] when accessing [GET https://compute.mycloud.example.com/v2.1/0693e2bb538c92b79a49fe6d2e61b0fc/servers/d7f95d73-7a68-423d-89a7-1bf0031f7016], but got 406 instead
{"computeFault": {"code": 406, "message": "Version 2.71 is not supported by the API. Minimum is 2.1 and maximum is 2.60."}}
```

On a recent enough could, here's the matrix of my results on **OpenStack ~16** ([code](https://gist.github.com/pierreprinetti/c2011ba2d41a9efc51576963f57b9b0e)):
|call microversion|server has SG|ServerGroups is nil |len(ServerGroups)|
|-|-|-|-|
|2.60|false|true|0|
|2.60|true|true|0|
|2.71|false|false|0|
|2.71|true|false|1|

My OpenStack13 cloud is currently down, but based on what I've seen yesterday, I would bet that this is the result matrix for **OpenStack 13**:
|call microversion|server has SG|ServerGroups is nil |len(ServerGroups)|
|-|-|-|-|
|2.60|false|true|0|
|2.60|true|true|0|
|2.71|false|error|error|
|2.71|true|error|error|

It seems that the results are distinct and deterministic based on the microversion in the Gophercloud call.
Incidentally, this slice is only `nil` when the microversion is below threshold (but I understand that this `nil` distinction can not be generalised to any type).

As I have already mentioned, however, this is just a thought and I am happy to conform to the rule.

pierreprinetti:
> In addition, for updates like this, we need to see the upstream Nova code that implements this feature. For this case, I believe this is the following:
> 
>     * https://github.com/openstack/nova/blob/d2a5fe5621d6ff1ae8ba5087049e0c4347592cf6/nova/api/openstack/compute/servers.py#L454-L455
> 
>     * https://github.com/openstack/nova/blob/52cae8801de9229d6cfb7871c2073b83b3e41b81/nova/api/openstack/compute/views/servers.py#L420-L423
> 
> 
> Does that look correct to you?

This looks perfectly correct to me, thank you for the research :pray:

openstack/compute/v2/servers/results.go
jtopjian: > However I wonder what case does this "pointer rule" address? 
>
> Incidentally, this slice is only nil when the microversion is below threshold

Yes, this is the reason to have these fields as pointers. It provides _some_ kind of way to denote that a sufficient microversion wasn't used. It's not great, but it's the most approachable way of balancing a typed language like Go and a dynamic result set like what "microversions" are doing.

jtopjian (approved):
LGTM - thank you!

Issue #2220, opened 2021-09-23
image service v2 should support updating image protection status


dustymabe:
It supports updating a few other fields (like [hidden](https://github.com/gophercloud/gophercloud/blob/94f0a20116d479db31a92860f61cdd9a02a25cd0/openstack/imageservice/v2/images/requests.go#L292-L304)). Let's add convenience wrappers for `protected` as well.

PR #2221, opened 2021-09-23
Image Service v2: Add Updating image.Protected field

dustymabe:
Now we should be able to do something like:

```
updateOpts := images.UpdateOpts{
    images.ReplaceImageProtected{
        NewProtected: false,
    },
}
image, err = images.Update(client, imageID, updateOpts).Extract()
```

Fixes https://github.com/gophercloud/gophercloud/issues/2220

### DIFF
--- a/acceptance/openstack/imageservice/v2/images_test.go
+++ b/acceptance/openstack/imageservice/v2/images_test.go
@@ -154,6 +154,7 @@ func TestImagesUpdate(t *testing.T) {
 		images.ReplaceImageName{NewName: image.Name + "foo"},
 		images.ReplaceImageTags{NewTags: newTags},
 		images.ReplaceImageMinDisk{NewMinDisk: 21},
+		images.ReplaceImageProtected{NewProtected: true},
 		images.UpdateImageProperty{
 			Op:    images.AddOp,
 			Name:  "hw_disk_bus",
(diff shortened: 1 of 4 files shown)


dustymabe:
I'm confident in this change working but not confident in my changes to the tests, can someone please help me verify those changes are appropriate and fail if tweaked. For some reason locally I couldn't get them to fail even if I put in bad values.

dustymabe:
> Build succeeded.
🎉

kayrus:
@dustymabe can you extend functional test to handle set true -> set false checks?

dustymabe:
> @dustymabe can you extend functional test to handle set true -> set false checks?

which file is that in?

jtopjian:
@dustymabe @kayrus is suggesting to either modify the current acceptance test or create a new one that flips protection on and off and confirm the value is correct each time.

jtopjian:
> I'm confident in this change working but not confident in my changes to the tests, can someone please help me verify those changes are appropriate and fail if tweaked. For some reason locally I couldn't get them to fail even if I put in bad values.

Looks like you've got the hang of it 🙂 

I noticed with the first OpenLab failure, it was due to Glance not wanting to delete a protected image, so this change does work. It's just a matter of extending the test a little bit and we should be good to go. Thanks for working on this.

dustymabe:
Yay. worked this time!

jtopjian:
To fulfill the PR requirement of showing the API service code that allows this change: according to the Glance code, it looks like all Image attributes are able to be updated, so this change is good.

https://github.com/openstack/glance/blob/25e60aea8e2fe11937e4b4d94e89c35cbb43940e/glance/api/v2/images.py#L640-L661

jtopjian:
@dustymabe Let me know if you're ready to have this reviewed/merged.

dustymabe:
> @dustymabe Let me know if you're ready to have this reviewed/merged.

+1 ready for review/merge

jtopjian (approved):
LGTM - thank you!

Issue #2222, opened 2021-09-30
Display detailed message for 404 not found error


schirevko:
Hello.
Similarly to https://github.com/gophercloud/gophercloud/issues/243, some APIs may return detailed error message for 404 Not Found errors. For example:
`{"itemNotFound": {"code": 404, "message": "Resource ... cannot be found."}}`
It could be useful if `ErrDefault404.Error()` would show this detailed message instead of `"Resource not found"`
(https://github.com/gophercloud/gophercloud/blob/master/errors.go#L182)

mandre:
Fixed via https://github.com/gophercloud/gophercloud/pull/2223.

PR #2223, opened 2021-09-30
Expose the real error message of NotFound error

schirevko:
This change exposes the error message of NotFound error from openstack
API, this is helpful to find out the reason of request failure.

For #2222


jtopjian (approved):
LGTM - thank you!

PR #2224, opened 2021-10-03
Support HttpHeaders in create_subscription

iurygregory:
For #1429

This commit adds support for HttpHeaders in the create_subscription. (#2201)
It was added during the Xena release of Ironic in [1] after the
initial support for create subscriptions [2]

We don't need to modify the response of get_subscription because
redfish doesn't return the value of the HttpHeaders for the enduser.

[1] https://review.opendev.org/c/openstack/ironic/+/806859
[2] https://review.opendev.org/c/openstack/ironic/+/801064

### DIFF
--- a/openstack/baremetal/v1/nodes/doc.go
+++ b/openstack/baremetal/v1/nodes/doc.go
@@ -182,6 +182,7 @@ Example to create a subscription
 		Context:     "MyContext",
 		Protocol:    "Redfish",
 		EventTypes:  ["Alert"],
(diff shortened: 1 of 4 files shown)


iurygregory:
I've tested locally in a setup with Ironic.

dtantsur (approved):

jtopjian:
@iurygregory sounds good - thanks! The OpenLab failure can be ignored.

iurygregory:
Thanks!

Issue #2208, opened 2021-08-31
Neutron V2: BGP Dynamic Routing


fuhrmannb:
I'd like to add support for Neutron BGP Dynamic Routing:

* BGP speakers
* BGP peers

API doc: https://docs.openstack.org/api-ref/network/v2/#bgp-dynamic-routing

mandre:
@shhgs also had a couple of PRs implementing BGP support that we merged already:
* #2241
* #2388

PR #2229, opened 2021-10-07
Add ListBGPSpeakers (continued)

shhgs:
Hi @jtopjian, 

I pull rebase this branch against the origin/master, and then run `push -f` without knowing that will close the [original pull request](https://github.com/gophercloud/gophercloud/pull/2228). So I create this one to replace old one. 

Thanks
-Jon

For #2208


shhgs:
There was an empty line in fixtures.go.

I am using vim-go, which is suppose to call gofmt everytime it writes the file. Anyway it is fixed.

PR #2231, opened 2021-10-09
nova: add support for keypair type

### LINKED ISSUE #2230: Use --key-type to create keypair
Hi, I'm having a problem creating keypair using GopherCloud, In nova openstack, I can select --key-type when I use nova keypair-add, but I don't see this parameter when I use CreateOpts in GopherCloud. How do I create keypair using GopherCloud if I want to select --key-type?

kayrus:
resolves #2230

### DIFF
--- a/openstack/compute/v2/extensions/keypairs/requests.go
+++ b/openstack/compute/v2/extensions/keypairs/requests.go
@@ -81,6 +81,10 @@ type CreateOpts struct {
 	// This requires microversion 2.10 or higher.
 	UserID string `json:"user_id,omitempty"`
 
+	// The type of the keypair. Allowed values are ssh or x509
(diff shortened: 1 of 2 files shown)


EmilienM:
Even if the tests fail, this one is fine and I confirm that there are keypair types: https://docs.openstack.org/api-ref/compute/?expanded=create-flavor-detail,create-or-import-keypair-detail
https://github.com/openstack/nova/blob/260cc737d0235d4637f232e6341c3fc527c5b3e9/nova/objects/keypair.py#L103
https://github.com/openstack/nova/blob/50fdbc752a9ca9c31488140ef2997ed59d861a41/nova/api/openstack/compute/views/keypairs.py#L28

Since the param is optional this won't break anything.

Thanks!

Issue #2236, opened 2021-10-15
Octavia: Add support for tags on pools


nikParasyr:
Octavia supports tags on pools since 2.5.

[Code](https://github.com/openstack/octavia/blob/0b84631948bd3c5eb25a0921d5fbad657036304f/octavia/api/v2/types/pool.py#L82)

PR #2237, opened 2021-10-15
Octavia: Add tags on pools

nikParasyr:
Add support for tags on octavia pools.
For #2236

[Code](https://github.com/openstack/octavia/blob/0b84631948bd3c5eb25a0921d5fbad657036304f/octavia/api/v2/types/pool.py#L82)


jtopjian:
@nikParasyr The OpenLab failure can be ignored - there's something wrong with the test configuration. 

This looks good to me. Is this ready to be merged?

Issue #2244, opened 2021-10-25
Errors types to handle 502 and 504 server errors


pbasov:
In errors.go 502 and 504 HTTP errors are missing, which makes handling them a pain in downstream code.

Related: https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1303

mandre:
Fixed via https://github.com/gophercloud/gophercloud/pull/2245

PR #2245, opened 2021-10-25
Add ability to handle 502 and 504 errors

pbasov:
This change introduces additional types to handle 502 Bad Gateway and 504 Gateway Timeout http errors

For #2244 
Related terraform-provider-openstack/terraform-provider-openstack#1303

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -156,7 +156,7 @@ type ErrDefault500 struct {
 	ErrUnexpectedResponseCode
 }
 
-// ErrDefault500 is the default error type returned on a 502 HTTP response code.
+// ErrDefault502 is the default error type returned on a 502 HTTP response code.
 type ErrDefault502 struct {
 	ErrUnexpectedResponseCode
 }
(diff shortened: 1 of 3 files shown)


pierreprinetti:
recheck

EmilienM:
The coverage decreased, could you please add tests in https://github.com/gophercloud/gophercloud/blob/master/testing/errors_test.go if that makes sense to you?

Thanks for your contribution!

EmilienM:
@pbasov hopefully this is fine for you, I pushed a patch to maintain unit testing coverage. Feel free to rework the patch that I sent.

Thanks

EmilienM:
oops sounds like coverage is still not happy. I'll take a look later unless you can check it as well.

errors.go
```diff
@@ -156,11 +156,21 @@ type ErrDefault500 struct {
 	ErrUnexpectedResponseCode
 }
 
+// ErrDefault500 is the default error type returned on a 502 HTTP response code.
```
mandre, line 159:
```suggestion
// ErrDefault502 is the default error type returned on a 502 HTTP response code.
```

mandre (commented):
Just a small typo, otherwise looks like it's in good shape to merge.

PR #2255, opened 2021-11-02
fix(octavia): PUT l7policy success response code is 202, not 200 (closes #2254)

### LINKED ISSUE #2254: PUT Octavia L7Policy should expect a 202 response code instead of 200
Per official docs https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=update-a-l7-policy-detail#update-a-l7-policy expected success status code should be 202 not 200.

> If the request is valid, the service returns the Accepted (202) response code.

Benzhaomin:
For #2254

https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=update-a-l7-policy-detail#update-a-l7-policy

### DIFF
--- a/openstack/loadbalancer/v2/l7policies/requests.go
+++ b/openstack/loadbalancer/v2/l7policies/requests.go
@@ -219,7 +219,7 @@ func Update(c *gophercloud.ServiceClient, id string, opts UpdateOptsBuilder) (r 
 		return
 	}
 	resp, err := c.Put(resourceURL(c, id), b, &r.Body, &gophercloud.RequestOpts{
-		OkCodes: []int{200},
+		OkCodes: []int{202},
 	})
 	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
 	return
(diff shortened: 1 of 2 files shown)


EmilienM:
LGTM

cc @pierreprinetti for final review before merge (and of course pending tests).
Note that acceptance tests are flacky, we won't take them seriously if they failed for another reason, not related to your patch.

jtopjian:
The original PR has the reference source code that was used for this implementation: #905. From there, the following block of code was used as the basis for the 200 return code:

https://github.com/openstack/octavia/blob/06bf5c58d5845f684fcaf933605ed112586eefc3/octavia/api/v2/controllers/l7policy.py#L201-L204

And it still looks to be a 200 code as of the master branch:

https://github.com/openstack/octavia/blob/master/octavia/api/v2/controllers/l7policy.py#L199-L202

This might want to be double-checked by looking at the raw HTTP requests in a devstack environment. If only a 202 return code was being returned, Gophercloud would have thrown an error within the past 4 years.

Benzhaomin:
right, docs are wrong, code is right :+1:

EmilienM:
@Benzhaomin @jtopjian I've pushed a patch into octavia to fix the doc: https://review.opendev.org/c/openstack/octavia/+/816393

Nice catch @jtopjian, I missed it while doing a first & fast review.

PR #2256, opened 2021-11-03
Create dependabot.yaml

EmilienM:
Get our dependencies automatically updates by Dependabot. It's free and easy to setup, let's just use it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+- package-ecosystem: gomod
+  directory: "/"
+  schedule:
+    interval: daily
+  open-pull-requests-limit: 10


EmilienM:
cc @pierreprinetti 
any thoughts?

EmilienM:
cc @mandre

mandre:
I'm OK giving it a try, but I'm a bit worried we'll get spammed with low value PRs bumping micro versions of libraries.

pierreprinetti:
I have mixed feelings WRT dependabot, but I have no strong arguments against.

/approve

EmilienM:
@mandre we don't have much libraries:

https://github.com/gophercloud/gophercloud/blob/master/go.mod#L6-L8

EmilienM:
@pierreprinetti :

> I have mixed feelings WRT dependabot, but I have no strong arguments against.

please elaborate, discussion is open.

jtopjian:
In my experience, dependabot for Go projects has had a really good signal/noise ratio. Combined with the low amount of external packages that Gophercloud needs, I don't think this is going to cause any problems.

mandre:
LGTM :+1:

PR #2238, opened 2021-10-20
Add List and Get method for bgp/speakers

shhgs:
For: Neutron V2: BGP Dynamic Routing #2208
 
https://github.com/shhgs/gophercloud/tree/20211020

```
 11:05:30 ~/ext/src/github.com/shhgs/gophercloud/openstack/networking/v2/extensions/bgp/speakers $ git diff --stat official/master                                                                      
 openstack/networking/v2/extensions/bgp/speakers/doc.go                   | 34 ++++++++++++++++++++++++++++++++++                                                                                                  
 openstack/networking/v2/extensions/bgp/speakers/requests.go              | 21 +++++++++++++++++++++
 openstack/networking/v2/extensions/bgp/speakers/results.go               | 20 +++++++++++++++++++-
 openstack/networking/v2/extensions/bgp/speakers/testing/doc.go           |  2 ++
 openstack/networking/v2/extensions/bgp/speakers/testing/fixture.go       | 63 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 openstack/networking/v2/extensions/bgp/speakers/testing/requests_test.go | 59 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 openstack/networking/v2/extensions/bgp/speakers/urls.go                  | 25 +++++++++++++++++++++++++
 7 files changed, 223 insertions(+), 1 deletion(-)
```

PR #2259, opened 2021-11-04
core: keep keystone reauth error on failed reauth

### LINKED ISSUE #2258: Core: reauthentication error is not reflected in original error
When reauthentication fails, the `ErrUnableToReauthenticate` struct contains only the initial openstack service error and the keystone error is missed:

https://github.com/gophercloud/gophercloud/blob/492ab2c54bfaf8cc328ae3dc0d04bd922b0a0d0c/provider_client.go#L474-L479

Since we don't catch the keystone error response, it is impossible to debug what was wrong during the reauth.

I propose to add `ErrReauth` struct member and print it along the service error, e.g.

```
func (e ErrUnableToReauthenticate) Error() string {
        e.DefaultErrString = fmt.Sprintf("Unable to re-authenticate: %s: %s", e.ErrOriginal, e.ErrReauth)
        return e.choseErrString()
}
```

Alternatively use the `Info` member:

```
                                err = client.Reauthenticate(prereqtok)
                                if err != nil {
                                        e := &ErrUnableToReauthenticate{}
                                        e.ErrOriginal = respErr
                                        e.Info = err.Error()
                                        return nil, e
                                }
```

kayrus:
resolves #2258

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -277,10 +277,11 @@ func (e ErrTimeOut) Error() string {
 type ErrUnableToReauthenticate struct {
 	BaseError
 	ErrOriginal error
+	ErrReauth   error
 }
 
 func (e ErrUnableToReauthenticate) Error() string {
-	e.DefaultErrString = fmt.Sprintf("Unable to re-authenticate: %s", e.ErrOriginal)
+	e.DefaultErrString = fmt.Sprintf("Unable to re-authenticate: %s: %s", e.ErrOriginal, e.ErrReauth)
 	return e.choseErrString()
 }
 
--- a/provider_client.go
+++ b/provider_client.go
@@ -475,6 +475,7 @@ func (client *ProviderClient) doRequest(method, url string, options *RequestOpts
 				if err != nil {
 					e := &ErrUnableToReauthenticate{}
 					e.ErrOriginal = respErr
+					e.ErrReauth = err
 					return nil, e
 				}
 				if options.RawBody != nil {


EmilienM:
LGTM, thanks!

kayrus:
@EmilienM 
The following tests failed:
* TestTrunk*
* TestStacks*

would you like to add skip flag for these tests?

EmilienM:
> @EmilienM The following tests failed:
> 
>     * TestTrunk*
> 
>     * TestStacks*
> 
> 
> would you like to add skip flag for these tests?

yes I'm working on restoring acceptance job into a good shape, here: https://github.com/gophercloud/gophercloud/pull/2252

I'll fix them up.

jtopjian:
One thing to keep in mind is that acceptance tests were passing less than 2 weeks ago (#2238, #2243). Either the issues are with the testing environment configuration or a recent upstream change.

EmilienM:
> One thing to keep in mind is that acceptance tests were passing less than 2 weeks ago (#2238, #2243). Either the issues are with the testing environment configuration or a recent upstream change.

yeah I tried to figure this out today and couldn't find the culprit. Did a bunch of investigations between jobs, like package diffs (APT + pip), etc. It would take more time to figure this out. I'll see what I can do.

kayrus:
@EmilienM I apologies for not creating a new issue, but I noticed that you don't update the https://github.com/gophercloud/gophercloud/blob/master/CHANGELOG.md after the PR is merged. I'd appreciate if you start maintaining this file. Thanks in advance.

EmilienM:
> @EmilienM I apologies for not creating a new issue, but I noticed that you don't update the https://github.com/gophercloud/gophercloud/blob/master/CHANGELOG.md after the PR is merged. I'd appreciate if you start maintaining this file. Thanks in advance.

yeah it's on my list

EmilienM:
https://github.com/gophercloud/gophercloud/pull/2264

PR #2261, opened 2021-11-05
octavia: add new pool monitor types

### LINKED ISSUE #2260: Octavia: add new pool members monitor types
New monitor types to use:
* TLS-HELLO
* UDP-CONNECT
* SCTP

code: https://github.com/openstack/octavia-lib/blob/3208f7fbcae31f84f0d8b0ff05ef63f5f7f2a075/octavia_lib/common/constants.py#L50-L56

kayrus:
resolves #2260


EmilienM:
this LGTM and safe to land without the acceptance test.
note that this won't work when the backend is OVN, as it doesn't support Health monitors. However, this would return a proper error if anyone tried to created a health monitor when LB is backed by OVN, so nothing to to in Gophercloud I think.
thanks!

PR #2268, opened 2021-11-10
Use regex to build list of acceptance tests

mandre:
This ensures new tests to be picked up.

Enable `magnum` in the compute acceptance test suite as the regex for
container now matches `containerinfra` which was entirely skipped
previously.


EmilienM:
recheck

mandre:
No longer needed.

PR #2272, opened 2021-11-10
Add coverage files into .gitignore

EmilienM:
We don't need them commited in git anyway.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 **/*.swp
 .idea
 .vscode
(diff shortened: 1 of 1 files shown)


EmilienM:
note: please ignore zuul failures. they should be gone by now.

mandre (commented):
If anyone else is wondering, it's the `script/coverage` script that generates the `testing_*.coverprofile` files.

https://github.com/gophercloud/gophercloud/blob/master/script/coverage#L8

Issue #1987, opened 2020-06-23
Core: DeepCopy method for the provider client


kayrus:
@jtopjian not sure if I miss this feature, but I need to initialize a provider client, then copy it for multiple go rotines. Does gophercloud support this?

jtopjian:
I haven't personally done that, so I'm not sure how easily it's natively supported.

What limitations are you hitting with creating a new provider client "from scratch"?

kayrus:
> What limitations are you hitting with creating a new provider client "from scratch"?

I don't want to issue an auth logic every time I create a provider. In addition it must support reauth functionality, because I don't know how long each go routine will run.

jtopjian:
> The only shared stuff I need is auth options + token

You're looking for a way to create a client that doesn't have to hit Keystone for the initial authorization, right? So you'd provide a scoped token and begin using the client?

That makes sense. I could have swore there was a way to set a scoped token, but maybe I'm mistaken. If I get some time, I can look into this, but as an aside, getting time to work on Gophercloud has been difficult for me this year.

kayrus:
I managed to solve the issue by creating a custom cloneAuth function, based on `v3auth`.

```
// global client
var client *gophercloud.ProviderClient

func cloneAuth(c *gophercloud.ProviderClient, endpoint string, opts tokens3.AuthOptionsBuilder, eo gophercloud.EndpointOpts, reauth bool) error {
        // Override the generated service endpoint with the one returned by the version endpoint.
        v3Client, err := openstack.NewIdentityV3(c, eo)
        if err != nil {
                return err
        }

        if endpoint != "" {
                v3Client.Endpoint = endpoint
        }

        if reauth {
                var result tokens3.CreateResult
                switch opts.(type) {
                case *ec2tokens.AuthOptions:
                        result = ec2tokens.Create(v3Client, opts)
                case *oauth1.AuthOptions:
                        result = oauth1.Create(v3Client, opts)
                default:
                        result = tokens3.Create(v3Client, opts)
                }

                err = c.SetTokenAndAuthResult(result)
                if err != nil {
                        return err
                }

                catalog, err := result.ExtractServiceCatalog()
                if err != nil {
                        return err
                }

                c.EndpointLocator = func(opts gophercloud.EndpointOpts) (string, error) {
                        return openstack.V3EndpointURL(catalog, opts)
                }
        }

        if opts.CanReauth() {
[...]

PR #2274, opened 2021-11-11
service_client: Accept per-call Context

pierreprinetti:
Adds the ability to pass per-call context.Context.
Adds a WithContext method to the ServiceClient, which will return a new
ServiceClient.

The context added to ServiceClient will be used in addition to the
ProviderClient's Context to make HTTP calls.

Example use:

```Go
	ctx := context.Background()
	allPages, err := servers.List(client.WithContext(ctx), servers.ListOpts{}).AllPages()
```

Most of the complexity of this patch is a consequence of the choice to merge the ServiceClient context with the ProviderClient's, instead of just overwriting it. This is another choice that is up for debate.

---

## Alternative B:

```Go
	ctx := context.Background()
	allPages, err := servers.List(client, servers.ListOpts{Context: ctx}).AllPages()
```

However I see one issue with this alternative: not all verbs currently accept an "options". See for example [`servers.Get(client *gophercloud.ServiceClient, id string)`](https://github.com/gophercloud/gophercloud/blob/1069770c51f3c757c6fc065ce3a43414b9db6825/openstack/compute/v2/servers/requests.go#L324). Such a change would not then be backwards-compatible.

---

## Alternative C:

Duplicate every user-facing function with a `WithContext` version

---

## Alternative D:

Wait until the next Major release to change every single user-facing function to accept a Context as the first argument.

---

Any more ideas?

For #1987

### DIFF
--- a/context.go
+++ b/context.go
@@ -1,0 +1,100 @@
+package gophercloud
+
+import (
+	"context"
+	"sync"
+	"time"
+)
+
+type mergeContext struct {
+	child  context.Context
+	parent context.Context
+	ch     chan struct{}
+
+	sync.Mutex
+	err error
+}
+
+func idempotentlyClose(ch chan struct{}) {
+	select {
+	case <-ch:
+	default:
+		close(ch)
+	}
+}
+
+// MergeContext returns a copy of ctx that also obeys a second parent.
+// MergeContext always returns a non-nil context.Context.
+func MergeContext(child, parent context.Context) (context.Context, context.CancelFunc) {
+	if child == nil {
+		if parent == nil {
+			return context.WithCancel(context.Background())
+		}
+		return context.WithCancel(parent)
+	}
+	if parent == nil {
+		return context.WithCancel(child)
+	}
+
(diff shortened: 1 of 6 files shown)


pierreprinetti:
The test failures look related; I'm investigating those next week. I am not merging anything anyway until there's some buy-in on the general idea of adding a context at the service level.

zhsj:
Hi, @pierreprinetti is this still on your radar? I'm still interested in this feature.

pierreprinetti:
> Hi, @pierreprinetti is this still on your radar? I'm still interested in this feature.

Yes it is! We definitely want to land support for call-scoped context, however we will probably rather opt for a non-breaking solution like new `XxxxWithContext` functions

pierreprinetti:
I now believe that this PR is not the way to go.

The ProviderClient's context provides cancellation for the lifespan of authentication. What we need is the ability to cancel a single call independently. There is no need to merge the two contexts.

The next proposal should focus on adding XxxWithContext functions that just pass their ctx to the http.Request.

PR #2283, opened 2021-11-26
blockstorage/v3/qos Add Get, Update and Delete Keys

nikParasyr:
Partially implements: #2139 

Implements `Get`, `Update` and `DeleteKeys`

code in this PR:

`Get`: [api docs](https://docs.openstack.org/api-ref/block-storage/v3/?expanded=show-a-qos-specification-details-detail#quality-of-service-qos-specifications-qos-specs), [code](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/volume/qos_specs.py#L241)

`Update`: [api docs](https://docs.openstack.org/api-ref/block-storage/v3/?expanded=set-keys-in-a-qos-specification-detail#quality-of-service-qos-specifications-qos-specs) [schema](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/api/schemas/qos_specs.py#L40), [code](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/volume/qos_specs.py#L70)

`DeleteKeys`: [api docs](https://docs.openstack.org/api-ref/block-storage/v3/?expanded=unset-keys-in-a-qos-specification-detail#quality-of-service-qos-specifications-qos-specs), [schema](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/api/schemas/qos_specs.py#L50), [code](https://github.com/openstack/cinder/blob/393c2e4ad90c05ebf28cc3a2c65811d7e1e0bc18/cinder/volume/qos_specs.py#L127)

### DIFF
--- a/acceptance/openstack/blockstorage/v3/qos_test.go
+++ b/acceptance/openstack/blockstorage/v3/qos_test.go
@@ -28,6 +28,24 @@ func TestQoS(t *testing.T) {
 	th.AssertNoErr(t, err)
 	th.AssertDeepEquals(t, qos2, getQoS2)
 
+	updateOpts := qos.UpdateOpts{
+		Consumer: qos.ConsumerBack,
+		Specs: map[string]string{
+			"read_iops_sec":  "40000",
+			"write_iops_sec": "40000",
+		},
+	}
+
+	expectedQosSpecs := map[string]string{
+		"consumer":       "back-end",
+		"read_iops_sec":  "40000",
+		"write_iops_sec": "40000",
+	}
+
+	updatedQosSpecs, err := qos.Update(client, qos2.ID, updateOpts).Extract()
+	th.AssertNoErr(t, err)
+	th.AssertDeepEquals(t, updatedQosSpecs, expectedQosSpecs)
+
 	listOpts := qos.ListOpts{
 		Limit: 1,
 	}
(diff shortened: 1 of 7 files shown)


nikParasyr:
@EmilienM not sure if i should ping you on MRs, now with the change of guard. In any case :point_down: 

**Note** for `Update`: This api call is a bit strange. This was discussed a bit during the [Create](https://github.com/gophercloud/gophercloud/pull/2140#issuecomment-814485392) where it was decided to have `Consumer` as a seperate field and the rest under `Specs`. `Name` cannot be updated. I decided to implement it in a similar way with Create. Unfortunately the response doesn't return the full QoS. The links above should be helpful.

EmilienM:
> @EmilienM not sure if i should ping you on MRs, now with the change of guard. In any case point_down
> 
> **Note** for `Update`: This api call is a bit strange. This was discussed a bit during the [Create](https://github.com/gophercloud/gophercloud/pull/2140#issuecomment-814485392) where it was decided to have `Consumer` as a seperate field and the rest under `Specs`. `Name` cannot be updated. I decided to implement it in a similar way with Create. Unfortunately the response doesn't return the full QoS. The links above should be helpful.

Hello!
You can ping me anytime, I have no problem with it.
Thanks for the PR, I'll add it to my review list.

PR #2293, opened 2021-12-10
actions: add a go mod tidy check

pierreprinetti:
Errors if dependencies were modified and `go mod tidy` was not run
before push.

### DIFF
--- a/.github/workflows/gomod.yml
+++ b/.github/workflows/gomod.yml
@@ -1,0 +1,11 @@
+on: [push, pull_request]
+name: go mod
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


.github/workflows/gomod.yml
```diff
@@ -0,0 +1,11 @@
+on: [push, pull_request]
+name: go mod
+jobs:
+  test:
+    runs-on: ubuntu-20.04
```
EmilienM, line 5: switch to latest?

EmilienM (changes requested):

PR #2294, opened 2021-12-10
Bump Go to 1.14

pierreprinetti:
This PR contains several little changes to the test definitions:

* Test against Go 1.14 (the oldest supported version) and Go tip
* Let the Goveralls GH action deal with the Goveralls installation
* Use ubuntu-latest in tests instead of pinning to an arbitrary version
* Remove a (probably) unneeded explicit install of `crypto/ssh`
* Run scripts/unittest in a separate workflow, to increase readability and parallelise work

Additionally, it turns out that we are using in `testhelper` a method that was introduced in Go 1.14. With this PR, I propose to bump the minimum acceptable version of Go to v1.14 in `go.mod`.

### DIFF
--- a/.github/workflows/reauth-retests.yaml
+++ b/.github/workflows/reauth-retests.yaml
@@ -1,0 +1,22 @@
+on: [push, pull_request]
+name: Reauth retests
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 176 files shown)


EmilienM:
LGTM
will wait on CI to merge it.

pierreprinetti:
@EmilienM Please note that I have now bumped Go to v1.14 in go.mod

PR #2296, opened 2021-12-14
Remove code for CDN (poppy)

EmilienM:
OpenStack CDN (Poppy) isn't maintained anymore, last commit was 5 years
ago:
https://opendev.org/x/poppy/commits/branch/master

This patch removes all the code that called CDN v1 API, which is not
tested, and the upstream project doesn't seem to be alive anymore.

This effort is made for helping to maintain Gophercloud and clean the
things we don't need/use anymore.


EmilienM:
@jtopjian @jrperritt would you be against it?

jrperritt:
> @jrperritt would you be against it?

No, sounds fine to me.

jtopjian:
Ditto - I am in favour of removing this.

PR #2297, opened 2021-12-14
Remove support for blockstorage v1

EmilienM:
This can't be tested today (it was deprecated in Juno!) and even the
Python SDK doesn't support it anymore:
https://github.com/openstack/openstacksdk/tree/master/openstack/block_storage

I think this is time to remove the code for Cinder v1.


EmilienM:
@jtopjian do you have thoughts on this one?

pierreprinetti:
Note: this is a backwards-incompatible change and can't be merged without a major version bump.

jtopjian:
> @jtopjian do you have thoughts on this one?

There are some discussions in older, closed issues where I am firmly against this - specifically the removal of blockstorage v1. With my (now very dusty) operator hat on, I tried to keep all sorts of older OpenStack environments in mind, regardless of what was officially supported. My goal was for those environments to be able to use tools like Terraform without putting more stress and pressure on the admins/operators to upgrade their environment.

For blockstorage v1 specifically, I think times are now different. It's beyond old and if there was an environment still leveraging v1, even I'd be comfortable saying "dude, you need to upgrade" 🙂 

So I am in favour of removing this. 

But I wanted to provide some previous guidelines of how older API packages were handled in Gophercloud. Even if the API version is no longer officially supported, I find it beneficial for the client-side API calls to still be kept in Gophercloud so developers can create tools that work with older environments.

PR #2307, opened 2022-01-06
[CI-v2] Add functional-basic job

EmilienM:
This runs a Github Action to:

* Deploy Devstack with defaults (minimum services)
* Run acceptance/openstack which is a minimal set of tests

### DIFF
--- a/.github/workflows/functional-basic.yaml
+++ b/.github/workflows/functional-basic.yaml
@@ -1,0 +1,59 @@
+name: functional-basic
+on:
+  pull_request:
+    paths-ignore:
+      - '^docs/|\.md$|^(?:.*/)?(?:\.gitignore|LICENSE)$'
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
+  functional-basic:
+    strategy:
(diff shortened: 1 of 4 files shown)


EmilienM:
@mandre @pierreprinetti ready for review, please

mandre:
Should we perhaps upload the artifacts for the unit tests in any cases? Otherwise it's hard to know which tests were run as part of the jobs.

EmilienM:
> Should we perhaps upload the artifacts for the unit tests in any cases? Otherwise it's hard to know which tests were run as part of the jobs.

Yeah I need to investigate that but for now we can rely on the logs that are in the Github actions, we can see which tests fail and not

mandre:
> > Should we perhaps upload the artifacts for the unit tests in any cases? Otherwise it's hard to know which tests were run as part of the jobs.
> 
> Yeah I need to investigate that but for now we can rely on the logs that are in the Github actions, we can see which tests fail and not

Sure, we can iterate on that. No objection from me for merging your PR as is.

PR #2312, opened 2022-01-08
acceptance: store results into a log file

EmilienM:
This log file will be collected by Github Actions artifacts.

### DIFF
--- a/script/acceptancetest
+++ b/script/acceptancetest
@@ -7,6 +7,12 @@ source `dirname $0`/stackenv
 timeout="60m"
 failed=
 
+if [[ -z "${LOG_DIR}" ]]; then
+    echo "LOG_DIR not set, will set a temp directory"
+    LOG_DIR=/tmp/devstack-logs
+fi
+mkdir -p ${LOG_DIR}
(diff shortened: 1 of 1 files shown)


script/acceptancetest
```diff
@@ -7,6 +7,12 @@ source `dirname $0`/stackenv
 timeout="60m"
 failed=
 
+if [[ -z "${LOG_DIR}" ]]; then
+    echo "LOG_DIR not set, will create a temp directory"
+    mkdir -p /tmp/devstack-logs
```
HappyFX, line 12: It can be safety to move mkdir outer the if state, in that case `${LOG_DIR}` included from job vars will be ready in any cases:

```bash
if [[ -z "${LOG_DIR}" ]]; then
    echo "LOG_DIR not set, will set a temp directory"
    LOG_DIR=/tmp/devstack-logs
fi
mkdir -p ${LOG_DIR}
```

EmilienM: Good idea!

mandre:
Looks good, thanks.

PR #2313, opened 2022-01-09
[CI-v2] Add compute job with Github Action

EmilienM:
This runs a Github Action to:

* Deploy Devstack with Nova and Cinder
* Run acceptance/openstack/compute
* Fix aggregate test to a hostname instead of the FQDN. Nova takes
  hostname: https://github.com/openstack/nova/blob/master/nova/compute/api.py#L6454
* Skip UsageTest and will fix later. The test doesn't pass for now, the
  usage is null everywhere we need to figure out why.
* Switch QuotaSet tests to use API v3 of Keystone instead of v2, which
  was removed in previous versions of OpenStack.
* Unskip all tests except `TestUsageSingleTenant` which seems complicated and potentially broken at API level, so we test almost everything :-)

### DIFF
--- a/.github/workflows/functional-compute.yaml
+++ b/.github/workflows/functional-compute.yaml
@@ -1,0 +1,61 @@
+name: functional-compute
+on:
+  pull_request:
+    paths:
+      - '^.*compute.*$'
+      - '.github/workflows/functional-compute.yaml'
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
+  functional-compute:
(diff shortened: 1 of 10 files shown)


mandre (approved):
LGTM

Just to capture what I was discussing with @EmilienM, we didn't notice the error previously in the aggregates acceptance tests because we got lucky on openlab, where the hostname matched the FQDN. This is no longer the case on the VMs started by github action.

Issue #2215, opened 2021-09-15
Swift V1: Cannot delete ACL rules


kayrus:
I noticed that it is impossible to delete ACL rules. Swift accepts empty header values for this, e.g.:

```
curl hsot -H 'X-Container-Read: '
```

I found at least two problems and I made a local fix:

```diff
diff --git a/vendor/github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers/requests.go b/vendor/github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers/requests.go
index bea0d95..d9670a6 100644
--- a/vendor/github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers/requests.go
+++ b/vendor/github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers/requests.go
@@ -154,18 +154,18 @@ type UpdateOptsBuilder interface {
 type UpdateOpts struct {
        Metadata               map[string]string
        RemoveMetadata         []string
-       ContainerRead          string `h:"X-Container-Read"`
-       ContainerSyncTo        string `h:"X-Container-Sync-To"`
-       ContainerSyncKey       string `h:"X-Container-Sync-Key"`
-       ContainerWrite         string `h:"X-Container-Write"`
-       ContentType            string `h:"Content-Type"`
-       DetectContentType      bool   `h:"X-Detect-Content-Type"`
-       RemoveVersionsLocation string `h:"X-Remove-Versions-Location"`
-       VersionsLocation       string `h:"X-Versions-Location"`
-       RemoveHistoryLocation  string `h:"X-Remove-History-Location"`
-       HistoryLocation        string `h:"X-History-Location"`
-       TempURLKey             string `h:"X-Container-Meta-Temp-URL-Key"`
-       TempURLKey2            string `h:"X-Container-Meta-Temp-URL-Key-2"`
+       ContainerRead          *string `h:"X-Container-Read"`
+       ContainerSyncTo        *string `h:"X-Container-Sync-To"`
+       ContainerSyncKey       *string `h:"X-Container-Sync-Key"`
+       ContainerWrite         *string `h:"X-Container-Write"`
+       ContentType            string  `h:"Content-Type"`
+       DetectContentType      bool    `h:"X-Detect-Content-Type"`
+       RemoveVersionsLocation string  `h:"X-Remove-Versions-Location"`
+       VersionsLocation       string  `h:"X-Versions-Location"`
[...]

PR #2315, opened 2022-01-10
Identity v2&v3: add RequestOpts.OmitHeaders, remove blank header

### LINKED ISSUE #2314: OpenStack V2 Authentication Failed since RequestOpts.MoreHeaders does not make sense
In `provider_client.go:322`, RequestOpts.Header says it will be omitted if it is a blank value. 
```
// RequestOpts customizes the behavior of the provider.Request() method.
type RequestOpts struct {
	// MoreHeaders specifies additional HTTP headers to be provide on the request. If a header is
	// provided with a blank value (""), that header will be *omitted* instead: use this to suppress
	// the default Accept header or an inferred Content-Type, for example.
	MoreHeaders map[string]string
}
```

But in commit 29581a2fc7076bae722d4b681f9cfaa467aa3e09 [Swift V1: fix setting empty headers (#2218)], the method locate in  ​`provider_client.go:411` has been changed, and blank header no longer been removed.

This caused an `Authentication Failed` error when trying to get an v2auth client due to an empty `X-Auth-Token` header.
This symptom is imported after v0.21.0.

`openstack/identity/v2/tokens/requests.go:94`:
```
resp, err := client.Post(CreateURL(client), b, &r.Body, &gophercloud.RequestOpts{
	OkCodes:     []int{200, 203},
	MoreHeaders: map[string]string{"X-Auth-Token": ""},
})
```

How can I make it correct? Is there any suggestion or work around?

freedywu:
Fixes #2314

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -325,10 +325,12 @@ type RequestOpts struct {
 	// OkCodes contains a list of numeric HTTP status codes that should be interpreted as success. If
 	// the response has a different code, an error will be returned.
 	OkCodes []int
-	// MoreHeaders specifies additional HTTP headers to be provide on the request. If a header is
-	// provided with a blank value (""), that header will be *omitted* instead: use this to suppress
-	// the default Accept header or an inferred Content-Type, for example.
+	// MoreHeaders specifies additional HTTP headers to be provided on the request.
+	// MoreHeaders will be overridden by OmitHeaders
 	MoreHeaders map[string]string
+	// OmitHeaders specifies the HTTP headers which should be omitted.
+	// OmitHeaders will override MoreHeaders
+	OmitHeaders []string
 	// ErrorContext specifies the resource error type to return if an error is encountered.
 	// This lets resources override default error messages based on the response status code.
 	ErrorContext error
@@ -396,7 +398,8 @@ func (client *ProviderClient) doRequest(method, url string, options *RequestOpts
 		req = req.WithContext(client.Context)
 	}
 
-	// Populate the request headers. Apply options.MoreHeaders last, to give the caller the chance to
+	// Populate the request headers.
+	// Apply options.MoreHeaders and options.OmitHeaders, to give the caller the chance to
 	// modify or omit any header.
 	if contentType != nil {
 		req.Header.Set("Content-Type", *contentType)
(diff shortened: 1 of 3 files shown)


provider_client.go
```diff
@@ -412,6 +412,10 @@ func (client *ProviderClient) doRequest(method, url string, options *RequestOpts
 		}
 	}
 
+	for _, v := range options.OmitHeaders {
+		req.Header.Del(v)
```
pierreprinetti, line 416: Is there any concern that the headers from `client.AuthenticatedHeaders()` will override this setting (line 420-422)?

freedywu: Before commit 29581a2fc7076bae722d4b681f9cfaa467aa3e09, `client.AuthenticatedHeaders()` is prior to `options.MoreHeaders`. In my case, `options.OmitHeaders` makes up for the previous `options.MoreHeaders` and should inherit the original order

freedywu:
@pierreprinetti

provider_client.go
kayrus: yeah, we expected that something could be broken, see https://github.com/gophercloud/gophercloud/issues/2215#issuecomment-921114591

jtopjian: I'm not sure that something _is_ broken yet, though. I think there needs to be some investigation and comparison to @freedywu's use-case that is triggering this and the basic acceptance test that passes without changes required.

jtopjian: > Is there any concern that the headers from client.AuthenticatedHeaders() will override this setting

I _think_ it should be okay given that the order of the existing code has `MoreHeaders` first adding an empty `X-Auth-Token` and then `client.AuthenticatedHeaders` adds headers on top of it. Definitely warrants running this PR through the Keystone suite of acceptance tests, though.

jtopjian:
Per discussion in #2314, it looks like the root cause is a _very_ old version of Keystone triggering this issue. 

In my opinion, if the Keystone suite of acceptance tests passes with this PR, it should be fine to merge. It's not a lot of work to help get this functionality fixed and adds a feature that may be useful in other areas. And right now, the only place `OmitHeaders` is used is in the token creation functions, so that limits the scope of what's being modified.

EmilienM:
@freedywu could you please rebase on master, so we get the new CI job for Keystone? Thanks

freedywu:
> @freedywu could you please rebase on master, so we get the new CI job for Keystone? Thanks

Sure, thanks for noticing.

freedywu:
Hi @EmilienM , my commit has been rebased, could you have a look at it?

mandre (approved):
LGTM, the code looks safe, and the identity acceptance tests worked as expected. 
Any objection to merge?

PR #2318, opened 2022-01-11
[CI-v2] Add networking job with Github Action

EmilienM:
This runs a Github Action to:

* Deploy Devstack with Neutron
* Run acceptance/openstack/networking
* Unskip a lot of tests
* Do not test FWaaS v1, too old and will be removed from Gophercloud via #2320
* Skip FWAAS v2 testing after Ussuri (service was removed); and deploy it only on
  Ussuri and Train, so it can still be tested.
* Fix a test for FWAAS, where ICMP rule can't have a port in source
* Skip TestAgentsRUD because it can't work with OVN (default in devstack
  now). This test will have to be reworked in the future.
* Fix TestDNSPortCRUDL test (some typos)
* Skip TestLayer3RouterAgents because it can't work with OVN (default
  in devstack now). This test will have to be rewored in the future.
* Skip TestDNSFloatingIPCRDL which doesn't work with ML2/OVN
* Skip LBAAS (v1 and v2) from Neutron API, it was removed and replaced by Octavia.
  We'll need to deprecate and remove it from Gophercloud as well.
* In TestMTUNetworkCRUDL, reduce the MTU to 1440. With OVN + Geneve, we
  can't go above 1442, so let's set 1440 for the test to be safe.
* Skip VlanTransparent, Trunk and QoS testing if the extension is not there.

### DIFF
--- a/.github/workflows/functional-networking.yaml
+++ b/.github/workflows/functional-networking.yaml
@@ -1,0 +1,72 @@
+name: functional-networking
+on:
+  pull_request:
+    paths:
+      - '**networking**'
+  schedule:
+    - cron: '0 0 * * *'
+jobs:
+  functional-networking:
+    strategy:
+      fail-fast: false
+      matrix:
+        name: ["master"]
+        openstack_version: ["master"]
+        ubuntu_version: ["20.04"]
+        devstack_conf_overrides: [""]
+        include:
+          - name: "xena"
+            openstack_version: "stable/xena"
+            ubuntu_version: "20.04"
+            devstack_conf_overrides: ""
+          - name: "wallaby"
+            openstack_version: "stable/wallaby"
+            ubuntu_version: "20.04"
+            devstack_conf_overrides: ""
+          - name: "victoria"
+            openstack_version: "stable/victoria"
+            ubuntu_version: "20.04"
+            devstack_conf_overrides: ""
+          - name: "ussuri"
+            openstack_version: "stable/ussuri"
+            ubuntu_version: "18.04"
+            devstack_conf_overrides: |
+              enable_plugin neutron-fwaas https://git.openstack.org/openstack/neutron-fwaas stable/ussuri
+          - name: "train"
+            openstack_version: "stable/train"
+            ubuntu_version: "18.04"
+            devstack_conf_overrides: |
+              enable_plugin neutron-fwaas https://git.openstack.org/openstack/neutron-fwaas stable/train
+    runs-on: ubuntu-${{ matrix.ubuntu_version }}
+    name: Deploy OpenStack ${{ matrix.name }} with Neutron and run networking acceptance tests
+    steps:
+      - name: Checkout Gophercloud
+        uses: actions/checkout@v2
+      - name: Deploy devstack
+        uses: EmilienM/devstack-action@v0.6
+        with:
+          branch: ${{ matrix.openstack_version }}
+          conf_overrides: |
+            Q_ML2_PLUGIN_EXT_DRIVERS=qos,port_security,dns_domain_keywords
+            enable_plugin neutron-vpnaas https://opendev.org/openstack/neutron-vpnaas ${{ matrix.openstack_version }}
+            ${{ matrix.devstack_conf_overrides }}
+          enabled_services: 'neutron-dns,neutron-qos,neutron-segments,neutron-trunk,neutron-uplink-status-propagation,neutron-network-segment-range,neutron-port-forwarding'
+      - name: Checkout go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15'
+      - name: Run Gophercloud acceptance tests
+        run: ./script/acceptancetest
+        env:
+          DEVSTACK_PATH: ${{ github.workspace }}/devstack
+          ACCEPTANCE_TESTS_FILTER: "^.*networking.*$"
+          OS_BRANCH: ${{ matrix.openstack_version }}
+      - name: Generate logs on failure
+        run: ./script/collectlogs
+        if: failure()
+      - name: Upload logs artifacts on failure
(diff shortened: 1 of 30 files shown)


acceptance/openstack/networking/v2/extensions/agents/agents_test.go
```diff
@@ -70,9 +70,6 @@ func TestAgentsRUD(t *testing.T) {
 	th.AssertNoErr(t, err)
 	th.AssertEquals(t, agent.Description, allAgents[0].Description)
 
-	// skip this part
```
mandre, line 73: Any reason you didn't remove the `t.Skip("Skip DHCP agent deletion")` below?

.github/workflows/functional-networking.yaml
```diff
@@ -0,0 +1,59 @@
...
+          branch: ${{ matrix.openstack_version }}
+      - name: Checkout go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15'
+      - name: Run Gophercloud acceptance tests
+        run: ./script/acceptancetest
+        env:
+          DEVSTACK_PATH: ${{ github.workspace }}/devstack
+          ACCEPTANCE_TESTS_FILTER: "^.*networking.*$"
```
mandre, line 50: Something's wrong, if you look at the jobs it's passing a different `ACCEPTANCE_TESTS_FILTER` env var to the acceptancetest, with a value of `^acceptance/openstack$`.

https://github.com/gophercloud/gophercloud/pull/2318/checks

.github/workflows/functional-networking.yaml
```diff
@@ -0,0 +1,59 @@
...
+      - name: Checkout Gophercloud
+        uses: actions/checkout@v2
+      - name: Deploy devstack
+        uses: EmilienM/devstack-action@v0.6
+        with:
+          branch: ${{ matrix.openstack_version }}
+      - name: Checkout go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15'
```
mandre, line 45: This step is also weird. It installs go 1.17 where I would expect it to install 1.15:

```
Run actions/setup-go@v2
  with:
    go-version: ^1.15
    stable: true
    token: ***
Setup go stable version spec ^1.15
Found in cache @ /opt/hostedtoolcache/go/1.17.5/x64
Added go to the path
Successfully setup go version ^1.15
go version go1.17.5 linux/amd64
```

https://github.com/gophercloud/gophercloud/runs/4781723151?check_suite_focus=true#step:4:1

acceptance/openstack/networking/v2/extensions/agents/agents_test.go
EmilienM: I think it's causing problem because there is no DHCP agent when deploying with OVN (the default in recent devstack). Also another issue is that if you remove the agent, later in the tests, you'll have networking issue if you start a VM.
Honestly, this can be re-worked later. But I think it's fine to skip it now like it was the case already.

acceptance/openstack/networking/v2/extensions/qos/policies/policies_test.go
```diff
@@ -16,6 +17,12 @@ func TestPoliciesCRUD(t *testing.T) {
 	client, err := clients.NewNetworkV2Client()
 	th.AssertNoErr(t, err)
 
+	extension, err := extensions.Get(client, "qos").Extract()
```
mandre, line 20: Yes! I like that we check the capability of the cloud to enable tests dynamically. It's a nice pattern to follow.
Much better than what I originally did in https://github.com/gophercloud/gophercloud/pull/2268/commits/22e65202a3f4837701247d4eeaa35e5ddd308262

.github/workflows/functional-networking.yaml
```diff
@@ -0,0 +1,73 @@
...
+        uses: actions/checkout@v2
+      - name: Deploy devstack
+        uses: EmilienM/devstack-action@v0.6
+        with:
+          branch: ${{ matrix.openstack_version }}
+          conf_overrides: |
+            Q_ML2_PLUGIN_EXT_DRIVERS=qos,port_security,dns_domain_keywords
+            enable_plugin neutron-vpnaas https://opendev.org/openstack/neutron-vpnaas ${{ matrix.openstack_version }}
+            ${{ matrix.devstack_conf_overrides }}
+          enabled_services: 'neutron-dns,neutron-qos,neutron-segments,neutron-trunk,neutron-uplink-status-propagation,neutron-network-segment-range,neutron-port-forwarding'
```
mandre, line 54: It seems like the `port-forwarding` extension is not enabled: https://github.com/gophercloud/gophercloud/runs/4928771066?check_suite_focus=true#step:5:3055
Same goes for the `vlan-transparent` extension: https://github.com/gophercloud/gophercloud/runs/4928771066?check_suite_focus=true#step:5:4069

On train, in addition to the above extensions, we're also missing:
- dns-integration: https://github.com/gophercloud/gophercloud/runs/4928771304?check_suite_focus=true#step:5:2589
- qos: https://github.com/gophercloud/gophercloud/runs/4928771304?check_suite_focus=true#step:5:3353
- trunk: https://github.com/gophercloud/gophercloud/runs/4928771304?check_suite_focus=true#step:5:3574

EmilienM: Honestly, I'm not sure how to enable VLAN transparent, and I'm not even sure it works for OVN.
As for Train, I'm sure it's because of the default Neutron plugins (ML2/OVS). We could change that in the future if you don't mind, but I think the coverage is pretty good now.

.github/workflows/functional-networking.yaml
EmilienM: I'll investigate these issues later.

PR #2319, opened 2022-01-11
acceptance: add pipefail to `acceptancetest` script

EmilienM:
Since we now redirect the output to a log file, we need to enable
pipefail otherwise RC will always be 0 which is wrong if tests fail.

### DIFF
--- a/script/acceptancetest
+++ b/script/acceptancetest
@@ -1,6 +1,7 @@
 #!/bin/bash
 #
 set -x
+set -o pipefail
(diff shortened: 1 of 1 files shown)


EmilienM:
@mandre @pierreprinetti @jtopjian if you can look this one quickly please.

mandre (approved):
Ah, it didn't occur to me during the last review we would have a problem with return codes. LGTM

PR #2334, opened 2022-01-25
Github workflows: update REGEX to trigger jobs

EmilienM:
We thought Github supported usual REGEX syntax but it's not, see:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

So we update the functional workflows so the jobs run when the right
files are modified.

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -2,8 +2,7 @@ name: functional-baremetal
 on:
   pull_request:
     paths:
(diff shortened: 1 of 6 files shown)


EmilienM:
ping @mandre

mandre (approved):
Nice catch. We're missing the `functional-basic.yaml` and `functional-basic.yaml` workflows that I'm going to take care of in a subsequent patch.

PR #2339, opened 2022-01-31
Add network provider extension

### LINKED ISSUE #2338: Add support for network provider extended attributes
Currently it is not possible to set [Provider extended attributes](https://docs.openstack.org/api-ref/network/v2/index.html?expanded=show-network-details-detail#provider-extended-attributes). This is only allowed via `segments` by [Multiple provider extension](https://docs.openstack.org/api-ref/network/v2/index.html?expanded=show-network-details-detail#multiple-provider-extension).
Let's extend it and support both options.

[CODE](https://github.com/openstack/neutron/blob/master/neutron/extensions/providernet.py)
[API](https://docs.openstack.org/api-ref/network/v2/index.html)

ZdenekPesek:
Fixes #2338

### DIFF
--- a/openstack/networking/v2/extensions/provider/requests.go
+++ b/openstack/networking/v2/extensions/provider/requests.go
@@ -7,7 +7,10 @@ import (
 // CreateOptsExt adds a Segments option to the base Network CreateOpts.
 type CreateOptsExt struct {
 	networks.CreateOptsBuilder
-	Segments []Segment `json:"segments,omitempty"`
+	NetworkType     string    `json:"provider:network_type,omitempty"`
+	PhysicalNetwork string    `json:"provider:physical_network,omitempty"`
+	SegmentationID  int       `json:"provider:segmentation_id,omitempty"`
+	Segments        []Segment `json:"segments,omitempty"`
 }
 
 // ToNetworkCreateMap adds segments to the base network creation options.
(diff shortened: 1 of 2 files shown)


EmilienM:
CI failed on stable versions of OpenStack. Not related to this patch so I retriggered the jobs to see if it passes this time.

openstack/networking/v2/extensions/provider/results.go
```diff
@@ -24,7 +19,7 @@ type NetworkProviderExt struct {
 	// segment depends on the segmentation model defined by network_type. For
 	// instance, if network_type is vlan, then this is a vlan identifier;
 	// otherwise, if network_type is gre, then this will be a gre key.
-	SegmentationID string `json:"-"`
+	SegmentationID int `json:"provider:segmentation_id"`
```
EmilienM, line 22: This is not backward compatible, right?
@mandre @pierreprinetti what do you think?

mandre: Correct, this isn't backward compatible. If we were to do this, it should be in its own PR for which we add the `breaking` label.

However, digging into this a bit, it seems the int vs string issue was previously discussed in #69 and we settled on the string type.

To quote @jtopjian :
> another argument for a string is in the case of zero value.
>
> Neutron is returning "null" rather than 0. Having a segmentation_id of 0 might cause [unexpected results](https://www.google.ca/?ion=1&espv=2#q=vlan%200).

jtopjian: Yeah, there are three reasons a string might be used as the return/result type:

1. It truly is a string.
2. Duct tape to handle a wide array of return values (null vs 0).
3. Supporting multiple OpenStack versions and return types. I think the [hypervisors API extension](https://github.com/gophercloud/gophercloud/blob/master/openstack/compute/v2/extensions/hypervisors/results.go) is the winner for this. 🙂

pierreprinetti:
Labeling as semver:major as the current implementation introduces a breaking change.

openstack/networking/v2/extensions/provider/results.go
ZdenekPesek: Hello @EmilienM , @mandre , @jtopjian 
I have updated PR to keep the `SegmentationID string` in `NetworkProviderExt` struct.
Although I am little bit confused we have string here but inside `Segment` struct there is `SegmentationID  int`. 

Hopefully now it all make sense together with the test :)

I would appreciate some review.
Thank you

EmilienM:
This PR has been staled for a very long time, I'll close it. Feel free to re-open!

PR #2344, opened 2022-02-08
Add disk/by_id to introspection data

redixin:
Disk by-id (e.g. /dev/disk/by-id/nvme-eui.0134314821d7dd35)

ById is most reliable way to identify disk because disk Name and
ByPath may be changed after reboot. At the same time ById is
always the same for same disk because it based on device serial
number.

### DIFF
--- a/openstack/baremetalintrospection/v1/introspection/results.go
+++ b/openstack/baremetalintrospection/v1/introspection/results.go
@@ -236,6 +236,7 @@ type RootDiskType struct {
 	Model              string `json:"model"`
 	Name               string `json:"name"`
 	ByPath             string `json:"by_path"`
+	ById               string `json:"by_id"`
 	Rotational         bool   `json:"rotational"`
 	Serial             string `json:"serial"`
 	Size               int64  `json:"size"`


EmilienM:
@redixin could you please provide a link of the API where the parameter is document? I didn't find it (quickly)
thanks

redixin:
@EmilienM it's brand new and not yet merged into upstream `ipa`

Also those parameters are completely up to `ipa` and I'm not sure if they are even documented somewhere in bare metal api docs.

EmilienM:
@jtopjian I'm not sure what to do in this situation. Any thoughts?

EmilienM:
@redixin do you have an update on the upstream code? Could you link it please? If it has merged let us know.

dtantsur:
> I'm not sure if they are even documented somewhere in bare metal api docs.

Not in the API docs, but they are documented in https://docs.openstack.org/ironic-python-agent/latest/admin/how_it_works.html#hardware-inventory.

I cannot find any reference to by-id in the code, nor in any open patches.

pierreprinetti:
I am closing this PR as stale. @redixin (or anyone else) feel free to reopen (or ask us to reopen) in case you can bring the information that was required by the reviewers. Thank you!

PR #2329, opened 2022-01-25
[CI-v2] Add dns job with Github Action

mandre:
* Deploy Designate with Devstack
* Run dns acceptance tests
* Logs failures if any

PR #2351, opened 2022-02-14
zuul: remove networking job

EmilienM:
we replaced it by github actions, coverage is better now and more
stable.


.zuul.yaml
```diff
@@ -360,48 +239,8 @@
...
     vars:
-      # the list of all available tests can generated by:
-      # find acceptance/openstack -name '*_test.go' -exec dirname {} \; | sort -n | uniq -c | awk '{print $2}'
       acceptance_tests:
         - acceptance/openstack
-        - acceptance/openstack/blockstorage
-        - acceptance/openstack/blockstorage/extensions
-        - acceptance/openstack/blockstorage/v3
-        - acceptance/openstack/compute/v2
-        - acceptance/openstack/container/v1
```
mandre, line 371: Shouldn't we leave in the legacy job services that haven't yet been ported to github action?

From that list, I think we're missing:
- container
- orchestration
- placement
- sharedfilesystems

mandre (commented):
Cool, the dns, loadbalancer and networking jobs were moved to github actions in #2329, #2328 and #2318 respectively.

I'm less sure about removing all services from the `gophercloud-acceptance-test-legacy` job. We're decreasing test coverage until we port the remaining services to github action. IMO we should keep the ones we haven't ported yet in that job.

.zuul.yaml
EmilienM: done

PR #2361, opened 2022-02-25
Adding CI job for testing sharedfilesystems (Manila)

EmilienM:
Adding a Github Action to cover Manila testing.


mandre (approved):
Great, with this patch we should be able to retire the `gophercloud-acceptance-test-storage` test suite running on openlab too.

mandre:
> Great, with this patch we should be able to retire the `gophercloud-acceptance-test-storage` test suite running on openlab too.

Removing zuul storage jobs with https://github.com/gophercloud/gophercloud/pull/2362

PR #2379, opened 2022-04-08
[CI-v2] Add containerinfra job with Github Action

mandre:
- Deploy Magnum with Devstack
- Run messaging acceptance tests
- Logs failures if any


mnaser:
FWIW, swarm is probably not really supported so a functional test that tries to do that will not go very far :)

PR #2384, opened 2022-04-20
baremetal: add API to set maintenance and maintenance reason

### LINKED ISSUE #2383: Missing maintenance API for baremetal Node
For very historical reasons, Ironic requires a special API to set maintenance and maintenance_reason.
While the maintenance itself can be set directly (for even more historical reasons), the reason cannot.

dtantsur:
Fixes #2383

See https://docs.openstack.org/api-ref/baremetal/?expanded=set-maintenance-flag-detail,clear-maintenance-flag-detail#set-maintenance-flag for API reference.

### DIFF
--- a/acceptance/openstack/baremetal/v1/nodes_test.go
+++ b/acceptance/openstack/baremetal/v1/nodes_test.go
@@ -68,6 +68,38 @@ func TestNodesUpdate(t *testing.T) {
 	th.AssertEquals(t, updated.Maintenance, true)
 }
 
+func TestNodesMaintenance(t *testing.T) {
+	clients.RequireLong(t)
+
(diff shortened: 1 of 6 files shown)


EmilienM:
@dtantsur thanks for your PR :-)
Would you be able to add tests (fixtures and acceptance) if possible?

dtantsur:
Yep, working on exactly that (struggling to even compile anything locally, meh)

dtantsur:
Okay, this has acceptance tests (that I couldn't verify locally, so may not even compile). I'll look into fixtures tomorrow, need to understand what they are.

EmilienM:
LGTM

PR #2404, opened 2022-05-19
Adjusted Blockstorage V3 Snapshot List Endpoint

### LINKED ISSUE #2403: V3 blockstorage snapshot list is using the wrong endpoint
### Currently used endpoint and the correct endpoint
The snapshot list of blockstorage v3 is using the following endpoint:
```
https://URL/v3/ADMIN_PROJECT_ID/snapshots?all_tenants=true
```
But this endpoint is missing some values (for example the project ID). To include all attributes of the snapshot the following endpoint should be used:
```
https://URL/v3/ADMIN_PROJECT_ID/snapshots/detail?all_tenants=true
```

### Code cause for this issue
This is because a error in [https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v3/snapshots/urls.go](https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v3/snapshots/urls.go). With the following code:
```
...
func createURL(c *gophercloud.ServiceClient) string {
	return c.ServiceURL("snapshots")
}
...
func listURL(c *gophercloud.ServiceClient) string {
	return createURL(c)
}
...
```

As comparison the url generation for the blockstorage volumes (which use the correct `detail` endpoint) [https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v3/volumes/urls.go](https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/v3/volumes/urls.go)
```
func listURL(c *gophercloud.ServiceClient) string {
	return c.ServiceURL("volumes", "detail")
}
```

### Potential fix for this issue
This can be fixed by using the same logic for the snapshots like used for the volumes. That would look something like this:
```
func listURL(c *gophercloud.ServiceClient) string {
	return c.ServiceURL("snapshots", "detail")
}
```

TaminoWalter:
Replaced the normal list snapshots with the detailed list endpoint url.

Fixes #2403

code in this PR:

[https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=list-accessible-snapshots-detail,list-snapshots-and-details-detail#list-snapshots-and-details](https://docs.openstack.org/api-ref/block-storage/v3/index.html?expanded=list-accessible-snapshots-detail,list-snapshots-and-details-detail#list-snapshots-and-details)

### DIFF
--- a/openstack/blockstorage/v3/snapshots/testing/fixtures.go
+++ b/openstack/blockstorage/v3/snapshots/testing/fixtures.go
@@ -11,7 +11,7 @@ import (
 
 // MockListResponse provides mock responce for list snapshot API call
 func MockListResponse(t *testing.T) {
-	th.Mux.HandleFunc("/snapshots", func(w http.ResponseWriter, r *http.Request) {
+	th.Mux.HandleFunc("/snapshots/detail", func(w http.ResponseWriter, r *http.Request) {
 		th.TestMethod(t, r, "GET")
 		th.TestHeader(t, r, "X-Auth-Token", fake.TokenID)
 
--- a/openstack/blockstorage/v3/snapshots/urls.go
+++ b/openstack/blockstorage/v3/snapshots/urls.go
@@ -15,7 +15,7 @@ func getURL(c *gophercloud.ServiceClient, id string) string {
 }
 
 func listURL(c *gophercloud.ServiceClient) string {
-	return createURL(c)
+	return c.ServiceURL("snapshots", "detail")
 }
 
 func updateURL(c *gophercloud.ServiceClient, id string) string {


mandre (changes requested):
These are two different API calls, both valid:
- [list snapshots](https://docs.openstack.org/api-ref/block-storage/v3/?expanded=#list-accessible-snapshots) on `/v3/{project_id}/snapshots`
- [list snapshots and details](https://docs.openstack.org/api-ref/block-storage/v3/?expanded=#list-snapshots-and-details) on `/v3/{project_id}/snapshots/detail`

In consequence, we should create a new function (`ListDetail()`?) in [/v3/snapshots/requests.go](https://github.com/shiftstack/gophercloud/blob/master/openstack/blockstorage/v3/snapshots/requests.go) for the new `/v3/{project_id}/snapshots/detail` endpoint, rather than updating the `List()` function to return the detailed list of snapshots.

PR #2405, opened 2022-05-19
workflow/unit: update goveralls to 1.6

EmilienM:
Update from v1 to v1.6.


mandre (changes requested):
I do not thing that's necessary as shogo82148/actions-goveralls@v1 points to the latest v1 release, see https://github.com/shogo82148/actions-goveralls/tree/releases/v1

PR #2407, opened 2022-05-22
Add missing lb listener fields

mandre:
The loadbalancer listener was missing a few fields.

https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=list-listeners-detail#list-listeners

Fixes #2402

PR #2410, opened 2022-05-23
Octavia: Run the L7 policies tests against HTTP listener

mandre:
Use HTTP listener rather than TCP listener for the L7 policies tests.

Jobs would otherwise fail with:

    Provider 'amphora' does not support a requested option: TCP protocol
    listeners do not support L7 policies

Which is in-line with the documentation [1]:

    Pools of type SCTP, TCP or UDP cannot be used in L7 policies at this
    time.

Fixes #2049.

[1] https://docs.openstack.org/api-ref/load-balancer/v2/index.html?expanded=create-an-l7-policy-detail#create-an-l7-policy


mandre:
It was merged as part of https://github.com/gophercloud/gophercloud/pull/2407, we can close this as it's no longer needed.

PR #2411, opened 2022-05-24
Identity v3: add role assignments list param include_subtree

saydamir:
Add to Identity v3 role assignments list request `include_subtree` parameter, which allows to get a list of assignments for all projects in the project hierarchy below specified projectID or domainID.

`scope.project.id` is required if `include_subtree` is specified.

[Keystone code with include_subtree query parameter](https://github.com/openstack/keystone/blob/stable/yoga/keystone/api/role_assignments.py#L43) .

[API documentation](https://docs.openstack.org/api-ref/identity/v3/?expanded=id627-detail#id627).


EmilienM:
LGTM

Issue #2408, opened 2022-05-23
Ussuri job fails installing devstack


mandre:
Devstack doesn't respect the `INSTALL_TEMPEST=False` setting and and still tries to install tempest.

Additionally, it tries to install tempest from master, as tempest does not create stable branches:
```
++ extras.d/80-tempest.sh:source:9          :   install_tempest
++ lib/tempest:install_tempest:750          :   git_clone https://opendev.org/openstack/tempest.git /opt/stack/tempest master
```

This causes it to fail find dependencies satisfying requirements:
```
[...]

PR #2412, opened 2022-05-24
CI: Bump devstack-action to v0.7

EmilienM:
That will help to disable Tempest (not needed in our jobs) and therefore
bring back the Ussuri job in good shape.

Issue #2408


mandre (approved):
LGTM

Issue #2418, opened 2022-06-02
Adding Support for User Domain Name/ID and Default Domain


ching-kuo:
According to [OpenStack document](https://docs.openstack.org/python-openstackclient/latest/cli/authentication.html#authenticating-using-identity-server-api-v3), users should be able to use `OS_USER_DOMAIN_NAME`, `OS_USER_DOMAIN_ID` or `OS_DEFAULT_DOMAIN` to specify the DOMAIN of user. 

Currently gophercloud only supports `OS_DOMAIN_NAME` and `OS_DOMAIN_ID`, which in the documents stated as for domain scope token.

majewsky:
My understanding is that `AuthOptionsFromEnv` is frozen to preserve backwards-compatibility. You should use [`clientconfig.AuthOptions`](https://pkg.go.dev/github.com/gophercloud/utils@master/openstack/clientconfig#AuthOptions) instead.

vooon:
@majewsky , @pierreprinetti i think that should be fixed in v2.
Personally i think it's unlikely to see Keystone API v2, so it will be better to support `OS_USER_DOMAIN_NAME/ID` and `OS_PROJECT_DOMAIN_NAME/ID` similar way as python library does.

PR #2419, opened 2022-06-02
Add Support for User Domain Name/ID/Default Domain

ching-kuo:
users should be able to use OS_USER_DOMAIN_NAME, OS_USER_DOMAIN_ID or
OS_DEFAULT_DOMAIN to specify the DOMAIN of user.

Fixes: #2418

Reference:
https://docs.openstack.org/python-openstackclient/latest/cli/authentication.html#authenticating-using-identity-server-api-v3


majewsky:
My understanding is that `AuthOptionsFromEnv` is frozen to preserve backwards-compatibility. You should use [`clientconfig.AuthOptions`](https://pkg.go.dev/github.com/gophercloud/utils@master/openstack/clientconfig#AuthOptions) instead.

PR #2422, opened 2022-06-14
Use newer OS images for train and ussuri LB jobs

mandre:
Diskimage-builder now requires at least python 3.8.

Unfortunately the amphora image building in octavia's devstack plugin
doesn't respect the python version set via devstack's PYTHON3_VERSION
environment variable. Instead, let's switch to Ubunto 20.04 that ships
with python 3.8 as the default python interpreter.

https://github.com/openstack/diskimage-builder/commit/fe0e5324d4248d114660ec35111ae9601e05b95b

### DIFF
--- a/.github/workflows/functional-loadbalancer.yaml
+++ b/.github/workflows/functional-loadbalancer.yaml
@@ -28,10 +28,10 @@ jobs:
             ubuntu_version: "20.04"
           - name: "ussuri"
             openstack_version: "stable/ussuri"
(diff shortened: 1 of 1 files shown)


mandre:
Closing in favor of https://review.opendev.org/c/openstack/octavia/+/845812 (stable/train) and https://review.opendev.org/c/openstack/octavia/+/845979 (stable/ussuri).

Issue #2421, opened 2022-06-14
Add support for deleting nova service 


Robin5911:
Adding support for deleting nova service in Gophercloud.

There is no support for following APIs related to delete nova service.

https://docs.openstack.org/python-openstackclient/latest/cli/command-objects/compute-service.html

PR #2427, opened 2022-06-20
Add support for deleting nova service

CloudCourierStation:
Add support for deleting nova service

Fixes https://github.com/gophercloud/gophercloud/issues/2421

Reference:
https://docs.openstack.org/python-openstackclient/latest/cli/command-objects/compute-service.html

### DIFF
--- a/openstack/compute/v2/extensions/services/requests.go
+++ b/openstack/compute/v2/extensions/services/requests.go
@@ -80,3 +80,12 @@ func Update(client *gophercloud.ServiceClient, id string, opts UpdateOpts) (r Up
 	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
 	return
 }
+
+// Delete will delete the existing service with the provided ID.
+func Delete(client *gophercloud.ServiceClient, id string) (r UpdateResult) {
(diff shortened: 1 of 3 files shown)


CloudCourierStation:
cc @kayrus @jtopjian

EmilienM:
Thanks for your contribution, I guess this one can't be tested in acceptance, because once a service is removed, this can't be re-added from API, AFIK.

openstack/compute/v2/extensions/services/requests.go
```diff
@@ -80,3 +80,12 @@ func Update(client *gophercloud.ServiceClient, id string, opts UpdateOpts) (r Up
 	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
 	return
 }
+
+// Delete will delete the existing service with the provided ID.
+func Delete(client *gophercloud.ServiceClient, id string) (r DeleteResult) {
+	resp, err := client.Delete(updateURL(client, id), &gophercloud.RequestOpts{
+		OkCodes: []int{204},
```
EmilienM, line 87: https://docs.openstack.org/api-ref/compute/?expanded=delete-compute-service-detail#delete-compute-service

I confirm this is 204.

PR #2445, opened 2022-08-02
Add UpdatedAt to ports.Port

### LINKED ISSUE #2444: ports.Port missing UpdatedAt field
All neutron core resources have an `updated_at` field, see <https://docs.openstack.org/api-ref/network/v2/index.html?expanded=show-port-details-detail#show-port-details>.

boyvinall:
Fixes #2444

code in this PR:

- https://github.com/openstack/neutron/blob/b7fad3dd35ad23303a5f52ec64478860a137775e/neutron/extensions/timestamp.py#L33

### DIFF
--- a/openstack/networking/v2/ports/results.go
+++ b/openstack/networking/v2/ports/results.go
@@ -1,6 +1,8 @@
 package ports
 
 import (
+	"time"
+
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
 )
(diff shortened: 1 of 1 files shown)


mandre (commented):
This looks good.
While looking at the Network resource, it looks like neutron used to use a different time format and we extra steps when unmarshalling the JSON response, as seen in https://github.com/gophercloud/gophercloud/issues/1670. However, the linked PR from Aug 2019 mentions "older" openstack version. At the time, Train was still the development version (released on 2019-10-16), and is now the oldest version we support in gophercloud. We should be fine with your change.

However, while you're at it, `updated_at` comes in pair with `created_at`, would you mind adding it also?

boyvinall:
Sure, added that now, thanks.

mandre (approved):
Thanks!

PR #2459, opened 2022-09-01
actions: Add semver tools

pierreprinetti:
* Define three new labels: `semver:major`, `semver:minor`, `semver:patch`
* Only allow merge if one of the `semver:` labels is present
* Remove the `semver:` label upon new push to the pull request

### DIFF
--- a/.github/workflows/semver-require.yaml
+++ b/.github/workflows/semver-require.yaml
@@ -1,0 +1,17 @@
+name: Pull Request Labels
+on:
+  pull_request:
+    types:
+      - opened
+      - labeled
+      - unlabeled
+      - synchronize
+jobs:
+  label:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mheap/github-action-required-labels@v2
+        with:
+          mode: exactly
+          count: 1
+          labels: "semver:patch, semver:minor, semver:major"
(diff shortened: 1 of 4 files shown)


.github/workflows/semver-require.yaml
```diff
@@ -0,0 +1,16 @@
+name: Pull Request Labels
+on:
+  pull_request:
+    types:
+      - opened
+      - labeled
+      - unlabeled
```
mandre, line 7: The example at https://github.com/mheap/github-action-required-labels#complete-example also runs on `synchronize`.
What does it mean for us not to have the `synchronize` type? In particular, are we certain it runs after someone pushes a new commit to a PR?
Take this scenario for instance: someone opens a PR and does not set labels, this person then add a new commit. The `semver-unlabel` workflow won't do anything, and `semver-require` won't kick-in. Will the PR be allowed to merged without a label?

mandre (commented):
LGTM on principle, however I'm not sure how the `semver-require` workflow currently covers all cases. Please see my question inline.

.github/workflows/semver-require.yaml
pierreprinetti: nice catch!

mandre (approved):
LGTM

PR #2463, opened 2022-09-06
Unlabeler: Fix permissions issue

pierreprinetti:
The action must be given explicit permissions to operate on a PR from a
forked repository.


.github/workflows/semver-unlabel.yaml
```diff
@@ -11,6 +11,8 @@ on:
 
 jobs:
   remove-semver-label:
+    permissions:
```
mandre, line 14: You should instead operate on the `pull_request_target` trigger.

See https://github.com/gophercloud/gophercloud/pull/2378 where I fixed a similar issue for the greetings workflow.

mandre (changes requested):

.github/workflows/semver-unlabel.yaml
pierreprinetti: Thanks! Updated.

I have also kept the `permissions` stanza, which in combination with your suggestion, acts as a restriction rather than a grant. See the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target):

> the GITHUB_TOKEN is granted read/write repository permission unless the permissions key is specified

mandre (approved):
LGTM

PR #2477, opened 2022-09-12
Port CreatedAt and UpdatedAt: add back JSON tags

pierreprinetti:
Removing them in 48581dacb7ec726e2ea5ed5094e0bf91db20751b was not necessary. Removing them might break client-side serialisation.

### DIFF
--- a/openstack/networking/v2/ports/results.go
+++ b/openstack/networking/v2/ports/results.go
@@ -115,10 +115,10 @@ type Port struct {
 	RevisionNumber int `json:"revision_number"`
 
 	// Timestamp when the port was created
-	CreatedAt time.Time `json:"-"`
+	CreatedAt time.Time `json:"created_at"`
 
 	// Timestamp when the port was last updated
-	UpdatedAt time.Time `json:"-"`
+	UpdatedAt time.Time `json:"updated_at"`
 }
 
 func (r *Port) UnmarshalJSON(b []byte) error {


mandre (commented):
I copied what was done in https://github.com/gophercloud/gophercloud/pull/1671. Should we go back and also add the `created_at` and `updated_at` JSON tags for the other networking resources?

pierreprinetti:
> I copied what was done in #1671. Should we go back and also add the `created_at` and `updated_at` JSON tags for the other networking resources?

@jtopjian Do you happen to remember why the `json:"created_at"` tags were removed  in https://github.com/gophercloud/gophercloud/pull/1671 while overriding `UnmarshalJSON`? What am I missing?

pierreprinetti:
One reason might be that when sending the JSON representation of a Port to the OpenStack API, maybe if you send the "wrong" time representation you get an error. For a field that would be ignored anyway. Just guessing.

jtopjian:
@pierreprinetti Sorry for the late reply.

This area of code is only for results returned from the API. Requests being sent to the API shouldn't be hitting this code.

The tags were removed so that parsing was picked up by the `UnmarshalJSON` method. This method is not called if the struct has an explicit JSON tag. 

In that method, two different time formats are supported, noted in `s1` and `s2`. If the JSON result contained a time format that wasn't parseable by `gophercloud.JSONRFC3339NoZ` in `s1`, then the code attempts to parse the time as `time.Time` in `s2`. 

However, if `s1` was successful, the type is converted to `time.Time`.

The goal being to support an older style timestamp and a more general/generic timestamp simultaneously. 

Let me know if this helps.

pierreprinetti:
@jtopjian 
Thank you for taking the time to explain. I however am not yet clear how the tag removal helps.

In a Go playground I have implemented a simplified Port type, and copy-pasted the UnmarshalJSON method. CreatedAt and UpdatedAt have tags. Decoding seems to work: https://go.dev/play/p/FlTrhozjqZr

What am I missing?

jtopjian:
hm... So keeping the JSON tags still causes those fields to be processed in the `UnmarshalJSON` method. This might be a change in behaviour with Golang. There are _lots_ of other places in Gophercloud where a JSON tag of `-` is used. We would have run into issues years ago if those fields were being omitted from the response body.

pierreprinetti:
Thanks again @jtopjian.
My only concern is about changing the shape of an exported struct between `v1.0.0` and `v1.y.z`. The unit tests run against Go 1.14 and seem to confirm that it's fine to add the tags back. I propose to merge this just to adhere to our promise of compatibility before we ship v1.1.0.

WDYT @mandre?

mandre:
> Thanks again @jtopjian. My only concern is about changing the shape of an exported struct between `v1.0.0` and `v1.y.z`. The unit tests run against Go 1.14 and seem to confirm that it's fine to add the tags back. I propose to merge this just to adhere to our promise of compatibility before we ship v1.1.0.
> 
> WDYT @mandre?

No strong feeling either way. Let's merge it then, so that the struct doesn't change.

PR #2480, opened 2022-09-21
Add support for volume type for db/v1/instance

### LINKED ISSUE #2340: Add support for volume types in DBaaS
The OpenStack [API](https://docs.openstack.org/api-ref/database/?expanded=create-database-instance-detail) has support for both a volume Size and volume Type parameter, when declaring a new database instance.

The [current provider](https://github.com/gophercloud/gophercloud/blob/a1ac165819cd057e431dd1d37f05534ca355fdbe/openstack/db/v1/instances/requests.go#L72) only uses Size as an integer, and does not allow a volume Type field to be provided.

pawcykca:
Fixes #2340

Openstack API Ref: 
https://docs.openstack.org/api-ref/database/?expanded=create-database-instance-detail#create-database-instance

### DIFF
--- a/openstack/db/v1/instances/requests.go
+++ b/openstack/db/v1/instances/requests.go
@@ -53,6 +53,8 @@ type CreateOpts struct {
 	// Specifies the volume size in gigabytes (GB). The value must be between 1
 	// and 300. Required.
 	Size int
+	// Specifies the volume type.
+	VolumeType string
 	// Name of the instance to create. The length of the name is limited to
 	// 255 characters and any characters are permitted. Optional.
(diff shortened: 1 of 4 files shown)


mandre:
Similar to https://github.com/gophercloud/gophercloud/pull/2484, it would be nice to have a CI job for the DB service.

pawcykca:
@mandre shouldn't this issue be labeled as 'semver:minor' as well as #2484?

EmilienM (approved):
I think we can live without DBaaS CI job for now, as nobody has the time to do it.

Issue #2483, opened 2022-09-23
Add support for availability zone in DBaaS 


pawcykca:
The OpenStack [API](https://docs.openstack.org/api-ref/database/?expanded=create-database-instance-detail,show-database-instance-details-detail#create-database-instance) has support for availability zone parameter, when declaring a new database instance.

Currently we can't provide this field while creating database instance.

zhekazuev:
@pawcykca Hi,

Maybe you forget to close that issue.

pawcykca:
Fixed by #2871

PR #2484, opened 2022-09-23
Add AvailabilityZone field to DBaaS instance

pawcykca:
Fixes https://github.com/gophercloud/gophercloud/issues/2483

Openstack API Ref:
https://docs.openstack.org/api-ref/database/?expanded=create-database-instance-detail#create-database-instance

### DIFF
--- a/openstack/db/v1/instances/requests.go
+++ b/openstack/db/v1/instances/requests.go
@@ -56,6 +56,8 @@ type CreateOpts struct {
 	// Name of the instance to create. The length of the name is limited to
 	// 255 characters and any characters are permitted. Optional.
 	Name string
+	// The availability zone of the instance.
+	AvailabilityZone string
 	// A slice of database information options.
 	Databases db.CreateOptsBuilder
 	// A slice of user information options.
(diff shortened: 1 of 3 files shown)


mandre:
Thanks for the PR.
Ideally we should have a CI for DBaaS before merging this change so that we're sure we don't break anything.
Also, since we modify the CreateOpts struct, it's also a breaking change so adding the `semver:major` label.

mandre:
> Thanks for the PR. Ideally we should have a CI for DBaaS before merging this change so that we're sure we don't break anything. Also, since we modify the CreateOpts struct, it's also a breaking change so adding the `semver:major` label.

I talked to @pierreprinetti and we decided to relax this rule so that we don't categorize changes as breaking changes when you're not impacted with a normal usage of the library. We'll rely on `gorelease` to let us know if changes are to be considered backward-compatible or not.

In this case:
```
# github.com/gophercloud/gophercloud/openstack/db/v1/instances
## compatible changes
CreateOpts.AvailabilityZone: added
```

EmilienM:
@mandre I'll be fine merging as is, even if we don't have CI. I don't think any of us currently has the time to add CI for DBaaS.

EmilienM (approved):

EmilienM:
Please rebase

pawcykca:
@EmilienM what about this PR?

pawcykca:
@mandre @EmilienM 
Can we go on with this PR?

pawcykca:
@mandre @EmilienM 
Once again, can we go on with this PR?

EmilienM:
@pawcykca sorry we have been fixing a few CI things and produced a release. Could you please rebase on master. Thanks

pawcykca:
Can by closed. Done in PR #2871

PR #2486, opened 2022-10-07
Fix BareMetalV1 version

### LINKED ISSUE #2485: Add support of baremetal endpoint without /v1 in keystone catalog
Hello, 
Our baremetal endpoints are declared without /v1 at the end of the URL, and gophercloud don't add it in the url.
Looks like some other users have it because it's merged like that on gophercloud ?
Can we add a fix to add the suffix if it's not present ?

snigle:
Fixes #2485

There is no documentation to link as it's only a configuration on keystone.

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -363,7 +363,11 @@ func initClientOpts(client *gophercloud.ProviderClient, eo gophercloud.EndpointO
 // NewBareMetalV1 creates a ServiceClient that may be used with the v1
 // bare metal package.
 func NewBareMetalV1(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
-	return initClientOpts(client, eo, "baremetal")
+	sc, err := initClientOpts(client, eo, "baremetal")
+	if !strings.HasSuffix(sc.Endpoint, "v1/") {
+		sc.ResourceBase = sc.Endpoint + "v1/"
+	}
+	return sc, err
 }
 
 // NewBareMetalIntrospectionV1 creates a ServiceClient that may be used with the v1


openstack/client.go
```diff
@@ -363,7 +363,11 @@ func initClientOpts(client *gophercloud.ProviderClient, eo gophercloud.EndpointO
 // NewBareMetalV1 creates a ServiceClient that may be used with the v1
 // bare metal package.
 func NewBareMetalV1(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
-	return initClientOpts(client, eo, "baremetal")
+	sc, err := initClientOpts(client, eo, "baremetal")
+	if !strings.HasSuffix(sc.Endpoint, "v1/") {
```
pierreprinetti, line 367: Is it possible that we have suffix `v1` without the `/`?

snigle: looks like others services have the / at the end like network : 
https://github.com/gophercloud/gophercloud/pull/2486/files/535d1003686b256713d5a47eb4cfd97e90c0fb73#diff-5643779347824e0ec4258ef187be1f23c3b1eeca504fcfe48e47efe0e051b4bfR395

snigle: yes you are right maybe we nee double check v1/ and v1 ?

reenjii: Maybe it is best to check both ?

```
if !strings.HasSuffix(strings.TrimSuffix(sc.Endpoint, "/"), "v1")
```

mandre: Looking at the initialization for the other clients, it seems unlikely the suffix has `v1` without the `/`, but better be safe than sorry.

EmilienM (approved):

PR #2487, opened 2022-10-12
Add Prometheus protocol for octavia listeners

nikParasyr:
Prometheus protocol was added for octavia listeners on 2.25

[docs](https://docs.openstack.org/api-ref/load-balancer/v2/#create-listener)
[code](https://github.com/openstack/octavia/blob/master/octavia/api/root_controller.py#L142)


nikParasyr:
@mandre this is rdy from my side. Very minor update

mandre (approved):
LGTM, thanks!

PR #2488, opened 2022-10-12
Add Persistance for octavia pools.UpdateOpts

nikParasyr:
Session persistance can be updated on Octavia. Add it to UpdateOpts
[Docs](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=update-a-pool-detail#update-a-pool)


nikParasyr:
@mandre this is ready from my side as well

mandre (approved):
Thanks again.

PR #2489, opened 2022-10-12
Update github workflow for Openstack Zed

nikParasyr:
Openstack Zed was released recently. Update github workflows to include testing for the stable/zed verson. ~~Furthermore remove stable/train from the versions that are tested.~~

I was updating today the ci on TPO and thought to push it here as well. I "assumed" that removing stable/train was the way to go, but maybe i was wrong

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -14,6 +14,9 @@ jobs:
         openstack_version: ["master"]
         ubuntu_version: ["20.04"]
         include:
+          - name: "zed"
+            openstack_version: "stable/zed"
+            ubuntu_version: "20.04"
           - name: "yoga"
             openstack_version: "stable/yoga"
             ubuntu_version: "20.04"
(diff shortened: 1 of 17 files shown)


mandre (commented):
Thanks for updating the CI jobs to latest version. I'm less sure about removing Train just yet, as it's still the version the current RHOSP is based on, and possibly other supported distros. On the other hand, Train is now EOL upstream, so might as well drop support for it.

@EmilienM, @pierreprinetti WDYT?

EmilienM:
@mandre sorry for my late comment.

I strongly disagree to remove testing on Train, major distros are still using this version (I know...) and I don't think it's a problem to keep testing it for us now.
Please change your patch to just add Zed. @nikParasyr

nikParasyr:
I'll close this one tbh. Unfortunately I dont have time to look into it right now. There are a lot of deprecations on various services that are not tested daily/weekly and are triggered now. A couple i could figure out how to fix but some others not.

PR #2491, opened 2022-10-13
Add VipQosPolicyID to loadbalancer Create and Update

nikParasyr:
VipQosPolicyID can be defined for loadbalancers during creation and update. Moreover add the `neutron-qos` service to loadbalancer workflow.
[Docs](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=create-a-load-balancer-detail,update-a-load-balancer-detail#create-a-load-balancer)


mandre (approved):
LGTM, thanks.

Issue #2381, opened 2022-04-13
L7Policy : Support Tags


hbw222:
Hello

Would really appreciate it if the L7Policy-create,L7Rule-create,pool-list could support Tags.

https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=create-an-l7-policy-detail#create-an-l7-policy

Thanks in advance!

Kind regards

hbw222:
L7Rule create missing Tags as well

winiciusallan:
Completed by https://github.com/gophercloud/gophercloud/commit/706597e94d5316815a4dc1159f1bc42bf69b0dc0. Feel free to re-open or create a new issue if you need something else.

PR #2492, opened 2022-10-13
Add tags for loadbalancer l7policy and l7rule

nikParasyr:
Tags are supported for octavia l7policy and l7rules since version 2.5. Add support for Create and Update.
[Docs](https://docs.openstack.org/api-ref/load-balancer/v2/?expanded=create-an-l7-policy-detail,update-a-l7-policy-detail,create-an-l7-rule-detail,update-a-l7-rule-detail#create-an-l7-policy)

partially implemets: #2381

### DIFF
--- a/acceptance/openstack/loadbalancer/v2/loadbalancer.go
+++ b/acceptance/openstack/loadbalancer/v2/loadbalancer.go
@@ -419,7 +419,7 @@ func CreatePoolHTTP(t *testing.T, client *gophercloud.ServiceClient, lb *loadbal
 // CreateL7Policy will create a l7 policy with a random name with a specified listener
 // and loadbalancer. An error will be returned if the l7 policy could not be
 // created.
-func CreateL7Policy(t *testing.T, client *gophercloud.ServiceClient, listener *listeners.Listener, lb *loadbalancers.LoadBalancer) (*l7policies.L7Policy, error) {
+func CreateL7Policy(t *testing.T, client *gophercloud.ServiceClient, listener *listeners.Listener, lb *loadbalancers.LoadBalancer, tags []string) (*l7policies.L7Policy, error) {
 	policyName := tools.RandomString("TESTACCT-", 8)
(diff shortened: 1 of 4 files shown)


nikParasyr:
@mandre  this is rdy from my side as well

mandre:
I'll try my best to clear my backlog of reviews before the end of the week, sorry for the time it takes.

mandre:
`gorelease` reports this change to not be backward compatible and I'm not sure why... @pierreprinetti, do you have an idea?
```
# github.com/gophercloud/gophercloud/openstack/loadbalancer/v2/l7policies
## incompatible changes
CreateRuleOpts: old is comparable, new is not
Rule: old is comparable, new is not
## compatible changes
CreateOpts.Tags: added
CreateRuleOpts.Tags: added
L7Policy.Tags: added
Rule.Tags: added
UpdateOpts.Tags: added
UpdateRuleOpts.Tags: added
```

pierreprinetti:
This PR changes the signature of the exported functions `CreateL7Policy` and `CreateL7Rule`. I am surprised that `gorelease` does not mention that as a breaking change.

Anyway: structs containing []string [cannot be compared](https://go.dev/play/p/fJB3oA4BOag). Before this change, the structs `Rule` and `CreateRuleOpts` only contained comparable types (string, bool) and a client could have used `==` on them. Adding a slice breaks such clients.

mandre:
> This PR changes the signature of the exported functions `CreateL7Policy` and `CreateL7Rule`. I am surprised that `gorelease` does not mention that as a breaking change.

It does, but since they're in acceptance tests we can safely ignore.

> Anyway: structs containing slices [cannot be compared](https://go.dev/play/p/fJB3oA4BOag). Before this change, the structs `Rule` and `CreateRuleOpts` only contained comparable types (string, bool) and a client could have used `==` on them. Adding a `[]string` breaks such clients.

Ack, thanks for your insight.

mandre (approved):

PR #2503, opened 2022-11-07
Bump golang.org/x/crypto

pierreprinetti:
Update the depedency to the last commit that is compatible with Go v1.14. The [next commit][1] replaces calls to `io/ioutil` to its replacements in the `io` and `os` packages, which are not available in Go v1.14.

[1]: https://go-review.googlesource.com/c/crypto/+/430797

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,6 @@ module github.com/gophercloud/gophercloud
 go 1.14
 
 require (
-	golang.org/x/crypto v0.0.0-20211202192323-5770296d904e
+	golang.org/x/crypto v0.0.0-20220829220503-c86fa9a7ed90
 	gopkg.in/yaml.v2 v2.4.0
 )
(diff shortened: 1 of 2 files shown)


pierreprinetti:
The `go vet` command embedded in our unit tests somehow triggers a bump of the crypto library, which then fails to compile under Go v1.14 (happens [here](https://github.com/gophercloud/gophercloud/pull/2502) as well).

mandre (approved):
LGTM. We should consider dropping go 1.14 support at some point so we can upgrade to a newer version of golang.org/x/crypto.

mandre:
Not sure why the `go vet` command tries to bump the lib either...

pierreprinetti:
> Not sure why the `go vet` command tries to bump the lib either...

Fixed [here](https://github.com/gophercloud/gophercloud/pull/2504)

PR #2508, opened 2022-11-18
Add Github Workflow for checking Go API compatibility on PRs

EmilienM:
It'll automatically set the semver label for a given PR.

Limitation: files under `/acceptance` or named `doc.go`, `fixtures.go`
and `fixture.go` which don't count as semver, will count in this new
job.

This is because we would need to add a commit that exclude these files,
then rebase the PR on top of that and run `go-apidiff`. This is too
complex for now and might be addressed later.


mandre:
I think I fixed the permissions issue by switching to labeler, however for some reason it didn't apply the label (or at least I'm not seeing it yet).
But even if setting a label worked, I wonder if we should add one automatically... we probably need to perform some sort of filtering on the api-diff output since we don't care about API compatibility in the acceptance tests for example. We also want to differentiate between `no API changes` and `backward compatible API change`.

EmilienM:
@mandre are you good to close this one and merge https://github.com/gophercloud/gophercloud/pull/2513 ?

EmilienM:
@mandre no, the job hasn't run. Probably because you changed the trigger.

EmilienM:
meh

EmilienM:
The old workflow is still running because a user can't remove a workflow like this, this PR has to be merged first.

Same thing for the new workflow: because we're using `pull_request_target` ([reason](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/)) - we can't see the job output here, we would have to merge it first. However I tested it manually and it seems to be working.

@pierreprinetti I'll let you comment on this one.

Thanks

.github/workflows/semver-auto.yaml
```diff
@@ -0,0 +1,37 @@
...
+      uses: actions-ecosystem/action-remove-labels@v1
+      with:
+        labels: |
+          semver:patch
+          semver:minor
+          semver:major
+        github_token: ${{ github.token }}
+
+    - name: Checking Go API Compatibility
+      uses: joelanford/go-apidiff@main
```
pierreprinetti, line 23: Please pin to the commit sha of the latest release:

```suggestion
      uses: joelanford/go-apidiff@7037613b8a1dfa41270919a854e3703953c36d8f
```

Dependabot will keep it up to date, while letting us review every change.

.github/workflows/semver-auto.yaml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - uses: actions/setup-go@v2
+      with:
+        go-version: '1'
+
+    - name: Remove the semver label
+      uses: actions-ecosystem/action-remove-labels@v1
```
pierreprinetti, line 14:
```suggestion
      uses: actions-ecosystem/action-remove-labels@2ce5d41b4b6aa8503e285553f75ed56e0a40bae0
```

.github/workflows/semver-auto.yaml
```diff
@@ -0,0 +1,37 @@
...
+          semver:minor
+          semver:major
+        github_token: ${{ github.token }}
+
+    - name: Checking Go API Compatibility
+      uses: joelanford/go-apidiff@main
+
+    - name: Add semver:minor label
+      if: ${{ success() }}
+      uses: actions-ecosystem/action-add-labels@v1
```
pierreprinetti, line 27:
```suggestion
      uses: actions-ecosystem/action-add-labels@bd52874380e3909a1ac983768df6976535ece7f8
```

.github/workflows/semver-auto.yaml
```diff
@@ -0,0 +1,37 @@
...
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        labels: semver:minor
+
+    - name: Add semver:major label
+      if: ${{ failure() }}
+      uses: actions-ecosystem/action-add-labels@v1
```
pierreprinetti, line 34:
```suggestion
      uses: actions-ecosystem/action-add-labels@bd52874380e3909a1ac983768df6976535ece7f8
```

.github/workflows/semver-auto.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: Add semver:minor label
+      if: ${{ success() }}
+      uses: actions-ecosystem/action-add-labels@v1
+      with:
+        github_token: ${{ secrets.GITHUB_TOKEN }}
+        labels: semver:minor
+
+    - name: Add semver:major label
+      if: ${{ failure() }}
```
pierreprinetti, line 33: Are you sure that the tool exits with a failure in case of incompatible changes? I can't find that in the docs

.github/workflows/semver-auto.yaml
```diff
@@ -0,0 +1,37 @@
+name: Add PR semver labels on push
+on: [pull_request_target]
```
pierreprinetti, line 2: Will this workflow not be triggered by any comment, label and such? I thought `types: [synchronize]` was needed here

EmilienM: me too, let me re-add it. I'm not sure why I removed it.

pierreprinetti (commented):
There’s a sad news. The tool will check API changes in all non-_test.go files. That includes many Gophercloud test files that live under ‘testing’ directories. If the tool checks all packages, we’re going to get many false positives.

EmilienM:
> There’s a sad news. The tool will check API changes in all non-_test.go files. That includes many Gophercloud test files that live under ‘testing’ directories. If the tool checks all packages, we’re going to get many false positives.

Are you sure we'll get that many false positives?
I checked and the non-_test.go in 'testing' directories are: fixtures.go (or fixture), doc.go and that's it.

pierreprinetti:
>Are you sure we'll get that many false positives?

I am also fine if we just try and see and iterate.

EmilienM:
> > Are you sure we'll get that many false positives?
> 
> I am also fine if we just try and see and iterate.

I'll update it to support the semver:patch, and ask you for a review.

EmilienM:
More complicated than I thought, joelanford/go-apidiff doesn't return any output from stdout so we can't parse the output of the commands, which means for now we can't make the difference between semver:patch and semver:minor :(

EmilienM:
https://github.com/joelanford/go-apidiff/pull/28 should help

pierreprinetti:
> > There’s a sad news. The tool will check API changes in all non-_test.go files. That includes many Gophercloud test files that live under ‘testing’ directories. If the tool checks all packages, we’re going to get many false positives.
> 
> Are you sure we'll get that many false positives? I checked and the non-_test.go in 'testing' directories are: fixtures.go (or fixture), doc.go and that's it.

and `/acceptance`

EmilienM:
Tested via https://github.com/EmilienM/gophercloud/pull/2
Results (working): https://github.com/EmilienM/gophercloud/actions/runs/6004309227/job/16284616345?pr=2

pierreprinetti (approved):
Let's try it!

PR #2510, opened 2022-11-21
objectstorage: Treat empty content-type as text/plain

mandre:
This allows retrieving objects list when Swift speaks HTTP2.

Fixes #2509

Issue #2509, opened 2022-11-21
Failure to list objects when swift speaks HTTP2


mandre:
Some OpenStack environments may have Swift speaking HTTP2, possibly when behind a reverse proxy. It would fail with:

```
Cannot extract names from response with content-type: []
```

This error affected gophercloud v0.17.0 and below for listing containers, and was fixed in a seemingly unrelated patch via https://github.com/gophercloud/gophercloud/commit/b7d5b2cdd7ffc13e79d924f61571b0e5f74ec91c#diff-775ab0e4f42bef31b7fc4b024f3be57b813ff2756645efab1869777bf589eefbR75

The error still affects listing objects. We may want to also treat an empty content type as `text/plain` there.

mandre:
Fixed with https://github.com/gophercloud/gophercloud/pull/2513.

PR #2513, opened 2022-11-23
objectstorage: Do not parse NoContent responses

pierreprinetti:
Some Swift instances respond with HTTP status code 204 when asked to list containers and there are no containers, or when asked to list objects on an empty container. Before this patch, Gophercloud would error because the header `content-type` is absent on the response.

With this patch, objectstorage responses with HTTP status code 204 are immediately recognised as empty and their body is not read.

Potentially supersedes https://github.com/gophercloud/gophercloud/pull/2510

### DIFF
--- a/openstack/objectstorage/v1/containers/results.go
+++ b/openstack/objectstorage/v1/containers/results.go
@@ -30,6 +30,10 @@ type ContainerPage struct {
 
 // IsEmpty returns true if a ListResult contains no container names.
 func (r ContainerPage) IsEmpty() (bool, error) {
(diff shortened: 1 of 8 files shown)


mandre (approved):
LGTM. This should solve https://github.com/gophercloud/gophercloud/issues/2509.

Does #2510 still makes sense?

pierreprinetti:
> Does https://github.com/gophercloud/gophercloud/pull/2510 still makes sense?

Ideally, no. But we might take it out of the hat again in the future.

PR #2514, opened 2022-11-23
objectstorage testing: Fix order of assertions

pierreprinetti:
Fix confusing error messages on failing tests. The assertions all required "expected" before "actual" in their arguments.


mandre:
Looks like the problem is quite widespread in the acceptance tests. I've reported https://github.com/gophercloud/gophercloud/issues/2515 to track the cleanup.

mandre (approved):

PR #2519, opened 2022-11-29
Modify user-agent header to ensure current gophercloud version is provided

### LINKED ISSUE #2511: Should the user-agent always contain `gophercloud/2.0.0`
### Discussed in https://github.com/gophercloud/gophercloud/discussions/2505

<div type='discussions-op-text'>

<sup>Originally posted by **UmBsublime** November 10, 2022</sup>
Hi,

I am parsing my openstack api logs and trying to identify what API-clients and SDKs our user-base implement to consume our services. To do this I can usually check the HTTP user-agent for the specific  sdk and version being used.

When parsing for usage of gophercloud the user-agent is set statically to `gophercloud/2.0.0`
I am curious as to what this represents since it does not relate to the SDK version.

https://github.com/gophercloud/gophercloud/blob/master/provider_client.go#L17</div>

UmBsublime:
Fixes #2511

In addition to this change a new github action will be required to ensure there is no version drift.
This version should stay in-line with the current release of gophercloud

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -14,7 +14,7 @@ import (
 
 // DefaultUserAgent is the default User-Agent string set in the request header.
 const (
-	DefaultUserAgent         = "gophercloud/2.0.0"
+	DefaultUserAgent         = "gophercloud/v1.1.1"
 	DefaultMaxBackoffRetries = 60
 )
 
--- a/testing/provider_client_test.go
+++ b/testing/provider_client_test.go
@@ -32,17 +32,17 @@ func TestUserAgent(t *testing.T) {
 	p := &gophercloud.ProviderClient{}
 
 	p.UserAgent.Prepend("custom-user-agent/2.4.0")
-	expected := "custom-user-agent/2.4.0 gophercloud/2.0.0"
+	expected := "custom-user-agent/2.4.0 " + gophercloud.DefaultUserAgent
 	actual := p.UserAgent.Join()
 	th.CheckEquals(t, expected, actual)
 
 	p.UserAgent.Prepend("another-custom-user-agent/0.3.0", "a-third-ua/5.9.0")
-	expected = "another-custom-user-agent/0.3.0 a-third-ua/5.9.0 custom-user-agent/2.4.0 gophercloud/2.0.0"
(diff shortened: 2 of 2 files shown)


pierreprinetti:
Unless you have reasons to be more specific, I would propose to only mention the major version number: `"gophercloud/v1"`. The maintenance burden would then be much lower.
I also suggest the `v` prefix so that once we hit `v2` it's going to be different from the `2.0.0` we've used so far.

UmBsublime:
@pierreprinetti That would for sure be suitable. If you prefer this approach I will adapt accordingly.

My logic for wanting the exact version comes from an operator perspective.
Having the specific version helps specifically with this point.

- Identify clients using known "bugged" version of any SDK (has happened with other SDKs in the past)

Other than that it's simply for statistics (and personal curiosity about how often people update).

UmBsublime:
Latest test failures appear to be related to a hiccup on CI apt mirrors. Not sure how to re-trigger them

```
E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/e/edk2/qemu-efi-arm_0~20191122.bd85bf54-2ubuntu3.3_all.deb  404  Not Found [IP: 52.147.219.192 80
```

UmBsublime:
@pierreprinetti any blockers with this PR or are we simply awaiting next release for merge ?

pierreprinetti:
@mandre 
What if we merge this, and add instructions in a release process document? I am working on it right now.

UmBsublime:
Hi @pierreprinetti,

I see this PR missed the v1.1.1 release and also that v1.2.0 is coming up.

Would it be possible to include this in the 1.2.0 release or does it need it's own `patch` release?

pierreprinetti:
@UmBsublime 
I am proposing to add "bump the user-agent" in our new [release document](https://github.com/gophercloud/gophercloud/pull/2539). It would then be part of the release process, and we could build some muscle memory around that. I have added (a slightly modified version of) this commit to [the v1.2.0 preparation PR](https://github.com/gophercloud/gophercloud/pull/2538). Let's see what the other maintainers think.

pierreprinetti:
LGTM

pierreprinetti (approved):

PR #2521, opened 2022-12-02
Bump gopkg.in/yaml to v3

pierreprinetti:
Note that this new version treats booleans in a slightly different way. As per the package's docs, as of v3 of the yaml package:

    YAML 1.1 bools (yes/no, on/off) are supported as long as they are
    being decoded into a typed bool value. Otherwise they behave as a
    string. Booleans in YAML 1.2 are true/false only.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/gophercloud/gophercloud/v2
 go 1.21.6
 
 require (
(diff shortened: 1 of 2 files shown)


pierreprinetti:
Removing milestone v2.0.0 as this is not expected to give us any improvement in Gophercloud

PR #2523, opened 2022-12-14
CI: cleanup some jobs for containerinfra & dns

EmilienM:
* containerinfra: train and ussuri aren't supported anymore in upstream.
* dns: train isn't supported anymore in upstream.

Let's remove these jobs, since they can't work anyway.

### DIFF
--- a/.github/workflows/functional-containerinfra.yaml
+++ b/.github/workflows/functional-containerinfra.yaml
@@ -26,12 +26,6 @@ jobs:
           - name: "victoria"
             openstack_version: "stable/victoria"
             ubuntu_version: "20.04"
-          - name: "ussuri"
-            openstack_version: "stable/ussuri"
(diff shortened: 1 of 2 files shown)


.github/workflows/functional-containerinfra.yaml
```diff
@@ -26,12 +26,6 @@ jobs:
           - name: "victoria"
             openstack_version: "stable/victoria"
             ubuntu_version: "20.04"
-          - name: "ussuri"
-            openstack_version: "stable/ussuri"
-            ubuntu_version: "18.04"
-          - name: "train"
-            openstack_version: "stable/train"
```
bshephar, line 33: I'm not too familiar with this CI environment. But does this support using tags instead of branches? ie, could we use `train-eol` here instead of just removing it entirely?

PR #2529, opened 2022-12-22
Added target state "rebuild" for Ironic nodes

BirknerAlex:
Added missing target state "rebuild" for Ironic nodes.


mandre (approved):
LGTM, thanks.

For reference, the docs are at https://docs.openstack.org/api-ref/baremetal/?expanded=change-node-provision-state-detail#change-node-provision-state.

And code at https://opendev.org/openstack/ironic/src/commit/4704c74b4d028fa73bb465d43f7147702519dc2b/ironic/common/states.py#L168

Issue #2461, opened 2022-09-06
Add support for identity v3 OS-FEDERATION mappings extension


emilmaruszczak:
As part of the `OS-FEDERATION` keystone extension has added a `mappings` api.
Add support for it.

[docs](https://docs.openstack.org/api-ref/identity/v3-ext/index.html#os-federation-api)
[code](https://github.com/openstack/keystone/blob/stable/yoga/keystone/api/os_federation.py#L248)

- [x] List mappings
- [x] Create mapping
- [x] Get mapping
- [x] Update mapping
- [x] Delete mapping

emilmaruszczak:
Answering the question asked in the opened MR #2468 : yes I do have all the operations implemented, just decided to push them one by one so they are not dependent on each other and it's easier for the merger to handle ;).

I will add acceptance tests soon as I figure them out.

shaardie:
We are also interested in this feature and thought about implementing it our-self and bringing it upstream. And were happy to see, that there was already work done for that.

What is the current status of this? I see the List Mappings is already present. What about the other calls?

@emilmaruszczak Is it still your plan to create PR for them?

emilmaruszczak:
@shaardie The logic and unit tests are implemented, I will PR them in shortly, after adding the acceptance tests.

shaardie:
> @shaardie The logic and unit tests are implemented, I will PR them in shortly, after adding the acceptance tests.

Thanks, we are looking forward to it :+1:

mandre:
@emilmaruszczak looks like you've implemented it all. Is there anything left, or can we close this issue now?

emilmaruszczak:
> @emilmaruszczak looks like you've implemented it all. Is there anything left, or can we close this issue now?

Yes, all is implemented now. Thank you for reviews and your work.

PR #2530, opened 2022-12-22
keystone: add v3 OS-FEDERATION mappings create operation

emilmaruszczak:
For https://github.com/gophercloud/gophercloud/issues/2461

[docs](https://docs.openstack.org/api-ref/identity/v3-ext/index.html#create-a-mapping)
[code](https://github.com/openstack/keystone/blob/stable/yoga/keystone/api/os_federation.py#L275)


mandre (approved):
LGTM.

Issue #2532, opened 2023-01-05
Enhancement: Support value_specs for ports


lentzi90:
We have a need for using [value_specs](https://docs.openstack.org/heat/zed/template_guide/openstack.html#OS::Neutron::Port-prop-value_specs) for Ports. This is an optional map of extra parameters to send as part of the request, and it is part of the official API as seen in the link.

This has been previously proposed and discarded in https://github.com/gophercloud/gophercloud/issues/2180 but I think it is worth reconsidering since it is actually in the API. I'm opening a new issue to hopefully get some attention to this since there was no answer on the closed issue.

PR #2537, opened 2023-01-24
Support value_specs for Ports

lentzi90:
Fixes https://github.com/gophercloud/gophercloud/issues/2532

code in this PR:

Docs: https://docs.openstack.org/heat/zed/template_guide/openstack.html#OS::Neutron::Port-prop-value_specs

This replaces #2535 because a branch name with `/` is [hard to use with go modules](https://github.com/golang/go/issues/32955#issuecomment-511860346)...


pierreprinetti:
LGTM! @mandre any concerns?

mandre (approved):
> LGTM! @mandre any concerns?

No, I don't see any concers with implementing value_specs for ports.

Thanks for the patch @lentzi90.

PR #2540, opened 2023-01-26
[all] IsEmpty to check for HTTP status 204

pierreprinetti:
With this commit, all `IsEmpty() (bool, error)` functions now check the status code of the result. If it's `204`, they immediately confirm that the page is empty.

This change was introduced in commit 64ed1bc1 in `objectstorage` to support Swift instances running behind a reverse proxy that was truncating `content-type` headers from `204 Empty` responses. With this change, the same check is applied to all services.

While no functional impact is expected on non-proxied OpenStack modules, this patch will prevent issues on proxied clouds. In general, it also seems reasonable to expect `204 Empty` responses to be empty.

---

This change is the result of running:

```shell
find -name 'results.go' -exec sed -i 's|^\(func (\(.\+\) .\+) IsEmpty() (bool, error) {\)$|\1\nif \2.StatusCode==204 {\nreturn true,nil\n}\n|' {} \;
go fmt ./...
```

### DIFF
--- a/docs/contributor-tutorial/.template/results.go
+++ b/docs/contributor-tutorial/.template/results.go
@@ -44,6 +44,10 @@ type ResourcePage struct {
 
 // IsEmpty determines whether or not a page of RESOURCES contains any results.
 func (r ResourcePage) IsEmpty() (bool, error) {
+	if r.StatusCode == 204 {
+		return true, nil
+	}
+
 	resources, err := ExtractResources(r)
 	return len(resources) == 0, err
 }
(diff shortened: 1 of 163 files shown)


pierreprinetti:
@mandre you modified the milestone to a minor bump. Is that because you think this PR needs a semver:minor label?

mandre:
> @mandre you modified the milestone to a minor bump. Is that because you think this PR needs a semver:minor label?

Not at all, it's because I've merge a couple of PRs in main last week that requires a minor bump. Because we don't maintain stable branches yet, this supposed the next version is going to be 1.3.0.

PR #2545, opened 2023-02-02
baremetal: add inspection_{started,finished}_at to Node

### LINKED ISSUE #2544: baremetal: missing inspection_{started,finished}_at in Node
Not sure why we didn't add them, probably just overlooked.

dtantsur:
Fixes #2544

code in this PR:

https://docs.openstack.org/api-ref/baremetal/?expanded=show-node-details-detail#show-node-details

### DIFF
--- a/openstack/baremetal/v1/nodes/results.go
+++ b/openstack/baremetal/v1/nodes/results.go
@@ -245,6 +245,12 @@ type Node struct {
 
 	// The UTC date and time when the provision state was updated, ISO 8601 format. May be “null”.
 	ProvisionUpdatedAt time.Time `json:"provision_updated_at"`
+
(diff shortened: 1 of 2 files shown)


dtantsur:
Re unit test: the ancient Go you use is buggy and apparently compares time.Location by pointer not by value. I'm not sure what to do: if I fix it, I'll probably break recent Go (which passes this test).

mandre (commented):
LGTM, just waiting for the Victoria job to finish running before merging. The train job is currently broken, and we might just remove it as upstream has EOL-ed it and we don't see how we can fix it otherwise.

mandre (approved):
LGTM, thanks.
On a related news, we've also dropped the permafailing train job in https://github.com/gophercloud/gophercloud/pull/2546.

PR #2555, opened 2023-02-08
nova: introduce servers.ListDetails along with a simple servers.List

### LINKED ISSUE #2553: nova: List request always uses detailed server listing
There is no way to list servers without details, see https://docs.openstack.org/api-ref/compute/#list-servers vs https://docs.openstack.org/api-ref/compute/#list-servers-detailed
https://github.com/gophercloud/gophercloud/blob/49ee5b4f954aa1b1693feef63d3e841923a2685f/openstack/compute/v2/servers/requests.go#L98-L99

A simple comparison for 1k servers listing:

```
$ time curl -s https://nova:443/v2.1/servers -H "X-Auth-Token: ${OS_AUTH_TOKEN}" | jq -r '.servers|length'
985

real    0m1,907s
user    0m0,097s
sys     0m0,015s
$ time curl -s https://nove:443/v2.1/servers/detail -H "X-Auth-Token: ${OS_AUTH_TOKEN}" | jq -r '.servers|length'
985

real    0m20,425s
user    0m0,117s
sys     0m0,039s
```

simple servers listing takes 10 times faster results.

Proposal:

1) implement a NEW `ListNoDetail` function to list servers using `listURL(client)` call, keeping the backward compatibility for gophercloud
2) rename current `List` func to `ListDetail` and implement a new `List` func using NO-detail URL. This will break backward compatibility. Nevertheless I prefer this solution, since other API calls do have `ListDetail` functions, e.g. nova images, flavors, manila shares and baremetals.

see also: https://github.com/kubernetes/cloud-provider-openstack/issues/2089

kayrus:
Fixes #2553

### DIFF
--- a/openstack/compute/v2/servers/doc.go
+++ b/openstack/compute/v2/servers/doc.go
@@ -11,6 +11,26 @@ Example to List Servers
 		AllTenants: true,
 	}
 
+	allPages, err := servers.ListSimple(computeClient, listOpts).AllPages()
+	if err != nil {
+		panic(err)
+	}
+
+	allServers, err := servers.ExtractServers(allPages)
+	if err != nil {
+		panic(err)
+	}
+
+	for _, server := range allServers {
+		fmt.Printf("%+v\n", server)
+	}
+
+Example to List Detail Servers
+
+	listOpts := servers.ListOpts{
+		AllTenants: true,
+	}
+
 	allPages, err := servers.List(computeClient, listOpts).AllPages()
 	if err != nil {
 		panic(err)
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -94,7 +94,22 @@ func (opts ListOpts) ToServerListQuery() (string, error) {
 	return q.String(), err
 }
 
-// List makes a request against the API to list servers accessible to you.
+// ListSimple makes a request against the API to list servers accessible to you.
+func ListSimple(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pager {
+	url := listURL(client)
+	if opts != nil {
+		query, err := opts.ToServerListQuery()
+		if err != nil {
+			return pagination.Pager{Err: err}
+		}
+		url += query
+	}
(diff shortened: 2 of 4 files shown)


pierreprinetti:
Can you think of a way of adding this functionality in a non-breaking fashion?

kayrus:
@pierreprinetti I'd like to, but it won't be aligned with other services... See my thoughts in the original ticket.

pierreprinetti:
I don't think it's reasonable to merge a fix that breaks existing users for efficiency. If it's not possible to achieve what you need in a backwards-compatible manner, I'd say it's a wontfix.

kayrus:
I know it's a breaking change. But it's breaks for good reasons: align with other `ListDetails` functions in a project:
```
$ grep -rl 'func ListDetail(' openstack/
openstack/compute/v2/servers/requests.go
openstack/compute/v2/images/requests.go
openstack/compute/v2/flavors/requests.go
openstack/compute/v2/extensions/availabilityzones/requests.go
openstack/sharedfilesystems/v2/schedulerstats/requests.go
openstack/sharedfilesystems/v2/sharenetworks/requests.go
openstack/sharedfilesystems/v2/shares/requests.go
openstack/sharedfilesystems/v2/snapshots/requests.go
openstack/baremetal/v1/ports/requests.go
openstack/baremetal/v1/nodes/requests.go
openstack/blockstorage/extensions/backups/requests.go
openstack/containerinfra/v1/clusters/requests.go
```
It's not the first time gophercloud introduces breaking changes, see https://github.com/gophercloud/gophercloud/blob/master/CHANGELOG.md

pierreprinetti:
>It's not the first time gophercloud introduces breaking changes

I am not against breaking changes. However they have a cost. Increased efficiency is rarely going to balance that cost.

kayrus:
@pierreprinetti ok. I appreciate if you can suggest a name for a _simple_ list function.

pierreprinetti:
ehr... ListSimple?

I hope you are more creative than I am :grimacing:

kayrus:
@pierreprinetti done

EmilienM (approved):

mandre:
I renamed the PR to reflect the latest code change.

PR #2560, opened 2023-02-16
loadbalancer: Use CreateMemberOpts instead of BatchUpdateMemberOpts in PoolCreateOpts

### LINKED ISSUE #2559: Loadbalancer create fully populated LB broken for empty subnet id in members
When creating a fully populated loadbalancer, the listener, health-manager as well as the pool with its members can be declared in one API call.
However in some cases the API call can include subnet_id: "" for the members, which results in an failed API call, since it is not a valid UUID as expected by octavia.

I would like to propose changing the datatype of the Members variable in the pool CreateOpts in openstack/loadbalancer/v2/pools/requests.go to be of type CreateMemberOpts instead of BatchUpdateMemberOpts.

For more explanation:
In the kubernetes openstack-cloud-provider (https://github.com/kubernetes/cloud-provider-openstack/blob/88ddaa260bf2b2f9a7cf60ab2fb41086f87b3e6e/pkg/openstack/loadbalancer.go#L1229) a subnet-id with empty string is passed in case the value is not defined. 
gophercloud uses the BatchUpdateMemberOpts in the pool CreateOpts struct instead of CreateMemberOpts.
The difference between CreateMemberOpts and BatchUpdateMemberOpts are three strings, that are declared as pointers in order to be able to delete those values during an update call.
In order to still support an empty string, the BatchUpdateMemberOpts have a secondary method called ToBatchMemberUpdateMap, which should translate an empty subnet_id to nil.
Since the ToBatchMemberUpdateMap is not called in all cases, it can happen, that the subnet_id is transmitted to the octavia API with value "" (empty string). This is not recognised as a valid UUID and thus the API call to octavia fails.

danfai:
In order to support members without any subnet, the struct for the pool members need to be CreateMemberOpts.

Fixes #2559

code in this PR:
API validation for UUID in octavia:
https://opendev.org/openstack/octavia/src/commit/c2c59f4c9eb9f9ef6081e386bf2bc1badc145241/octavia/api/v2/types/member.py#L127

### DIFF
--- a/acceptance/openstack/loadbalancer/v2/loadbalancer.go
+++ b/acceptance/openstack/loadbalancer/v2/loadbalancer.go
@@ -195,7 +195,7 @@ func CreateLoadBalancerFullyPopulated(t *testing.T, client *gophercloud.ServiceC
 				Description: poolDescription,
 				Protocol:    pools.ProtocolHTTP,
 				LBMethod:    pools.LBMethodLeastConnections,
-				Members: []pools.BatchUpdateMemberOpts{{
+				Members: []pools.CreateMemberOpts{{
 					Name:         &memberName,
 					ProtocolPort: memberPort,
 					Weight:       &memberWeight,
--- a/acceptance/openstack/loadbalancer/v2/loadbalancer.go
+++ b/acceptance/openstack/loadbalancer/v2/loadbalancer.go
@@ -196,11 +196,11 @@ func CreateLoadBalancerFullyPopulated(t *testing.T, client *gophercloud.ServiceC
 				Protocol:    pools.ProtocolHTTP,
 				LBMethod:    pools.LBMethodLeastConnections,
 				Members: []pools.CreateMemberOpts{{
-					Name:         &memberName,
+					Name:         memberName,
 					ProtocolPort: memberPort,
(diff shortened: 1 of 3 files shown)


EmilienM:
Could you please update the tests accordingly, so at least we get happy CI jobs. Thanks

EmilienM:
also FYI I marked this PR as non backward compatible since we're changing the type of an existing parameter in a struct.

dulek:
I'm not sure about the CI, but the change makes sense to me.

dulek:
Wait, are we allowed to break API like this? With this change `SubnetID` becomes `string` instead of `*string`. Same goes with `Name` and `MonitorAddress`.

dulek:
Alright, cool, ten works for me.

mandre (approved):

pierreprinetti:
@dulek @kayrus 
Since we tagged Gophercloud `v1.0.0`, we do not allow breaking changes in the `v1` branch. @mandre merged this PR to the development branch (`master`) with the `semver:major` label, which means that we won't backport to `v1`. We will only release this patch in a `v2` branch and tag it `v2.y.z`.

@mandre 
In the future, please make sure that PRs are well-squashed before merging them.

Issue #2565, opened 2023-02-23
CI: Many jobs fail cloning repos from opendev.org


mandre:
Many CI jobs fail with GnuTLS error, for example:
```
+ functions-common:git_timed:639           :   timeout -s SIGINT 0 git clone https://opendev.org/openstack/glance.git /opt/stack/glance --branch stable/train
Cloning into '/opt/stack/glance'...
error: RPC failed; curl 56 GnuTLS recv error (-110): The TLS connection was non-properly terminated.
fatal: expected 'packfile'
```

https://github.com/gophercloud/gophercloud/actions/runs/4248650726/jobs/7388053337

Many of the answers on stackoverflow for this error message point that it's a known issue with git compiled against gnutls and suggest compiling it against openssl. The github action image installs git from the [git-core ppa](https://launchpad.net/~git-core/+archive/ubuntu/ppa/) and is using gnutls. The official package from the ubuntu repositories also uses gnutls.
We might want to build git ourselves, however that seems odd, instead let's try to use github mirrors to pull the code instead of opendev.org.

mandre:
Switching devstack-action to use github repos by default with https://github.com/EmilienM/devstack-action/pull/12.

EmilienM:
@mandre your PR landed and I released v0.10.0, we can update our CI jobs if you want.

PR #2568, opened 2023-02-24
Prefer github mirrors over opendev repos

mandre:
The jobs are running in the Github infra and we should prefer github repos when possible.

Bump devstack-action to v0.10 that changes the default GIT_BASE to point to github repos rather than opendev.
Also change the URL of devstack-plugins to prefer github mirrors over opendev repos.

Hopefully mitigates #2565.


mandre:
This looks a lot greener. Zaqar and Senlin jobs fail on master because of non-compatible mongodb (`Zaqar needs Mongo DB version >= 6.0 to run.`).

EmilienM (approved):

PR #2576, opened 2023-03-01
keystone: add v3 OS-FEDERATION mappings delete operation

emilmaruszczak:
For https://github.com/gophercloud/gophercloud/issues/2461

[docs](https://docs.openstack.org/api-ref/identity/v3-ext/index.html#delete-a-mapping)
[code](https://github.com/openstack/keystone/blob/stable/yoga/keystone/api/os_federation.py#L301)

### DIFF
--- a/acceptance/openstack/identity/v3/federation_test.go
+++ b/acceptance/openstack/identity/v3/federation_test.go
@@ -6,6 +6,7 @@ package v3
 import (
 	"testing"
 
+	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/acceptance/clients"
 	"github.com/gophercloud/gophercloud/acceptance/tools"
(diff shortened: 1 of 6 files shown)


acceptance/openstack/identity/v3/federation_test.go
```diff
@@ -109,4 +109,7 @@ func TestMappingsCRUD(t *testing.T) {
 	th.AssertNoErr(t, err)
 	th.AssertEquals(t, len(updateOpts.Rules), len(updatedMapping.Rules))
 	th.CheckDeepEquals(t, updateOpts.Rules[0], updatedMapping.Rules[0])
+
+	err = federation.DeleteMapping(client, mappingName).ExtractErr()
+	th.AssertNoErr(t, err)
```
mandre, line 114: Can you perhaps also check that the mapping effectively no longer exists?

mandre (approved):
LGTM.

PR #2578, opened 2023-03-04
Add periodic jobs for OpenStack zed release and reduce periodic jobs frequency

mandre:
This PR adds periodic jobs for the zed release that came out in oct 22. It also reduce the frequency of periodic jobs to run once every 3 days rather than every day.


EmilienM:
FYI @mandre I released a new devstack-action including your PR to force the deployment on devstack, for any OS.

EmilienM (approved):
LGTM, thanks

PR #2583, opened 2023-03-06
Add missing rule protocol constants for IPIP

### LINKED ISSUE #2582: Missing SG rule protocol constant for IPIP
I noticed that Neutron (on Wallaby) returned the following list of protocols when providing a missing/unsupported protocol:
```
{"NeutronError": {"type": "SecurityGroupRuleInvalidProtocol", "message": "Security group rule protocol -1 not supported. Only protocol values [None, 'ah', 'dccp', 'egp', 'esp', 'gre', 'hopopt', 'icmp', 'igmp', 'ip', 'ipip', 'ipv6-encap', 'ipv6-frag', 'ipv6-icmp', 'icmpv6', 'ipv6-nonxt', 'ipv6-opts', 'ipv6-route', 'ospf', 'pgm', 'rsvp', 'sctp', 'tcp', 'udp', 'udplite', 'vrrp'] and integer representations [0 to 255] are supported.", "detail": ""}
```

By comparing the lists of Neutron supported protocols i saw that we miss:
* `ipip`

In the `CreateOpts` constant list here: https://github.com/gophercloud/gophercloud/blob/7edb0d08b18f7ea2ca39b0d3fd1a86bf9fee5faa/openstack/networking/v2/extensions/security/rules/requests.go#L51-L78

Links:
* From the `neutron-lib`: https://github.com/openstack/neutron-lib/blob/master/neutron_lib/constants.py#L153-L180
* From the official OpenStack Neutron API ref: https://docs.openstack.org/api-ref/network/v2/?expanded=create-security-group-rule-detail#create-security-group-rule

mikejoh:
Fixes #2582


mandre (commented):
The `ip` and `hopopt` protocols are not documented in https://docs.openstack.org/api-ref/network/v2/?expanded=create-security-group-rule-detail#create-security-group-rule. They also share the same [number 0](https://github.com/openstack/neutron-lib/blob/9d58d4a/neutron_lib/constants.py#L187). Are they aliases to the `any` protocol?

mikejoh:
> The `ip` and `hopopt` protocols are not documented in https://docs.openstack.org/api-ref/network/v2/?expanded=create-security-group-rule-detail#create-security-group-rule. They also share the same [number 0](https://github.com/openstack/neutron-lib/blob/9d58d4a/neutron_lib/constants.py#L187). Are they aliases to the `any` protocol?

Good catch, my bad, i didn't check the protocols numbers.. I think that `ipip`, which has it's owned assigned unique number would make sense to add though. I think i'll change the scope of this PR and the related issue to only add `ipip`.

mandre (approved):
LGTM, thanks.

PR #2584, opened 2023-03-07
CI: workaround mongodb dependency for messaging and clustering master jobs

mandre:
The zaqar devstack plugin has a workaround for the failing dependency
for ubuntu version 22.04. Let's use this as the runner in those jobs.

### DIFF
--- a/.github/workflows/functional-clustering.yaml
+++ b/.github/workflows/functional-clustering.yaml
@@ -12,7 +12,7 @@ jobs:
       matrix:
         name: ["master"]
         openstack_version: ["master"]
-        ubuntu_version: ["20.04"]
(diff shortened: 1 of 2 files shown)


mandre:
My conditional run is not working :-/

mandre:
It seems the zaqar upstream CI uses openSUSE to workaround the issue. We don't have that possibility since github action [only offers Ubuntu  for Linux runners](https://docs.github.com/en/actions/using-jobs/choosing-the-runner-for-a-job#choosing-github-hosted-runners). Instead, I'm trying https://review.opendev.org/c/openstack/zaqar/+/876773 to fix the jobs.

EmilienM (approved):
LGTM, thanks

PR #2585, opened 2023-03-08
Use ubuntu 20.04 instead of 18.04 for train and ussuri jobs

mandre:
The Ubuntu 18.04 image for github action runner was deprecated. https://github.blog/changelog/2022-08-09-github-actions-the-ubuntu-18-04-actions-runner-image-is-being-deprecated-and-will-be-removed-by-12-1-22/

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -28,17 +28,14 @@ jobs:
             ubuntu_version: "20.04"
           - name: "ussuri"
             openstack_version: "stable/ussuri"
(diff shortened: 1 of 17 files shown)


mandre:
Requires https://github.com/EmilienM/devstack-action/pull/13

mandre:
This feels like trying to run train and ussuri jobs on ubuntu 20.04 is bound to fail, and even if we get it to work, it's going to be a nightmare to maintain.
We'll likely have to drop the jobs once github retires the 18.04 image.

PR #2597, opened 2023-03-22
Add validation and optimise objects.BulkDelete

pierreprinetti:
This PR makes the following changes to `objects.BulkDelete`:
* validate that the container name does not contain `/`;
* validate that all provided objects are non-empty;
* optimise how the request body is built.

Note that the function mentions urlencoding in a comment, but there is no urlencoding. I have removed the comment and kept the behaviour. We might want to consider add urlencoding in a follow-up.

---

WRT the optimisation:

Use `bytes.Buffer` instead of repeatedly copying strings to compose the request body for `objects.BulkDelete`.

This PR adds, and in a separate commit, removes the benchmark tests. To execute the benchmark:

```shell
git reset --hard f282ec49
cd openstack/objectstorage/v1/objects/testing
go test -bench=.
```

Locally I get a marginal gain when bulk-deleting only a few objects, and
good gains (to around 3/4 of the time per execution) with large batches:

```plaintext
goos: linux
goarch: amd64
pkg: github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects/testing
cpu: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
BenchmarkBulkDelete-8                                246           4536244 ns/op
BenchmarkBulkDeleteFewerObjects-8                  11943             97338 ns/op
BenchmarkBulkDeleteWithBuffer-8                      303           3463102 ns/op
BenchmarkBulkDeleteWithBufferFewerObjects-8        11587             92153 ns/op
PASS
ok      github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects/testing   19.943s
```

### DIFF
--- a/openstack/objectstorage/v1/objects/requests.go
+++ b/openstack/objectstorage/v1/objects/requests.go
@@ -576,11 +576,21 @@ func CreateTempURL(c *gophercloud.ServiceClient, containerName, objectName strin
 
 // BulkDelete is a function that bulk deletes objects.
 func BulkDelete(c *gophercloud.ServiceClient, container string, objects []string) (r BulkDeleteResult) {
+	err := containers.CheckContainerName(container)
+	if err != nil {
+		r.Err = err
+		return
+	}
+
 	// urlencode object names to be on the safe side
 	// https://github.com/openstack/swift/blob/stable/train/swift/common/middleware/bulk.py#L160
 	// https://github.com/openstack/swift/blob/stable/train/swift/common/swob.py#L302
 	encodedObjects := make([]string, len(objects))
 	for i, v := range objects {
+		if v == "" {
+			r.Err = fmt.Errorf("object names must not be the empty string")
+			return
+		}
 		encodedObjects[i] = strings.Join([]string{container, v}, "/")
 	}
 	b := strings.NewReader(strings.Join(encodedObjects, "\n") + "\n")
--- a/openstack/objectstorage/v1/objects/requests.go
+++ b/openstack/objectstorage/v1/objects/requests.go
@@ -604,3 +604,34 @@ func BulkDelete(c *gophercloud.ServiceClient, container string, objects []string
 	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
 	return
 }
+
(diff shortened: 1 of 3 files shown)


kayrus:
@pierreprinetti I wonder whether it will work with the new versioning system
https://github.com/gophercloud/gophercloud/blob/master/acceptance/openstack/objectstorage/v1/versioning_test.go

kayrus:
@pierreprinetti I don't know. Just thinking out loud. It may happen that this function consumers may face 409 on objects or container deletion :\

openstack/objectstorage/v1/objects/requests.go
```diff
@@ -576,15 +576,25 @@ func CreateTempURL(c *gophercloud.ServiceClient, containerName, objectName strin
 
 // BulkDelete is a function that bulk deletes objects.
```
stephenfin, line 577: Unrelated to this change, but I wonder if we should mention the (configurable) limit of 10,000 objects per request?

stephenfin (approved):
Looks sane to me

openstack/objectstorage/v1/objects/requests.go
pierreprinetti: I wasn't able to find a reference to a limit [in the API docs](https://docs.openstack.org/api-ref/object-store/#delete-object). Did you find that in Swift or Ceph?

EmilienM (approved):

openstack/objectstorage/v1/objects/requests.go
EmilienM: I think it's defined by the user in Swift config:

https://github.com/openstack/swift/blob/6d3d4197151f44bf28b51257c1a4c5d33411dcae/etc/proxy-server.conf-sample#L1029-L1034
https://github.com/openstack/swift/blob/e8cecf7fcc1630ee83b08f9a73e1e59c07f8d372/swift/common/middleware/bulk.py#L309

pierreprinetti: wonderful, thank you @stephenfin @EmilienM. I have added that information in the function docs. PTAL

mandre (approved):

PR #2599, opened 2023-03-23
Allows modification of the 'Host' field in the 'req.headers'.

Leeclow-ops:
For the codes At https://github.com/gophercloud/gophercloud/blob/master/provider_client.go#L412-L416
Hope enables the modification of the 'Host' field in the 'req.headers'.

Fixes #[2598]


pierreprinetti:
Did you really mean to file a pull request? You changed the import everywhere.

pierreprinetti:
Why do you want to modify the Host field of a request? What is the use case?

kayrus:
@Leeclow-ops for these cases openstack has internal and public endpoints. Have you considered to pass the `gophercloud.AvailabilityInternal` into `gophercloud.EndpointOpts` while creating a service client?

PR #2606, opened 2023-04-05
Add convenient access to error's original body

### LINKED ISSUE #2605: Add convenient access to error's original body
I propose to add a convenient access to error's original body. It will help in a case when it is necessary not to expose such information as the request URL to the end user.

skunkie:
Fixes #2605

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -94,9 +94,24 @@ func (e ErrUnexpectedResponseCode) Error() string {
 	return e.choseErrString()
 }
 
+// GetBody returns the original error message
+func (e ErrUnexpectedResponseCode) GetBody() []byte {
+	return e.Body
+}
+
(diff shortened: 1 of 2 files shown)


pierreprinetti:
I have tested your patch by trying to access the object storage with a user that doesn't have the swiftoperator role. I wasn't able to get the response body. Do you have an example of your patch working as intended, so that I can compare?

This is what I tried: https://codeberg.org/pierreprinetti/test_gophercloud_errorbody

Thank you.

skunkie:
It is a better approach, thank you!

pierreprinetti:
Here's the full proposal: https://github.com/gophercloud/gophercloud/pull/2639

@skunkie 
Please close this PR if (and only if!) you think that #2639 one meets your needs.

Issue #2612, opened 2023-05-08
RFE: new baremetal inventory API


dtantsur:
The ironic team is working [1] on merging Inspector into Ironic. There will be several work items related to this, the first being the new node inventory API [2] based on the existing introspection data API. A tricky part will be to avoid hardcoding data structures that are not formally defined. Currently, gophercloud takes a pretty relaxed approach to this problem, simply assuming the internal Inspector data does not change. With the migration to Ironic, it will change, and we need to take that into account.

[1] https://specs.openstack.org/openstack/ironic-specs/specs/approved/merge-inspector.html
[2] https://docs.openstack.org/api-ref/baremetal/?expanded=show-node-inventory-detail#node-inventory

PR #2613, opened 2023-05-08
Migrate baremetal inventory to a common location

dtantsur:
With the introduction of the Ironic inventory API, these types are now shared between Ironic and Inspector. No new functionality here yet, posting it to get the refactoring out of my way. Only the documented parts of the introspection data is migrated, not the Inspector-specific ones.

This change is technically breaking since the code is moved around and can no longer be imported at the old locations.

Part of #2612


EmilienM:
@dtantsur thanks for your patch, this is good to me, just rebase it please and I'll merge it.

dtantsur:
I will, just need to find some time. Rebase won't be trivial, unfortunately.

dtantsur:
@EmilienM done!

EmilienM (approved):

PR #2619, opened 2023-05-17
Fix version comparison for acceptance tests

mandre:
Some acceptance tests rely on a version comparison to conditionally skip tests. It turns out we were not consistently setting the relevant `OS_BRANCH` environment variable in all the jobs, meaning the version comparison would fail and skip the test.

This PR does two things:
- consistently set the `OS_BRANCH` variable for all our acceptance jobs
- fail the test when the required `OS_BRANCH` variable is not set

### DIFF
--- a/.github/workflows/functional-identity.yaml
+++ b/.github/workflows/functional-identity.yaml
@@ -47,6 +47,7 @@ jobs:
         env:
           DEVSTACK_PATH: ${{ github.workspace }}/devstack
           ACCEPTANCE_TESTS_FILTER: "^.*identity.*$"
+          OS_BRANCH: ${{ matrix.openstack_version }}
       - name: Generate logs on failure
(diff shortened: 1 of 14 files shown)


.github/workflows/functional-identity.yaml
```diff
@@ -47,6 +47,7 @@ jobs:
         env:
           DEVSTACK_PATH: ${{ github.workspace }}/devstack
           ACCEPTANCE_TESTS_FILTER: "^.*identity.*$"
+          # OS_BRANCH: ${{ matrix.openstack_version }}
```
mandre, line 50: This is intentional, to verify the test fail when OS_BRANCH isn't set.
I'll change it back after I confirmed it works as intended.

pierreprinetti (approved):

PR #2620, opened 2023-05-17
Pin goimport dep to a version that works with go 1.14

mandre:
Fixes the unit tests when running with go 1.14.


pierreprinetti (approved):
Great, thanks

PR #2629, opened 2023-05-24
[manila]: Add share from snapshot restore functional test

kayrus:
This PR adds a functional test that restores a snapshot into a bigger share. Primarily this test is required to check whether Manila in devstack is capable to perform such an action.

### DIFF
--- a/acceptance/openstack/sharedfilesystems/v2/shares_test.go
+++ b/acceptance/openstack/sharedfilesystems/v2/shares_test.go
@@ -45,7 +45,7 @@ func TestShareExportLocations(t *testing.T) {
 
 	defer DeleteShare(t, client, share)
 
-	err = waitForStatus(t, client, share.ID, "available")
+	_, err = waitForStatus(t, client, share.ID, "available")
 	if err != nil {
 		t.Fatalf("Share status error: %v", err)
 	}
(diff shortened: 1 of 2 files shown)


acceptance/openstack/sharedfilesystems/v2/shares_test.go
```diff
@@ -353,6 +353,73 @@ func TestRevert(t *testing.T) {
 	t.Logf("Share %s successfuly reverted", share.ID)
 }
 
+func TestShareRestoreFromSnapshot(t *testing.T) {
+	client, err := clients.NewSharedFileSystemV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create a shared file system client: %v", err)
+	}
+	client.Microversion = "2.27"
```
mandre, line 361: The microversion doesn't seem right, [2.27 is where revert to snapshot was added](https://docs.openstack.org/manila/latest/contributor/api_microversion_history.html#id25), and it looks like that's not what we're using here. Not a big deal really, since we're way past microversion 2.27 (Ocata) anyway.

mandre (approved):

acceptance/openstack/sharedfilesystems/v2/shares_test.go
gouthampacha: but agree with the point here that "restoring from snapshot" here is really "creating share from snapshot" in manila terminology; and confusingly, manila has a feature called "revert_to_snapshot" which allows in-place reverts.. and this "2.27" version seems to come from that.

PR #2639, opened 2023-06-02
Implement errors.Unwrap() on unexpected status code errors

### LINKED ISSUE #2605: Add convenient access to error's original body

pierreprinetti:
With this patch, we add a method with signature `Unwrap() error` to all errors signaling an unexpected status code. All of them contain a value of type `gophercloud.ErrUnexpectedStatusCode` that exposes, among other useful values, the body of the response sent by the server.

To access the response body of a request that resulted in an unexpected response code, try to unwrap the returned error to `gophercloud.ErrUnexpectedStatusCode`:

```Go
pages, err := containers.List(client, nil).AllPages()
if err != nil {
	var responseCodeError gophercloud.ErrUnexpectedResponseCode
	if errors.As(err, &responseCodeError) {
		log.Printf("unexpected response code. Response body:\n---\n%s\n---", responseCodeError.Body)
	} else {
		log.Printf("unknown error")
	}
}
```

Fixes #2605


pierreprinetti:
Tested here --> https://codeberg.org/pierreprinetti/test_gophercloud_errorbody

mandre (approved):

Issue #2507, opened 2022-11-18
Manila may choke on dates with UTC offset


mandre:
While reviewing https://github.com/gophercloud/gophercloud/pull/2496 we noticed that Manila can return dates as ISO 8601 with an UTC offset that the sharedfilesystems module doesn't know how to parse.

We expect the `2006-01-02T15:04:05.999999` format [for JSONRFC3339MilliNoZ](https://github.com/gophercloud/gophercloud/blob/0721d75e876f20e0f3f9be83b3133fa3476ee992/results.go#L306), which is different from what Manila claims to return:

> The date and time stamp format is [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601):
> 
> CCYY-MM-DDThh:mm:ss±hh:mm
> 
> The ±hh:mm value, if included, returns the time zone as an offset from UTC.
> 
> For example, 2019-03-27T09:49:58-05:00.

Playing a little bit with it, it seems to break as soon as the date include an UTC offset: https://go.dev/play/p/pJh7jLgonx6. It looks like we're just being lucky.

pierreprinetti:
@mandre is there outstanding work required besides #2496?

mandre:
Yes, we identified the issue while reviewing https://github.com/gophercloud/gophercloud/pull/2496. We'll need to change the parsing to account for UTC offset, as this is something Manila can return, according to the docs.
This is also potentially an issue with other modules.

PR #2648, opened 2023-06-14
[manila]: implement share transfer API

### LINKED ISSUE #2647: [manila]: implement share transfer API
https://docs.openstack.org/api-ref/shared-file-system/?expanded=#share-transfer-since-api-v2-77

Mostly it's the same as transferring the Cinder Volumes, but for shares: https://github.com/gophercloud/gophercloud/blob/master/openstack/blockstorage/extensions/volumetransfers/requests.go

kayrus:
Fixes #2647

### DIFF
--- a/acceptance/clients/conditions.go
+++ b/acceptance/clients/conditions.go
@@ -147,7 +147,7 @@ func IsReleasesAbove(t *testing.T, release string) bool {
 func IsReleasesBelow(t *testing.T, release string) bool {
 	current_branch := getReleaseFromEnv(t)
 
-	if current_branch != "master" && current_branch < release {
+	if current_branch != "master" || current_branch < release {
 		return true
 	}
 	t.Logf("Target release %s is above the current branch %s", release, current_branch)
--- a/openstack/sharedfilesystems/v2/sharetransfers/results.go
+++ b/openstack/sharedfilesystems/v2/sharetransfers/results.go
@@ -1,0 +1,170 @@
+package sharetransfers
+
+import (
+	"encoding/json"
+	"net/url"
+	"strconv"
+	"time"
+
+	"github.com/gophercloud/gophercloud"
+	"github.com/gophercloud/gophercloud/pagination"
+)
+
+const (
+	invalidMarker = "-1"
+)
+
+// Transfer represents a Share Transfer record.
+type Transfer struct {
+	ID                   string              `json:"id"`
+	Accepted             bool                `json:"accepted"`
+	AuthKey              string              `json:"auth_key"`
+	Name                 string              `json:"name"`
+	SourceProjectID      string              `json:"source_project_id"`
+	DestinationProjectID string              `json:"destination_project_id"`
+	ResourceID           string              `json:"resource_id"`
+	ResourceType         string              `json:"resource_type"`
+	CreatedAt            time.Time           `json:"-"`
+	ExpiresAt            time.Time           `json:"-"`
+	Links                []map[string]string `json:"links"`
+}
+
+// UnmarshalJSON is our unmarshalling helper.
+func (r *Transfer) UnmarshalJSON(b []byte) error {
+	type tmp Transfer
+	var s struct {
+		tmp
+		CreatedAt gophercloud.JSONRFC3339MilliNoZ `json:"created_at"`
+		ExpiresAt gophercloud.JSONRFC3339MilliNoZ `json:"expires_at"`
+	}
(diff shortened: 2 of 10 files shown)


acceptance/clients/conditions.go
```diff
@@ -147,7 +147,7 @@ func IsReleasesAbove(t *testing.T, release string) bool {
 func IsReleasesBelow(t *testing.T, release string) bool {
 	current_branch := getReleaseFromEnv(t)
 
-	if current_branch != "master" && current_branch < release {
+	if current_branch != "master" || current_branch < release {
```
kayrus, line 150: This fix allows to run tests on master branch only while using `clients.SkipReleasesBelow(t, "master")`

kayrus:
@mandre @pierreprinetti ready for review

openstack/sharedfilesystems/v2/sharetransfers/results.go
```diff
@@ -0,0 +1,167 @@
...
+type Transfer struct {
+	ID                   string              `json:"id"`
+	Accepted             bool                `json:"accepted"`
+	AuthKey              string              `json:"auth_key"`
+	Name                 string              `json:"name"`
+	SourceProjectID      string              `json:"source_project_id"`
+	DestinationProjectID string              `json:"destination_project_id"`
+	ResourceID           string              `json:"resource_id"`
+	ResourceType         string              `json:"resource_type"`
+	CreatedAt            time.Time           `json:"-"`
```
mandre, line 27: Perhaps we could also add [`expires_at`](https://docs.openstack.org/api-ref/shared-file-system/?expanded=show-share-transfer-detail-detail#id774), WDYT?

openstack/sharedfilesystems/v2/sharetransfers/results.go
```diff
@@ -0,0 +1,167 @@
...
+	Links                []map[string]string `json:"links"`
+}
+
+// UnmarshalJSON is our unmarshalling helper.
+func (r *Transfer) UnmarshalJSON(b []byte) error {
+	type tmp Transfer
+	var s struct {
+		tmp
+		CreatedAt gophercloud.JSONRFC3339MilliNoZ `json:"created_at"`
```
mandre, line 36: Which makes me think, we still have https://github.com/gophercloud/gophercloud/issues/2507 :sweat_smile:

kayrus:
You're right in all comments :) Thanks. As for the UTC offset, for some reason I cannot access debug logs for intermediate push :\ I don't have this API locally, therefore I rely on API reference and CI tests :)

kayrus:
According to debug logs the time format is fine:

```
2023/06/21 07:48:00 [DEBUG] OpenStack Response Body: {
  "transfer": {
    "accepted": false,
    "auth_key": "4fbbb3ba15e6549e",
    "created_at": "2023-06-21T07:48:00.589548",
    "destination_project_id": null,
    "expires_at": "2023-06-21T07:53:00.587670",
    "id": "787da824-1b89-46f5-9ec7-e0b5afa2f28c",
    "links": [
      {
        "href": "http://10.1.120.0/share/v2/share-transfer/787da824-1b89-46f5-9ec7-e0b5afa2f28c",
        "rel": "self"
      },
      {
        "href": "http://10.1.120.0/share/share-transfer/787da824-1b89-46f5-9ec7-e0b5afa2f28c",
        "rel": "bookmark"
      }
    ],
    "name": "123",
    "resource_id": "dbae4463-25f3-42c3-9b18-46bc29ff733d",
    "resource_type": "share",    
    "source_project_id": "98df2063d7a8467eadc137dd4c5f4ca7"
  }
}
```

mandre:
Yep, because our devstack returns the date without the UTC offset, which is why we're able to parse it. The `±hh:mm` part is optional in the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format. It would break if manila included it in it's responses.
It's not a big deal, as far as I know nobody reported issues, so that's probably not common at all in the wild, or perhaps just a documentation problem.

mandre (approved):

kayrus:
@mandre BTW, does it make sense to turn on the `OS_DEBUG=true` for all tests?

mandre:
> @mandre BTW, does it make sense to turn on the `OS_DEBUG=true` for all tests?

I'm afraid this is going to generate too much logs. The errors are generally obvious when we debug a CI job. We can always manually add it when needed, like you did.

PR #2653, opened 2023-06-22
[manila]: introduce WaitForStatus helpers for manila Shares, Replicas…

### LINKED ISSUE #2652: [manila]: introduce `WaitForStatus` helpers for manila Shares, Replicas, Snapshots
Introduce `WaitForStatus` helpers for manila Shares, Replicas, Snapshots.

Like it's already done for cinder block storage.

kayrus:
Resolves #2652

### DIFF
--- a/openstack/sharedfilesystems/v2/replicas/util.go
+++ b/openstack/sharedfilesystems/v2/replicas/util.go
@@ -1,0 +1,39 @@
+package replicas
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// WaitForStatus will continually poll the resource, checking for a particular
+// status. It will do this for the amount of seconds defined.
+func WaitForStatus(c *gophercloud.ServiceClient, id, status string, secs int) error {
+	return gophercloud.WaitFor(secs, func() (bool, error) {
+		current, err := Get(c, id).Extract()
+		if err != nil {
+			return false, err
+		}
+
+		if current.Status == status {
+			return true, nil
+		}
+
+		return false, nil
+	})
+}
+
+// WaitForState will continually poll the resource, checking for a particular
+// state. It will do this for the amount of seconds defined.
+func WaitForState(c *gophercloud.ServiceClient, id, state string, secs int) error {
+	return gophercloud.WaitFor(secs, func() (bool, error) {
+		current, err := Get(c, id).Extract()
+		if err != nil {
+			return false, err
+		}
+
+		if current.State == state {
+			return true, nil
+		}
+
+		return false, nil
+	})
+}
--- a/openstack/sharedfilesystems/v2/shares/util.go
+++ b/openstack/sharedfilesystems/v2/shares/util.go
@@ -1,0 +1,22 @@
+package shares
+
+import (
+	"github.com/gophercloud/gophercloud"
+)
+
+// WaitForStatus will continually poll the resource, checking for a particular
+// status. It will do this for the amount of seconds defined.
+func WaitForStatus(c *gophercloud.ServiceClient, id, status string, secs int) error {
+	return gophercloud.WaitFor(secs, func() (bool, error) {
+		current, err := Get(c, id).Extract()
+		if err != nil {
+			return false, err
+		}
+
+		if current.Status == status {
+			return true, nil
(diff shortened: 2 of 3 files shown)


kayrus:
cc @mandre @pierreprinetti kindly ping.

pierreprinetti:
Thank you for your addition. I finally find a moment to look into it.

I appreciate the need for polling logic in a consumer application. However, I wonder if Gophercloud is the right place where to code it.

[The `WaitFor` function](https://github.com/gophercloud/gophercloud/blob/b98471812a77380b2bb25a28b04bb1be9fc2e72e/util.go#L92) doesn't do anything OpenStack-specific and I honestly would like to deprecate and remove it from Gophercloud at some point. A function like that should accept a `context.Context` instead, and there are a gazillion libraries out there that do that job.

So my question is: would this addition solve a problem that you can't reasonably solve using a retry function from a specialised library?

kayrus:
@pierreprinetti I was under impression that once `WaitForStatus` functions are already in the codebase, then it makes sense to add them to another packages. I understand that `WaitFor` doesn't do anything openstack-related, but gophercloud is an SDK and helps people to build their applications in a more convenient and clean way, and I'd not remove this functionality from gophercloud.

There was another initiative, when `IDFromName` were moved to `gophercloud/utils`. I cannot say that it really helped, from my point of view it just added more confusion. But if you think that these utils have to be removed from gophercloud, I'd just moved them to utils then...

pierreprinetti:
I think I understand your point of view. You are perfectly right: other packages use the same `WaitFor` function and they have been since forever. Your addition looks consistent with Gophercloud as it is.

My point is: if given the choice, I'd rather not maintain a retry library in this repository. `utils` may have been sort of a compromise. However if it were for me, I'd consider removing `WaitFor` in Gophercloud `v2`. I understand that consumer applications need that logic; on the other hand, they can find equivalent (or better) code somewhere else, or implement it on their own. I can imagine that based on the application you write, or based on the cloud you operate with, you will want a different polling interval, or you may want to be able to cancel operation if your caller is not interested in the result any more (e.g. when your application acts upon an HTTP-induced trigger and the connection is closed). You may even want exponential backoff in some situations. That's a whole bunch of use cases we're unlikely to satisfy, and for the reason that it's not really Gophercloud's business. That's why I would immediately deprecate `WaitFor` and plan its removal.

So my question is: why do you think it's best to have this function here, rather than directly in you application's code?

However, I am not the sole decision maker here! Let's see what others think @mandre @EmilienM

kayrus:
> why do you think it's best to have this function here, rather than directly in you application's code?

I work with a number of OpenStack consumer projects that use these shared functions. Moving them into each project means: more code duplication, complicated maintainability, potential license copyright problems, less attention from the community for code improvement.

I'd rather move them to utils. If necessary, the `context` support can be added latyer.

pierreprinetti:
>I'd rather move them to utils.

Let's see what the other Gophercloud maintainers think. I don't think your function should go into `utils`: either we keep consistent and accept them here in gophercloud/gophercloud, or we deprecate the wait functions altogether. After a quick search on Github, I didn't really find that gazillion alternative modules I had imagined.

mandre:
I personally don't see a problem having convenience functions in gophercloud to make it easier to consume. Let's also remember we are also consumers, via the acceptance tests, although we're using a [_different_ implementation of WaitFor()](https://github.com/gophercloud/gophercloud/blob/157270b09bb3e14c233f313c8c6db7a9ebdb4c4a/acceptance/tools/tools.go#L15-L37) :sweat_smile: 
The other one being [in util.go](https://github.com/gophercloud/gophercloud/blob/157270b09bb3e14c233f313c8c6db7a9ebdb4c4a/util.go#L86-L130).

EmilienM:
I'll like to see this function in https://github.com/gophercloud/gophercloud/blob/master/util.go if possible and use it from acceptance, so we can remove https://github.com/gophercloud/gophercloud/blob/157270b09bb3e14c233f313c8c6db7a9ebdb4c4a/acceptance/tools/tools.go#L15-L37 at some point.

kayrus:
I can also implement something that uses context.

Issue #1330, opened 2018-11-29
What would it look like if gophercloud wanted to attempt best-effort support for Rackspace's OpenStack public cloud?


clayg:
Gophercloud appears to partially work with Rackspace's OpenStack public cloud and has been used with some success via terraform [1].

One of the "known issues" relates to the network api [2].

The foundation seems to support the notion that Rackspace *is* an OpenStack cloud [3] - does that mean I "should" be able to use openstack sdk's like python-openstacksdk [4] and gophercloud with Rackspace?

If that's a worthwhile goal, what would progress toward that goal look like?

@emonty @jkoelker Do you have any color to add to the version end-point mangling?  I'm not sure exactly how to frame the "issue" - I mean the url is wrong and the code that makes the url would need to change for the url to be correct?  :man_shrugging:

1. https://www.terraform.io/docs/providers/openstack/index.html#openstack-releases-and-versions
2. incomplete attempt at a "fix" #1329
3. https://www.openstack.org/marketplace/public-clouds/rackspace/rackspace-public-cloud
4. some sanding of rough edges is not without precedence in openstack sdk's https://storyboard.openstack.org/#!/story/2004310

emonty:
openstacksdk TOTALLY works on Rackspace - but yes, there are many of the hacks in openstacksdk to make it that way.

We do what you're doing in #1329 in openstacksdk. To answer @jtopjian question from the PR, Rackspace is the only cloud I am aware of putting a versioned neutron endpoint into the catalog. Most clouds correctly put the unversioned endpoint into the catalog. You'd think this would be great - but up until recent releases, neutron version discovery didn't work. Nobody noticed though, since python-neutronclient didn't actually do discovery but instead just appended v2.0 to the unversioned endpoint from the catalog. Even though I find this behavior reprehensible, it winds up being the thing that works on the widest number of clouds. In order to work with Rackspace *and* the older clouds that don't support version discover, openstacksdk does a hack very similar to the one in #1329.

I'm not happy about it - but it allows the rest of the things to work.

In general, Rackspace works quite well as a pure OpenStack cloud. We use it extensively and in no circumstances do we use Rackspace-specific drivers for anything. (Also, if anyone ever tells you you need to use the silly rackspace auth token they are lying - password auth works just fine)

jtopjian:
@clayg Thanks for starting this discussion. @emonty thank you for the info, too!

[...]

PR #2655, opened 2023-06-22
identity: use AuthOptionsBuilder for v2 auth

### LINKED ISSUE #2651: use Identity v2 AuthOptionsBuilder
So for Identity v3, gophercloud takes in an `AuthOptionsBuilder` from the `token3` which ensures that the data to authenticate with will conform to the Identity v3 APIs. This allows people to implement their own custom auth mechanisms like exist in the Python ecosystem which is great. This same support/ecosystem exists for Identity v2. In fact gophercloud originated from https://github.com/rackspace/gophercloud which has their own Identity v2 auth mechanism using an "API key". When gophercloud moved from under the Rackspace umbrella it rightfully removed all the "Rackspace-isms", however there's no mechanism by which to add this back.

I'm proposing using `tokens2.AuthOptionsBuilder` similar to how the Identity v3 auth works in gophercloud. This would improve #1330. However I'm aware of #2577 wanting to remove Identity v2 support, however Rackspace is still using Identity v2. I'd like to get Rackspace contributing more to gophercloud and using it more in the future.

As a first attempt of getting Rackspace to use upstream gophercloud I implemented https://github.com/os-pc/cloud-provider-rackspace/commit/56cf8a12d36c0a039edade9ba42457d9adf33faf#diff-b5b02d252ffde072fea7841ee6419db056b027d7f365193f0d33fad8294ffedc but I've now seen how the interfaces can be used and implemented and will improve it.

cardoe:
Rather than creating a tokens2.AuthOptions inside the authentication flow for identity v2, allow any structure that implements the tokens2.AuthBuilderOptions interface to be used. The interface is modified to add a CanReauth function like the v3 version has but provides a default implementation. The gophercloud.AuthOptions already implements this interface so there is no change to its API. Lastly add an AuthenticateV2Ext function which will allow an out of tree identity v2 auth mechanism to be implemented. fixes #2651. ref #1330.

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -99,7 +99,7 @@ func Authenticate(ctx context.Context, client *gophercloud.ProviderClient, optio
 
 	switch chosen.ID {
 	case v2:
-		return v2auth(ctx, client, endpoint, options, gophercloud.EndpointOpts{})
+		return v2auth(ctx, client, endpoint, &options, gophercloud.EndpointOpts{})
 	case v3:
 		return v3auth(ctx, client, endpoint, &options, gophercloud.EndpointOpts{})
 	default:
@@ -109,11 +109,17 @@ func Authenticate(ctx context.Context, client *gophercloud.ProviderClient, optio
 }
 
 // AuthenticateV2 explicitly authenticates against the identity v2 endpoint.
-func AuthenticateV2(ctx context.Context, client *gophercloud.ProviderClient, options gophercloud.AuthOptions, eo gophercloud.EndpointOpts) error {
+func AuthenticateV2(ctx context.Context, client *gophercloud.ProviderClient, options tokens2.AuthOptionsBuilder, eo gophercloud.EndpointOpts) error {
 	return v2auth(ctx, client, "", options, eo)
 }
 
-func v2auth(ctx context.Context, client *gophercloud.ProviderClient, endpoint string, options gophercloud.AuthOptions, eo gophercloud.EndpointOpts) error {
+type v2TokenNoReauth struct {
+	tokens2.AuthOptionsBuilder
+}
+
+func (v2TokenNoReauth) CanReauth() bool { return false }
+
+func v2auth(ctx context.Context, client *gophercloud.ProviderClient, endpoint string, options tokens2.AuthOptionsBuilder, eo gophercloud.EndpointOpts) error {
 	v2Client, err := NewIdentityV2(client, eo)
 	if err != nil {
 		return err
@@ -123,17 +129,7 @@ func v2auth(ctx context.Context, client *gophercloud.ProviderClient, endpoint st
 		v2Client.Endpoint = endpoint
 	}
 
-	v2Opts := tokens2.AuthOptions{
-		IdentityEndpoint: options.IdentityEndpoint,
-		Username:         options.Username,
-		Password:         options.Password,
-		TenantID:         options.TenantID,
-		TenantName:       options.TenantName,
-		AllowReauth:      options.AllowReauth,
-		TokenID:          options.TokenID,
-	}
-
-	result := tokens2.Create(ctx, v2Client, v2Opts)
+	result := tokens2.Create(ctx, v2Client, options)
 
 	err = client.SetTokenAndAuthResult(result)
 	if err != nil {
@@ -145,7 +141,7 @@ func v2auth(ctx context.Context, client *gophercloud.ProviderClient, endpoint st
 		return err
 	}
 
-	if options.AllowReauth {
+	if options.CanReauth() {
 		// here we're creating a throw-away client (tac). it's a copy of the user's provider client, but
 		// with the token and reauth func zeroed out. combined with setting `AllowReauth` to `false`,
 		// this should retry authentication only once
@@ -156,10 +152,8 @@ func v2auth(ctx context.Context, client *gophercloud.ProviderClient, endpoint st
 		if err != nil {
 			return err
 		}
-		tao := options
-		tao.AllowReauth = false
 		client.ReauthFunc = func(ctx context.Context) error {
-			err := v2auth(ctx, &tac, endpoint, tao, eo)
+			err := v2auth(ctx, &tac, endpoint, &v2TokenNoReauth{options}, eo)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


cardoe:
I tried my best not to make an API change here. The unfortunate result is that `tokens2.AuthOptionsBuilder` does have a function added that mirrors what `tokens3.AuthOptionsBuilder` has. So anyone that implemented their own would need to add that method but it'd literally be a 1 liner.

I do think that my approach to handling the setting of `AllowReauth` when creating the `ReauthFunc` is something that the v3 side of the code could use and then out of tree auth plugins could be used for v3 as well. But I'm happy to do that in a separate PR.

cardoe:
Just wondering if you've had any thoughts on this approach @pierreprinetti ?

EmilienM:
This LGTM but I'll like to see Pierre's review on this one.

cardoe:
I noticed @mandre marked this as a major semver bump. If that's the case and I'll have to wait for 2.0 for this then would it make more sense to not add a method and instead change the existing one?

cardoe:
Hoping you've got some feedback here @pierreprinetti

openstack/identity/v2/tokens/requests.go
```diff
@@ -53,7 +54,7 @@ type AuthOptions struct {
 }
 
 // ToTokenV2CreateMap builds a token request body from the given AuthOptions.
-func (opts AuthOptions) ToTokenV2CreateMap() (map[string]interface{}, error) {
+func (opts *AuthOptions) ToTokenV2CreateMap() (map[string]interface{}, error) {
```
pierreprinetti, line 57: Why have you changed this method to have a pointer receiver?

openstack/identity/v2/tokens/requests.go
```diff
@@ -77,6 +78,10 @@ func (opts AuthOptions) ToTokenV2CreateMap() (map[string]interface{}, error) {
 	return b, nil
 }
 
+func (opts *AuthOptions) CanReauth() bool {
```
pierreprinetti, line 81: nit: this method may not need to have a pointer receiver

pierreprinetti (commented):
This change looks beautifully done. I have just a couple questions inline.

We will have to test the hell out of this PR when it's ready to merge.

openstack/identity/v2/tokens/requests.go
cardoe: I thought it was required to allow it to take a structure that implements the `AuthOptions` interface? Is that wrong?

openstack/client.go
```diff
@@ -113,10 +113,15 @@ func Authenticate(client *gophercloud.ProviderClient, options gophercloud.AuthOp
 
 // AuthenticateV2 explicitly authenticates against the identity v2 endpoint.
 func AuthenticateV2(client *gophercloud.ProviderClient, options gophercloud.AuthOptions, eo gophercloud.EndpointOpts) error {
+	return v2auth(client, "", &options, eo)
+}
+
+// AuthenticateV2Ext explicitly authenticates against the identity v2 endpoint.
+func AuthenticateV2Ext(client *gophercloud.ProviderClient, options tokens2.AuthOptionsBuilder, eo gophercloud.EndpointOpts) error {
```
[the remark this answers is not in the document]
cardoe: It stood for Extended. Just trying to come up with a different name to not break the API of the existing function.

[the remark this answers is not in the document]
cardoe: I dropped it since you guys flagged this as being a major version bump and incorporated it into the regular V2 call. Since you're reading gopher cloud 2.x I figured it would be okay.

openstack/identity/v2/tokens/requests.go
cardoe: You're correct.

openstack/identity/v2/tokens/requests.go
cardoe: Agreed.

cardoe:
I've opened #3030 as well which performs the same changes in a non `semver:major` way.

pierreprinetti:
Thank you for this, and sorry for the time it's taken to get a review.

I am not convinced that we need a type assertion, and I would only use that as a last resort. For example: how do you prevent an infinite reauth loop with custom types?

One alternative could be embedding and overloading the CanReauth method. Can you please see if this makes sense?

```diff
diff --git a/openstack/client.go b/openstack/client.go
index dfe5e53e..07cc6d40 100644
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -113,6 +113,12 @@ func AuthenticateV2(ctx context.Context, client *gophercloud.ProviderClient, opt
 	return v2auth(ctx, client, "", options, eo)
 }
 
+type reauthToken struct {
+	tokens2.AuthOptionsBuilder
+}
+
+func (rt *reauthToken) CanReauth() bool { return false }
+
 func v2auth(ctx context.Context, client *gophercloud.ProviderClient, endpoint string, options tokens2.AuthOptionsBuilder, eo gophercloud.EndpointOpts) error {
 	v2Client, err := NewIdentityV2(client, eo)
 	if err != nil {
@@ -143,21 +149,8 @@ func v2auth(ctx context.Context, client *gophercloud.ProviderClient, endpoint st
 		tac.SetThrowaway(true)
 		tac.ReauthFunc = nil
 		tac.SetTokenAndAuthResult(nil)
-		var tao tokens2.AuthOptionsBuilder
-		switch ot := options.(type) {
-		case *gophercloud.AuthOptions:
-			o := *ot
-			o.AllowReauth = false
-			tao = &o
-		case *tokens2.AuthOptions:
-			o := *ot
-			o.AllowReauth = false
-			tao = &o
-		default:
-			tao = options
-		}
 		client.ReauthFunc = func(ctx context.Context) error {
-			err := v2auth(ctx, &tac, endpoint, tao, eo)
+			err := v2auth(ctx, &tac, endpoint, &reauthToken{options}, eo)
 			if err != nil {
 				return err
 			}
diff --git a/openstack/client_test.go b/openstack/client_test.go
new file mode 100644
index 00000000..68a9fdb0
--- /dev/null
+++ b/openstack/client_test.go
@@ -0,0 +1,5 @@
+package openstack
+
+import tokens2 "github.com/gophercloud/gophercloud/v2/openstack/identity/v2/tokens"
+
+var _ tokens2.AuthOptionsBuilder = &reauthToken{}
```

pierreprinetti (approved):

pierreprinetti:
Thank you, now on to testing :grinning:

PR #2656, opened 2023-06-28
Instrument backporting to v1

pierreprinetti:
The stable branch is now `v1`. Non-breaking changes will be backported by applying a `backport-v1` label.

### DIFF
--- a/.github/backport-labels.yaml
+++ b/.github/backport-labels.yaml
@@ -1,0 +1,3 @@
+- name: backport-v1
+  description: This PR will be backported to v1
+  color: '30ABB9'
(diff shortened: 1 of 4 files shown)


EmilienM (approved):

EmilienM:
I don't know if/how you are able to test it but from a first look, it's good to me.
Let me know if you want to merge it as is and I'm happy to press the button.

pierreprinetti:
@EmilienM 

> I don't know if/how you are able to test it

I have tested it in github.com/pierreprinetti/gophtest and it seemed OK. I’d say that if you agree with the change in principle, we can merge and go ahead. I guess we can deal with unexpected issues as they come

PR #2673, opened 2023-07-05
CI/semver: only run on master

EmilienM:
We don't want these workflows to run on other branches than master.

### DIFF
--- a/.github/workflows/semver-require.yaml
+++ b/.github/workflows/semver-require.yaml
@@ -1,6 +1,8 @@
 name: Verify PR Labels
 on:
   pull_request:
+    branches:
+      - master
     types:
       - opened
       - labeled
--- a/.github/workflows/semver-unlabel.yaml
+++ b/.github/workflows/semver-unlabel.yaml
@@ -6,6 +6,8 @@ name: Reset PR labels on push
 
 on:
   pull_request_target:
+    branches:
+      - master
     types:
       - synchronize
 


pierreprinetti:
> We don't want these workflows to run on other branches than master.

Why not? I think they should run on all branches

pierreprinetti:
The corollary, of course, is that we should also set up a bot that applies the semver label.

EmilienM:
> > We don't want these workflows to run on other branches than master.
> 
> Why not? I think they should run on all branches

I'm not sure why we would run it on a stable branch.

EmilienM:
I don't have a strong opinion on this one, I'll let @mandre give his opinion as well and then we can make a decision.

mandre:
I think it's still useful to check for API changes on the v1 branch in case someone submits a PR directly to v1. It could be a direct fix for code that is no longer in master, or a backport that does not apply cleanly for example. In these cases, we would still want to ensure we're not breaking the API compatibility in the stable branch.

mandre:
I believe we should close this PR in favor of https://github.com/gophercloud/gophercloud/pull/2674.

mandre (changes requested):
I believe we should close this PR in favor of https://github.com/gophercloud/gophercloud/pull/2674.

pierreprinetti:
>I think it's still useful to check for API changes on the v1 branch in case someone submits a PR directly to v1

Also once the branches diverge, there is a chance that a PR that is "patch" on main becomes "minor" or "major" on v1

PR #2674, opened 2023-07-06
[CI] Drop branch filters for 'Ensure Labels' and 'CodeQL' jobs

mandre:
These jobs should run on all branches, so drop the branch filter that were in place.

For the CodeQL job, it seems like the github action syntax does not allow for a simplified `on` syntax, while at the same time using a `on.schedule`, so drop the periodic that we weren't paying attention to anyway. We have enough signal with just the PR jobs.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,13 +1,6 @@
 name: "CodeQL"
 
-on:
-  push:
-    branches: [ master ]
-  pull_request:
-    # The branches below must be a subset of the branches above
-    branches: [ master ]
-  schedule:
-    - cron: '18 8 * * 6'
+on: [push, pull_request]
 
 jobs:
   analyze:
--- a/.github/workflows/ensure-labels.yaml
+++ b/.github/workflows/ensure-labels.yaml
@@ -1,8 +1,6 @@
 name: Apply labels in .github/labels.yaml
 on:
   push:
-    branches:
-      - master
     paths:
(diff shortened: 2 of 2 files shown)


EmilienM (approved):

pierreprinetti:
“ensure labels” should only run on the main branch. The labels are repo-wide and shouldn’t be controlled by multiple branches. Also note that each time the job runs, it removes non-listed labels

mandre:
Also, note for later, the automated backport [failed](https://github.com/gophercloud/gophercloud/actions/runs/5470887140/jobs/9961456096) because of a lack of permission:
```
 Error:  Error: To https://github.com/gophercloud/gophercloud.git
!	refs/heads/bp-v1-66ac5e4c8ad6899f05b80485f10f17ec624b57e6:refs/heads/bp-v1-66ac5e4c8ad6899f05b80485f10f17ec624b57e6	[remote rejected] (refusing to allow a GitHub App to create or update workflow `.github/workflows/semver-labels.yaml` without `workflows` permission)
```

Looking at the doc, it seems we'll have to create a Personal Access Token to workaround the issue: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#granting-additional-permissions

pierreprinetti:
>Also, note for later, the automated backport [failed](https://github.com/gophercloud/gophercloud/actions/runs/5470887140/jobs/9961456096) because of a lack of permission

Do you think we need to enable the backport-bot to modify workflows? Maybe that's a security feature that we could use.a

PR #2708, opened 2023-08-01
backport-v1: use PAT token when creating the PR

EmilienM:
We need to use a Personal Access Token (PAT) when creating the backport pull request
because we expect GitHub Actions workflows to be triggered when the pull request is opened.


pierreprinetti (approved):
Thank you!

PR #2711, opened 2023-08-01
backport: use github app instead of PAT

EmilienM:
I created the secrets and the right permissions for this app.

### DIFF
--- a/.github/workflows/backport_v1.yaml
+++ b/.github/workflows/backport_v1.yaml
@@ -23,14 +23,18 @@ jobs:
       )
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


mandre (commented):
I looked at https://github.com/tibdex/github-app-token and nothing threw me off. It's a tiny app that uses the official SDK for github API.

The only thing I noticed what that is uses the deprecated https://github.com/actions/github while it should now use https://github.com/actions/github-script.

edit: it looks like it's [using version 5.1.1](https://github.com/tibdex/github-app-token/blob/v1.8.0/package.json#L18), so I guess it's using https://github.com/actions/github-script under the hood.

mandre (approved):

PR #2740, opened 2023-08-30
workflow/semver-auto: rebase PR

EmilienM:
When testing if a PR is API backward compatible or not, we need to test
against the base branch, not the head against the PR branched is based
on.

Otherwise it can lead to the case where a breaking change was merged
after that a PR was proposed, go-apidiff will fail because it'll test
the repo against base_ref.

The PR needs to be rebased on base_ref so the actual compatibility can be
verified.

If the PR can't be rebase for any reason, we'll return a failure and add
the label `semver:unknown`.

### DIFF
--- a/.github/workflows/semver-auto.yaml
+++ b/.github/workflows/semver-auto.yaml
@@ -12,6 +12,17 @@ jobs:
         ref: ${{ github.event.pull_request.head.sha }}
         token: ${{ secrets.GITHUB_TOKEN }}
 
+    - uses: peter-evans/rebase@56c359b35ff7ba8426d0fdb842958b35b1db827
+      with:
+        base: ${{ github.base_ref }}
+
+    - name: Add semver:unknown label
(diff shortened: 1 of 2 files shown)


EmilienM:
This was tested on my fork via https://github.com/EmilienM/gophercloud/pull/14 and https://github.com/EmilienM/gophercloud/actions/runs/6028891076/job/16357163968?pr=14#step:4:13

.github/workflows/semver-auto.yaml
```diff
@@ -12,6 +12,17 @@ jobs:
         ref: ${{ github.event.pull_request.head.sha }}
         token: ${{ secrets.GITHUB_TOKEN }}
 
+    - uses: peter-evans/rebase@v2
+      with:
+        base: master
```
EmilienM, line 17: let me see if I can substitute "master" by `${{ github.base_ref }}`.

PR #2759, opened 2023-09-05
ci/functional: switch master jobs to run on ubuntu 22.04

EmilienM:
Since https://github.com/openstack/devstack/commit/427a4e1a9b7f20a8be0ad5091f2229945ce711a8
We don't support focal anymore in devstack-master. We need to run 22.04
from now.

Note: some workflows already had it right but it wasn't consistent
across all services.


EmilienM:
Is replaced by https://github.com/gophercloud/gophercloud/pull/2762

PR #2761, opened 2023-09-06
CI: automatically reject backward incompatible backports

### LINKED ISSUE #2758: Automatically reject backports if semver is not `minor` or `patch`.
Now that we have the automation to:
* backport PRs
* set the semvers (`major`, `minor`, `patch`)

We want to reject any backport that would not be backward compatible to prevent breakages in the stable branch.

EmilienM:
If a backport is labeled as `semver-unknown` or `semver-major`, we
consider it unsafe to backport until someone has verified otherwise and
applied `semver-patch` or `semver-minor`.

We create a commit with an error to make it visible if that happens.

Fixes #2758

### DIFF
--- a/.github/workflows/backport_v1.yaml
+++ b/.github/workflows/backport_v1.yaml
@@ -31,6 +31,9 @@ jobs:
           private_key: ${{ secrets.BACKPORT_APP_PRIVATE_KEY }}
 
       - name: Backporting
+        if: >
+          contains(github.event.pull_request.labels.*.name, 'semver-patch')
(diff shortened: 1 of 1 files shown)


pierreprinetti (commented):
Why not leaving the root-level `if` as-is, and only add the semver check to the individual steps?

EmilienM:
> Why not leaving the root-level `if` as-is, and only add the semver check to the individual steps?

sure, I can do that.

EmilienM:
done @pierreprinetti

pierreprinetti (approved):

PR #2764, opened 2023-09-07
Acceptance: Handle numerical version names in version comparison helpers

mandre:
There are two changes here:
- rename the `IsReleasesBelow()` and `IsReleasesAbove()` functions to `IsCurrentBelow()` and `IsCurrentAbove()` respectively.
- fix their behavior to account for numerical release names, after zed, and ensure that
comparing the same version always yields false. Also fixing bugs when comparing to master and add comprehensive tests.

### DIFF
--- a/acceptance/clients/conditions.go
+++ b/acceptance/clients/conditions.go
@@ -111,7 +111,7 @@ func SkipRelease(t *testing.T, release string) {
 func SkipReleasesBelow(t *testing.T, release string) {
 	current_branch := getReleaseFromEnv(t)
 
-	if IsReleasesBelow(t, release) {
+	if IsCurrentBelow(t, release) {
 		t.Skipf("this is not supported below %s, testing in %s", release, current_branch)
 	}
 }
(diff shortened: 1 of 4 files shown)


EmilienM:
I changed the label manually before our workflow doesn't know (yet) that it's acceptance and that API is not maintained here.
I also fixed the failure via https://github.com/gophercloud/gophercloud/pull/2765 .

acceptance/clients/conditions_test.go
```diff
@@ -0,0 +1,66 @@
+//go:build acceptance
+// +build acceptance
```
EmilienM, line 2: Please put that file into a `testing directory` or this will be skipped by the unittest (in fact coverage) job. See https://github.com/gophercloud/gophercloud/blob/4f6d1451acac9260a6901451e146b2bee4742bb5/script/coverage#L6

mandre:
The [go-apidiff failure](https://github.com/gophercloud/gophercloud/actions/runs/6118191513/job/16606621557?pr=2764) should be fixed with https://github.com/joelanford/go-apidiff/commit/05117854c081149093a1282693d26426dc5b3d0e. Waiting for a new release.

EmilienM:
> The [go-apidiff failure](https://github.com/gophercloud/gophercloud/actions/runs/6118191513/job/16606621557?pr=2764) should be fixed with [joelanford/go-apidiff@0511785](https://github.com/joelanford/go-apidiff/commit/05117854c081149093a1282693d26426dc5b3d0e). Waiting for a new release.

no, you'll need to rebase your PR on top of https://github.com/gophercloud/gophercloud/pull/2773 once merged, and the job will be happy. However the job will set the label to `semver:major` because again we don't make a difference of API chances in the directories, for now.

EmilienM (approved):

PR #2765, opened 2023-09-07
ci/semver-auto: return a failure only when needed

EmilienM:
When go-apidiff returns semver=major, it returns RC=1 which causes the
workflow to return an error while we don't want that, we just want the
label to be applied.

Instead, let's ignore that error so the label can be set later and add a
step to check if an error occured and semver is not major, then it's an
actual problem that we should check in the logs (e.g. an issue when
running go-apidiff).

### DIFF
--- a/.github/workflows/semver-auto.yaml
+++ b/.github/workflows/semver-auto.yaml
@@ -43,24 +43,33 @@ jobs:
 
     - name: Checking Go API Compatibility
       id: go-apidiff
+      # if semver=major, this will return RC=1, so let's ignore the failure so label
+      # can be set later. We check for actual errors in the next step.
+      continue-on-error: true
       uses: joelanford/go-apidiff@v0.7.0
 
+    # go-apidiff returns RC=1 when semver=major, which makes the workflow to return
+    # a failure. Instead let's just return a failure if go-apidiff failed to run.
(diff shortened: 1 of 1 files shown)


EmilienM:
Note: this won't be backported because in v1 we don't want semver:major anyway.

.github/workflows/semver-auto.yaml
```diff
@@ -43,24 +43,33 @@ jobs:
...
-      if: always() && steps.go-apidiff.outputs.semver-type == 'minor'
+      if: steps.go-apidiff.outputs.semver-type == 'minor'
       uses: actions-ecosystem/action-add-labels@18f1af5e3544586314bbe15c0273249c770b2daf
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
         labels: semver:minor
 
     - name: Add semver:major label
-      if: always() && steps.go-apidiff.outputs.semver-type == 'major'
+      if: steps.go-apidiff.outputs.semver-type == 'major'
```
pierreprinetti, line 72: Will this step ever run? If I understand correctly, when go-apidiff returns `major` it also exits with 1

EmilienM: Yes it'll run thanks to:

```golang
 continue-on-error: true
```

on L#48

PR #2779, opened 2023-09-20
acceptance/volumeattachments: use nova API for most of it

### LINKED ISSUE #2768: acceptance: we should not use volume attachment API
Volume attachment API shouldn't be used to attach a volume into a server in our tests, unless we want to test the volume attach API specifically.
The reason is that we should be using the compute API to attach volumes and the volume attachment API was designed for debug purposes:

```
This command will only create a volume attachment in the Volume service. It will not invoke the
necessary Compute service actions to actually attach the volume to the server at the hypervisor level. As a result, it should typically
only be used for troubleshooting issues with an existing server in combination with other tooling. For all other use cases, the 'server
volume add' command should be preferred.
```

EmilienM:
It is not recommended [1] to use volume attachment API from Cinder
directly when we want to test real user workflows, like creating a
server, a volume and attach the volume to the server.

Indeed, it's advised to use the Nova API [2] instead.

This patch doesn't bring any feature, however it brings best practice in
our tests, instead of using "low level" volume APIs, which aren't
advised to our users.

[1] https://docs.openstack.org/python-openstackclient/2023.1/cli/command-objects/volume-attachment.html#volume-attachment-create
[2] https://docs.openstack.org/api-ref/compute/#servers-with-volume-attachments-servers-os-volume-attachments

Fixes #2768

### DIFF
--- a/acceptance/openstack/blockstorage/v3/volumeattachments.go
+++ b/acceptance/openstack/blockstorage/v3/volumeattachments.go
@@ -7,6 +7,7 @@ import (
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/openstack/blockstorage/v3/attachments"
 	v3 "github.com/gophercloud/gophercloud/openstack/blockstorage/v3/volumes"
+	"github.com/gophercloud/gophercloud/openstack/compute/v2/extensions/volumeattach"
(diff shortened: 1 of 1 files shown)


EmilienM:
I'll abandon this effort, the right APIs are used in `acceptance/openstack/compute/v2/compute.go` and I think it's ok that the blockstorage tests use the cinder API instead.

PR #2781, opened 2023-09-21
baremetal: support ironic native PluginData

### LINKED ISSUE #2612: RFE: new baremetal inventory API

dtantsur:
Add logic to detect which PluginData is provided for the node.

Fixes #2612

### DIFF
--- a/openstack/baremetal/inventory/plugindata.go
+++ b/openstack/baremetal/inventory/plugindata.go
@@ -71,3 +71,44 @@ func (r *LLDPTLVType) UnmarshalJSON(data []byte) error {
 	r.Value = value
 	return nil
 }
+
+type HardwareManager struct {
+	Name    string `json:"name"`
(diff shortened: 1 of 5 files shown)


EmilienM (approved):
please add acceptance at some point, but LGTM.

PR #2782, opened 2023-09-21
[v1] (manual clean backport) Add tag field to compute block_device_v2

EmilienM:
- Add tag field to compute block_device_v2
- Add requests test for new field
- acceptance/compute: remove flavor ID check
- Add acceptance tests for tag field to compute block_device_v2

This is a clean backport with no conflict.
I had to do it manually because the workflow has been fixed after these commits so it didn't run.


EmilienM:
For CI to work we need https://github.com/gophercloud/gophercloud/pull/2783

mandre (approved):

PR #2786, opened 2023-09-22
Make acceptance tests internal

mandre:
Do not expose acceptance tests packages externally. This has the added benefit of go-apidiff no longer reporting false positive when analyzing for backward incompatible changes.


EmilienM (approved):

kayrus (approved):
Nice decision, thanks!

SuperSandro2000:
We are extending the acceptance tests for our sap specific additions from some of those packages. Could those be moved back?

Specifically ``github.com/gophercloud/gophercloud/acceptance/clients`` https://github.com/sapcc/gophercloud-sapcc/blob/master/acceptance/automation/v1/client.go#L22 and ``github.com/gophercloud/gophercloud/acceptance/tools`` https://github.com/sapcc/gophercloud-sapcc/blob/master/acceptance/automation/v1/automations.go#L21

pierreprinetti:
> We are extending the acceptance tests for our sap specific additions from some of those packages. Could those be moved back?
> 
> Specifically `github.com/gophercloud/gophercloud/acceptance/clients` https://github.com/sapcc/gophercloud-sapcc/blob/master/acceptance/automation/v1/client.go#L22 and `github.com/gophercloud/gophercloud/acceptance/tools` https://github.com/sapcc/gophercloud-sapcc/blob/master/acceptance/automation/v1/automations.go#L21

If I understand your case correctly, you maintain a domain-specific extension of Gophercloud; to test it, you extend Gophercloud's acceptance tests in parallel. Is that what's happening here?

What we want is to remove the acceptance tests from the API surface of Gophercloud. Moving the packages to an `/internal` directory is just the way we've gone to achieve that.

One alternative to using `/internal` packages would be to move all acceptance-related code to either `_test.go` files or to testing packages. @SuperSandro2000 would this solution fit your case?

SuperSandro2000:
> If I understand your case correctly, you maintain a domain-specific extension of Gophercloud; to test it, you extend Gophercloud's acceptance tests in parallel. Is that what's happening here?

Yes

> would this solution fit your case?

Yeah, that should work, too.

PR #2791, opened 2023-10-04
Add microversion utilities

lentzi90:
This adds GetSupportedMicroversions and MicroversionSupported utility functions for working with microversions.
The doc string for ChooseVersion is also updated to clarify how it works.

Fixes #2791

code in this PR:

TODO. I am not sure what is needed here

### DIFF
--- a/openstack/utils/choose_version.go
+++ b/openstack/utils/choose_version.go
@@ -2,6 +2,7 @@ package utils
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/gophercloud/gophercloud"
@@ -20,9 +21,12 @@ var goodStatus = map[string]bool{
 	"stable":    true,
 }
 
-// ChooseVersion queries the base endpoint of an API to choose the most recent non-experimental alternative from a service's
-// published versions.
-// It returns the highest-Priority Version among the alternatives that are provided, as well as its corresponding endpoint.
+// ChooseVersion queries the base endpoint of an API to choose the identity service version.
+// It will pick a version among the recognized, taking into account the priority and avoiding
+// experimental alternatives from the published versions. However, if the client specifies a full
+// endpoint that is among the recognized versions, it will be used regardless of priority.
+// It returns the highest-Priority Version, OR exact match with client endpoint,
+// among the alternatives that are provided, as well as its corresponding endpoint.
 func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*Version, string, error) {
 	type linkResp struct {
 		Href string `json:"href"`
@@ -108,4 +112,124 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
 	}
 
 	return highest, endpoint, nil
+}
+
+type SupportedMicroversions struct {
+	MaxMajor int
+	MaxMinor int
+	MinMajor int
+	MinMinor int
+}
+
+// GetSupportedMicroversions returns the minimum and maximum microversion that is supported by the ServiceClient Endpoint.
+func GetSupportedMicroversions(client *gophercloud.ServiceClient) (SupportedMicroversions, error) {
+	type valueResp struct {
+		ID         string `json:"id"`
+		Status     string `json:"status"`
+		Version    string `json:"version"`
+		MinVersion string `json:"min_version"`
+	}
+
+	type response struct {
+		Version  valueResp   `json:"version"`
+		Versions []valueResp `json:"versions"`
+	}
+	var minVersion, maxVersion string
+	var supportedMicroversions SupportedMicroversions
+	var resp response
+	_, err := client.Get(client.Endpoint, &resp, &gophercloud.RequestOpts{
+		OkCodes: []int{200, 300},
+	})
+
+	if err != nil {
+		return supportedMicroversions, err
+	}
+
+	if len(resp.Versions) > 0 {
+		// We are dealing with an unversioned endpoint
+		// We only handle the case when there is exactly one, and assume it is the correct one
+		if len(resp.Versions) > 1 {
+			return supportedMicroversions, fmt.Errorf("unversioned endpoint with multiple alternatives not supported")
+		}
+		minVersion = resp.Versions[0].MinVersion
+		maxVersion = resp.Versions[0].Version
+	} else {
+		minVersion = resp.Version.MinVersion
+		maxVersion = resp.Version.Version
+	}
+
+	// Return early if the endpoint does not support microversions
+	if minVersion == "" && maxVersion == "" {
+		return supportedMicroversions, fmt.Errorf("microversions not supported by ServiceClient Endpoint")
+	}
+
+	supportedMicroversions.MinMajor, supportedMicroversions.MinMinor, err = ParseMicroversion(minVersion)
+	if err != nil {
+		return supportedMicroversions, err
+	}
+
+	supportedMicroversions.MaxMajor, supportedMicroversions.MaxMinor, err = ParseMicroversion(maxVersion)
+	if err != nil {
+		return supportedMicroversions, err
+	}
+
+	return supportedMicroversions, nil
+}
+
+// RequireMicroversion checks that the required microversion is supported and
+// returns a ServiceClient with the microversion set.
+func RequireMicroversion(client gophercloud.ServiceClient, required string) (gophercloud.ServiceClient, error) {
+	supportedMicroversions, err := GetSupportedMicroversions(&client)
+	if err != nil {
+		return client, fmt.Errorf("unable to determine supported microversions: %w", err)
+	}
+	supported, err := supportedMicroversions.IsSupported(required)
+	if err != nil {
+		return client, err
+	}
+	if !supported {
+		return client, fmt.Errorf("microversion %s not supported. Supported versions: %v", required, supportedMicroversions)
+	}
+	client.Microversion = required
+	return client, nil
+}
+
+// IsSupported checks if a microversion falls in the supported interval.
+// It returns true if the version is within the interval and false otherwise.
+func (supported SupportedMicroversions) IsSupported(version string) (bool, error) {
+	// Parse the version X.Y into X and Y integers that are easier to compare.
+	vMajor, vMinor, err := ParseMicroversion(version)
+	if err != nil {
+		return false, err
+	}
+
+	// Check that the major version number is supported.
+	if (vMajor < supported.MinMajor) || (vMajor > supported.MaxMajor) {
+		return false, nil
+	}
+
+	// Check that the minor version number is supported
+	if (vMinor <= supported.MaxMinor) && (vMinor >= supported.MinMinor) {
+		return true, nil
+	}
+
+	return false, nil
+}
+
+// ParseMicroversion parses the version major.minor into separate integers major and minor.
+// For example, "2.53" becomes 2 and 53.
+func ParseMicroversion(version string) (major int, minor int, err error) {
+	parts := strings.Split(version, ".")
+	if len(parts) != 2 {
+		return 0, 0, fmt.Errorf("invalid microversion format: %q", version)
+	}
+	major, err = strconv.Atoi(parts[0])
+	if err != nil {
+		return 0, 0, err
+	}
+	minor, err = strconv.Atoi(parts[1])
+	if err != nil {
+		return 0, 0, err
+	}
+	return major, minor, nil
 }
--- a/openstack/utils/testing/choose_version_test.go
+++ b/openstack/utils/testing/choose_version_test.go
@@ -3,6 +3,7 @@ package testing
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"testing"
 
 	"github.com/gophercloud/gophercloud"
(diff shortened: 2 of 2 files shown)


openstack/utils/choose_version.go
```diff
@@ -109,3 +113,94 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
 
 	return highest, endpoint, nil
 }
+
+// GetSupportedMicroversions returns the minimum and maximum microversion that is supported by the ServiceClient Endpoint.
+func GetSupportedMicroversions(client gophercloud.ServiceClient) (string, string, error) {
```
dtantsur, line 118: (Thinking aloud) maybe we could return a wrapper structure with something like
```go
type SupportedMicroversions struct {
    MaxMajor int
    MaxMinor int
    MinMajor int
    MinMinor int
}
```

Then some of these functions become methods on it, like
```go
func (vs SupportedMicroversions) Require(version string) error
func (vs SupportedMicroversions) IsSupported(version string) bool
func (vs SupportedMicroversions) Choose(versions []string) string
```

etc

dtantsur: Also: you want to pass the client as a pointer, not by value.

openstack/utils/choose_version.go
```diff
@@ -109,3 +113,94 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
...
+		Status     string `json:"status"`
+		Version    string `json:"version"`
+		MinVersion string `json:"min_version"`
+	}
+
+	type response struct {
+		Version valueResp `json:"version"`
+	}
+	var resp response
+	_, err := client.Request("GET", client.Endpoint, &gophercloud.RequestOpts{
```
dtantsur, line 130: When I try this, I get a segfault
```
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/gophercloud/gophercloud.(*ServiceClient).Request(0xc00010fa90, {0x6efecb, 0x3}, {0xc00001e100, 0x1a}, 0xc?)
        /home/dtantsur/gophercloud/service_client.go:150 +0x147
github.com/gophercloud/gophercloud/openstack/utils.GetSupportedMicroversions({0xc00013c000, {0xc00001e100, 0x1a}, {0x0, 0x0}, {0x6f117e, 0x9}, {0x0, 0x0}, 0xc00002ac90})
        /home/dtantsur/gophercloud/openstack/utils/choose_version.go:130 +0xd1
main.main()
        /home/dtantsur/gophercloud/main/main.go:26 +0x98
```

openstack/utils/choose_version.go
```diff
@@ -20,9 +21,12 @@ var goodStatus = map[string]bool{
 	"stable":    true,
 }
 
-// ChooseVersion queries the base endpoint of an API to choose the most recent non-experimental alternative from a service's
-// published versions.
-// It returns the highest-Priority Version among the alternatives that are provided, as well as its corresponding endpoint.
+// ChooseVersion queries the base endpoint of an API to choose the identity service version.
```
dulek, line 24: You're sure this is limited to "identity service", i.e. Keystone?

openstack/utils/choose_version.go
```diff
@@ -109,3 +113,94 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
...
+		return false, err
+	}
+	maxMajor, max, err := parseMicroversion(maxVersion)
+	if err != nil {
+		return false, err
+	}
+
+	// Check that the major version number is supported.
+	if (vMajor < minMajor) || (vMajor > maxMajor) {
+		return false, err
```
dulek, line 179: This should be `return false, nil`.

dulek (changes requested):

openstack/utils/choose_version.go
dtantsur: It has a reference to `IdentityBase` down the code, dunno why.

lentzi90: Yes, I got fooled by the doc string into believing this could be used for Nova. I was very confused about why it never could find a good version to use until @mdbooth pointed out that it is taking the ProviderClient as input, not the ServiceClient. It is indeed only checking the identity endpoint and picks a version for Keystone.

openstack/utils/choose_version.go
lentzi90: Would the idea be to hold on to this wrapper and pass it around? I'm worried it would become hard to use. If I need to check/require microversions in many different places in a code base it would probably be cumbersome. Alternatively, I would end up calling `GetSupportedMicroversions` all over the place.

dtantsur: I only thing it will be more convenient to use methods on the wrapper rather than passing two strings around and parsing them each time. There is also a risk of people using the wrong order (what goes first, min or max)?

lentzi90: Ah good point! I was still in my head thinking about how I did it in CAPO, where the min and max is stored in the ComputeClient and then these functions become methods on that. Maybe I can find some way to integrate this better with the ServiceClient :thinking:

openstack/utils/choose_version.go
dtantsur: Given that gophercloud is working on a major release, maybe it's time to change that?

lentzi90: Yeah that would be nice!

openstack/utils/choose_version.go
dtantsur: https://github.com/gophercloud/gophercloud/pull/2799

dtantsur: You can avoid it by using Get instead of Request for now.

openstack/utils/choose_version.go
mdbooth: It would also be really good to have this in the v1 branch though, if possible. We have a bunch of potential near term use cases for it, one of which is described in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1715#issuecomment-1752923531.

openstack/utils/choose_version.go
```diff
@@ -109,3 +113,94 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
...
+	if err != nil {
+		return "", "", err
+	}
+
+	return resp.Version.MinVersion, resp.Version.Version, nil
+}
+
+// RequireMicroversion checks that the required microversion is supported and sets it for
+// the ServiceClient.
+func RequireMicroversion(client *gophercloud.ServiceClient, required string) error {
```
mdbooth, line 144: Would it be possible to change this API to return a copy of the ServiceClient rather than modifying the existing one? The reason for this is because I would like to ensure we always use the same 'base' microversion, and bump it explicitly only for the scope of code sections which require it. This is much simpler to do if you get a copy of the ServiceClient, because you don't have to worry about undoing it under all circumstances when you exist the relevant scope. Something like:

```go
hostnameServiceClient, err := WithMicroversion(serviceClient, "2.90")
if err != nil {
  if IsMicroversionNotSupported(err) {
    ...plan B...
  } else {
    return err
  }
}

// hostnameServiceClient has microversion 2.90
// serviceClient still has its original microversion, so no accidental microversion leaks
```

...or whatever. I'm not so bothered about the actual API so much as the principal of leaving the original unmodified so we don't have to worry about leaks.

openstack/utils/choose_version.go
lentzi90: Fixed

openstack/utils/choose_version.go
lentzi90: @mdbooth I thought you did not want this kind of negotiation of the version? I have reworked the CAPO PR for that reason to set an explicit version in all cases instead of using a function like this. In which cases do you think we should use this?

openstack/utils/choose_version.go
lentzi90: Done :slightly_smiling_face:

openstack/utils/choose_version.go
```diff
@@ -109,3 +113,103 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
...
+	if !supported {
+		return client, fmt.Errorf("microversion %s not supported. Supported versions: %v", required, supportedMicroversions)
+	}
+	client.Microversion = required
+	return client, nil
+}
+
+// MicroversionSupported checks if a microversion falls in the supported interval.
+// It returns true if the version is within the interval and false otherwise.
+func MicroversionSupported(version string, supportedMicroversions SupportedMicroversions) (bool, error) {
```
dtantsur, line 179: Why not make this a method on SupportedMicroversions?

openstack/utils/choose_version.go
```diff
@@ -109,3 +113,103 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
...
+func GetSupportedMicroversions(client *gophercloud.ServiceClient) (SupportedMicroversions, error) {
+	type valueResp struct {
+		ID         string `json:"id"`
+		Status     string `json:"status"`
+		Version    string `json:"version"`
+		MinVersion string `json:"min_version"`
+	}
+
+	type response struct {
+		Version valueResp `json:"version"`
```
dtantsur, line 134: This assumes a version endpoint, like `http://example.com/compute/v2`. It fails for me locally with Ironic since I have an unversioned one, which looks like this:

```json
{
  "name": "OpenStack Ironic API",
  "description": "Ironic is an OpenStack project which enables the provision and management of baremetal machines.",
  "default_version": {
    "id": "v1",
    "links": [
      {
        "href": "http://192.168.122.1:6385/v1/",
        "rel": "self"
      }
    ],
    "status": "CURRENT",
    "min_version": "1.1",
    "version": "1.87"
  },
  "versions": [
    {
      "id": "v1",
      "links": [
        {
          "href": "http://192.168.122.1:6385/v1/",
          "rel": "self"
        }
      ],
      "status": "CURRENT",
      "min_version": "1.1",
      "version": "1.87"
    }
  ]
}
```

dtantsur: Maybe we can check the `Versions` list, and if it contains only one item, assume it's the correct one? Otherwise, fail with a clear error message?

openstack/utils/testing/choose_version_test.go
```diff
@@ -117,3 +179,146 @@ func TestChooseVersionFromSuffix(t *testing.T) {
...
+	defer testhelper.TeardownHTTP()
+	setupVersionHandler()
+
+	tests := []getSupportedServiceMicroversions{
+		// v2 does not support microversions and returns error
+		{
+			Endpoint:    testhelper.Endpoint() + "compute/v2/",
+			ExpectedMax: "",
+			ExpectedMin: "",
+			ExpectedErr: true,
```
dtantsur, line 201: Could you test for a reasonable error message? With Ironic I get `invalid microversion format: ""` which is quite confusing.

openstack/utils/choose_version.go
mdbooth: I probably got confused and thought I was commenting on the PR in general :)

You are correct, I don't want to do negotiation in CAPO: we either get a specific version or we don't execute the code. There may be cases where some code might run with one of a small number of specific microversions (e.g. GET server) but we should never, e.g. request microversion N and get microversion N+1.

That said, it might be useful to be able to specify a microversion limit for testing so we can test old microversions against a newer server. The microversion requested would still be at the point of use, though, and we would still either get it or not get it.

However, CAPO hopefully isn't the only consumer of this code. Negotiation might be useful elsewhere, e.g. a CLI utility.

openstack/utils/choose_version.go
lentzi90: Done (and renamed to `IsSupported`)

openstack/utils/choose_version.go
lentzi90: I have updated with something I believe does what you suggested. I even added a test with a made up Ironic endpoint that returns multiple versions to check the error case. Not sure if it makes sense :sweat_smile:

dtantsur (approved):
Works with Ironic in my testing, thank you!

lentzi90:
Thank you for testing!
I have also confirmed that the code works with CAPO in https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1567 (although I had to backport it to the v1 branch first since that is what CAPO is using)

EmilienM (approved):
Nice work, I'll wait for others to approve or request changes.

pierreprinetti:
I want to look at this this afternoon; please go ahead and merge if I don’t

pierreprinetti:
> I want to look at this this afternoon; please go ahead and merge if I don’t

I still believe I can get to that today :crossed_fingers:

openstack/utils/choose_version.go
```diff
@@ -109,3 +113,123 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
...
+		ID         string `json:"id"`
+		Status     string `json:"status"`
+		Version    string `json:"version"`
+		MinVersion string `json:"min_version"`
+	}
+
+	type response struct {
+		Version  valueResp   `json:"version"`
+		Versions []valueResp `json:"versions"`
+	}
```
pierreprinetti, line 136: I wonder if we can assume that all services will either use `version`, `versions`, or no versions. What if we have service-specific logic instead? Have you considered having service-specific logic, maybe exposed in the ServiceClient through an interface?

openstack/utils/choose_version.go
```diff
@@ -109,3 +113,123 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
...
+	if err != nil {
+		return supportedMicroversions, err
+	}
+
+	return supportedMicroversions, nil
+}
+
+// RequireMicroversion checks that the required microversion is supported and
+// returns a ServiceClient with the microversion set.
+func RequireMicroversion(client gophercloud.ServiceClient, required string) (gophercloud.ServiceClient, error) {
```
pierreprinetti, line 181: I like this function signature. You give me a version, and I tell you if it's OK. This requires many less assumptions. Can we make it so that this is the only new exported symbol, and unexport all the rest?

pierreprinetti (changes requested):

pierreprinetti:
Thank you for this addition. I think that the functionality is useful.

openstack/utils/choose_version.go
```diff
@@ -109,3 +113,123 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
 
 	return highest, endpoint, nil
 }
+
+type SupportedMicroversions struct {
+	MaxMajor int
+	MaxMinor int
+	MinMajor int
+	MinMinor int
+}
```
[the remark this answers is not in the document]
dtantsur: "the fact that they're split into exactly major and minor" is the hard fact that is embedded into the definition of microversions. I'm fine with hiding private details as long as I can still access them via accessors (not just the string representations - actual numbers).

openstack/utils/choose_version.go
```diff
@@ -109,3 +113,123 @@ func ChooseVersion(client *gophercloud.ProviderClient, recognized []*Version) (*
...
+
+// GetSupportedMicroversions returns the minimum and maximum microversion that is supported by the ServiceClient Endpoint.
+func GetSupportedMicroversions(client *gophercloud.ServiceClient) (SupportedMicroversions, error) {
```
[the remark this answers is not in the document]
dtantsur: Your example is too hypothetical :) First, it assumes having two major versions on one endpoint, which is not how things tend to be done. Second, it assumes two major versions *with microversions* which is a thing that is quite unlikely to happen because the very idea of microversions implies that you can do breaking changes without creating major versions. We do ponder Ironic v2 from time to time, but that's more of a wishful thinking (file it under "rewrite Ironic in Go").

openstack/utils/choose_version.go
dtantsur: Please don't. GopherCloud is already bare bones enough to justify another high-level client built upon it.

openstack/utils/choose_version.go
dtantsur: In theory - no. In practice, everyone just copied what Nova did :)

dtantsur: As a data point, I've implemented a similar thing in rust-osauth long ago, and nobody complained so far: https://github.com/dtantsur/rust-osauth/blob/master/src/protocol.rs#L31-L39

openstack/utils/choose_version.go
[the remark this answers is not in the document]
dtantsur: Reference: https://specs.openstack.org/openstack/api-sig/guidelines/microversion_specification.html

openstack/utils/choose_version.go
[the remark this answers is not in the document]
pierreprinetti: The way I see it, the assumption is to think that my example won’t happen. Which may or may not be a safe assumption.

openstack/utils/choose_version.go
pierreprinetti: How is Swift versioning its API?

openstack/utils/choose_version.go
pierreprinetti: I may be failing to see your intended use case. Do you have one in mind?

The use case I’m most inclined to think: a client has two max three behaviours coded for a specific action, each corresponding to a specific API. The client would ask for its preferred version, then if not available it would test the fallback behaviour, and so on. In this case the client knows what exact microversion it needs and asks Goohercloud whether Itshak OK. Do you have a case where a client needs something different, like a range?

dtantsur: Largely, I want to know *in advance* what features I can and cannot use, so that I don't need to delay this decision until much later in the code. Our code is organized in the way that the controller layer does not know about the Ironic implementation details, and we'd rather keep it that way.

openstack/utils/choose_version.go
dtantsur: I don't think it actually does. Of the core services, I think only Nova, Keystone and Cinder (and Ironic if you count it core) use microversioning.

openstack/utils/choose_version.go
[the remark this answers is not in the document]
dtantsur: It's a question of whether you want to try covering 100% of potential future cases. It's not a terrible idea if you do, but that also mean you will do much more work than you need to. With rust-openstack, I took the approach of "I'll think about it when it happens" (fwiw, openstacksdk does the same), which has its own pros and cons.

pierreprinetti:
I will not have the opportunity to come back to this PR in the next week; @EmilienM please go ahead and do as you see fit

openstack/utils/choose_version.go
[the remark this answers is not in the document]
lentzi90: From what I can see, this (min and max) is the only information we can get from a single endpoint so there is nothing more we can do.

lentzi90:
> A more general comment: the most Gophercloud-y solution for this problem would probably be to have a `versions` package in each relevant service, with a `List` function that makes the call to the service and returns a typed response. The caller would then adjust their ServiceClient microversion accordingly, with or without a helper function in Gophercloud.
> 
> I wonder if this "generic" solution that aims at encapsulating the calls and solving negotiation in one swipe, isn't going to feel like a trap going forward.

Thank you for the comment! I appreciate the insights and would ask that the maintainers decide on what would be the preferred way forward here. I can see pros and cons witch both approaches.

I want to also highlight this function in the utils repo: https://github.com/gophercloud/utils/blob/80377eca5d56d1f07d0ebdd4c2a5ae000a66a90e/openstack/clientconfig/requests.go#L739
Would it make more sense to move the microversion utilities there? It also handles all services in one function so perhaps it would fit in better there.

dtantsur:
> the most Gophercloud-y solution for this problem would probably be to have a versions package in each relevant service, with a List function that makes the call to the service and returns a typed response

Could be, but currently Gophercloud is handling Microversions as a standard feature (which they are in a sense), e.g. they handling is hardcoded in the ServiceClient: https://github.com/gophercloud/gophercloud/blob/69b57d9428c814942f444f049b6d6a1aebd82600/service_client.go#L116. If we start treating Microversions as service features, we need to change that as well. Which is probably a lot of work and will make the change no longer backportable..

dtantsur:
@lentzi90 @EmilienM @pierreprinetti so folks, what's the path forward here? I can leave with a reduced scope if that allows this change to move forward.

EmilienM:
I feel like this is a great iteration moving forward.

From Pierre:
> I wonder if this "generic" solution that aims at encapsulating the calls and solving negotiation in one swipe, isn't going to feel like a trap going forward.

I agree but I also think this would be doable in a future version. I want to unblock these folks for now.

PR #2799, opened 2023-10-06
Fix options initialization in ServiceClient.Request (fixes #2798)

### LINKED ISSUE #2798: Assignment to entry in nil map in ServiceClient.Request
The code writes to `MoreHeaders` without initializing it. The actual initialization happens in `Get`/`Post`/etc, so if you use `Request` directly, you can get:
```
panic: assignment to entry in nil map

goroutine 1 [running]:
github.com/gophercloud/gophercloud.(*ServiceClient).Request(0xc00010fa90, {0x6efecb, 0x3}, {0xc00001e100, 0x1a}, 0xc?)
        /home/dtantsur/gophercloud/service_client.go:150 +0x147
github.com/gophercloud/gophercloud/openstack/utils.GetSupportedMicroversions({0xc00013c000, {0xc00001e100, 0x1a}, {0x0, 0x0}, {0x6f117e, 0x9}, {0x0, 0x0}, 0xc00002ac90})
        /home/dtantsur/gophercloud/openstack/utils/choose_version.go:130 +0xd1
main.main()
        /home/dtantsur/gophercloud/main/main.go:26 +0x98
```

dtantsur:
Request is a part of the public API, but it relies on being called by
Get/Post/etc to properly initialize its options. Namely:
1) it may crash on a nil map assignment if there are MoreHeaders,
2) it does not handle microversions.

This change moves the relevant code to Request.

### DIFF
--- a/service_client.go
+++ b/service_client.go
@@ -47,7 +47,7 @@ func (client *ServiceClient) ServiceURL(parts ...string) string {
 	return client.ResourceBaseURL() + strings.Join(parts, "/")
 }
 
-func (client *ServiceClient) initReqOpts(url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) {
+func (client *ServiceClient) initReqOpts(JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) {
 	if v, ok := (JSONBody).(io.Reader); ok {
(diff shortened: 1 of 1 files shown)


pierreprinetti (approved):
Thanks! Good catch.

PR #2805, opened 2023-10-11
Fix baremetal jobs on Ubuntu 20.04

iurygregory:
- Workaround to fix the grub-efi-amd64-signed happening in ubuntu 20.04

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -40,6 +40,10 @@ jobs:
     steps:
       - name: Checkout Gophercloud
         uses: actions/checkout@v4
+      - name: Workaround for grub-efi-amd64-signed
+        run: sudo apt update && sudo apt -y upgrade
+        shell: bash
+        if: ${{ matrix.ubuntu_version == "20.04" }}
       - name: Deploy devstack
         uses: EmilienM/devstack-action@v0.11
         with:


.github/workflows/functional-baremetal.yaml
```diff
@@ -40,6 +40,11 @@ jobs:
     steps:
       - name: Checkout Gophercloud
         uses: actions/checkout@v4
+      - name: Workaround for grub-efi-amd64-signed
```
EmilienM, line 43: I'm wondering if we could add a comment here to explain a bit why we're doing this so next time it breaks we know why we are apt-get uprading here.

This also makes me wonder if we should just enforce an apt-get upgrade in here: https://github.com/EmilienM/devstack-action/blob/main/action.yaml so we always get the latest packages.

And then I'll tag a new release and we can simply update the action tag in all workflows here.

.github/workflows/functional-baremetal.yaml
```diff
@@ -91,9 +96,7 @@ jobs:
           OS_BRANCH: ${{ matrix.openstack_version }}
       - name: Generate logs on failure
         run: ./script/collectlogs
-        if: failure()
```
EmilienM, line 94: we did this to save log artifacts, we found that it wasn't super useful to store all these files if the jobs passed. WDYT?

.github/workflows/functional-baremetal.yaml
iurygregory: So, my idea to execute this was after reading 
https://forum.hestiacp.com/t/ubuntu-20-04-installation-error/4323/5 and https://askubuntu.com/questions/1276111/error-upgrading-grub-efi-amd64-signed-special-device-old-ssd-does-not-exist

I do think it makes sense to run an upgrade before trying to install devstack (normally I do this when trying to deploy).

Yup, we can just add to https://github.com/EmilienM/devstack-action/blob/main/action.yaml =)

.github/workflows/functional-baremetal.yaml
iurygregory: yeah, I just removed in case things failed again I wanted to double check the pip-freeze from a green job (sometimes it makes sense to collect a few things I would say)

.github/workflows/functional-baremetal.yaml
EmilienM: Sounds good; so you can submit a PR there, I'll merge and right away tag a new version. Then you can update this PR with updating all tags for the action (on all workflows using it) and we'll see how that works!

.github/workflows/functional-baremetal.yaml
EmilienM: cool, yeah let's just remove it.

.github/workflows/functional-baremetal.yaml
iurygregory: ack, doing now

.github/workflows/functional-baremetal.yaml
```diff
@@ -41,7 +41,7 @@ jobs:
       - name: Checkout Gophercloud
         uses: actions/checkout@v4
       - name: Deploy devstack
-        uses: EmilienM/devstack-action@v0.11
+        uses: EmilienM/devstack-action@v0.12
```
mandre, line 44: For anyone looking, the fix consists in getting the ubuntu packages at their latest version prior to installing devstack.
The proposed fix in devstack-action is at https://github.com/EmilienM/devstack-action/pull/15.

EmilienM (approved):

.github/workflows/functional-baremetal.yaml
```diff
@@ -40,6 +40,10 @@ jobs:
     steps:
       - name: Checkout Gophercloud
         uses: actions/checkout@v4
+      - name: Workaround for grub-efi-amd64-signed
+        run: sudo apt update && sudo apt -y upgrade
+        shell: bash
+        if: ${{ matrix.ubuntu_version == "20.04" }}
```
EmilienM, line 46:
```
if: matrix.ubuntu_version == "20.04"
```

PR #2806, opened 2023-10-12
Fix list ports with multiple fixedip parameters

### LINKED ISSUE #2807: Invalid query generated for port list with fixed IPs
The following ListOpts generates an incorrect query:
```golang
ports.ListOpts{
	FixedIPs: []ports.FixedIPOpts{
		{SubnetID: "41701fe5-decf-4ab4-b3bb-410e7404a9cf", IPAddress: "10.0.128.1"},
	},
},
```

It generates:
```
&fixed_ips=ip-address=10.0.128.1,subnet_id=41701fe5-decf-4ab4-b3bb-410e7404a9cf
```

OpenStack client generates:
```
&fixed_ips=subnet_id=41701fe5-decf-4ab4-b3bb-410e7404a9cf&fixed_ips=ip_address=10.0.128.1
```

OpenStack client's query correctly returns the matched port. Gophercloud's query does not.

mdbooth:
Fixes #2807

code in this PR:

Absolutely no idea! I can't even find the entry point for the ports api handler. This is reverse engineered from the behaviour from the cli client, which works.

### DIFF
--- a/internal/acceptance/openstack/networking/v2/ports_test.go
+++ b/internal/acceptance/openstack/networking/v2/ports_test.go
@@ -4,6 +4,7 @@
 package v2
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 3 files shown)


mdbooth:
I'm anticipating that "Subnet ID and IP Address" will fail. If it does I'll open an issue.

mdbooth:
Test failed successfully 😁

mdbooth:
Semver arguments:

It doesn't affect semver at all because it's a bugfix. It previously did not work, now it works.

A user using gophercloud correctly would previously get no returned ports, where now they get returned ports. This is a breaking change in the previous behaviour of the gophercloud API, so is semver:major.

While the latter may be technically correct, I believe the former is more useful: it's just a bugfix.

pierreprinetti:
WRT semver: bug fixes can absolutely be major. If the same function call returns different results before and after given the same environment, then it’s major. There is a 100% chance that someone somewhere for some wicked reason is relying on the old broken behaviour.
This PR is another good reason to issue v2!

internal/acceptance/openstack/networking/v2/ports_test.go
```diff
@@ -70,6 +71,92 @@ func TestPortsCRUD(t *testing.T) {
...
+			name: "Subnet ID and IP Address",
+			opts: ports.ListOpts{
+				FixedIPs: []ports.FixedIPOpts{
+					{SubnetID: subnet.ID, IPAddress: ipAddress},
+				},
+			},
+			expectedPorts: 1,
+		},
+	} {
+		t.Run(fmt.Sprintf("List ports by %s", tt.name), func(t *testing.T) {
```
pierreprinetti, line 127: U.U.o.fmt

internal/acceptance/openstack/networking/v2/ports_test.go
```diff
@@ -70,6 +71,92 @@ func TestPortsCRUD(t *testing.T) {
...
+			}
+			func() {
+				for _, port := range allPorts {
+					if port.ID == newPort.ID {
+						return
+					}
+				}
+				logPorts()
+				t.Fatalf("Returned ports did not contain expected port")
+			}()
```
pierreprinetti, line 157: lol why

pierreprinetti: oh I get it. you're saving us one "found" bool

internal/acceptance/openstack/networking/v2/ports_test.go
mdbooth: What does that mean?

internal/acceptance/openstack/networking/v2/ports_test.go
mdbooth: Yeah, the early return.

mdbooth:
@pierreprinetti Did you need any changes before approval?

internal/acceptance/openstack/networking/v2/ports_test.go
pierreprinetti: Useless Use of fmt :grin:

pierreprinetti:
> @pierreprinetti Did you need any changes before approval?

Not on my side.

pierreprinetti (approved):
Thank you for fixing this, LGTM.

PR #2814, opened 2023-10-16
Make fixtures part of tests

mandre:
By renaming the fixtures to fixtures_test, we mark them as test, and no longer part of the public API.

Some files were left out as they were included in other tests:

    $ grep "/testing" -R openstack | cut -d " " -f 2 | sort | uniq
    "github.com/gophercloud/gophercloud/openstack/common/extensions/testing"
    "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens/testing"
    "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips/testing"
    "github.com/gophercloud/gophercloud/openstack/networking/v2/networks/testing"
    "github.com/gophercloud/gophercloud/openstack/networking/v2/ports/testing"
    "github.com/gophercloud/gophercloud/openstack/objectstorage/v1/accounts/testing"
    "github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers/testing"

This should prevent go-apidiff from complaining when modifying fixtures.

This commit mirrors 7f1d07519ffc99fdc74e1dafc92d56be52624989 that merged in master.


pierreprinetti:
Can you please mention the corresponding commit on the main branch, in the PR description and possibly in the commit itself?

pierreprinetti (approved):
This is technically a major change we're pushing to the stable branch.

The reason for this is: we're removing most test fixtures from the public API surface. Many changes that should have been "patch" or "minor" fell into the "major" bucket because they changed the test fixtures while adding coverage; and this is bad.

I expect disruption from this change. However, I believe that is necessary and that test code shouldn't have been part of the public API in the first place. Sorry about that.

PR #2815, opened 2023-10-17
Fix blind replacement of paths in Heat template referencing other files

dverbeir:
This change fixes an issue in the handling of Heat templates that reference other files (child templates or files included by means of `get_file:`) and that also contain, in some other context, the same string as the relative path being referenced (that must indeed be substituted for the absolute local path).

The existing fixFileRefs() does a simple blind string replacement inside the referencing template, regardless of the context. This is not appropriate as the same string as the path being replaced could possibly appear outside of the valid template/file reference cases (`type: ...` and `get_file: ...`), as in the below example:

      init-cmd:
        type: OS::Heat::CloudConfig
        properties:
          cloud_config:
            write_files:
              - path: /etc/somefile
                content: { get_file: somefile }

Here "somefile" in "{ get_file: somefile }" should be replaced by the absolute local path to 'somefile', but the "path: /etc/somefile" part should not be modified as that path refers here to the target destination in a completely different context (in this example, a cloud-init target path on a to-be-created VM).

The change also fixes another issue when using go v1.20 where the usage of `filepath.Dir()` on a URL removes any duplicate path separator when they should in fact be preserved (in "file:///some_abs_path/here", the "///" must remain).

Finally, the stack template tests have been slightly cleaned up (removing duplications, more concise code).

### DIFF
--- a/openstack/orchestration/v1/stacks/requests.go
+++ b/openstack/orchestration/v1/stacks/requests.go
@@ -52,7 +52,6 @@ func (opts CreateOpts) ToStackCreateMap() (map[string]interface{}, error) {
 	if err := opts.TemplateOpts.getFileContents(opts.TemplateOpts.Parsed, ignoreIfTemplate, true); err != nil {
 		return nil, err
 	}
-	opts.TemplateOpts.fixFileRefs()
 	b["template"] = string(opts.TemplateOpts.Bin)
 
 	files := make(map[string]string)
@@ -146,7 +145,6 @@ func (opts AdoptOpts) ToStackAdoptMap() (map[string]interface{}, error) {
 	if err := opts.TemplateOpts.getFileContents(opts.TemplateOpts.Parsed, ignoreIfTemplate, true); err != nil {
 		return nil, err
 	}
-	opts.TemplateOpts.fixFileRefs()
 	b["template"] = string(opts.TemplateOpts.Bin)
 
 	files := make(map[string]string)
@@ -371,7 +369,6 @@ func toStackUpdateMap(opts UpdateOpts) (map[string]interface{}, error) {
 		if err := opts.TemplateOpts.getFileContents(opts.TemplateOpts.Parsed, ignoreIfTemplate, true); err != nil {
 			return nil, err
 		}
-		opts.TemplateOpts.fixFileRefs()
 		b["template"] = string(opts.TemplateOpts.Bin)
 
 		for k, v := range opts.TemplateOpts.Files {
@@ -479,7 +476,6 @@ func (opts PreviewOpts) ToStackPreviewMap() (map[string]interface{}, error) {
 	if err := opts.TemplateOpts.getFileContents(opts.TemplateOpts.Parsed, ignoreIfTemplate, true); err != nil {
 		return nil, err
 	}
-	opts.TemplateOpts.fixFileRefs()
 	b["template"] = string(opts.TemplateOpts.Bin)
 
 	files := make(map[string]string)
--- a/openstack/orchestration/v1/stacks/requests.go
+++ b/openstack/orchestration/v1/stacks/requests.go
@@ -66,7 +66,6 @@ func (opts CreateOpts) ToStackCreateMap() (map[string]interface{}, error) {
 		if err := opts.EnvironmentOpts.getRRFileContents(ignoreIfEnvironment); err != nil {
 			return nil, err
 		}
-		opts.EnvironmentOpts.fixFileRefs()
 		for k, v := range opts.EnvironmentOpts.Files {
 			files[k] = v
 		}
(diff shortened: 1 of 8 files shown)


mandre:
Thanks for the PR @dverbeir. I'll look at your changes today.

You just made us realize we were skipping these tests in CI because they're not in the `testing` package. That's something else we'll have to fix.

Running the test locally indeed shows the failure:
```
❯ go test -tags 'fixtures' ./openstack/orchestration/v1/...
?       github.com/gophercloud/gophercloud/openstack/orchestration/v1/apiversions       [no test files]
?       github.com/gophercloud/gophercloud/openstack/orchestration/v1/buildinfo [no test files]
?       github.com/gophercloud/gophercloud/openstack/orchestration/v1/resourcetypes     [no test files]
ok      github.com/gophercloud/gophercloud/openstack/orchestration/v1/apiversions/testing       (cached)
ok      github.com/gophercloud/gophercloud/openstack/orchestration/v1/buildinfo/testing (cached)
?       github.com/gophercloud/gophercloud/openstack/orchestration/v1/stackevents       [no test files]
ok      github.com/gophercloud/gophercloud/openstack/orchestration/v1/resourcetypes/testing     (cached)
?       github.com/gophercloud/gophercloud/openstack/orchestration/v1/stackresources    [no test files]
ok      github.com/gophercloud/gophercloud/openstack/orchestration/v1/stackevents/testing       (cached)
ok      github.com/gophercloud/gophercloud/openstack/orchestration/v1/stackresources/testing    (cached)
?       github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacktemplates    [no test files]
--- FAIL: TestGetFileContentsComposeRelativePath (0.00s)
    convenience.go:36: Failure in template_test.go, line 269: unexpected error "error fetching file:/home/martin/go/src/github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacks/templates/my_nova.yaml: 404 Not Found"
FAIL
FAIL    github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacks    0.012s
ok      github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacks/testing    (cached)
ok      github.com/gophercloud/gophercloud/openstack/orchestration/v1/stacktemplates/testing    (cached)
FAIL

```

openstack/orchestration/v1/stacks/requests.go
```diff
@@ -52,7 +52,6 @@ func (opts CreateOpts) ToStackCreateMap() (map[string]interface{}, error) {
 	if err := opts.TemplateOpts.getFileContents(opts.TemplateOpts.Parsed, ignoreIfTemplate, true); err != nil {
 		return nil, err
 	}
-	opts.TemplateOpts.fixFileRefs()
```
mandre, line 55: Was that the only call to `fixFileRefs()` that was problematic, or should we drop more of them?
A quick grep shows more occurrences in the same file.

mandre (commented):
Thanks for breaking your work in multiple commits, this makes it easier to review. I just have a simple question, otherwise lgtm.

EmilienM:
please rebase

openstack/orchestration/v1/stacks/requests.go
dverbeir: Good point. I had looked into that at some point but because I didn't know what kind of manipulation was expected on the `Environment` part (just a few lines below, and where `getRRFileContents()` is used instead of `getFileContents()`), I decided to only touch the `Template` part that I was interested in and understood.
But I missed the fact that `Adopt`, `Update` and `Preview` also call `fixFileRefs()` for their template parts (after doing `getFileContents()`). I have now removed the calls from those other operations so they will also benefit from the fix.

Looking into `getRRFileContents()`, I saw it also uses `getFileContents()` but simply with a different filter function (`ignoreIfEnvironment()`). Hence, it is in fact already affected by my change since some of the files it loads could be re-generated with full paths expanded, as now done by `getFileContents()` without the need for `fixFileRefs()`.
So I went ahead and made a separate commit that completes similar changes for the env part and completely removes `fixFileRefs()`. I have however not tested this part other than with the included unit tests and am somewhat unsure as to whether it's okay to re-generate env content from the parsed structure as the code is now doing. Note that this was already the case with the previous commit (templates), just not for the top-level env content. I hope my explanation is not too unclear....

dverbeir:
Rebased to have the tests run on the modified code.

openstack/orchestration/v1/stacks/requests.go
mandre: Your explanation is perfectly clear, and I believe it is fine. It looks like a bug that was never noticed.

mandre (approved):
Really good work, thanks a lot.

PR #2818, opened 2023-10-20
Test files alongside code

mandre:
Prior to this patch, all test files sitting alongside the code, outside of a `testing` directory were ignored by the CI scripts.


mandre:
The unit tests failed as expected, they're fixed in https://github.com/gophercloud/gophercloud/pull/2815.

PR #2821, opened 2023-10-23
objects: Escape names in Gophercloud

### LINKED ISSUE #2547: Cannot download object if its name contains a % sign
Hello :wave: 

I have an object named `my%20file.txt` (not `my file.txt`!).
Unfortunately, If I do `objects.Download(storageClient, containerName, "my%20file.txt", nil)` I get a 404...

In order to be able to download the file I need to use `"my%2520file.txt"` (note the `%` was changed to `%25`).

The Download documentation doesn't mention anything about it: https://pkg.go.dev/github.com/gophercloud/gophercloud/openstack/objectstorage/v1/objects#Download

But in the library I'm using in another language, if I pass `['objectName' => 'my%20file.txt']` it returns 200 OK.

I see 2 possible fixes:
1. Update the doc
2. (breaking change) being able to download file with the correct name without having to encode it manually

Thanks for your attention :)

### LINKED ISSUE #2816: BREAKING: Urlencode Swift names
Gophercloud currently expects all incoming Swift container and object names to be pre-escaped.

I believe that Gophercloud should accept utf-8 strings and urlencode them as part of their handling.

This is a breaking change.

### LINKED ISSUE #2827: [Swift] Can't create a temp URL for an object with `/v1/` in its path
If an object's path contains the string `/v1/`, then `objects.CreateTempURL` will generate a URL that can only return 404.

pierreprinetti:
With this change, container and object names passed to the Gophercloud functions in `objectstorage/v1/containers` and `objectstorage/v1/objects` are now URL-escaped. If you were escaping them prior to passing them to Gophercloud, you should now remove escaping and pass the intended name to Gophercloud directly.

**This is a breaking change.**

API CHANGES:
* Before this change, container and object names were sent as-is to Swift. With this change, **Gophercloud now escapes container and object names in all `objects` and `containers` functions**. If you were previously escaping names ( with, for example: `url.PathEscape` or `url.QueryEscape`), then you should REMOVE that and pass the intended names to Gophercloud directly
* The **`containers.ListOpts#Full` and `objects.ListOpts#Full` properties are REMOVED** from the Gophercloud API. The reason for that is: plaintext listing is unfixably wrong and won't handle special characters reliably (i.e. `\n`).
* Empty container names, container names containing a slash (`/`), and empty object names are now rejected in Gophercloud before any call to Swift.
* `containers.ErrInvalidContainerName` is now `v1.ErrInvalidContainerName`
* New name validation errors:
  * `v1.ErrEmptyContainerName`
  * `v1.ErrEmptyObjectName`
* In `objects.Copy`: the `destination` field (e.g. `objects.CopyOpts#Destination`) must be in the form `/container/object`: the function will reject a destination path if it doesn't start with a slash (`/`).

CODE CHANGES:
* `containers.List` and `objects.List` are now always setting `accept: application/json`. There is no plaintext listing any more, for the reason detailed above.
* `containers` and `objects` acceptance tests are now run with nasty Unicode code points (including European characters, a space, "\n", backslash, slash where appropriate, emojis, emojis with modifiers, Chinese traditional characters), instead of only using nice ASCII characters.
* Container and object names are now systematically validated before being sent to Swift.

INCIDENTAL FIX:
* it is now possible to successfuly call `objects.CreateTempURL` with a container name or object name containing the string `/v1/`.

Fixes #2816 
Fixes #2547
Fixes #2827


pierreprinetti:
TODO:
* [ ] acceptance tests
* [ ] unit tests

internal/acceptance/openstack/objectstorage/v1/objects_test.go
```diff
@@ -135,7 +136,7 @@ func TestObjects(t *testing.T) {
 
 	// Copy the contents of one object to another.
 	copyOpts := objects.CopyOpts{
-		Destination: cName + "/" + oNames[1],
+		Destination: "/" + cName + "/" + oNames[1],
```
pierreprinetti, line 139: As per [the docs](https://docs.openstack.org/api-ref/object-store/#copy-object): the Copy `Destination` header must be in the form of `/container/object`.

internal/acceptance/openstack/objectstorage/v1/objects_test.go
```diff
@@ -68,7 +70,6 @@ func TestObjects(t *testing.T) {
 
 	// List all created objects
 	listOpts := objects.ListOpts{
-		Full:   true,
```
pierreprinetti, line 71: List option `Full: false` (`accept: text/plain`) doesn't work and should not be used

internal/acceptance/openstack/objectstorage/v1/objects_test.go
```diff
@@ -22,19 +22,21 @@ import (
...
 	}
 
 	// Make a slice of length numObjects to hold the random object names.
 	oNames := make([]string, numObjects)
-	for i := 0; i < len(oNames); i++ {
-		oNames[i] = tools.RandomString("test-object-", 8)
+	for i := 0; i < len(oNames)-1; i++ {
+		oNames[i] = "test-object-" + tools.RandomFunnyString(8)
 	}
+	oNames[len(oNames)-1] = "test-object-with-/v1/-in-the-name"
```
pierreprinetti, line 36: There were two test cases. Now there are three: the last one tests that we correctly split the object path in `CreateTempURL`.

internal/acceptance/openstack/objectstorage/v1/objects_test.go
```diff
@@ -301,7 +302,6 @@ func TestObjectsListSubdir(t *testing.T) {
 	}()
 
 	listOpts := objects.ListOpts{
-		Full:      true,
```
pierreprinetti, line 304: List option `Full: false` (`accept: text/plain`) doesn't work and should not be used. We always list "full" now (`accept: application/json`).

internal/acceptance/openstack/objectstorage/v1/objects_test.go
```diff
@@ -330,7 +330,6 @@ func TestObjectsListSubdir(t *testing.T) {
 	}
 
 	listOpts = objects.ListOpts{
-		Full:      true,
```
pierreprinetti, line 333: List option `Full: false` (`accept: text/plain`) doesn't work and should not be used. We always list "full" now (`accept: application/json`).

internal/acceptance/openstack/objectstorage/v1/objects_test.go
```diff
@@ -416,7 +415,6 @@ func TestObjectsBulkDelete(t *testing.T) {
 
 	// Verify deletion
 	listOpts := objects.ListOpts{
-		Full:      true,
```
pierreprinetti, line 419: List option `Full: false` (`accept: text/plain`) doesn't work and should not be used. We always list "full" now (`accept: application/json`).

internal/acceptance/tools/tools.go
```diff
@@ -1,10 +1,10 @@
 package tools
 
 import (
-	"crypto/rand"
```
pierreprinetti, line 4: cryptographically secure randomness is not required here. and the `math/rand` package has lots more convenient functions

internal/acceptance/tools/tools.go
```diff
@@ -47,22 +47,36 @@ func MakeNewPassword(oldPass string) string {
...
 // RandomString generates a string of given length, but random content.
 // All content will be within the ASCII graphic character set.
-// (Implementation from Even Shaw's contribution on
-// http://stackoverflow.com/questions/12771930/what-is-the-fastest-way-to-generate-a-long-random-string-in-go).
 func RandomString(prefix string, n int) string {
-	const alphanum = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz"
-	var bytes = make([]byte, n)
-	rand.Read(bytes)
-	for i, b := range bytes {
-		bytes[i] = alphanum[b%byte(len(alphanum))]
```
pierreprinetti, line 57: I have changed how we generate random strings because this algorithm only works with ASCII chars: most non-ASCII chars take more than one byte to represent.

internal/acceptance/tools/tools.go
```diff
@@ -47,22 +47,36 @@ func MakeNewPassword(oldPass string) string {
...
+	charset := []rune("0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz")
+	return prefix + randomString(charset, n)
```
pierreprinetti, line 52: The old `RandomString` function retains this prefix concatenation for compatibility. The new funny generators don't accept a prefix.

internal/acceptance/tools/tools.go
```diff
@@ -47,22 +47,36 @@ func MakeNewPassword(oldPass string) string {
...
+	for i := 0; i < length; i++ {
+		s.WriteRune(charset[rand.Intn(len(charset))])
 	}
-	return prefix + string(bytes)
+	return s.String()
 }
 
 // RandomInt will return a random integer between a specified range.
 func RandomInt(min, max int) int {
-	mrand.Seed(time.Now().Unix())
```
pierreprinetti, line 64: The global `Rand` of `math/rand` [is automatically seeded](https://pkg.go.dev/math%2Frand#Seed) as of Go v1.20.

openstack/objectstorage/v1/containers/errors.go
```diff
@@ -1,13 +0,0 @@
-package containers
-
-import "github.com/gophercloud/gophercloud"
-
-// ErrInvalidContainerName signals a container name containing an illegal
-// character.
-type ErrInvalidContainerName struct {
```
pierreprinetti, line 7: Moved to the parent directory `v1` to share it with `objects`

openstack/objectstorage/v1/containers/requests.go
```diff
@@ -123,15 +124,18 @@ func Create(c *gophercloud.ServiceClient, containerName string, opts CreateOptsB
 
 // BulkDelete is a function that bulk deletes containers.
 func BulkDelete(c *gophercloud.ServiceClient, containers []string) (r BulkDeleteResult) {
-	// urlencode container names to be on the safe side
-	// https://github.com/openstack/swift/blob/stable/train/swift/common/middleware/bulk.py#L160
-	// https://github.com/openstack/swift/blob/stable/train/swift/common/swob.py#L302
-	encodedContainers := make([]string, len(containers))
-	for i, v := range containers {
-		encodedContainers[i] = v
```
pierreprinetti, line 131: The comment says we encoded, but we didn't! The strings were just copied into a new slice :shrug:

openstack/objectstorage/v1/containers/testing/fixtures.go
```diff
@@ -82,38 +82,14 @@ func HandleListContainerInfoSuccessfully(t *testing.T) {
 	})
 }
 
-// HandleListContainerNamesSuccessfully creates an HTTP handler at `/` on the test handler mux that
-// responds with a `ListNames` response when only container names are requested.
-func HandleListContainerNamesSuccessfully(t *testing.T) {
```
pierreprinetti, line 87: This fixture is not used any more, because we don't send `accept: text/plain` any more on listing.

openstack/objectstorage/v1/containers/testing/requests_test.go
```diff
@@ -18,18 +19,17 @@ func TestContainerNames(t *testing.T) {
...
+		expectedError error
 	}{
 		{
 			"rejects_a_slash",
 			"one/two",
+			v1.ErrInvalidContainerName{},
 		},
 		{
-			"rejects_an_escaped_slash",
-			"one%2Ftwo",
```
pierreprinetti, line 28: The name validation is always run right before encoding in Gophercloud. Thus if the caller sent us an encoded slash, we can happily take it as it will be double-encoded and will not expand as a slash in Swift.

openstack/objectstorage/v1/containers/urls.go
```diff
@@ -1,40 +1,21 @@
...
-	"strings"
+	"net/url"
 
 	"github.com/gophercloud/gophercloud"
+	v1 "github.com/gophercloud/gophercloud/openstack/objectstorage/v1"
 )
 
-const forbiddenContainerRunes = "/"
-
-func CheckContainerName(s string) error {
```
pierreprinetti, line 12: Moved to the parent directory `v1`

openstack/objectstorage/v1/objects/requests.go
```diff
@@ -328,22 +329,41 @@ func (opts CopyOpts) ToObjectCopyQuery() (string, error) {
 
 // Copy is a function that copies one object to another.
 func Copy(c *gophercloud.ServiceClient, containerName, objectName string, opts CopyOptsBuilder) (r CopyResult) {
-	url, err := copyURL(c, containerName, objectName)
```
pierreprinetti, line 331: moving this just a bit down below, just because `url` shadows the `net/url` package which we need for escaping.

openstack/objectstorage/v1/objects/requests.go
```diff
@@ -574,6 +594,10 @@ func CreateTempURL(c *gophercloud.ServiceClient, containerName, objectName strin
 	if err != nil {
 		return "", err
 	}
+	urlToBeSigned, err := tempURL(c, containerName, objectName)
```
pierreprinetti, line 597: We need two different URLs: one with non-escaped names (which will be signed), and one with escaped names (which will be used as the actual URL to send back to the caller). Documented here: https://docs.openstack.org/swift/latest/api/temporary_url_middleware.html#hmac-signature-for-temporary-urls

openstack/objectstorage/v1/objects/requests.go
```diff
@@ -611,8 +635,10 @@ func CreateTempURL(c *gophercloud.ServiceClient, containerName, objectName strin
 	}
 
 	secretKey := []byte(tempURLKey)
-	splitPath := strings.Split(url, opts.Split)
```
pierreprinetti, line 614: incidental fix: `Split` splits `url` each time `opts.Split` (default: `/v1/`) is found. But below, we only joined the first two parts back into a full URL. Which means that a path like `https://swift/v1/AUTH_123/container/my/v1/object` became `https://swift/v1/AUTH_123/container/my` (see https://github.com/gophercloud/gophercloud/issues/2827)

openstack/objectstorage/v1/objects/testing/fixtures_test.go
```diff
@@ -94,7 +94,7 @@ var ExpectedListSubdir = []objects.Object{
 
 // ExpectedListNames is the result expected from a call to `List` when just
 // object names are requested.
-var ExpectedListNames = []string{"hello", "goodbye"}
+var ExpectedListNames = []string{"goodbye", "hello"}
```
pierreprinetti, line 97: I am not sure how this ever worked. Objects are always returned lexicographically sorted.

EmilienM (approved):
impressive work, LGTM

PR #2822, opened 2023-10-24
Add more godoc to GuestFormat

EmilienM:
I recently found out that when not providing the GuestFormat to a block
device mapping, the device will be formated to `vfat` anyway.

https://opendev.org/openstack/nova/src/commit/d0b459423dd81644e8d9382b6c87fabaa4f03ad4/nova/privsep/fs.py#L257

I think it's valuable that we document that so the users know what to
expect.


pierreprinetti (approved):
Thank you: that is extremely valuable input for the user.

PR #2840, opened 2023-11-30
unit tests: Fix the installation of tools

pierreprinetti:
Install Go tools with new newest Go version before switching to the target version of Go to run the unit tests.

This commit fixes an issue where the Go tools we use in unit tests can't be compiled with Go v1.14 any more.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -20,22 +20,23 @@ jobs:
       GO111MODULE: "on"
 
     steps:
-      - name: Setup Go ${{ matrix.go-version }}
+      - uses: actions/checkout@v3
+
+      - name: Setup Go v1
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1
+
+      - name: Install tools
+        run: |
+          go install github.com/wadey/gocovmerge@master
+          go install golang.org/x/tools/cmd/goimports@latest
+
+      - if: ${{ matrix.go-version != '1' }}
+        name: Setup Go ${{ matrix.go-version }}
         uses: actions/setup-go@v4
         with:
           go-version: ${{ matrix.go-version }}
-
-      - uses: actions/checkout@v3
-
-      - name: Setup environment
-        run: |
-          # Changing into a different directory to avoid polluting go.sum with "go get"
-          cd "$(mktemp -d)"
-          go mod init unit_tests
-
-          # we use "go get" for Go v1.14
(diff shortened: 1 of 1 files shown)


.github/workflows/unit.yml
```diff
@@ -20,29 +20,38 @@ jobs:
       GO111MODULE: "on"
 
     steps:
-      - name: Setup Go ${{ matrix.go-version }}
+      - if: ${{ matrix.go-version == '1.14' }}
```
mandre, line 23: It feels like this step shouldn't be a conditional, we want to always install the latest 1.x go.

.github/workflows/unit.yml
```diff
@@ -20,29 +20,38 @@ jobs:
...
+        name: Setup Go to install tools
         uses: actions/setup-go@v4
         with:
-          go-version: ${{ matrix.go-version }}
+          go-version: 1
+
+      - if: ${{ matrix.go-version == '1.14' }}
```
mandre, line 29: You can remove the conditional here too.

.github/workflows/unit.yml
```diff
@@ -20,29 +20,38 @@ jobs:
...
-          go mod init unit_tests
+      - name: Setup Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v4
+        with:
+          go-version: ${{ matrix.go-version }}
 
-          # we use "go get" for Go v1.14
-          go install github.com/wadey/gocovmerge@master || go get github.com/wadey/gocovmerge
-          go install golang.org/x/tools/cmd/goimports@latest || go get golang.org/x/tools/cmd/goimports@v0.8.0
+      - if: ${{ matrix.go-version == '1' }}
```
mandre, line 42: This step can go away, we're installing the deps for all versions above.

.github/workflows/unit.yml
```diff
@@ -20,29 +20,38 @@ jobs:
...
+          go install golang.org/x/tools/cmd/goimports@latest
 
       - uses: actions/checkout@v3
 
-      - name: Setup environment
-        run: |
-          # Changing into a different directory to avoid polluting go.sum with "go get"
-          cd "$(mktemp -d)"
-          go mod init unit_tests
+      - name: Setup Go ${{ matrix.go-version }}
```
mandre, line 37: Here this should be a conditional to install go at version 1.14 when needed.

EmilienM:
Unit tests have passed we don't have to wait for the functional tests to pass, this is good to me.
Thanks for doing that.

PR #2863, opened 2024-01-11
Add octavia pool member batch update BatchUpdateMembersAdditiveOnly

AdrienPensart:
Fixes #[2862]

Related to https://docs.openstack.org/api-ref/load-balancer/v2/#batch-update-members


openstack/loadbalancer/v2/pools/requests.go
```diff
@@ -492,6 +493,35 @@ func BatchUpdateMembers(c *gophercloud.ServiceClient, poolID string, opts []Batc
...
+	for _, opt := range opts {
+		b, err := opt.ToBatchMemberUpdateMap()
+		if err != nil {
+			r.Err = err
+			return
+		}
+		members = append(members, b)
+	}
+
+	b := map[string]interface{}{"members": members}
```
dulek, line 509: Looks like code duplication with `BatchUpdateMembers()` that you could minimize by extracting this into a separate function.

mandre (commented):
The code looks good, but I'd like to see acceptance tests showing that the new `BatchUpdateMembersAdditiveOnly()` function does what is claims to do. Can we add that?

AdrienPensart:
Can we have a review?

PR #2876, opened 2024-01-19
[wip][db/v1/instances] add access field

zhekazuev:

Fixes #[PUT ISSUE NUMBER HERE]

code in this PR:

[PUT URLS HERE]

### DIFF
--- a/openstack/db/v1/instances/requests.go
+++ b/openstack/db/v1/instances/requests.go
@@ -45,6 +45,22 @@ func (opts NetworkOpts) ToMap() (map[string]interface{}, error) {
 	return gophercloud.BuildRequestBody(opts, "")
 }
 
+// AccessOpts is used within CreateOpts to control database service exposing.
+type AccessOpts struct {
+	// Whether the database service is exposed to the public.
+	// Optional.
+	IsPublic       bool     `json:"is_public,omitempty"`
+	// A list of IPv4, IPv6 or mix of both CIDRs that restrict access to the database service.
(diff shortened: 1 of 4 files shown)


zhekazuev:
@EmilienM Hi,

What better to use for Access field, a pointer(`*AccessOpts`) or a value(`AccessOpts`)?
Commit with changes: https://github.com/gophercloud/gophercloud/pull/2876/commits/0b8b1e12aa3a4dc8cc8689da90226285582574b3

From the risks and style side.
I prefer to not use pointers where possible, to avoid null pointer issues, but maybe for current scenario there is no option with a value.

zhekazuev:
> @EmilienM Hi,
> 
> What better to use for Access field, a pointer(`*AccessOpts`) or a value(`AccessOpts`)? Commit with changes: [0b8b1e1](https://github.com/gophercloud/gophercloud/commit/0b8b1e12aa3a4dc8cc8689da90226285582574b3)
> 
> From the risks and style side. I prefer to not use pointers where possible, to avoid null pointer issues, but maybe for current scenario there is no option with a value.

@EmilienM Hi,

Could you review these changes?

I just added new field - Access. Without methods specific for that field, e.g.:
- UpdateInstanceAccessbility - update access options like [here]( https://github.com/gophercloud/gophercloud/pull/2831/files#diff-a5c2bfd4393cc9171dafbca6ff6ec3d87ff04b4f7a25cf38ddc737adb6fd9ca9R305).

And decided to use the value for Access struct instead of pointers as described below.
If it is important to use pointers, I can restore the changes to the pointer version.

PR #2882, opened 2024-01-28
db/v1/instance: Add configuration to createOpts

nikParasyr:
Add configuration group id to createOpts and update unit tests accordingly.

Docs: https://docs.openstack.org/api-ref/database/#create-database-instance


pierreprinetti:
@EmilienM Do you plan on backporting this patch to a release?

PR #2883, opened 2024-01-31
Context-aware methods to ProviderClient and ServiceClient

pierreprinetti:
Provide new methods to leverage per-call `context.Context`.

In the http.Request resulting from using these new context-aware calls, the per-call context values take precendence over the ProviderClient's.

Cancellation obeys both the per-call context and the ProviderClient's.

### DIFF
--- a/internal/ctxt/merge_test.go
+++ b/internal/ctxt/merge_test.go
@@ -1,0 +1,133 @@
+package ctxt_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/gophercloud/gophercloud/internal/ctxt"
+)
+
+func TestMerge(t *testing.T) {
+	t.Run("returns values from both parents", func(t *testing.T) {
+		ctx1 := context.WithValue(context.Background(),
+			"key1", "value1")
+
+		ctx2 := context.WithValue(context.WithValue(context.Background(),
+			"key1", "this value should be overridden"),
+			"key2", "value2")
+
+		ctx, cancel := ctxt.Merge(ctx1, ctx2)
+		defer cancel()
+
+		if v1 := ctx.Value("key1"); v1 != nil {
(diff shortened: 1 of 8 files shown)


pierreprinetti:
This PR has been mentioned in the Gophercloud Slack channel: https://kubernetes.slack.com/archives/C05G4NJ6P6X/p1706708314206099

EmilienM (approved):

EmilienM (approved):

pierreprinetti:
One last change: Moved the context-merging function to an `internal` package, so that it doesn't add to Gophercloud's API.

EmilienM (approved):

internal/ctxt/merge_test.go
```diff
@@ -0,0 +1,129 @@
...
+	"time"
+
+	"github.com/gophercloud/gophercloud/internal/ctxt"
+)
+
+func TestMerge(t *testing.T) {
+	t.Run("returns values from both parents", func(t *testing.T) {
+		ctx, cancel := ctxt.Merge(
+			context.WithValue(context.Background(), "key1", "value1"),
+			context.WithValue(context.Background(), "key2", "value2"),
```
mandre, line 15: Can you perhaps also check that the value from the first context takes precedence when the second context holds the same key? If this is a behavior that we're going to support, then we should have tests for it so that we avoid regressions.

mandre (commented):
Thanks @pierreprinetti, I don't think that dropping the values from the second context is a big problem in practice so I'm fine starting with the current implementation and revisit if we find this is a blocking issue. This is already a huge step forward.

internal/ctxt/merge_test.go
pierreprinetti: this is a good idea, thanks

EmilienM (approved):

PR #2884, opened 2024-02-01
[v1] Context-aware methods to ProviderClient and ServiceClient

pierreprinetti:
Provide new methods to leverage per-call `context.Context`.

Backports https://github.com/gophercloud/gophercloud/pull/2883

This patch, suitable to be merged to `v1`, vendors the context package from the v1.21 standard library, to make `AfterFunc` available.

### DIFF
--- a/internal/ctxt/merge_test.go
+++ b/internal/ctxt/merge_test.go
@@ -1,0 +1,133 @@
+package ctxt_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/gophercloud/gophercloud/internal/ctxt"
+)
+
+func TestMerge(t *testing.T) {
+	t.Run("returns values from both parents", func(t *testing.T) {
+		ctx1 := context.WithValue(context.Background(),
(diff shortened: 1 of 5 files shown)


EmilienM (approved):

pierreprinetti:
One last change: Moved the context-merging function to an `internal` package, so that it doesn't add to Gophercloud's API.

EmilienM (dismissed):

internal/ctxt/merge_test.go
```diff
@@ -0,0 +1,129 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/gophercloud/gophercloud/internal/ctxt"
+)
+
+func TestMerge(t *testing.T) {
+	t.Run("returns values from both parents", func(t *testing.T) {
```
mandre, line 12: Let's just add a test for values that take precedence, as in https://github.com/gophercloud/gophercloud/pull/2883.

EmilienM (approved):

PR #2686, opened 2023-07-11
Introduce context usage in the service client

pkenchap:
Context usage in the gopher cloud gives more request handling and goroutines working on a request typically needs access to request-specific values such as the identity of the end user, authorization tokens, and the request’s deadline. if we enable the context which provide easy integration external tools .

why we are proposing to have a context in the client is to integration of **Opentelemetry (**https://github.com/open-telemetry/opentelemetry-go) in the Gophercloud sdk (https://github.com/gophercloud/gophercloud) . provides the tracing of the API can be monitored with the timelines, which gives immense use for the performance test for calculation individual api time taken for the operation. Also gives a nice UI based graphs using the JaegerUI .
Prior to starting a PR, please make sure you have read our
[contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial).

we may need to pass the context parameter to functions in the service_client.go file .
For example : Adding the ctx context.Context to the function which has these REST API endpoints .
func (client *ServiceClient) Get(ctx context.Context, url string, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error)

But service client will call the Request method in the Provider_client it does not contain to have a context , only when we call the do_request 
https://github.com/pkenchap/gophercloud/blob/cf659998b191737a8e9116dcf2f31918635558af/service_client.go#L153

if we use the https://github.com/pkenchap/gophercloud/blob/cf659998b191737a8e9116dcf2f31918635558af/provider_client.go#L363

do request method for the context in the service client then we can provide the context to the provider_client . 

https://github.com/gophercloud/gophercloud/issues/2670

Fixes #2670

PR #2886, opened 2024-02-02
Fix value_specs implementation

### LINKED ISSUE #2870: ValueSpecs in port CreateOpts is a no-op
I was reading https://github.com/gophercloud/gophercloud/pull/2537, which adds `value_specs` to ports. The doc link in the PR points to the Heat documentation, where it is a parameter. It is documented as:
> Extra parameters to include in the request.

The gophercloud implementation includes a `value_specs` parameter in the port creation request. However, I couldn't find any evidence that neutron supports this parameter. Digging in to the implementation in Heat, I think the purpose of this field is to add arbitrary top-level fields to the request. i.e. There is no `value_specs` field.
 I found some validation of `value_specs` which supports that:
https://github.com/openstack/heat/blob/2df46d4c59140255ece640896019bfca03437388/heat/engine/resources/openstack/neutron/neutron.py#L49-L62

Also, the only use of it I could find uses it in this way:
https://github.com/openstack/heat/blob/2df46d4c59140255ece640896019bfca03437388/heat/engine/resources/openstack/nova/server_network_mixin.py#L131-L133

I think we should re-examine the use case for this API. It is possible that it's required to support out-of-tree neutron drivers. If so, we should fix it to behave the same way as it does in Heat. However, it's clear that nobody is currently using it. This also opens an argument for removing it[^1], especially as it's an unusual API backdoor.

I have a suspicion that it was just an escape-hatch in Heat so they didn't have to add every port parameter to Heat. I note that in Heat ports have binding:vnic_type, but not binding:(host_id|profile|vif_details|vif_type). Gophercloud does have these, though:
https://pkg.go.dev/github.com/gophercloud/gophercloud@v1.8.0/openstack/networking/v2/extensions/portsbinding. Is it possible this isn't required in gophercloud?

[^1]: Or marking it deprecated with a note that it's a no-op, as API stability guarantees require.

lentzi90:
The current implementation does not do anything useful. The key-pairs must be extracted and added directly in the body instead of under "value_specs".

This commit fixes the implementation and also adds validation of the value_specs based on what is done in Heat. There are some banned keys that are not allowed in value_specs, and values are not allowed to overwrite existing fields either.

Fixes #2870

### DIFF
--- a/openstack/networking/v2/ports/requests.go
+++ b/openstack/networking/v2/ports/requests.go
@@ -3,6 +3,7 @@ package ports
 import (
 	"fmt"
 	"net/url"
+	"slices"
 
 	"github.com/gophercloud/gophercloud"
 	"github.com/gophercloud/gophercloud/pagination"
@@ -130,7 +131,37 @@ type CreateOpts struct {
 
 // ToPortCreateMap builds a request body from CreateOpts.
 func (opts CreateOpts) ToPortCreateMap() (map[string]interface{}, error) {
-	return gophercloud.BuildRequestBody(opts, "port")
+	body, err := gophercloud.BuildRequestBody(opts, "port")
+	if err != nil {
+		return nil, err
+	}
+
+	return AddValueSpecs(body)
+}
+
+// AddValueSpecs expands the 'value_specs' object and removes 'value_specs'
+// from the request body. It will return error if the value specs would overwrite
+// an existing field or contains forbidden keys.
+func AddValueSpecs(body map[string]interface{}) (map[string]interface{}, error) {
+	// Banned the same as in heat. See https://github.com/openstack/heat/blob/dd7319e373b88812cb18897f742b5196a07227ea/heat/engine/resources/openstack/neutron/neutron.py#L59
+	bannedKeys := []string{"shared", "tenant_id"}
+	port := body["port"].(map[string]interface{})
+
+	if port["value_specs"] != nil {
+		for k, v := range port["value_specs"].(map[string]interface{}) {
+			if slices.Contains(bannedKeys, k) {
+				return nil, fmt.Errorf("forbidden key in value_specs: %s", k)
+			}
+			if _, ok := port[k]; ok {
+				return nil, fmt.Errorf("value_specs would overwrite key: %s", k)
+			}
+			port[k] = v
+		}
+		delete(port, "value_specs")
+	}
+	body["port"] = port
+
+	return body, nil
 }
 
 // Create accepts a CreateOpts struct and creates a new network using the values
@@ -173,7 +204,11 @@ type UpdateOpts struct {
 
 // ToPortUpdateMap builds a request body from UpdateOpts.
 func (opts UpdateOpts) ToPortUpdateMap() (map[string]interface{}, error) {
-	return gophercloud.BuildRequestBody(opts, "port")
+	body, err := gophercloud.BuildRequestBody(opts, "port")
+	if err != nil {
+		return nil, err
+	}
+	return AddValueSpecs(body)
 }
 
 // Update accepts a UpdateOpts struct and updates an existing port using the
--- a/openstack/networking/v2/ports/results.go
+++ b/openstack/networking/v2/ports/results.go
@@ -114,9 +114,6 @@ type Port struct {
 	// PropagateUplinkStatus enables/disables propagate uplink status on the port.
 	PropagateUplinkStatus bool `json:"propagate_uplink_status"`
 
-	// Extra parameters to include in the request.
-	ValueSpecs map[string]string `json:"value_specs"`
-
 	// RevisionNumber optionally set via extensions/standard-attr-revisions
 	RevisionNumber int `json:"revision_number"`
 
--- a/openstack/networking/v2/ports/testing/fixtures.go
+++ b/openstack/networking/v2/ports/testing/fixtures.go
@@ -299,9 +299,7 @@ const CreateValueSpecRequest = `
             "mac_address": "fa:16:3e:c9:cb:f0"
           }
         ],
-        "value_specs": {
-            "key": "value"
-        }
+        "test": "value"
     }
 }
 `
@@ -332,9 +330,6 @@ const CreateValueSpecResponse = `
             "mac_address": "fa:16:3e:c9:cb:f0"
           }
         ],
-        "value_specs": {
-            "key": "value"
-        },
         "device_id": ""
     }
 }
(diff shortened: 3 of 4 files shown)


openstack/networking/v2/ports/requests.go
```diff
@@ -1,9 +1,12 @@
 package ports
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
 
+	"golang.org/x/exp/slices"
```
lentzi90, line 8: This is in the standard library from go 1.21 so I hope it is ok to use already.
https://pkg.go.dev/slices

pierreprinetti: what you are importing here is the experimental package that went on being added to the stdlib as `"slices"` in Go v1.21. I am bumping the Go version in https://github.com/gophercloud/gophercloud/pull/2686; once it merges you can rebase and use `"slices"` directly.

pierreprinetti: actually, the package you import uses generics. This patch is not backportable anyway as-is.

openstack/networking/v2/ports/requests.go
```diff
@@ -130,7 +133,36 @@ type CreateOpts struct {
...
+// from the reqeust body. It will return error if the value specs would overwrite
+// an existing field or contains forbidden keys.
+func AddValueSpecs(body map[string]interface{}) (map[string]interface{}, error) {
+	bannedKeys := []string{"shared", "tenant_id"}
+	port := body["port"].(map[string]interface{})
+
+	if port["value_specs"] != nil {
+		for k, v := range port["value_specs"].(map[string]interface{}) {
+			if slices.Contains(bannedKeys, k) {
+				return nil, errors.New(fmt.Sprintf("forbidden key in value_specs: %s", k))
```
pierreprinetti, line 154:
```suggestion
				return nil, fmt.Errorf("forbidden key in value_specs: %q", k)
```

openstack/networking/v2/ports/requests.go
```diff
@@ -130,7 +133,36 @@ type CreateOpts struct {
...
+			}
+			if _, ok := port[k]; ok {
+				return nil, errors.New(fmt.Sprintf("value_specs would overwrite key: %s", k))
```
pierreprinetti, line 157:
```suggestion
				return nil, fmt.Errorf("value_specs would overwrite key: %q", k)
```

openstack/networking/v2/ports/requests.go
```diff
@@ -130,7 +133,36 @@ type CreateOpts struct {
...
+	}
+
+	return AddValueSpecs(body)
+}
+
+// AddValueSpecs expands the 'value_specs' object and removes 'value_specs'
+// from the reqeust body. It will return error if the value specs would overwrite
+// an existing field or contains forbidden keys.
+func AddValueSpecs(body map[string]interface{}) (map[string]interface{}, error) {
+	bannedKeys := []string{"shared", "tenant_id"}
```
pierreprinetti, line 148: Why only these two?

lentzi90: I copied what Heat is doing: https://github.com/openstack/heat/blob/dd7319e373b88812cb18897f742b5196a07227ea/heat/engine/resources/openstack/neutron/neutron.py#L59
Not sure about why exactly these two were banned though

pierreprinetti:
Did we ever return anything in the `ValueSpecs` field of `ports.Port`?

EmilienM:
@lentzi90 don't heat remove value_specs from the body before sending to neutron? shouldn't we do the same?
https://github.com/openstack/heat/commit/d4b6f37ab57226f1e4d90ebb8ac0ba127d4ad973

lentzi90:
> @lentzi90 doesn't Heat remove value_specs from the body before sending to neutron? shouldn't we do the same? [openstack/heat@d4b6f37](https://github.com/openstack/heat/commit/d4b6f37ab57226f1e4d90ebb8ac0ba127d4ad973)

Yes! I hope it is what I'm doing with

```go
delete(port, "value_specs")
```

So first we take all keys under `value_specs`, check if they are banned, check if they would overwrite an existing field, if not we add them to the body and finally remove the `value_specs` field.

lentzi90:
> Did we ever return anything in the `ValueSpecs` field of `ports.Port`?

I don't think so

lentzi90:
Any way to re-trigger that one failed test besides pushing again?
Everything passed already before I changed `errors.New` -> `fmt.Errorf`, which is trivial, so it is probably a flake...

pierreprinetti:
I have manually retriggered the run

openstack/networking/v2/ports/requests.go
EmilienM: Please rebase on main and use `slices` directly since https://github.com/gophercloud/gophercloud/pull/2883 has merged now. Thanks!

openstack/networking/v2/ports/requests.go
EmilienM: Fine by me, the list can evolve later. However I'll like to see a comment explaining that we took it from Heat without knowing much about why these 2 :smile:

EmilienM (changes requested):
Minor change is required before we can merge, I think today.

.gitignore
```diff
@@ -2,3 +2,5 @@
 .idea
 .vscode
 testing_*.coverprofile
+# Development container configuration (https://containers.dev/)
```
EmilienM, line 5: unrelated change, maybe via separate commit

lentzi90: I can remove it, no worries :slightly_smiling_face:

EmilienM (approved):

EmilienM:
unit test job is broken but it's not your PR

EmilienM:
LGTM for me. I'll let a chance to other reviewers to have another look as well (cc @pierreprinetti).

EmilienM:
might need a rebase in fact

openstack/networking/v2/ports/requests.go
```diff
@@ -130,7 +131,37 @@ type CreateOpts struct {
...
+	}
+
+	return AddValueSpecs(body)
+}
+
+// AddValueSpecs expands the 'value_specs' object and removes 'value_specs'
+// from the request body. It will return error if the value specs would overwrite
+// an existing field or contains forbidden keys.
+func AddValueSpecs(body map[string]interface{}) (map[string]interface{}, error) {
+	// Banned the same as in heat. See https://github.com/openstack/heat/blob/dd7319e373b88812cb18897f742b5196a07227ea/heat/engine/resources/openstack/neutron/neutron.py#L59
```
dulek, line 146: Would probably be better to link to opendev.org instead of GitHub mirrors, no big deal.

dulek (commented):
Looks okay from my perspective.

openstack/networking/v2/ports/requests.go
lentzi90: Ah my bad :facepalm: 
Just say the word and I'll update...

PR #2891, opened 2024-02-02
test: Restructure actions

pierreprinetti:
gofmt and go vet should now be enough for linting.


EmilienM:
failing with `/home/runner/work/_temp/1ae76895-cf5a-47fc-80eb-4bbe7bbc4815.sh: line 2: ./script/format: No such file or directory` but that's normal because the github action didn't take the change from Pierre yet since we need it merge first.

EmilienM (approved):

EmilienM:
I'll love to see that backported, eventually. I think the bot will fail to cherry-pick due to conflict though. WDYT?

EmilienM:
oops, that broke :( we probably want the file back, @pierreprinetti

PR #2892, opened 2024-02-05
Authenticate with a clouds.yaml

pierreprinetti:
This commit imports the clouds.yaml parsing code from the utils module, and exposes it in a way that fits the natural Gophercloud authentication flow.

Unlike the code from utils, this new `clouds.Parse` function keeps the separation between the ProviderClient (holding the cloud coordinates and a Keystone token) and the ServiceClient (holding specific endpoint configuration).

By default, `clouds.Parse` fetches its configuration from the environment, just like the openstack client would do.

Example use:

```Go
func main() {
	ctx := context.Background()
	ao, eo, tlsConfig, err := clouds.Parse()
	if err != nil {
		panic(err)
	}

	providerClient, err := config.NewProviderClient(ctx, ao, config.WithTLSConfig(tlsConfig))
	if err != nil {
		panic(err)
	}

	networkClient, err := openstack.NewNetworkV2(providerClient, eo)
	if err != nil {
		panic(err)
	}
}
```

The `clouds.Parse` function accepts several functional options that can modify its behaviour. For example, to use a `clouds.yaml` that exists in a non-standard path:

```Go
ao, eo, tlsConfig, err := clouds.Parse(clouds.WithLocations("/my/path/clouds2.yaml"))
```

It is also possible to pass a reader directly. Note that any number of options can be passed, with each of them taking precedence of the previous if there is conflict.

```Go
const exampleClouds = `clouds:
  openstack:
    auth:
      auth_url: https://example.com:13000`

ao, eo, tlsConfig, err := clouds.Parse(
	clouds.WithCloudsYAML(strings.NewReader(exampleClouds)),
	clouds.WithIdentityEndpoint("https://example.com:13001"),
	clouds.WithCloudName("osp1"),
	clouds.WithUsername("alice"),
)

```

### DIFF
--- a/openstack/config/clouds/clouds.go
+++ b/openstack/config/clouds/clouds.go
@@ -1,0 +1,271 @@
+// package clouds provides a parser for OpenStack credentials stored in a clouds.yaml file.
+//
+// Example use:
+//
+//	ctx := context.Background()
+//	ao, eo, tlsConfig, err := clouds.Parse()
+//	if err != nil {
+//		panic(err)
+//	}
+//
+//	providerClient, err := config.NewProviderClient(ctx, ao, config.WithTLSConfig(tlsConfig))
+//	if err != nil {
+//		panic(err)
+//	}
+//
+//	networkClient, err := openstack.NewNetworkV2(providerClient, eo)
+//	if err != nil {
+//		panic(err)
+//	}
+package clouds
+
+import (
+	"crypto/tls"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"path"
+	"reflect"
+
+	"github.com/gophercloud/gophercloud"
+	"gopkg.in/yaml.v2"
+)
+
+// Parse fetches a clouds.yaml file from disk and returns the parsed
+// credentials.
+//
+// By default this function mimics the behaviour of python-openstackclient, which is:
+//
+//   - if the environment variable `OS_CLIENT_CONFIG_FILE` is set and points to a
+//     clouds.yaml, use that location as the only search location for `clouds.yaml` and `secure.yaml`;
+//   - otherwise, the search locations for `clouds.yaml` and `secure.yaml` are:
+//     1. the current working directory (on Linux: `./`)
+//     2. the directory `openstack` under the standatd user config location for
+//     the operating system (on Linux: `${XDG_CONFIG_HOME:-$HOME/.config}/openstack/`)
+//     3. on Linux, `/etc/openstack/`
+//
+// Once `clouds.yaml` is found in a search location, the same location is used to search for `secure.yaml`.
+//
+// Like in python-openstackclient, relative paths in the `clouds.yaml` section
+// `cacert` are interpreted as relative the the current directory, and not to
+// the `clouds.yaml` location.
+//
+// Search locations, as well as individual `clouds.yaml` properties, can be
+// overwritten with functional options.
+func Parse(opts ...func(*cloudOpts)) (gophercloud.AuthOptions, gophercloud.EndpointOpts, *tls.Config, error) {
+	options := cloudOpts{
+		cloudName:    os.Getenv("OS_CLOUD"),
+		region:       os.Getenv("OS_REGION_NAME"),
+		endpointType: os.Getenv("OS_INTERFACE"),
+		locations: func() []string {
+			if path := os.Getenv("OS_CLIENT_CONFIG_FILE"); path != "" {
+				return []string{path}
+			}
+			return nil
+		}(),
+	}
+
+	for _, apply := range opts {
+		apply(&options)
+	}
+
+	if options.cloudName == "" {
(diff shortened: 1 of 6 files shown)


EmilienM (commented):
I haven't spot any problems with it, although I haven't tested it yet. Before approving, I'll give it a try. In the meantime, I was wondering if you could have a least one e2e test for this one?

pierreprinetti:
Replaced generic code with typed code in the function `coalesce`, as we currently only use it for strings. Generic code was preventing us from backporting this patch to `v1`.

pierreprinetti:
Replaced a call to [`strings.CutPrefix`](https://pkg.go.dev/strings#CutPrefix) with `strings.TrimPrefix`, which is compatible with Go v1.14.

pierreprinetti:
There may be more work required here, if e.g. we want `ProviderClient` and `ServiceClient` from v1 work with v2 (and vice-versa). This would require us to add getter methods to them (non-breaking change) and always accept an interfaces in the function calls (potentially breaking change).

EmilienM:
Finally managed to test this, really cool work!

```bash
~/gophercloud-playground via 🐹 v1.21.6 
❯ go run main.go
panic: cloud "" not found in clouds.yaml

goroutine 1 [running]:
main.main()
	/home/emilien/gophercloud-playground/main.go:20 +0x7f5
exit status 2

~/gophercloud-playground via 🐹 v1.21.6 
❯ export OS_CLOUD=foch_openshift

~/gophercloud-playground via 🐹 v1.21.6 on ☁️  foch_openshift(openshift) 
❯ go run main.go 

~/gophercloud-playground via 🐹 v1.21.6 on ☁️  foch_openshift(openshift) 
❯ openstack server list
+--------------------------------------+--------------+--------+----------------------------------------+--------------------------+-----------+
| ID                                   | Name         | Status | Networks                               | Image                    | Flavor    |
+--------------------------------------+--------------+--------+----------------------------------------+--------------------------+-----------+
| 17833e16-466d-4fbc-875a-cd36bc235f5f | test-emacchi | ACTIVE | hostonly=192.168.68.120, 2001:db8::221 | N/A (booted from volume) | m1.medium |
+--------------------------------------+--------------+--------+----------------------------------------+--------------------------+-----------+
```

pierreprinetti:
@EmilienM is that panic coming from your code, or from gophercloud?

EmilienM:
> @EmilienM is that panic coming from your code, or from gophercloud?

gophercloud I think, my code is here: https://paste.centos.org/view/65f08709

openstack/config/clouds/clouds.go
```diff
@@ -0,0 +1,267 @@
...
+//
+// Like in python-openstackclient, relative paths in the `clouds.yaml` section
+// `cacert` are interpreted as relative the the current directory, and not to
+// the `clouds.yaml` location.
+//
+// Search locations, as well as individual `clouds.yaml` properties, can be
+// overwritten with functional options.
+func Parse(opts ...func(*cloudOpts)) (gophercloud.AuthOptions, gophercloud.EndpointOpts, *tls.Config, error) {
+	options := cloudOpts{
+		cloudName:    os.Getenv("OS_CLOUD"),
```
EmilienM, line 58: please add a check for OS_CLOUD to be set, otherwise we end up with:

```
panic: cloud "" not found in clouds.yaml
```

pierreprinetti: On my machine, this code doesn't panic:

```Go
package main

import (
	"log"
	"os"

	"github.com/gophercloud/gophercloud/openstack/config/clouds"
)

func main() {
	os.Unsetenv("OS_CLOUD")
	_, _, _, err := clouds.Parse()
	if err != nil {
		log.Fatal(err)
	}
}
```

Is it different for you?

pierreprinetti:
> > @EmilienM is that panic coming from your code, or from gophercloud?
> 
> gophercloud I think, my code is here: https://paste.centos.org/view/65f08709
> 
> it's because I didn't export `OS_CLOUD`.

line 20?

openstack/config/clouds/clouds.go
EmilienM: it's the same for me. What I'm saying is that I would have improved the error to explicitly say that the user needs to export `OS_CLOUD`.

pierreprinetti: I have added an error:

```Go
		return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, fmt.Errorf("the empty string \"\" is not a valid cloud name")

```

Do you think it's better now?

EmilienM: works for me

EmilienM:
> There may be more work required here, if e.g. we want `ProviderClient` and `ServiceClient` from v1 work with v2 (and vice-versa). This would require us to add getter methods to them (non-breaking change) and always accept an interfaces in the function calls (potentially breaking change).

I think this could be done later, right? I'm going to approve this PR for now.

EmilienM (approved):

mandre (commented):
This is really cool :sunglasses: 
LGTM

Issue #2902, opened 2024-02-09
Please consider fixing more bad APIs in v2.0


majewsky:
I saw that you put out a v2.0.0 beta, and the breaking changes are very tame. Since this is a unique opportunity, I wonder how many thoughts have been spent on improving the core APIs. Specifically, I have two requests:

- Please consider removing all the `Err4xxer` interfaces and `ErrDefault4xx` structs without replacement, and just returning `ErrUnexpectedResponseCode` instead. Or at the very least, these types should be documented much better to explain what their purpose is. As far as I can tell, all they do is [ruin good error messages unless you know how to get rid of them](https://github.com/sapcc/limes/blob/f25253aacf7849836af9a08d28278e4e8ae6dd60/internal/util/errors.go#L28-L36).

- The `context.Context` support added in the previous release is pretty much worthless since all the structured API methods do not take `ctx` arguments. This is the best possible moment to fix that, and add a `ctx` argument to every method that wraps `ProviderClient.Request` or `ServiceClient.Request`. The variants of those methods without `ctx` arguments should be removed.

This just off the top of my head, there's probably more. Please consider using this opportunity if at all possible.

pierreprinetti:
Hi @majewsky, thank you for coming forward.

We have chosen to take a backwards-compatible approach to most changes, including the adoption of `context.Context`. Now that the main library is ready, we want to over time add `WithContext` versions of, ideally, every single call in Gophercloud. This is a time-consuming effort that you're welcome to join, and it can be done within `v2` (or even `v1`) without a major version bump.

The conversation on errors is a bit more complicated. I agree that Gophercloud errors need restructuring, and at the same time I'm afraid of breaking users in nasty ways. We didn't get complaints, so we assumed that we could just postpone the resolution of this issue. That's why I'm glad to hear from you. I would like all errors in Gophercloud to implement `Unwrap()` and expose strategic error types to facilitate their handling (in particular, but not limited to, the 404s). But I am afraid that we might be targeting v3 here.

Note that you're welcome to join [the #gophercloud channel in the Kubernetes Slack workspace](https://kubernetes.slack.com/archives/C05G4NJ6P6X/p1707471033444039).

pierreprinetti:
@majewsky do you have a proposal for errors?

majewsky:
> @majewsky do you have a proposal for errors?

[...]

PR #2904, opened 2024-02-09
Remove error code types

majewsky:
This is a breaking change intended for the v2.0 release, spun off from the discussion in #2902. I am posting this as a bit of a hot take. If this ends up being a discussion about why the ErrDefaultXXX types and the ErrorContext field are actually useful, and we just end up with actual real documentation on these, that would be a good outcome as well.

## Overview and rationale

The ErrDefault series of types (ErrDefault400, ErrDefault401, etc.) serve the apparent purpose of allowing to assert on a specific response code (e.g. to handle 404 responses differently from other 4xx/5xx responses). However, because of their design, they often tend to obscure the real underlying error message in many cases, reporting just something like "Method not allowed" instead of the default error string on ErrUnexpectedResponseCode that explains which request failed and includes the actual error message returned by the OpenStack service.

This commit removes the types, and all of the associated boilerplate, and replaces them with a single small helper function for quickly asserting that an error is an ErrUnexpectedResponseCode with a specific response code. A similar design has been in productive use in [my Swift client library](https://github.com/majewsky/schwift/blob/v1.3.0/errors.go#L152-L157) for several years.

This also gets rid of the undocumented and untested ProviderClient.ErrorContext interface that I have legitimately no idea what it's supposed to be good for, besides maybe allowing users to skip the generation of ErrDefaultXXX-type errors. If that's the only purpose, that would be yet more incentive to just get rid of them.

Finally, as a small (but still breaking) change, this removes some newlines from the error messages generated by ErrUnexpectedResponseCode. Newlines in error messages cause several log aggregators to choke (e.g. I have seen problems with this myself in syslog and fluentd), so we now try a bit harder to avoid them if possible.


pierreprinetti:
I am working on leveraging errors.Unwrap, which should bring the same benefits: https://github.com/gophercloud/gophercloud/pull/2906

pierreprinetti:
I completely agree that (RequestOpts).ContextError must go, since error unwrapping should provide the same capability in a more idiomatic way.

I am finding myself in deeper a rabbit hole than I can handle with #2906. If this change fits your needs, then I'd be inclined to merge.

pierreprinetti:
@gophercloud/core what do you think?

mandre (approved):
I'm very much in favor of this change as it makes consuming gophercloud easier.

pierreprinetti:
@majewsky Can you please rebase? I think we can then merge :rocket:

majewsky:
rebase done

pierreprinetti:
hmm what happened to the CI

pierreprinetti:
@majewsky something wrong happened on the Github side. When checking out this branch locally, I get the old commits and not the ones you just force-pushed. Do you mind `git commit --amend -no-edit && git push --force` please?

majewsky:
I rebased as requested three weeks ago, but apparently this was not seen (maybe because I did not also comment "done"?). Because new merge conflicts have arisen, I have rebased again.

majewsky:
Rebased onto current master to match 2.0.0-beta.4. Please take another look.

mandre:
There's a failed conflict resolution in `internal/acceptance/openstack/identity/v3/federation_test.go`.

majewsky:
One error remains in a testcase that I did not touch (`TestBGPAgentRUD`). The concrete failure does not seem to relate to my changes, so I'll have to leave it to you to decide if this is caused by the PR.

mandre (approved):

PR #2905, opened 2024-02-09
Add AvailabilityFromEndpointType

pierreprinetti:
We have a function in utils that returns "public" when given "publicURL", "internal" with "internalURL" and "admin" with "adminURL". If it is of any use, I think it has a place in the main Gophercloud repository.


pierreprinetti:
we are considering issuing utils/v2 instead
cc @EmilienM

PR #2906, opened 2024-02-09
Draft: Restructure errors

pierreprinetti:
Use error.Unwrap() to provide ergonomic errors.


pierreprinetti:
Closing. This change is too invasive; I suspect there may be simpler solutions.

PR #2907, opened 2024-02-09
add ctx argument on all Result-returning API methods

majewsky:
Please do not be scared by the size of this. This is not intended to be merged in this form, but to serve as a demonstration to make a point and maybe get a discussion going. @pierreprinetti said in #2902:

> We have chosen to take a backwards-compatible approach to most
changes, including the adoption of context.Context. Now that the main library is ready, we want to over time add WithContext versions of, ideally, every single call in Gophercloud. This is a time-consuming effort that you're welcome to join, and it can be done within v2 (or even v1) without a major version bump.

I felt that the part about this being a time-consuming effort is a misconception, so I tried how long this actually takes. What is in this PR took about 1.5 hours to produce and is probably 80-90% of the way there in terms of actual code changes, so I think this is something that should be considered for not just the 2.x line, but specifically the 2.0 release. When I say "1.5 hours", this obviously does not consider the time spent reviewing it, but if the aforementioned decision to introduce context-awareness slowly was taken based on the assumption that it requires a lot of person-hours that need to be spread out, I don't think that's really true. In fact, a lot of pain and manual labor can probably be saved by ripping the band-aid off right here and now, instead of maintaining to separate versions of each request function.

As a downstream user of this library, I have waited for a long time for it to become context-aware, then got excited seeing it on the 1.9.0 release notes, then immediately disappointed upon seeing that the API surface in 1.9.0 is not useful in practical applications. Now hearing that it will be another entire major version before I can fully convert my programs to context-aware for good is quite frustrating, especially if it could be done sooner. This is my motivation for proposing this alternate path. And besides, I cannot bear the thought of you folks possibly wasting dozens of hours on writing these changes manually in a tedious backwards-compatible manner when it could be done through scripting in a few minutes.

I do not intend for this PR to be merged:

- It has test failures because I only checked `go vet ./openstack/...`. Specifically, additional code changes will be required in the acceptance tests.
- It is ludicrously large and flies in the face of the "small PRs" requirement in the contribution guidelines.

If you guys decide that you want to move forward on this change as presented here, I can try and split it up into one PR per service and add the required changes to the acceptance tests etc. For now, this is a Request for Comments in the original sense of the term, not a literal Pull Request.

---

For reproducibility, here is how I prepared this changeset. Most of the auto changes are through `sed -i`, each time followed by quick manual confirmation through `git add -p`. I also ran `goimports -w` on all the files touched by `sed -i` to add `import "context"` as needed. Then I planned the next auto change by running `go vet ./openstack/...`, picking an error at random, and generalizing it into a regex.

```
sed -i 's/\<client\.\(Get\|Post\|Put\|Patch\|Delete\|Head\|Request\)(/client.\1WithContext(ctx, /g' openstack/**/*.go
sed -i '!/\<Pager\>/s/^func \(\w\+\)(\(\w\+\) \*gophercloud.ServiceClient\>/func \1(ctx context.Context, \2 *gophercloud.ServiceClient/' openstack/**/requests.go openstack/**/request.go
```

This adds the `ctx` argument to most Result-returning functions, and uses it in the `client.Get` etc. calls within.

```
sed -i 's/(\(\w\+\).ServiceClient()/(context.TODO(), \1.ServiceClient()/' openstack/**/*_test.go
sed -i 's/(\(&\?\(c\|sc\|client\)\), /(context.TODO(), \1, /' openstack/**/*_test.go
```

This adds a `context.TODO()` argument to most test invocations of those functions.

```
sed -i 's/List\(\w*\)(ctx, /List\1(/' openstack/**/*.go
sed -i 's/List\(\w*\)(context.TODO(), /List\1(/' openstack/**/*_test.go
```

Functions called `List...` usually return a Pager instead of a Result, so this reverts the previous changes there.

At this point, `go vet ./openstack/...` was only showing a few dozen things, so I flipped through those files manually with

```
vim $(go vet ./openstack/... |& sed -n 's/^vet: \(.*\)\.go:[0-9].*/\1.go/p')
```

To clean that up, I changed:
- the various `func WaitForStatus` in `openstack/**/util.go`
- the various functions in `openstack/utils/choose_version.go`
- some specialty interfaces that my regexes and/or fileglobs did not catch
- two instances where regexes replaces a `client.Get()` that was not a ServiceClient, but a `http.Client` (reverted those changes)
- about a dozen instances where Pager-returning functions had a superfluous `ctx` argument because they were not called `List...`, and Result-returning functions that did not have `ctx` because they were called `List...`

---

Some final observations:

- This changeset has been made a lot easier by Gophercloud following a relatively strict style guide throughout, so I have gained a new-found appreciation of this strict style.
- If it is decided to merge something like this for 2.0, I also suggest to consider removing all the non-context-aware versions of functions with `WithContext` versions. If we're going to rip the band-aid off, let's go all the way.


pierreprinetti:
Thank you for this PoC and for the detailed instructions how you got there.

I am absolutely open to this conversation.

I would be more than happy to merge context-aware functions across the entire API in v2.0.0 already, provided that we can get reasonably sure that the context is being effectively used in the function logic, and that nothing is terribly broken.

The big question now is whether we want to _replace_ the current function calls with a signature that accepts a ctx, or instead just _add_ new `WithContext` functions is a backwards-compatible manner.

The single advantage to `WithContext` functions is the backwards compatibility; however it comes at the cost of double the API surface, many more lines of code, and an unending source of mess when we add new functions to the API.

I am sure that @majewsky's `sed` skills can produce `WithContext` functions with a nice godoc, with a small time overhead. The question is whether we actually need that, as opposed to how scared we are to force our users to add `context.TODO()` here and there.

What do you think, Internet?

pierreprinetti:
For what it's worth, I would be fine merging a complete replacement of all functions with their context-aware counterpart.
But it has to be all or nothing imho.

majewsky:
I will be on the road from tomorrow for the rest of the week, so likely won't have time to continue on this PR. But I will make a note to try and get `go vet` cleared next week.

majewsky:
Ok, so I had another half hour of spare time, so I cleared out the `go vet` errors on the acceptance tests. I just added `context.TODO()` on all of these since that's easy to do and achieves feature parity with the existing non-context-bearing function calls (which imply `context.Background()`, which is equivalent `context.TODO()`). If you would rather want `context.Background()` on these, I can do a `sed -i s/context.TODO/context.Background/ internal/acceptance/**/*.go`.

---

Since I started microblogging my process, here's how I did this: I opened all files with compile errors using

```
vim $(go vet ./... |& sed -n 's/^\(vet: \)\?\(internal.*\)\.go:[0-9].*/\2.go/p' | sort -u)
```

and used the following mappings inside vim:

```vimscript

nmap <F1> }gazzF("tp
nnoremap <F2> :Gw<CR>
nnoremap <F3> :n<CR>
```

The `t` register contains `context.TODO(), `, so F1 goes to the next compile error and adds a `context.TODO()` to that method call as the first argument. F2 saves the file (which causes my ALE to re-run `goimports` and `go vet`), and F3 advances to the next file. I had to do that several times because `go vet` does not report everything at once, but I was able to finish this after about 20 minutes.

vooon:
In my opinion better to have **only** context-aware calls. Users anyway would have to change the code, at least to add `/v2` into imports. So that's not a big deal to place `context.TODO()` if they don't have context already.

For example, now i do a lot of tricks to pass context by cloning ServiceClient, which i will be more than happy to drop and just use my ctx, like pretty much all other REST/gRPC client libraries does...

pierreprinetti:
@majewsky Do you also plan on updating the acceptance tests?

Removing the helper functions from the service client and provider client will might help track any residue work.

```diff
diff --git a/provider_client.go b/provider_client.go
index 76404319..489c3306 100644
--- a/provider_client.go
+++ b/provider_client.go
@@ -355,19 +355,14 @@ type requestState struct {
 
 var applicationJSON = "application/json"
 
-// RequestWithContext performs an HTTP request using the ProviderClient's
+// Request performs an HTTP request using the ProviderClient's
 // current HTTPClient. An authentication header will automatically be provided.
-func (client *ProviderClient) RequestWithContext(ctx context.Context, method, url string, options *RequestOpts) (*http.Response, error) {
+func (client *ProviderClient) Request(ctx context.Context, method, url string, options *RequestOpts) (*http.Response, error) {
 	return client.doRequest(ctx, method, url, options, &requestState{
 		hasReauthenticated: false,
 	})
 }
 
-// Request is a compatibility wrapper for Request.
-func (client *ProviderClient) Request(method, url string, options *RequestOpts) (*http.Response, error) {
-	return client.RequestWithContext(context.Background(), method, url, options)
-}
-
 func (client *ProviderClient) doRequest(ctx context.Context, method, url string, options *RequestOpts, state *requestState) (*http.Response, error) {
 	var body io.Reader
 	var contentType *string
diff --git a/service_client.go b/service_client.go
index b8e6fd1a..61834cc6 100644
--- a/service_client.go
+++ b/service_client.go
@@ -60,88 +60,58 @@ func (client *ServiceClient) initReqOpts(JSONBody interface{}, JSONResponse inte
 	}
 }
 
-// GetWithContext calls `Request` with the "GET" HTTP verb.
-func (client *ServiceClient) GetWithContext(ctx context.Context, url string, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
+// Get calls `Request` with the "GET" HTTP verb.
+func (client *ServiceClient) Get(ctx context.Context, url string, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
 	if opts == nil {
 		opts = new(RequestOpts)
 	}
 	client.initReqOpts(nil, JSONResponse, opts)
-	return client.RequestWithContext(ctx, "GET", url, opts)
+	return client.Request(ctx, "GET", url, opts)
 }
 
-// Get is a compatibility wrapper for GetWithContext.
-func (client *ServiceClient) Get(url string, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
-	return client.GetWithContext(context.Background(), url, JSONResponse, opts)
-}
-
-// PostWithContext calls `Request` with the "POST" HTTP verb.
-func (client *ServiceClient) PostWithContext(ctx context.Context, url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
+// Post calls `Request` with the "POST" HTTP verb.
+func (client *ServiceClient) Post(ctx context.Context, url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
 	if opts == nil {
 		opts = new(RequestOpts)
 	}
 	client.initReqOpts(JSONBody, JSONResponse, opts)
-	return client.RequestWithContext(ctx, "POST", url, opts)
-}
-
-// Post is a compatibility wrapper for PostWithContext.
-func (client *ServiceClient) Post(url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
-	return client.PostWithContext(context.Background(), url, JSONBody, JSONResponse, opts)
+	return client.Request(ctx, "POST", url, opts)
 }
 
-// PutWithContext calls `Request` with the "PUT" HTTP verb.
-func (client *ServiceClient) PutWithContext(ctx context.Context, url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
+// Put calls `Request` with the "PUT" HTTP verb.
+func (client *ServiceClient) Put(ctx context.Context, url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
 	if opts == nil {
 		opts = new(RequestOpts)
 	}
 	client.initReqOpts(JSONBody, JSONResponse, opts)
-	return client.RequestWithContext(ctx, "PUT", url, opts)
+	return client.Request(ctx, "PUT", url, opts)
 }
 
-// Put is a compatibility wrapper for PurWithContext.
-func (client *ServiceClient) Put(url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
-	return client.PutWithContext(context.Background(), url, JSONBody, JSONResponse, opts)
-}
-
-// PatchWithContext calls `Request` with the "PATCH" HTTP verb.
-func (client *ServiceClient) PatchWithContext(ctx context.Context, url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
+// Patch calls `Request` with the "PATCH" HTTP verb.
+func (client *ServiceClient) Patch(ctx context.Context, url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
 	if opts == nil {
 		opts = new(RequestOpts)
 	}
 	client.initReqOpts(JSONBody, JSONResponse, opts)
-	return client.RequestWithContext(ctx, "PATCH", url, opts)
-}
-
-// Patch is a compatibility wrapper for PatchWithContext.
-func (client *ServiceClient) Patch(url string, JSONBody interface{}, JSONResponse interface{}, opts *RequestOpts) (*http.Response, error) {
-	return client.PatchWithContext(context.Background(), url, JSONBody, JSONResponse, opts)
+	return client.Request(ctx, "PATCH", url, opts)
 }
 
-// DeleteWithContext calls `Request` with the "DELETE" HTTP verb.
-func (client *ServiceClient) DeleteWithContext(ctx context.Context, url string, opts *RequestOpts) (*http.Response, error) {
+// Delete calls `Request` with the "DELETE" HTTP verb.
+func (client *ServiceClient) Delete(ctx context.Context, url string, opts *RequestOpts) (*http.Response, error) {
 	if opts == nil {
 		opts = new(RequestOpts)
 	}
 	client.initReqOpts(nil, nil, opts)
-	return client.RequestWithContext(ctx, "DELETE", url, opts)
-}
-
-// Delete is a compatibility wrapper for DeleteWithContext.
-func (client *ServiceClient) Delete(url string, opts *RequestOpts) (*http.Response, error) {
-	return client.DeleteWithContext(context.Background(), url, opts)
+	return client.Request(ctx, "DELETE", url, opts)
 }
 
-// HeadWithContext calls `Request` with the "HEAD" HTTP verb.
-func (client *ServiceClient) HeadWithContext(ctx context.Context, url string, opts *RequestOpts) (*http.Response, error) {
+// Head calls `Request` with the "HEAD" HTTP verb.
+func (client *ServiceClient) Head(ctx context.Context, url string, opts *RequestOpts) (*http.Response, error) {
 	if opts == nil {
 		opts = new(RequestOpts)
 	}
 	client.initReqOpts(nil, nil, opts)
-	return client.RequestWithContext(ctx, "HEAD", url, opts)
-}
-
-// Head is a compatibility wrapper for HeadWithContext.
-func (client *ServiceClient) Head(url string, opts *RequestOpts) (*http.Response, error) {
-	return client.HeadWithContext(context.Background(), url, opts)
+	return client.Request(ctx, "HEAD", url, opts)
 }
 
 func (client *ServiceClient) setMicroversionHeader(opts *RequestOpts) {
@@ -164,7 +134,7 @@ func (client *ServiceClient) setMicroversionHeader(opts *RequestOpts) {
 }
 
 // Request carries out the HTTP operation for the service client
-func (client *ServiceClient) RequestWithContext(ctx context.Context, method, url string, options *RequestOpts) (*http.Response, error) {
+func (client *ServiceClient) Request(ctx context.Context, method, url string, options *RequestOpts) (*http.Response, error) {
 	if options.MoreHeaders == nil {
 		options.MoreHeaders = make(map[string]string)
 	}
@@ -182,12 +152,7 @@ func (client *ServiceClient) RequestWithContext(ctx context.Context, method, url
 			options.MoreHeaders[k] = v
 		}
 	}
-	return client.ProviderClient.RequestWithContext(ctx, method, url, options)
-}
-
-// Request is a compatibility wrapper for RequestWithContext.
-func (client *ServiceClient) Request(method, url string, options *RequestOpts) (*http.Response, error) {
-	return client.RequestWithContext(context.Background(), method, url, options)
+	return client.ProviderClient.Request(ctx, method, url, options)
 }
 
 // ParseResponse is a helper function to parse http.Response to constituents.
```

I agree that we should do the paginator in a separate PR (before shipping v2!) because I think that some other change is needed there.

pierreprinetti:
In this branch I added two commits on top of yours, where I remove the non-context functions from provider_client and service_client: https://github.com/shiftstack/gophercloud/commits/ctx_everywhere/

Feel free to include them in your PR.

pierreprinetti:
@majewsky Thank you for your work. I'd like to merge it as soon as possible, and before `v2.0.0`. Do you plan on tackling the acceptance tests (`./internal/acceptance`) as well?

pierreprinetti:
@majewsky If https://github.com/gophercloud/gophercloud/pull/2936 passes the tests, I would be inclined to merge. It is your work and mine combined in one commit. I have removed the non-context-aware functions in the provider and service client, so that compilation should fail in case we left something behind. What do you think?

Please feel free to incorporate the changes of that PR in this PR instead, if you're more comfortable with that.

pierreprinetti:
> Sure, go ahead. I'm on vacation this week, so thanks for picking this up and moving it towards the finish line. I'm fine with you continuing in your PR since that's probably easier for you to continue working on.

OK then, I am closing this in favour of https://github.com/gophercloud/gophercloud/pull/2936!

PR #2914, opened 2024-02-13
v2: Add ParseOption type to made clouds.Parse() more usable for optional With* funcs

### LINKED ISSUE #2913: v2: Export a slice of clouds config opts
Now it's impossible to make a code like that:

```go
func connect(cfg *Config) {

    opts := []cloudOpts{clouds.WithCloud(cfg.Cloud)}
    if cfg.CloudsFile != "" {
        opts = append(opts, clouds.WithLocations(cfg.CloudsFile))
    }

    ao, eo, tlsCfg, err := clouds.Parse(opts...)
    ...
}
```

So it's hard to provide optional parts to a parser.

https://github.com/gophercloud/gophercloud/blob/master/openstack/config/clouds/options.go

vooon:
So that makes easier to have conditional With like that:

https://github.com/sardinasystems/fleeting-plugin-openstack/blob/90c744fc9ee23d8a18dad96f2a677a9d1a54222c/provider.go#L44-L52

Fixes #2913

In addition it fixes search for default `/etc/openstack/clouds.yaml` file.

### DIFF
--- a/openstack/config/clouds/clouds.go
+++ b/openstack/config/clouds/clouds.go
@@ -86,7 +86,7 @@ func Parse(opts ...func(*cloudOpts)) (gophercloud.AuthOptions, gophercloud.Endpo
 			if err != nil {
 				return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, fmt.Errorf("failed to get the user config directory: %w", err)
 			}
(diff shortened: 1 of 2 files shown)


pierreprinetti:
I understand that it can be useful to conditionally add functional options, and indeed I couldn't find a straightforward way of allowing that without exposing the options type. Thank you @vooon for your contribution.

vooon:
@pierreprinetti that's quite common way doing With*, for example [grpc.DialOption](https://pkg.go.dev/google.golang.org/grpc#DialOption)

pierreprinetti (approved):

PR #2936, opened 2024-02-21
Make Gophercloud context-aware

pierreprinetti:
With this change, all functions and methods that trigger an HTTP call
now accept an instance of `context.Context` as their first argument.

This breaking change enables cancellation and tracing to be used across
all packages of the library.

Co-Authored-By: @majewsky
Co-Authored-By: @pierreprinetti


pierreprinetti:
This should be good to go @gophercloud/core

EmilienM (approved):
Nice work, let's ship it now and if something was missed we can fix it on the go.

PR #2964, opened 2024-03-04
Use system scope when running baremetal tests

dtantsur:
Starting with 2024.1, some actions no longer work with a project scoped
token.


dtantsur:
Let's make sure functional-baremetal passes before merging.

mandre (approved):
LGTM, thanks.

Issue #2967, opened 2024-03-05
Allow setting the hostname (when creating a new server)


till:
API docs/Python here:

- https://docs.openstack.org/api-ref/compute/#create-server
- https://github.com/openstack/python-openstackclient/blob/2eea916b35fd6f01947dce168c50c314ed9072db/openstackclient/compute/v2/server.py#L1447-L1457

---

Somewhat related: flatcar/Flatcar#1385

till:
@mandre thanks for merging, I'll close this. :)

PR #2968, opened 2024-03-05
compute: Allow setting the hostname when creating a server

till:
Related: #2967

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -241,6 +241,12 @@ type CreateOpts struct {
 	// Tags allows a server to be tagged with single-word metadata.
 	// Requires microversion 2.52 or later.
 	Tags []string `json:"tags,omitempty"`
+
+	// (Available from 2.90) Hostname specifies the hostname to configure for the
+	// instance in the metadata service. Starting with microversion 2.94, this can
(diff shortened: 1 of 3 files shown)


openstack/compute/v2/servers/requests.go
```diff
@@ -241,6 +241,12 @@ type CreateOpts struct {
 	// Tags allows a server to be tagged with single-word metadata.
 	// Requires microversion 2.52 or later.
 	Tags []string `json:"tags,omitempty"`
+
+	// (Available from 2.9.0) Hostname specifies the hostname to configure for the
```
mandre, line 245: Nit, a small dot slipped in.
```suggestion
	// (Available from 2.90) Hostname specifies the hostname to configure for the
```

till: Thank you @mandre! Fixed!

mandre (approved):

till:
@mandre can you tell why it fails? I can't find anything in the log output. But let me know if I broke anything.

mandre:
> @mandre can you tell why it fails? I can't find anything in the log output. But let me know if I broke anything.

Second time it hit a timeout when rebuilding the server. Looking at the periodic jobs, this looks like the compute job has been failing with the same error on zed for quite some time. Definitely something to investigate, but also clearly nothing related to your patch. Let's merge it.

till:
Thank you!

Issue #2969, opened 2024-03-06
Provider methods return `context canceled` error when using `ProviderClient.Context`


MichaelEischer:
Gophercloud 1.10 introduced context aware functions.

For requests sent by `ProviderClient.doRequest` the `ProviderClient.Context` field and the passed in `ctx` are merged:

https://github.com/gophercloud/gophercloud/blob/5b44c45ec947301d531178302c4cecc0a923ad96/provider_client.go#L401-L405

However, that merged context is also canceled using defer. This prevents reading from the `Body` of the returned `http.Response`. Thus when using `ProviderClient.Context`, every method that needs the request result currently fails with a context canceled error.

pierreprinetti:
Thank you, that looks pretty serious. I'm having a look right now.

pierreprinetti:
For lack of a better solution, in the attached PR I propose removing Context from v1.

If anyone has a better idea, please come forward.

pierreprinetti:
Note that v1.9 is also affected.

pierreprinetti:
Fixed in https://github.com/gophercloud/gophercloud/pull/2971. Any feedback is very welcome.

PR #2970, opened 2024-03-06
Remove context from the Provider client

pierreprinetti:
Now that Gophercloud has per-call context, the `context.Context` in the Provider client is redundant and difficult to get right.

This patch removes support for attaching a context to the Provider client.

See also https://github.com/gophercloud/gophercloud/issues/2969


dtantsur (approved):

mandre (approved):
LGTM

PR #2971, opened 2024-03-06
[v1] Remove context

pierreprinetti:
When the Provider client contains a non-nil Context, the [`doRequest` method](https://github.com/gophercloud/gophercloud/blob/5b44c45ec947301d531178302c4cecc0a923ad96/provider_client.go#L401-L405) of the Provider client generated a new context by merging the Provider client's and the per-call context. However the new context is too short-lived to let the callers consume a response body.

This version reverts the inclusion of Context in the v1 branch. The retraction of Go versions is only available in Go v1.16+, so I can't mark the version as retracted; **Gophercloud v1.10 should nonetheless NOT be used**.

Fixes https://github.com/gophercloud/gophercloud/issues/2969


mandre:
For context (haha), this effectively reverts:
- https://github.com/gophercloud/gophercloud/pull/2893
- https://github.com/gophercloud/gophercloud/pull/2894
- https://github.com/gophercloud/gophercloud/pull/2899
- https://github.com/gophercloud/gophercloud/pull/2917
- https://github.com/gophercloud/gophercloud/pull/2933

mandre (approved):
LGTM

Issue #2958, opened 2024-02-29
baremetal: add support for new virtual media API


dtantsur:
Introduced in https://review.opendev.org/c/openstack/ironic/+/894918 (API reference not rendered yet).

PR #2961, opened 2024-03-01
baremetal: fix handling of the "fields" query argument

dtantsur:
Ironic expects a comma-separated string. Add a new feature to
BuildQueryString to choose the format of slices.

Closes: #2960

PR #2974, opened 2024-03-07
baremetal: support for virtual media attach/detach

dtantsur:
Requires:
- https://github.com/gophercloud/gophercloud/pull/2961

Closes: #2958


mandre (commented):
This looks good. Were you able to test this against a real environment?

It would probably be good to add acceptance tests as well with SkipReleasesBelow("master"). Please also add a comment about what version master will represent so that we can use a proper version name in the skip later on.

dtantsur:
> This looks good. Were you able to test this against a real environment?

I was, yes.

Regarding acceptance tests... I need to see what the fake backend does on this API. Probably returns an error.

mandre (approved):
Thanks for adding the acceptance tests :rocket:

PR #2978, opened 2024-03-08
Add bootable filter to v3 cinder list options

### LINKED ISSUE #2977: Cannot filter volumes by bootable flag
Currently there is no way to filter cinder volumes by the "bootable" flag.

This should be added for our use case.

Wieneo:
Fixes #2977

This PR adds the ability to filter volumes by the "bootable" flag, when using Cinder API V3.

### DIFF
--- a/openstack/blockstorage/v3/volumes/requests.go
+++ b/openstack/blockstorage/v3/volumes/requests.go
@@ -149,7 +149,7 @@ type ListOpts struct {
 	Marker string `q:"marker"`
 
 	// Bootable will filter results based on if they are bootable volumes
-	Bootable bool `q:"bootable"`
+	Bootable *bool `q:"bootable,omitempty"`
 }
 
 // ToVolumeListQuery formats a ListOpts into a query string.
--- a/openstack/blockstorage/v3/volumes/requests.go
+++ b/openstack/blockstorage/v3/volumes/requests.go
@@ -147,6 +147,9 @@ type ListOpts struct {
 
 	// The ID of the last-seen item.
 	Marker string `q:"marker"`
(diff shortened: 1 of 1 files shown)


mandre (changes requested):
Thanks for the PR. We want to make this new boolean field a pointer otherwise we're sending `bootable=false` when it's unset.

Could you check that this still works when your openstack is configured without the [bootable resource filter](https://docs.openstack.org/api-ref/block-storage/v3/index.html#resource-filters-resource-filters)? If not, we probably also want to add an omitempty tag to the field so that we're not sending a filter that will confuse OpenStack.

Wieneo:
@mandre 
You are totally right, I made the bool a pointer.

Unfortunately I cannot test this change with an OS environment, without the bootable flag.
I added the requested "omitempty" flag :)

openstack/blockstorage/v3/volumes/requests.go
```diff
@@ -147,6 +147,9 @@ type ListOpts struct {
 
 	// The ID of the last-seen item.
 	Marker string `q:"marker"`
+
+	// Bootable will filter results based on if they are bootable volumes
```
mandre, line 151: Just a small nit
```suggestion
	// Bootable will filter results based on whether they are bootable volumes
```

mandre (commented):
Weirdly enough, @pierreprinetti showed me that the `omitempty` tag seems to [have no effect](https://go.dev/play/p/tsv0yOOowHq). But let's keep it anyway as it shows our intent to drop the field if it's empty.

mandre (approved):

PR #2995, opened 2024-03-14
functional-baremetal: remove dnsmasq-base on all versions

dtantsur:
Zed now fails with a similar error. Let's make sure we start without
dnsmasq-base in all baremetal jobs.


EmilienM (approved):
Thanks

PR #3000, opened 2024-03-20
Update the doc of openstack.AuthOptionsFromEnv function

hex2dec:
If OS_PROJECT_NAME is set, it requires the domain to be set as well to get unique project entity on the identity service (keystone).


pierreprinetti (approved):
Seems fair, thanks

PR #3014, opened 2024-03-26
baremetal: add support for servicing

dtantsur:
Adds provision states and parametes for the new "service" provisioning
action. Adds hold states introduced as part of the same effort.


jayofdoom:
LGTM from an Ironic perspective.

elfosardo:
not enough powers to approve this, but it also looks good to me

dtantsur:
/cc @mandre @mdbooth @pierreprinetti 

Please take a look when you have a minute.

EmilienM (approved):

EmilienM:
> Labels `semver-major`, `semver-minor` and `semver-unknown` block backports to the legacy branch `v1`.

@pierreprinetti should we have that message only if the label `v1` was added here? In this PR, this is just noise IMO.

PR #3030, opened 2024-04-14
[v1] identity: use AuthOptionsBuilder for v2 auth

### LINKED ISSUE #2651: use Identity v2 AuthOptionsBuilder

cardoe:
Rather than creating a tokens2.AuthOptions inside the authentication flow for identity v2, allow any structure that implements the tokens2.AuthBuilderOptions interface to be used. The gophercloud.AuthOptions already implements this interface so there is no change to the API. To avoid any API change, add AuthenticateV2Ext() as a version of AuthenticateV2() accepting the interface. This will allow an out of tree identity v2 auth mechanism to be implemented, just like the v3. fixes #2651. ref #1330.

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -116,7 +116,11 @@ func AuthenticateV2(client *gophercloud.ProviderClient, options gophercloud.Auth
 	return v2auth(client, "", options, eo)
 }
 
-func v2auth(client *gophercloud.ProviderClient, endpoint string, options gophercloud.AuthOptions, eo gophercloud.EndpointOpts) error {
+func AuthenticateV2Ext(client *gophercloud.ProviderClient, options tokens2.AuthOptionsBuilder, eo gophercloud.EndpointOpts) error {
+	return v2auth(client, "", options, eo)
+}
+
+func v2auth(client *gophercloud.ProviderClient, endpoint string, options tokens2.AuthOptionsBuilder, eo gophercloud.EndpointOpts) error {
 	v2Client, err := NewIdentityV2(client, eo)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


cardoe:
I believe this is a compatible implementation of #2655 that is not a major version breaking change. The `CanReauth()` call is probably the better approach but that's a major break which I've left in #2655.

cardoe:
I've implemented https://github.com/rackerlabs/goraxauth using this and verified that it works.

EmilienM:
cc @pierreprinetti for double check on this one.

EmilienM (approved):

pierreprinetti:
I'd rather not merge new features into the `v1` branch, now that `v2` is almost out.

PR #3035, opened 2024-04-22
Add ci jobs for openstack caracal

nikParasyr:
Openstack caracal has been released. Update functional workflows to add jobs that cover it


nikParasyr:
6 failing jobs but only senlin caracal seems to be due to this change as senlin doesnt have a `2024.1` branch (https://opendev.org/openstack/senlin)

EmilienM (commented):
Thanks
LGTM

PR #3040, opened 2024-05-06
clouds: Fix the clouds.yaml locations

### LINKED ISSUE #3038: clouds.Parse() ignores the local clouds.yaml
The new function `clouds.Parse()` picks the clouds.yaml locations in a reverse order, that is:
1. `/etc/openstack/clouds.yaml`
1. `~/.config/openstack/clouds.yaml`
1. `./clouds.yaml`

CC @itzikb-redhat

pierreprinetti:
Before this patch, all default locations were scanned for a `clouds.yaml` even when a match was found. The last found `clouds.yaml` replaced any previous finding, effectively reversing the order of priority for the locations.

Fixes #3038

### DIFF
--- a/openstack/config/clouds/clouds.go
+++ b/openstack/config/clouds/clouds.go
@@ -22,7 +22,6 @@ package clouds
 import (
 	"crypto/tls"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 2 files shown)


pierreprinetti:
I am also removing the check on `*os.PathError` because according to the docs, [that's the only error that can ever come from `os.Open()`](https://pkg.go.dev/os#Open).

openstack/config/clouds/clouds.go
```diff
@@ -90,27 +89,22 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
 		}
 
 		for _, cloudsPath := range options.locations {
-			var errNotFound *os.PathError
 			f, err := os.Open(cloudsPath)
-			if err != nil && !errors.As(err, &errNotFound) {
-				return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, fmt.Errorf("failed to open %q: %w", cloudsPath, err)
+			if err != nil {
```
EmilienM, line 93: random question (mostly to fulfill my curiosity), why didn't you do:

```golang
f, _ := os.Open(cloudsPath)
defer f.Close()
options.cloudsyamlReader = f
```

EmilienM: Nevermind my comment, I misread the next lines.

EmilienM (approved):
LGTM

PR #3041, opened 2024-05-06
clouds: Fix secure.yaml parsing

### LINKED ISSUE #3039: clouds.Parse() fails to parse secure.yaml
The contents of a `secure.yaml` are never merged with the data from `clouds.yaml`.

pierreprinetti:
Before this patch, no `secure.yaml` could ever be found by the Parse function, because a programming error caused Gophercloud to always look at a path that doesn't exist.

Fixes #3039


EmilienM (approved):
LGTM
Thanks for the test case.

PR #3067, opened 2024-06-05
loadbalancer pools: Simplify types and allow extension

pierreprinetti:
With this change, `pools.CreateMemberOpts` and
`pools.BatchUpdateMemberOpts` have been merged into a single struct called `MemberOpts` that implements both interfaces. Alias types have been set up to facilitate migration. As a consequence, those can be used interchangeably.

`(pools.CreateOpts).Members` and `(pools.CreateOpts).Monitor` are now interfaces (`pools.CreateMemberOptsBuilder` and
`monitors.CreateOptsBuilder`) rather than concrete types. This means that when using `pools.CreateMemberOpts`, its type must be explicitly set in the literal:

```go
opts := pools.CreateOpts{
    Members: pools.CreateMemberOptsBuilder{
        []pools.CreateMemberOpts{
            pools.MemberOpts{
                Name: "lb-member-1",
            },
        },
    }
}
```

### DIFF
--- a/docs/MIGRATING.md
+++ b/docs/MIGRATING.md
@@ -132,6 +132,21 @@ The `objectstorage/v1/objects.Copy#Destination` field must be in the form
 `/container/object`. The function will reject a destination path if it doesn't
 start with a slash (`/`).
 
+### Load balancer pools
+
+`(pools.CreateOpts).Members` and `(pools.CreateOpts).Monitor` are now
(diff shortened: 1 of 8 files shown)


pierreprinetti:
This makes no sense. We want to be able to not update a pool (or pool member) name, that's why there was a pointer.

PR #3074, opened 2024-06-26
add migration docs for error handling changes

### LINKED ISSUE #3062: docs/MIGRATING.md: Describe how to replace objectResult.Err.(gophercloud.ErrDefault404)
cf. https://github.com/gophercloud/gophercloud/pull/2904

majewsky:
Fixes #3062.

This is just the same instructions that I already provided in #2904, but which evidently were forgotten about.

### DIFF
--- a/docs/MIGRATING.md
+++ b/docs/MIGRATING.md
@@ -63,6 +63,29 @@ defer cancel()
 err = attachments.WaitForStatus(ctx, client, attachment.ID, "attached")
 ```
 
+### Error handling
+
+The error types for specific response codes (`ErrDefault400`, `ErrDefault401`, etc.) have been removed.
+All unexpected response codes will now return `ErrUnexpectedResponseCode` instead.
+For quickly checking whether a request resulted in a specific response code, use the new `ResponseCodeIs` function:
+
+```go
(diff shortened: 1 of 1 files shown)


pierreprinetti:
I was recently thinking about this exactly after [having some trouble with handling errors](https://kubernetes.slack.com/archives/C05G4NJ6P6X/p1719321166760059) coming from both gophercloud/gophercloud and gophercloud/utils.

The TL;DR is: util functions like `IDFromName` return `gophercloud.ErrResourceNotFound` rather than `gophercloud.ErrUnexpectedResponseCode`. Do you have a moment to address that case in the migration docs? That would be something like "if using the utils library, to check for the NotFound case you still must run the type assertion on `gophercloud.ErrResourceNotFound`" or something like that. What do you think?

Otherwise, I'll do that as a follow-up.

majewsky:
I'm not as familiar with these parts of the API. But I believe I have "allow commits from maintainers" on, so you can extend this PR as necessary.

pierreprinetti (approved):

pierreprinetti:
No worries, I'll take care of that. Thank you again for your contribution!

PR #3077, opened 2024-06-29
[WIP] Methods

pierreprinetti:


stephenfin:
I've closed this as it's seriously in conflict and hasn't been updated in over a year. It's a good idea but I suspect it would need to be recreated at this point.

PR #3085, opened 2024-07-05
MIGRATING.md: Provide a migration script

pierreprinetti:
Some steps in the migration from Gophercloud v1 to v2 can be automated. Provide a script that can take out some of the pain.

### Expected outcome

After running the script, your code will likely not compile. The idea is that at this point, you're only left with a few changes that can't reasonably be automated.

### What it does
* Add /v2 to all Gophercloud imports, except to the packages that have been removed without replacement
* Adjust the import path of moved packages
* Adjust the package identifier in the code where possible
* Add context.TODO() where required

### Limitations:
* doesn't fix the use of removed extensions. For example, if you used `openstack/blockstorage/extensions/availabilityzones`, you will have to manually put that back into e.g. `servers.CreateOpts`
* it will only put `context.TODO()` where a context is required to satisfy the function signature. It's up to you to actually replace that with a variable that you reuse for proper cancellation

---

Is this useful?

### DIFF
--- a/docs/MIGRATING.md
+++ b/docs/MIGRATING.md
@@ -493,3 +493,137 @@ Users that still rely on theses old services should continue using Gophercloud v
   (`openstack/networking/v2/extensions/fwaas`)
 - Poppy (CDNaaS) service (`openstack/cdn`)
 - Senlin (Clustering) service (`openstack/clustering`)
+
+### Script-assisted migration
+
+#### Expected outcome
+
+After running the script, your code may not compile. The idea is that at this point, you're only left with a few changes that can't reasonably be automated.
+
+#### What it does
+
+* Add `/v2` to all Gophercloud imports, except to the packages that have been removed without replacement
+* Adjust the import path of moved packages
+* Adjust the package identifier in the code where possible
+* Add `context.TODO()` where required
+
+#### Limitations
+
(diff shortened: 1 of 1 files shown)


pierreprinetti:
One known issue:
* for some reason this import disappears: `github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/fwaas/firewalls`

pierreprinetti:
@kayrus you said you used this script for https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1747. Do you have any comments? Was it useful? How could it be better in your opinion?

EmilienM (approved):

docs/MIGRATING.md
```diff
@@ -493,3 +493,137 @@ Users that still rely on theses old services should continue using Gophercloud v
...
+        s#\(\.AllPages(\)#\1context.TODO(), #g
+        s#\(\.EachPage(\)\(func(\)#\1context.TODO(), \2ctx, #g
```
kayrus, line 609:
```suggestion
        s#\(\.EachPage(\)\(func(\)#\1context.TODO(), \2ctx context.Context, #g
```

pierreprinetti:
That you for the reviews! I have applied all your suggestions.

kayrus (approved):
Thank you as well :)

PR #3100, opened 2024-07-15
[tests]: bump golang to 1.22

kayrus:
Since the go.mod and CHANGELOG.md already mentioned golang 1.22, we need to change the go version in tests as well.


EmilienM (approved):

mandre (approved):
Yep, makes sense. Thanks.

PR #3101, opened 2024-07-15
[gophercloud v2]: introduce v2 branch with a backport functionality

kayrus:
Introduce the logic to backport master changes to v2 branch.

### DIFF
--- a/.github/labels.yaml
+++ b/.github/labels.yaml
@@ -1,6 +1,9 @@
 - color: '30ABB9'
   description: This PR will be backported to v1
   name: backport-v1
+- color: 'E99695'
+  description: This PR will be backported to v2
+  name: backport-v2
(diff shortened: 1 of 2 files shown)


EmilienM (approved):

mandre (changes requested):
This takes care of the automation once we're setting a `backport-v2` label to a PR, and is identical to the `backport_v1` workflow except for the version. Good.

But we'll also need a `backport-v2` label that you can define in https://github.com/gophercloud/gophercloud/blob/master/.github/labels.yaml. Can you add it in the same PR?

We'll need to create a v2 branch, that must be a master right before we merged the first API breaking change. I'll take care of creating this `v2` branch.

mandre (approved):

PR #3115, opened 2024-07-17
actions: Add a label to hold merging

pierreprinetti:
Create the label 'hold' and add a check that fails when the PR has that label.

### DIFF
--- a/.github/labels.yaml
+++ b/.github/labels.yaml
@@ -25,3 +25,6 @@
 - color: 'EC0101'
   description: Unable to figure out the semver type
   name: semver:unknown
+- color: 'D73A4A'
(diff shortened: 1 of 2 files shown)


.github/labels.yaml
```diff
@@ -25,3 +25,6 @@
 - color: 'EC0101'
   description: Unable to figure out the semver type
   name: semver:unknown
+- color: '#D73A4A'
```
kayrus, line 28:
```suggestion
- color: 'D73A4A'
```

pierreprinetti: good catch!

kayrus (approved):

PR #3116, opened 2024-07-17
Block `semver:minor` backports to v1

pierreprinetti:
With `v2` out, we are freezing `v1` to only receive bug fixes.

Backports to `v1` can be forced by manually creating a PR against `v1`.


pierreprinetti:
@EmilienM @mandre @kayrus @stephenfin PTAL, this is just a proposal.

EmilienM (approved):

PR #2811, opened 2023-10-13
Make fixtures part of tests

mandre:
By renaming the fixtures to `fixtures_test`, we mark them as test, and no longer part of the public API.

Some files were left out as they were included in other tests:

    $ grep "/testing" -R openstack | cut -d " " -f 2 | sort | uniq
    "github.com/gophercloud/gophercloud/openstack/baremetalintrospection/v1/introspection/testing"
    "github.com/gophercloud/gophercloud/openstack/baremetal/inventory/testing"
    "github.com/gophercloud/gophercloud/openstack/common/extensions/testing"
    "github.com/gophercloud/gophercloud/openstack/identity/v3/tokens/testing"
    "github.com/gophercloud/gophercloud/openstack/networking/v2/extensions/layer3/floatingips/testing"
    "github.com/gophercloud/gophercloud/openstack/networking/v2/networks/testing"
    "github.com/gophercloud/gophercloud/openstack/networking/v2/ports/testing"
    "github.com/gophercloud/gophercloud/openstack/objectstorage/v1/accounts/testing"
    "github.com/gophercloud/gophercloud/openstack/objectstorage/v1/containers/testing"

This should prevent `go-apidiff` from complaining when modifying fixtures.

PR #3118, opened 2024-07-17
Make fixtures part of tests (redux)

stephenfin:
In commit c71fc9d7c2b95b8de308f34f2927203a7eac1e25, we renamed 'fixtures' to 'fixtures_test' to avoid them getting picked up as part of the public API by tools like go-apidiff. Unfortunately a few more of these have crept in in the meantime. Resolve them now.

    for f in $(find . -name fixtures.go); do
        git mv -- "$f" "$(dirname $f)/fixtures_test.go"
    done


EmilienM (approved):

mandre (changes requested):
This is not going to work for reasons explained in https://github.com/gophercloud/gophercloud/pull/2811. These files are imported in other tests and we need to keep them public.

PR #3092, opened 2024-07-09
[neutron]: introduce Stateful argument for the security groups

kayrus:
Prior to starting a PR, please make sure you have read our
[contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial).

Prior to a PR being reviewed, there needs to be a Github issue that the PR
addresses. Replace the brackets and text below with that issue number.

Fixes #3091 

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:

https://github.com/openstack/neutron-lib/blob/47e1736c10ffd6f1850eba9970acdc00373f9de7/neutron_lib/api/definitions/stateful_security_group.py#L36-L42

PR #3119, opened 2024-07-17
[v2] [neutron]: introduce Stateful argument for the security groups

### LINKED ISSUE #3091: Add stateful argument to security groups
https://docs.openstack.org/api-ref/network/v2/?expanded=create-port-detail#create-security-group

required by https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1735

kayrus:

Fixes #3091
Backport of the #3092 without changing the List function signature

code in this PR:

[PUT URLS HERE]

### DIFF
--- a/internal/acceptance/openstack/networking/v2/extensions/security_test.go
+++ b/internal/acceptance/openstack/networking/v2/extensions/security_test.go
@@ -20,6 +20,7 @@ func TestSecurityGroupsCreateUpdateDelete(t *testing.T) {
 	group, err := CreateSecurityGroup(t, client)
 	th.AssertNoErr(t, err)
 	defer DeleteSecurityGroup(t, client, group.ID)
+	th.AssertEquals(t, group.Stateful, true)
 
 	rule, err := CreateSecurityGroupRule(t, client, group.ID)
 	th.AssertNoErr(t, err)
@@ -32,6 +33,7 @@ func TestSecurityGroupsCreateUpdateDelete(t *testing.T) {
 	updateOpts := groups.UpdateOpts{
 		Name:        name,
 		Description: &description,
+		Stateful:    new(bool),
 	}
 
 	newGroup, err := groups.Update(context.TODO(), client, group.ID, updateOpts).Extract()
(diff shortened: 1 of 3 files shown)


stephenfin:
This requires the [`stateful-security-group` extension](https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/stateful_security_group.py). Do we want to document this in the help text? On a lesser note, do we want to check for the extension first in the functional test, or do we assume OVS/OVN?

kayrus:
@stephenfin looks like the initial PR was incorrect, and I had to put this option into the `openstack/networking/v2/extensions`. So what's the plan? Fix the master branch, then update v1 and v2 backports?

kayrus:
Hm, looking at the `allowed_address_pairs` for ports. It's also an extension, but it resides in the `openstack/networking/v2/ports`.

stephenfin:
> Hm, looking at the `allowed_address_pairs` for ports. It's also an extension, but it resides in the `openstack/networking/v2/ports`.

Yeah, I suspect there are quite a few cases of this: the beauties of having extensions that can change absolutely anything about the API :sweat_smile: At least `CreateOpts.stateful` is a pointer so someone can omit it entirely during creation if the networking backend doesn't support it. That's not the case for the result struct though :disappointed:

Perhaps we could just document that it relies on the extension and will be unset if that extension is absent? afaict it will always be present on the OVN, OVS and linuxbridge backends, but that's not necessarily true for other backends.

stephenfin:
I also missed the fact that this was a backport, so feel free to ignore my comments entirely here. It would be good to fix this on master though, IMO.

kayrus:
v1 backport was already approved, can we merge this so we can close this topic until v3?

stephenfin (approved):

kayrus:
@stephenfin thanks. I added comments about the `stateful` flag in TF provider docs: https://github.com/terraform-provider-openstack/terraform-provider-openstack/blob/fa1184eb117d93b42b24bfc2019e62b1689f214e/docs/resources/networking_secgroup_v2.md

PR #3120, opened 2024-07-17
tests: Remove AZ add/remove host integration test

stephenfin:
Go runs tests in parallel for different packages in parallel. The AZ add/remove host test was adding and removing hosts from an AZ and if this ran at the same time as another test that created a server then the host addition or removal would fail since Nova forbids it. There does not appear to be any way to resolve this. We could run all integration tests serially instead, but this would be a backwards step that will massively increase test run time. We could also rewrite the test to only add/remove an unused host, but that would require us to fence off this compute node in all other tests using non-default parameters to server create. Given the lack of better options, we simply remove the test and assume our testing to date on this method has been good enough and unit tests will be enough to carry us forward.


mandre:
I'm not at all convinced these issues are caused by parallelization. As far as I understand, we do not run the tests in parallel -- we don't use `t.Parallel()` anywhere. Have you seen this test cause issues?

PR #3128, opened 2024-07-18
actions: Refactor the semver assessment

pierreprinetti:
Using `gh` lets us drop two external dependencies, and condensate the steps a bit.

After this change:
* new pushes in the PR won't trigger relabeling if the assessment is unchanged
* failures in the workflow will be notified as comments in the PR rather than via the `semver:unknown` label

The next natural step is to set this job as mandatory, to that a failure to assess semver will block merging.

### DIFF
--- a/.github/labels.yaml
+++ b/.github/labels.yaml
@@ -22,9 +22,6 @@
 - color: '6E7624'
   description: No API change
   name: semver:patch
-- color: 'EC0101'
-  description: Unable to figure out the semver type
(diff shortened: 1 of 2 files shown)


EmilienM:
could you test it somewhere first?

EmilienM (approved):

pierreprinetti:
failure to backport to v1 is expected, as labels.yaml is not available there.

PR #3131, opened 2024-07-18
Align ServiceFail provisioning state value with Ironic

### LINKED ISSUE #3130: ServiceFail provisioning state value doesn't match the value used by Ironic
There is a mismatch between ServiceFail provisioning state value used by Gophercloud [1] and
Ironic [2]. Ironic uses "service fail**ed**" value, rather than "service fail".

[1] https://github.com/gophercloud/gophercloud/blob/v2.0.0/openstack/baremetal/v1/nodes/requests.go#L53
[2] https://github.com/openstack/ironic/blob/stable/2024.1/ironic/common/states.py#L248

pszczerbik:
There is a mismatch between ServiceFail values used by Gophercloud and Ironic. Ironic uses "service failed" value [1], rather than "service fail". This commit addresses this discrepancy.

Fixes #3130 

[1] https://github.com/openstack/ironic/blob/stable/2024.1/ironic/common/states.py#L248

### DIFF
--- a/openstack/baremetal/v1/nodes/requests.go
+++ b/openstack/baremetal/v1/nodes/requests.go
@@ -50,7 +50,7 @@ const (
 	Unrescuing   ProvisionState = "unrescuing"
 	Servicing    ProvisionState = "servicing"
 	ServiceWait  ProvisionState = "service wait"
-	ServiceFail  ProvisionState = "service fail"
+	ServiceFail  ProvisionState = "service failed"
 	ServiceHold  ProvisionState = "service hold"
 )
 


pszczerbik:
FYI @dtantsur

dtantsur (approved):
LGTM, thanks. The "major" semver tag is quite incorrect IMO.

pierreprinetti (approved):
Thank you!

pierreprinetti:
I consider this to be a patch, however the [support for servicing](https://github.com/gophercloud/gophercloud/pull/3014) was never backported to v1, so this patch only applies to v2.

PR #3133, opened 2024-07-19
actions: Label PRs based on the paths of files being changed

pierreprinetti:
To facilitate PR assignment and ultimately reviews, label the PRs based on what they touch.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,144 @@
+edit:dependencies:
+- changed-files:
+  - any-glob-to-any-file:
+    - 'go.mod'
+    - 'go.sum'
+    - '.github/dependabot.yml'
+edit:actions:
+- changed-files:
+  - any-glob-to-any-file:
+    - '.github/**'
+edit:gophercloud:
+- changed-files:
+  - any-glob-to-any-file:
+    - '*.go'
+    - 'testing/**'
+    - 'pagination/**'
+edit:openstack:
+- changed-files:
+  - any-glob-to-any-file:
+    - 'openstack/*'
+    - 'openstack/testing/**'
+edit:baremetal:
+- changed-files:
+  - any-glob-to-any-file:
+    - 'openstack/baremetal/**'
+    - 'internal/acceptance/openstack/baremetal/**'
+edit:baremetalintrospection:
+- changed-files:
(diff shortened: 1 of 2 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,144 @@
...
+- changed-files:
+  - any-glob-to-any-file:
+    - '*.go'
+    - 'testing/**'
+    - 'pagination/**'
+edit:openstack:
+- changed-files:
+  - any-glob-to-any-file:
+    - 'openstack/*'
+    - 'openstack/testing/**'
```
stephenfin, line 21: I don't see label definitions for these. I'm also not sure how useful they are (they other labels below serve a similar purpose). Did you mean to include these?

stephenfin (changes requested):

.github/labeler.yml
pierreprinetti: Thanks, I added these labels to `labels.yaml`. I didn't fully get your point about them being less useful than other labels. Do you mind explaining to me again?

stephenfin: We have filters for all the subdirectories of `openstack`, so if you make a change to `openstack/compute` you will get two labels, `openstack` and `compute`. The former seems redundant. Is it possible to scope this filter only to `.go` files in `openstack` and `openstack/testing` and not subdirectories of same. That might make it more useful.

pierreprinetti: Unless I made a mistake, `edit:openstack` will only be applied to files in the immediate directory `openstack/*` (one star), whereas a change in `openstack/compute/**` (two stars, meaning "any subdirectory) will only trigger `edit:compute` ([docs](https://github.com/actions/labeler#match-object), [example](https://github.com/actions/labeler#basic-examples))

stephenfin: Aha. Do we want to do the same for `openstack/testing/**` in that case (i.e. change to `openstack/testing/*`)?

stephenfin (approved):

pierreprinetti:
backport to v2 requires https://github.com/gophercloud/gophercloud/pull/3146

This patch changes the name of the go-apidiff check. which means that it must be backported, so that the same job name can be enforced across all branches

Issue #2309, opened 2022-01-07
Add options any 


HappyFX:
Add option "Any" in security group rules, according the [documentaion](https://docs.openstack.org/python-openstackclient/latest/cli/command-objects/security-group-rule.html#cmdoption-openstack-security-group-rule-create-protocol):

```bash
Network version 2:
IP protocol (ah, dccp, egp, esp, gre, icmp, igmp, ipv6-encap, ipv6-frag, ipv6-icmp, ipv6-nonxt, ipv6-opts, ipv6-route, ospf, pgm, rsvp, sctp, tcp, udp, udplite, vrrp and integer representations [0-255] or any; default: any (all protocols))
```

PR #3157, opened 2024-07-24
fix: create security group rule with any protocol

### LINKED ISSUE #2442: Fix name in protocolAny
In [documentation](https://docs.openstack.org/python-openstackclient/latest/cli/command-objects/security-group-rule.html#cmdoption-openstack-security-group-rule-create-protocol): 
valid value for openstack cli is 'any'. In fact it's converted in to `None` and then in `null`:

debug from openstack call:

`openstack security group rule create --egress --remote-ip 8.8.8.8/32 --protocol 'any' eece1c60-baf3-4ea9-88bf-54cc40290720`

`REQ: curl -g -i -X POST http://test:9696/v2.0/security-group-rules -H "Content-Type: application/json" -H "User-Agent: openstacksdk/0.99.0 keystoneauth1/4.6.0 python-requests/2.28.1 CPython/3.9.13" -H "X-Auth-Token: {SHA256}547fc1809352d6d9431590520fb0923bbf92d25fbe73af21eb488e1b71e67e52" -d '{"security_group_rule": {"security_group_id": "eece1c60-baf3-4ea9-88bf-54cc40290720", "ethertype": "IPv4", "protocol": null, "direction": "egress", "remote_ip_prefix": "8.8.8.8/32"}}'`

kmlebedev:
Fixes #2442 after #2309

### DIFF
--- a/openstack/networking/v2/extensions/security/rules/requests.go
+++ b/openstack/networking/v2/extensions/security/rules/requests.go
@@ -2,8 +2,6 @@ package rules
 
 import (
 	"context"
-	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/fwaas_v2/rules"
-
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/pagination"
 )
@@ -78,7 +76,7 @@ const (
 	ProtocolUDP       RuleProtocol  = "udp"
 	ProtocolUDPLite   RuleProtocol  = "udplite"
 	ProtocolVRRP      RuleProtocol  = "vrrp"
-	ProtocolAny       RuleProtocol  = "any"
+	ProtocolAny       RuleProtocol  = ""
 )
 
 // CreateOptsBuilder allows extensions to add additional parameters to the
@@ -135,14 +133,7 @@ type CreateOpts struct {
 
 // ToSecGroupRuleCreateMap builds a request body from CreateOpts.
 func (opts CreateOpts) ToSecGroupRuleCreateMap() (map[string]any, error) {
-	b, err := gophercloud.BuildRequestBody(opts, "security_group_rule")
-	if err != nil {
-		return nil, err
-	}
-	if m := b["security_group_rule"].(map[string]any); m["protocol"] == string(rules.ProtocolAny) {
-		m["protocol"] = nil
-	}
-	return b, err
+	return gophercloud.BuildRequestBody(opts, "security_group_rule")
 }
 
 // Create is an operation which adds a new security group rule and associates it
--- a/openstack/networking/v2/extensions/security/rules/requests.go
+++ b/openstack/networking/v2/extensions/security/rules/requests.go
@@ -2,6 +2,7 @@ package rules
 
 import (
 	"context"
+
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/pagination"
 )
--- a/openstack/networking/v2/extensions/security/rules/requests.go
+++ b/openstack/networking/v2/extensions/security/rules/requests.go
@@ -2,6 +2,7 @@ package rules
 
 import (
 	"context"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/fwaas_v2/rules"
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/pagination"
@@ -138,10 +139,8 @@ func (opts CreateOpts) ToSecGroupRuleCreateMap() (map[string]any, error) {
 	if err != nil {
 		return nil, err
 	}
-	if m, mOk := b["security_group_rule"].(map[string]any); mOk {
-		if p, ok := m["protocol"]; ok && p == string(ProtocolAny) {
-			m["protocol"] = nil
-		}
+	if m := b["security_group_rule"].(map[string]any); m["protocol"] == string(rules.ProtocolAny) {
+		m["protocol"] = nil
 	}
 	return b, err
 }
--- a/openstack/networking/v2/extensions/security/rules/requests.go
+++ b/openstack/networking/v2/extensions/security/rules/requests.go
@@ -146,7 +146,7 @@ func Create(ctx context.Context, c *gophercloud.ServiceClient, opts CreateOptsBu
 		return
 	}
 	if m, mOk := b["security_group_rule"].(map[string]any); mOk {
-		if p, ok := m["protocol"]; ok && p == "any" {
+		if p, ok := m["protocol"]; ok && p == string(ProtocolAny) {
 			m["protocol"] = nil
 		}
 	}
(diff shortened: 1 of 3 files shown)


pierreprinetti:
I'm afraid that I'm missing something here. What is the problem that this patch is solving?

A minimal reproducer in code would be great.

/hold

kmlebedev:
> I'm afraid that I'm missing something here. What is the problem that this patch is solving?
> 
> A minimal reproducer in code would be great.
> 
> /hold

I did the same https://github.com/gophercloud/gophercloud/blob/master/openstack/networking/v2/extensions/fwaas_v2/rules/requests.go#L141

openstack/networking/v2/extensions/security/rules/requests.go
```diff
@@ -145,6 +145,11 @@ func Create(ctx context.Context, c *gophercloud.ServiceClient, opts CreateOptsBu
 		r.Err = err
 		return
 	}
+	if m, mOk := b["security_group_rule"].(map[string]any); mOk {
```
kayrus, line 148: it'd be better to add this into the `ToSecRuleCreateMap`

kmlebedev: done

openstack/networking/v2/extensions/security/rules/requests.go
```diff
@@ -134,7 +134,16 @@ type CreateOpts struct {
...
-	return gophercloud.BuildRequestBody(opts, "security_group_rule")
+	b, err := gophercloud.BuildRequestBody(opts, "security_group_rule")
+	if err != nil {
+		return nil, err
+	}
+	if m, mOk := b["security_group_rule"].(map[string]any); mOk {
+		if p, ok := m["protocol"]; ok && p == string(ProtocolAny) {
+			m["protocol"] = nil
+		}
+	}
```
kayrus, line 145:
```suggestion
	if m, ok := b["security_group_rule"].(map[string]interface{}); ok && m["protocol"] == string(rules.ProtocolAny) {
		m["protocol"] = nil
	}
```

kayrus:
There is another issue. When you retrieve the rule, the protocol is a string, and for `any` it will be an empty string. External code would need to make some additional checks.
```go
        if sgRule.Protocol == "" {
                sgRule.Protocol = string(rules.ProtocolAny)
        }
```

openstack/networking/v2/extensions/security/rules/requests.go
kmlebedev: done

kmlebedev:
> There is another issue. When you retrieve the rule, the protocol is a string, and for `any` it will be an empty string. External code would need to make some additional checks.
> 
> ```go
>         if sgRule.Protocol == "" {
>                 sgRule.Protocol = string(rules.ProtocolAny)
>         }
> ```
> 
> Maybe it makes sense to change this directly in gophercloud.
> 
> see an example [terraform-provider-openstack/terraform-provider-openstack#1765 (comment)](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1765#issuecomment-2248677701)


@kayrus thank you for helping. Done

kayrus:
Terraform supports integer value for the protocol. What if an end user provides "0"? I think we also need to convert it to null.

kmlebedev:
> Terraform supports integer value for the protocol. What if an end user provides "0"? I think we also need to convert it to null.

I didn't change the value 0.
Just made the register comparison a little easier.
```
len(s.SecGroupRule.Protocol) == 0
```
eq
```
s.SecGroupRule.Protocol) == ""
```

kayrus:
@kmlebedev currently I'm testing corner edges with terraform provider. Just give me some time to post the results of my findings.

kayrus:
@kmlebedev and @HappyFX you both opened a can of worms.

It's possible to set the `protocol` to `"0"` and this will not work. There is also a way to set it to `0` and I this will also cause `"0"` in POST response. I'd like a confirmation that the `"0"` protocol doesn't work and we need to ensure that any `0`, `""` or `"0"` are detected converted to `null`.

kayrus:
There is also missing `hopopt`, `ip` and `icmpv6` protocols

pierreprinetti:
I know I may sound like a broken record, but... what is the problem that this patch is solving?

pierreprinetti:
For anyone following the conversation:
* [this](https://github.com/gophercloud/gophercloud/issues/2442#issuecomment-2249711253) is the problem that this patch is attempting to solve
* [this](https://github.com/gophercloud/gophercloud/issues/2442#issuecomment-2249711253) is why we might not need this patch (please advise)

pierreprinetti:
After some discussion in , I believe that the fix here should be to change `ProtocolAny` to `""` , so that Gophercloud omits the `protocol` property from the payload and Neutron correctly sets `None`.

I think that this patch should really be:

```diff
diff --git a/openstack/networking/v2/extensions/security/rules/requests.go b/openstack/networking/v2/extensions/security/rules/requests.go
index 78b67b0d..f2ea9f2d 100644
--- a/openstack/networking/v2/extensions/security/rules/requests.go
+++ b/openstack/networking/v2/extensions/security/rules/requests.go
@@ -77,7 +77,7 @@ const (
 	ProtocolUDP       RuleProtocol  = "udp"
 	ProtocolUDPLite   RuleProtocol  = "udplite"
 	ProtocolVRRP      RuleProtocol  = "vrrp"
-	ProtocolAny       RuleProtocol  = "any"
+	ProtocolAny       RuleProtocol  = ""
 )
 
 // CreateOptsBuilder allows extensions to add additional parameters to the
```

What do you think?

pierreprinetti:
Or maybe that should be `"0"` instead of `""`, so that the same value can also be used as a filter in `List`.

HappyFX:
> What do you think?

Empty looks correct if it will omitted, still neutron API will response as `"protocol": null` and how handle it? 

> Or maybe that should be "0"

"0" is act as protocol set with "0x00" HOPOPT - it's will be not expecting filter (not "what ever protocol") security group in openstack.

pierreprinetti:
@kmlebedev Are you OK with changing this PR to be something like [this](https://github.com/gophercloud/gophercloud/pull/3157#issuecomment-2250639959)? Would that fit your use-case?

kmlebedev:
> @kmlebedev Are you OK with changing this PR to be something like [this](https://github.com/gophercloud/gophercloud/pull/3157#issuecomment-2250639959)? Would that fit your use-case?

@pierreprinetti Done

openstack/networking/v2/extensions/security/rules/requests.go
```diff
@@ -2,7 +2,6 @@ package rules
 
 import (
 	"context"
-
```
kayrus, line 5: can you bring back an empty newline here?

kayrus:
I updated a corresponding https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1765

openstack/networking/v2/extensions/security/rules/requests.go
kmlebedev: > can you bring back an empty newline here?

done

kayrus (approved):
LGTM, but I'm confused with the `semver:major` label.

pierreprinetti:
`label:semver` is technically correct because we are changing an exported value. However, no valid behaviour is going to be impacted by this change since `"any"` has never worked to the best of my knowledge.

Thank you

PR #3167, opened 2024-08-30
sharedfilesystems/schedulerstats: fix ListOpts, ListDetailOpts

### LINKED ISSUE #3166: Manila API: schedulerstats.ListOpts are completely broken
There are multiple issues with the `schedulerstats.{ListOpts,ListDetailOpts}` types that make them completely unusable.

- All fields are marked as `json:` when it should be `q:`.
- Several fields are using incorrect names (e.g. `host_name` instead of `host`) based on an incorrect reading of the API spec.
- The field `project_id` does not exist. It too is based on an incorrect reading of the API spec.

Spec: https://docs.openstack.org/api-ref/shared-file-system/#scheduler-stats-storage-pools

majewsky:
Ref: <https://docs.openstack.org/api-ref/shared-file-system/#scheduler-stats-storage-pools>

Closes #3166.

### DIFF
--- a/openstack/sharedfilesystems/v2/schedulerstats/requests.go
+++ b/openstack/sharedfilesystems/v2/schedulerstats/requests.go
@@ -14,17 +14,15 @@ type ListOptsBuilder interface {
 // ListOpts controls the view of data returned (e.g globally or per project).
 type ListOpts struct {
 	// The pool name for the back end.
-	ProjectID string `json:"project_id,omitempty"`
-	// The pool name for the back end.
-	PoolName string `json:"pool_name"`
+	PoolName string `q:"pool"`
 	// The host name for the back end.
-	HostName string `json:"host_name"`
+	HostName string `q:"host"`
 	// The name of the back end.
-	BackendName string `json:"backend_name"`
+	BackendName string `q:"backend"`
 	// The capabilities for the storage back end.
-	Capabilities string `json:"capabilities"`
+	Capabilities string `q:"capabilities"`
 	// The share type name or UUID. Allows filtering back end pools based on the extra-specs in the share type.
-	ShareType string `json:"share_type,omitempty"`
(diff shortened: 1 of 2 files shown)


kayrus (commented):
This is clearly a bugfix. Thanks for submitting the PR.
I think it makes sense to adjust the struct names as well so they correspond with the Pool struct. As for the `ListDetailOpts`, does it make sense to omit it, since it has the same members as the `ListOpts`?
I also wonder whether it's possible specify multiple capabilities (like [here](https://github.com/gophercloud/gophercloud/blob/18e5f3b351685851fee87d1b439d59303e8b6069/openstack/image/v2/images/requests.go#L77-L78)).

In addition I'd adjust unit tests to cover this change.

mandre (commented):
Ouch. In addition to @kayrus's comments, could you also add acceptance tests to show the filters working?

majewsky:
> I think it makes sense to adjust the struct names as well so they correspond with the Pool struct.

What do you mean?

majewsky:
> > Can you please clarify which renames you are thinking of?
> 
> `PoolName` -> `Pool`, `HostName` -> `Host`, etc.

Done.

> > how was this struct added in the first place?
> 
> unit tests were initially with empty `ListOpts`.

What would even be a useful test here? If a test is added that queries with `ListDetailOpts{Pool: "foo"}` and then checks against a fixture where `?pool=foo` is observed, that's just entirely redundant with the specification of the struct in the type. If this test had existed for the previous implementation, then the fixture would probably have looked for `?pool_name=foo` and just tested the wrong assumption anyway.

winiciusallan:
hey @majewsky, any plans to continue working on this? this still looks a valid gophercloud bug, so it'd be nice if we could address it. if you are not planning, we can try to do it ourselves :). thanks for proposing this!

PR #3171, opened 2024-09-11
Small improvements to v2 migration script

mandre:
These are a couple of small improvements to the automated v2 migration script:
- exclude vendor directory
- try converting all ErrDefaultXXX errors that existed in gophercloud v1 to their net/http equivalent


EmilienM:
Thanks, it was on my list as well!!

PR #3187, opened 2024-09-26
Add label: just-needs-tests

pierreprinetti:
Several pull requests in the backlog are code-complete according to their authors, but lack tests. This label will let us be more aware of the problem.


pierreprinetti:
also the horrible colour I have chosen for the label might help convincing everybody to gather the required energy to add tests

PR #3190, opened 2024-09-27
[core]: allow empty struct member updates

### LINKED ISSUE #1332: JSON marshalling and omitempty for empty strings and bools
Certain gophercloud implementations omit empty strings and false bools.

Examples:
1) [https://github.com/gophercloud/gophercloud/blob/3a7818a07cfc862267a0b03635075a444c92dcda/openstack/loadbalancer/v2/l7policies/requests.go#L212](invert) in loadbalancer/v2/l7policies/requests.go . You can set true to invert, but cannot set false.
2) Similar issue occurs for name and description fields, e.g. you can set non-empty [description](https://github.com/gophercloud/gophercloud/blob/3a7818a07cfc862267a0b03635075a444c92dcda/openstack/loadbalancer/v2/l7policies/requests.go#L162), but you cannot remove it.

ref: https://github.com/golang/go/issues/13284
#1314

kayrus:
Fixes #1332

### DIFF
--- a/internal/acceptance/openstack/compute/v2/aggregates_test.go
+++ b/internal/acceptance/openstack/compute/v2/aggregates_test.go
@@ -11,6 +11,7 @@ import (
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/clients"
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/tools"
+	"github.com/gophercloud/gophercloud/v2/internal/ptr"
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/aggregates"
 	"github.com/gophercloud/gophercloud/v2/openstack/compute/v2/hypervisors"
 	th "github.com/gophercloud/gophercloud/v2/testhelper"
@@ -47,8 +48,8 @@ func TestAggregatesCRUD(t *testing.T) {
 	tools.PrintResource(t, aggregate)
 
 	updateOpts := aggregates.UpdateOpts{
-		Name:             "new_aggregate_name",
-		AvailabilityZone: "new_azone",
+		Name:             ptr.To("new_aggregate_name"),
+		AvailabilityZone: ptr.To("new_azone"),
 	}
 
 	updatedAggregate, err := aggregates.Update(context.TODO(), client, aggregate.ID, updateOpts).Extract()
--- a/internal/acceptance/openstack/compute/v2/secgroup_test.go
+++ b/internal/acceptance/openstack/compute/v2/secgroup_test.go
@@ -48,7 +48,7 @@ func TestSecGroupsCRUD(t *testing.T) {
 	newName := tools.RandomString("secgroup_", 4)
 	description := ""
 	updateOpts := secgroups.UpdateOpts{
-		Name:        newName,
+		Name:        &newName,
 		Description: &description,
 	}
 	updatedSecurityGroup, err := secgroups.Update(context.TODO(), client, securityGroup.ID, updateOpts).Extract()
--- a/internal/ptr/ptr.go
+++ b/internal/ptr/ptr.go
@@ -1,0 +1,5 @@
+package ptr
+
+func To[T any](v T) *T {
+	return &v
+}
--- a/openstack/loadbalancer/v2/flavors/testing/fixtures.go
+++ b/openstack/loadbalancer/v2/flavors/testing/fixtures.go
@@ -44,6 +44,18 @@ const SingleFlavorBody = `
 }
 `
 
+const SingleFlavorDisabledBody = `
+{
+	"flavor": {
+		"id": "5548c807-e6e8-43d7-9ea4-b38d34dd74a0",
+		"name": "Basic",
+		"description": "A basic standalone Octavia load balancer.",
+		"enabled": false,
+		"flavor_profile_id": "9daa2768-74e7-4d13-bf5d-1b8e0dc239e1"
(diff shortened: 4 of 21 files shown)


EmilienM (approved):

EmilienM:
I've approved it but I need an extra pair of eyes on this one, ping @pierreprinetti

pierreprinetti (commented):
I understand the value of changing all types to pointer types in `UpdateOpts` structs while keeping the `omitempty` tag, to make a clear difference between "I want to change this property" and "I don't want to change this property", including the case where the required new value happens to be the zero value for the type (and as a consequence wasn't sent to the server before this patch).

This technically is a superfluous change in some instances. For example, let's take the `servers` package. It's useless to use `*string` in `Name`, because the API refuses the empty string as a name. Same goes for `accessIPv4` and `accessIPv6`. However, I agree with you that this change brings consistency and consistency is valuable.

My comments propose the addition of a convenience generic function that given a value, returns a pointer to it. It may make code more easily readable and reduce the LOC count of this patch. Please consider adding it in an internal package (I propose `internal/ptr`) to use it in both unit and acceptance tests.

openstack/loadbalancer/v2/flavors/testing/fixtures.go
```diff
@@ -119,7 +119,7 @@ func HandleFlavorCreationSuccessfully(t *testing.T, response string) {
 			"flavor": {
 				"name": "Basic",
 				"description": "A basic standalone Octavia load balancer.",
-				"enabled": true,
+				"enabled": false,
```
pierreprinetti, line 122: This can be reverted once you apply my suggestion above

openstack/loadbalancer/v2/flavors/testing/requests_test.go
```diff
@@ -62,7 +62,7 @@ func TestCreateFlavor(t *testing.T) {
 	actual, err := flavors.Create(context.TODO(), fake.ServiceClient(), flavors.CreateOpts{
 		Name:            "Basic",
 		Description:     "A basic standalone Octavia load balancer.",
-		Enabled:         true,
+		Enabled:         new(bool),
```
pierreprinetti, line 65:
```suggestion
		Enabled:         ptr.To(true),
```

openstack/loadbalancer/v2/flavors/testing/requests_test.go
```diff
@@ -106,10 +106,13 @@ func TestUpdateFlavor(t *testing.T) {
...
+	name := "Basic v2"
+	description := "Rename flavor"
+	enabled := true
 	actual, err := flavors.Update(context.TODO(), client, "5548c807-e6e8-43d7-9ea4-b38d34dd74a0", flavors.UpdateOpts{
-		Name:        "Basic v2",
-		Description: "Rename flavor",
-		Enabled:     true,
+		Name:        &name,
+		Description: &description,
+		Enabled:     &enabled,
```
pierreprinetti, line 115: consider using ptr.To

internal/acceptance/openstack/compute/v2/aggregates_test.go
```diff
@@ -46,9 +46,11 @@ func TestAggregatesCRUD(t *testing.T) {
 
 	tools.PrintResource(t, aggregate)
 
+	name := "new_aggregate_name"
+	availabilityZone := "new_azone"
 	updateOpts := aggregates.UpdateOpts{
-		Name:             "new_aggregate_name",
-		AvailabilityZone: "new_azone",
+		Name:             &name,
+		AvailabilityZone: &availabilityZone,
```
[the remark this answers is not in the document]
kayrus: why internal? this can be reused in other packages. also it would be helpful to have a helper to get a pointer value on-the-fly. and if it's a nil, return an empty struct.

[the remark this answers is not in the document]
pierreprinetti: Internal so that it doesn't increase the package's API. I don't think we want that function to be part of the exported Gophercloud funtions.

>it would be helpful to have a helper to get a pointer value on-the-fly. and if it's a nil, return an empty struct

That looks complex. Do you envision that as another convenience function for testing? Would you use that in this PR?

internal/acceptance/openstack/compute/v2/secgroup_test.go
```diff
@@ -48,7 +48,7 @@ func TestSecGroupsCRUD(t *testing.T) {
 	newName := tools.RandomString("secgroup_", 4)
 	description := ""
 	updateOpts := secgroups.UpdateOpts{
-		Name:        newName,
+		Name:        &newName,
 		Description: &description,
```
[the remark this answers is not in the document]
kayrus: I'll keep this as is, since `description` is used below in code.

openstack/loadbalancer/v2/flavors/testing/fixtures.go
pierreprinetti: OK; but then please add a new test to verify the new functionality, and leave this old test here doing its job of testing regressions :)

internal/ptr/ptr.go
```diff
@@ -0,0 +1,13 @@
...
+	return &v
+}
+
+func From[T any](p *T) T {
+	if p == nil {
+		var v T
+		return v
+	}
+	return *p
+}
```
pierreprinetti, line 13: This doesn't seem to be used anywhere. Please remove

pierreprinetti (approved):
Thanks! This is for v3

PR #3203, opened 2024-10-07
all: Accept ServiceClient as an interface

pierreprinetti:
Allow client libraries to modify the behaviour of ServiceClient or mock calls to the OpenStack API.

This change is targeted at `v3`.

### DIFF
--- a/openstack/baremetal/apiversions/requests.go
+++ b/openstack/baremetal/apiversions/requests.go
@@ -7,14 +7,14 @@ import (
 )
 
 // List lists all the API versions available to end users.
-func List(ctx context.Context, client *gophercloud.ServiceClient) (r ListResult) {
+func List(ctx context.Context, client gophercloud.Client) (r ListResult) {
 	resp, err := client.Get(ctx, listURL(client), &r.Body, nil)
 	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
 	return
 }
 
 // Get will get a specific API version, specified by major ID.
-func Get(ctx context.Context, client *gophercloud.ServiceClient, v string) (r GetResult) {
+func Get(ctx context.Context, client gophercloud.Client, v string) (r GetResult) {
 	resp, err := client.Get(ctx, getURL(client, v), &r.Body, nil)
 	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
 	return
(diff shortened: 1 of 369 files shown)


mdbooth:
How do you see client code using this?

Here's an example of how we're using mocks in CAPO: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/852272197ad100fecd9cc7e95a77e88897812520/pkg/cloud/services/compute/instance_test.go#L82-L96

```golang
		{
			testName: "Return image ID when name given",
			image: infrav1.ImageParam{
				Filter: &infrav1.ImageFilter{
					Name: ptr.To(imageName),
				},
			},
			want: ptr.To(imageID),
			expect: func(m *mock.MockImageClientMockRecorder) {
				m.ListImages(images.ListOpts{Name: imageName}).Return(
					[]images.Image{{ID: imageID, Name: imageName}},
					nil)
			},
			wantErr: false,
		},
```

`ListImages` is defined here: https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/852272197ad100fecd9cc7e95a77e88897812520/pkg/clients/image.go#L59-L66. It's just a mockable wrapper round gophercloud's `images.List` (plus some prometheus stuff which is probably a waste of time).

With my 'gophercloud library user' hat on, what I'm trying to achieve here is:
* I am asserting that we must call images.List with some arguments that I can match in flexible ways
* I am providing the return value of this call

I like that I can use gophercloud structs here. I'm asserting that the call uses an `images.ListOpts`, and I'm defining the return in terms of gophercloud `images.Image` objects. As a gophercloud library user, this is the level I want to interact with OpenStack at.

I could theoretically use the ServiceClient interface to mock my OpenStack calls, but this would take the gophercloud abstractions away from me. I would have to define my expected behaviours in terms of HTTP and raw json.

I can imagine that we could fairly easily add a 'recording' mock ServiceClient interface which we create in the tests. My test might have a 'record' phase something like:

```golang
mockServiceClient := MyMockServiceClient()
images.List(mockServiceClient ,images.ListOpts{Name: imageName})
... something with pages? ...
```

I can imagine that MyMockServiceClient could record any resulting invocation, however complicated, which could be used later in the 'execute' phase of the test. I can't immediately think of a good way to use an interface like this to provide a return value, though. The return value I'm interested in is `[]images.Image`. How can I make `mockServiceClient` return data that will be marshalled correctly into the `[]images.Image` I need for this test case?

pierreprinetti:
Let's park this.

PR #3209, opened 2024-10-10
Configure and respect service type aliases

### LINKED ISSUE #3207: Service type aliases are ignored
The [OpenStack Service Types Authority](https://service-types.openstack.org/) is the central authority for handing out service types to projects for their services. Per the docs for same:

> Each OpenStack service with a REST API must have a well known service type. The well known service type guarantees a documented API is available for that service. Users of this service can trust that it will be the same between different OpenStack environments.

We currently allow users to retrieve services by service type and service name, but we do not allow them to retrieve by a service type alias *and* we are using aliases in some places in place of the official type (for example, `volumev3` instead of `block-storage`). We should correct both issues.

stephenfin:
Fixes #3207

This means we can identify a service by either its type or any of its official aliases, which is the expectation when it comes to navigating the service catalog.

### DIFF
--- a/endpoint_search.go
+++ b/endpoint_search.go
@@ -1,5 +1,7 @@
 package gophercloud
 
+import "slices"
+
 // Availability indicates to whom a specific service endpoint is accessible:
 // the internet at large, internal networks only, or only to administrators.
 // Different identity services use different terminology for these. Identity v2
@@ -21,6 +23,31 @@ const (
 	// the cluster's internal network.
 	AvailabilityInternal Availability = "internal"
 )
+
+// ServiceTypeAliases contains a mapping of service types to any aliases, as
+// defined by the OpenStack Service Types Authority. Only service types that
+// we support are included.
+var ServiceTypeAliases = map[string][]string{
+	"application-container":               {"container"},
+	"baremetal":                           {"bare-metal"},
+	"baremetal-introspection":             {},
+	"block-storage":                       {"block-store", "volume", "volumev2", "volumev3"},
+	"compute":                             {},
+	"container-infrastructure-management": {"container-infrastructure", "container-infra"},
+	"database":                            {},
+	"dns":                                 {},
+	"identity":                            {},
+	"image":                               {},
+	"key-manager":                         {},
+	"load-balancer":                       {},
+	"message":                             {"messaging"},
+	"networking":                          {},
+	"object-store":                        {},
+	"orchestration":                       {},
+	"placement":                           {},
+	"shared-file-system":                  {"sharev2", "share"},
+	"workflow":                            {"workflowv2"},
+}
 
 // EndpointOpts specifies search criteria used by queries against an
 // OpenStack service catalog. The options must contain enough information to
@@ -80,6 +107,24 @@ func (eo *EndpointOpts) ApplyDefaults(t string) {
 	}
 	if eo.Availability == "" {
 		eo.Availability = AvailabilityPublic
+	}
+	if len(eo.Aliases) == 0 {
+		if aliases, ok := ServiceTypeAliases[eo.Type]; ok {
+			// happy path: user requested a service type by its official name
+			eo.Aliases = aliases
+		} else {
+			// unhappy path: user requested a service type by its alias or an
+			// invalid/unsupported service type
+			// TODO(stephenfin): This should probably be an error in v3
+			for t, aliases := range ServiceTypeAliases {
+				if slices.Contains(aliases, eo.Type) {
+					// we intentionally override the service type, even if it
+					// was explicitly requested by the user
+					eo.Type = t
+					eo.Aliases = aliases
+				}
+			}
+		}
 	}
 }
 
--- a/endpoint_search.go
+++ b/endpoint_search.go
@@ -30,14 +30,22 @@ const (
 // package, like "openstack.NewComputeV2()".
 type EndpointOpts struct {
 	// Type [required] is the service type for the client (e.g., "compute",
-	// "object-store"). Generally, this will be supplied by the service client
-	// function, but a user-given value will be honored if provided.
+	// "object-store"), as defined by the OpenStack Service Types Authority.
+	// This will generally be supplied by the service client function, but a
+	// user-given value will be honored if provided.
 	Type string
 
 	// Name [optional] is the service name for the client (e.g., "nova") as it
 	// appears in the service catalog. Services can have the same Type but a
 	// different Name, which is why both Type and Name are sometimes needed.
 	Name string
+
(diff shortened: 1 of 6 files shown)


stephenfin:
I'm not sure whether we want to expose the aliases via a new field in `EndpointOpts` like I've done here, or if it would be simpler to stick with a single `Type` field and do the mapping of service type to aliases inside `V2EndpointURL` / `V3EndpointURL` in `openstack/endpoint_location.go`. What we've done here is more obvious, but I'm not allowing users to override the aliases maybe exposing it is wholly unnecessary? :shrug: Open to thoughts here.

kayrus:
@stephenfin I think that `NewClient` functions also need to support aliases, so users can override them for custom service like `block-storage-beta`

stephenfin:
> @stephenfin I think that `NewClient` functions also need to support aliases, so users can override them for custom service like `block-storage-beta`

If I do that, it'll be a breaking API change though, which means I can't fix this until V3?

kayrus:
right. we can implement this in V3.

stephenfin:
Okay. So is what we have here good enough for now, in v2? What are your thoughts on [my questions above](https://github.com/gophercloud/gophercloud/pull/3209#issuecomment-2404775334)?

endpoint_search.go
```diff
@@ -22,6 +24,31 @@ const (
 	AvailabilityInternal Availability = "internal"
 )
 
+// ServiceTypeAliases contains a mapping of service types to any aliases, as
+// defined by the OpenStack Service Types Authority. Only service types that
```
EmilienM, line 28: It would have been nice to provide a link (for each service) of the code where we can find the alias. But I'm fine as is now.

EmilienM (approved):

EmilienM:
All the Cinder jobs have a failure, we need to check whether it's related here.

stephenfin:
> All the Cinder jobs have a failure, we need to check whether it's related here.

Fixed here https://github.com/gophercloud/gophercloud/pull/3209/commits/6428aced096b7e20306f2d25dd9fbebe8d82f1b9

Reported here https://lists.openstack.org/archives/list/openstack-discuss@lists.openstack.org/thread/QQ7L44VE4MOWOQVOFCZQPTIVI35A6SCK/

stephenfin:
> **EDIT:** Also, I'm surprised this has attracted the `semver:major` label. I only changed internal methods (various versions of `initClientOpts`) which I had assumed to be fair game. 😕

Turns out this is because of my addition of the `Aliases` field to `EndpointOpts`. It's a slice which is not comparable, thus `EndpointOpts` is no longer comparable.

EmilienM (approved):
Our CI is awesome.

endpoint_search.go
mnaser: FWIW, many months later, if someone is curious:

https://service-types.openstack.org/

openstack/client.go
```diff
@@ -402,17 +404,17 @@ func NewBlockStorageV1(client *gophercloud.ProviderClient, eo gophercloud.Endpoi
...
 // NewBlockStorageV3 creates a ServiceClient that may be used to access the v3 block storage service.
 func NewBlockStorageV3(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
-	return initClientOpts(client, eo, "volumev3")
+	return initClientOpts(client, eo, "block-storage")
 }
 
 // NewSharedFileSystemV2 creates a ServiceClient that may be used to access the v2 shared file system service.
 func NewSharedFileSystemV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
-	return initClientOpts(client, eo, "sharev2")
+	return initClientOpts(client, eo, "shared-file-system")
```
SuperSandro2000, line 417: I have the suspicion that this somehow reverted our usage to sharev1 from sharev2. Any hints how to properly debug this?

kayrus: see https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1848

SuperSandro2000: Would setting the Type on the endpoint opts also work?

```diff
diff --git a/internal/liquids/manila/liquid.go b/internal/liquids/manila/liquid.go
index 2a15d01a..b524ec9a 100644
--- a/internal/liquids/manila/liquid.go
+++ b/internal/liquids/manila/liquid.go
@@ -25,7 +25,6 @@
    "math"
    "regexp"
    "strconv"
-   "strings"
    "time"

    "github.com/gophercloud/gophercloud/v2"
@@ -74,12 +73,11 @@ func (l *Logic) Init(ctx context.Context, provider *gophercloud.ProviderClient,
    }

    // initialize connection to Manila
+   eo.Type = "sharev2"
    l.ManilaV2, err = openstack.NewSharedFileSystemV2(provider, eo)
    if err != nil {
        return err
    }
-   // Workaround for gophercloud 2.7.0 which gets the first found OpenStack URL for Manila.
-   l.ManilaV2.Endpoint = strings.Replace(l.ManilaV2.Endpoint, "v1", "v2", 1)
    microversion, err := l.findMicroversion(ctx, l.ManilaV2)
    if err != nil {
        return err
```

majewsky:
I'm confused why this patch was accepted at all. You're just aliasing `volume`, and `volumev2` and `volumev3` to each other, even though those are completely different APIs. How is this supposed to work on a theoretical level? If I'm asking for NewBlockStorageV3, giving back the v2 API is a categorically wrong answer.

majewsky:
So then the problem is only with including the v1 APIs in the list?

stephenfin:
The issue here is that we are missing service discovery, i.e. the equivalent of [`keystoneauth.discover`](https://github.com/openstack/keystoneauth/blob/43bc36dd59c34db64ab3305ba530ec16e0486c12/keystoneauth1/discover.py#L68). I am working on a fix plus a interim partial revert to tide us through until it's ready. A full revert is not an option as what gophercloud was doing previously was not compliant with deployments using the correct service type rather than an alias. The DevStack change merely exposed this incorrect behaviour.

stephenfin:
> The issue here is that we are missing service discovery, i.e. the equivalent of [`keystoneauth.discover`](https://github.com/openstack/keystoneauth/blob/43bc36dd59c34db64ab3305ba530ec16e0486c12/keystoneauth1/discover.py#L68). I am working on a fix plus a interim partial revert to tide us through until it's ready. A full revert is not an option as what gophercloud was doing previously was not compliant with deployments using the correct service type rather than an alias. The DevStack change merely exposed this incorrect behaviour.

The interim fix is available here https://github.com/gophercloud/gophercloud/pull/3350

PR #3221, opened 2024-10-21
Fix clear-needinfo

pierreprinetti:
The gh tool must be set to operate on an issue rather than on a PR.

cf https://docs.github.com/en/actions/use-cases-and-examples/project-management/adding-labels-to-issues

(thanks @stephenfin)


pierreprinetti:
Actually, this is expected to run on both issues and PRs. I might not be able to get away with this one liner

stephenfin:
Both `gh issue edit --help` and `gh pr edit --help` shows the `--remove-label` option, which makes sense since issues and PRs share some metadata (including their numbers). Are you sure that simply adding the permissions isn't sufficient?

EmilienM (approved):

PR #3265, opened 2024-12-06
labels: Add no-backport

pierreprinetti:
This label serves as manual triage for the PRs that legitimately shouldn't be backported.

Once we start labeling PRs that we actively don't want backported, it will be easier to identify which PRs we forgot to backport.


pierreprinetti:
I am not 100% sure this helps actually

PR #3267, opened 2024-12-06
SG rules: implement bulk create

mandre:
Implement bulk creation of security group rules [1].

[1] https://docs.openstack.org/api-ref/network/v2/#bulk-create-security-group-rule

### DIFF
--- a/openstack/networking/v2/extensions/security/rules/requests.go
+++ b/openstack/networking/v2/extensions/security/rules/requests.go
@@ -150,6 +150,23 @@ func Create(ctx context.Context, c *gophercloud.ServiceClient, opts CreateOptsBu
 	return
 }
 
+// CreateBulk is an operation which adds new security group rules and associates them
+// with an existing security group (whose ID is specified in CreateOpts).
+// As of Dalmatian (2024.2) neutron only allows bulk creation of rules when
+// they all belong to the same tenant and security group.
+// https://github.com/openstack/neutron/blob/6183792/neutron/db/securitygroups_db.py#L814-L828
+func CreateBulk(ctx context.Context, c *gophercloud.ServiceClient, opts []CreateOpts) (r CreateBulkResult) {
+	body, err := gophercloud.BuildRequestBody(opts, "security_group_rules")
+	if err != nil {
+		r.Err = err
+		return
+	}
+
+	resp, err := c.Post(ctx, rootURL(c), body, &r.Body, nil)
+	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
+	return
+}
+
 // Get retrieves a particular security group rule based on its unique ID.
 func Get(ctx context.Context, c *gophercloud.ServiceClient, id string) (r GetResult) {
 	resp, err := c.Get(ctx, resourceURL(c, id), &r.Body, nil)
--- a/internal/acceptance/openstack/networking/v2/extensions/extensions.go
+++ b/internal/acceptance/openstack/networking/v2/extensions/extensions.go
@@ -140,6 +140,44 @@ func CreateSecurityGroupRule(t *testing.T, client *gophercloud.ServiceClient, se
 	return rule, nil
 }
 
+// CreateSecurityGroupRulesBulk will create security group rules with a random name
+// and random port between 80 and 99.
+// An error will be returned if one was failed to be created.
+func CreateSecurityGroupRulesBulk(t *testing.T, client *gophercloud.ServiceClient, secGroupID string) ([]rules.SecGroupRule, error) {
+	t.Logf("Attempting to bulk create security group rules in group: %s", secGroupID)
+
+	sgRulesCreateOpts := make([]rules.CreateOpts, 3)
+	for i := range 3 {
+		description := "Rule description"
+		fromPort := tools.RandomInt(80, 89)
+		toPort := tools.RandomInt(90, 99)
+
+		sgRulesCreateOpts[i] = rules.CreateOpts{
+			Description:  description,
+			Direction:    "ingress",
+			EtherType:    "IPv4",
+			SecGroupID:   secGroupID,
+			PortRangeMin: fromPort,
+			PortRangeMax: toPort,
+			Protocol:     rules.ProtocolTCP,
+		}
+	}
+
+	rules, err := rules.CreateBulk(context.TODO(), client, sgRulesCreateOpts).Extract()
+	if err != nil {
+		return rules, err
+	}
+
+	for i, rule := range rules {
+		t.Logf("Created security group rule: %s", rule.ID)
+
+		th.AssertEquals(t, sgRulesCreateOpts[i].SecGroupID, rule.SecGroupID)
+		th.AssertEquals(t, sgRulesCreateOpts[i].Description, rule.Description)
+	}
+
+	return rules, nil
+}
+
 // DeleteSecurityGroup will delete a security group of a specified ID.
 // A fatal error will occur if the deletion failed. This works best as a
 // deferred function
--- a/internal/acceptance/openstack/networking/v2/extensions/security_test.go
+++ b/internal/acceptance/openstack/networking/v2/extensions/security_test.go
@@ -26,6 +26,12 @@ func TestSecurityGroupsCreateUpdateDelete(t *testing.T) {
 	th.AssertNoErr(t, err)
 	defer DeleteSecurityGroupRule(t, client, rule.ID)
 
+	rules, err := CreateSecurityGroupRulesBulk(t, client, group.ID)
+	th.AssertNoErr(t, err)
+	for _, r := range rules {
+		defer DeleteSecurityGroupRule(t, client, r.ID)
+	}
+
 	tools.PrintResource(t, group)
 
 	var name = "Update group"
--- a/openstack/networking/v2/extensions/security/rules/results.go
+++ b/openstack/networking/v2/extensions/security/rules/results.go
@@ -103,6 +103,10 @@ type commonResult struct {
 	gophercloud.Result
 }
 
+type bulkResult struct {
+	gophercloud.Result
+}
+
 // Extract is a function that accepts a result and extracts a security rule.
 func (r commonResult) Extract() (*SecGroupRule, error) {
 	var s struct {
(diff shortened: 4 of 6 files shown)


openstack/networking/v2/extensions/security/rules/requests.go
```diff
@@ -150,6 +150,25 @@ func Create(ctx context.Context, c *gophercloud.ServiceClient, opts CreateOptsBu
 	return
 }
 
+// CreateBulk is an operation which adds new security group rules and associates them
+// with an existing security group (whose ID is specified in CreateOpts).
+func CreateBulk(ctx context.Context, c *gophercloud.ServiceClient, opts []CreateOpts) (r CreateResult) {
+	createOpts := make([]map[string]any, len(opts))
```
mandre, line 156: Alternatively, we can modify [gophercloud.BuildRequestBody()](https://github.com/gophercloud/gophercloud/blob/de76eeb/params.go#L35) to accept an array/slice. Perhaps this would be more user friendly? Do we have a preference?

The nice thing if we decide to make gophercloud.BuildRequestBody() accept slices or arrays, is that it does not require changing the signature function, allowing us to backport the change.

pierreprinetti (commented):
What happens when you pass two rules.CreateOpts that refer each to a different, existing security group ID?

mandre:
> What happens when you pass two rules.CreateOpts that refer each to a different, existing security group ID?

I haven't tried but I understand they can be different, [from the doc](https://docs.openstack.org/api-ref/network/v2/#bulk-create-security-group-rule).

pierreprinetti:
nope.

```
$ curl -sS \
        -X POST \
        -H "X-Auth-Token: $TOKEN" \
        -H "content-type: application/json" \
        --data-binary @sg.json \
        "${NEUTRON_URL}/v2.0/security-group-rules" \
        | jq
{
  "NeutronError": {
    "type": "SecurityGroupNotSingleGroupRules",
    "message": "Only allowed to update rules for one security profile at a time",
    "detail": ""
  }
}
```

mandre:
Perhaps, but that's none of gophercloud concerns, is it?

pierreprinetti:
I also woulnd't preemptively block invalid calls; however it may be useful to note the limitation in a comment

openstack/networking/v2/extensions/security/rules/requests.go
mandre: I've moved the logic in BuildRequestBody(), as it makes more sense to me. PTAL

https://github.com/gophercloud/gophercloud/compare/324a1105f8e764e30e5483bd95176f347dfc07a9..b1f8bbb05730e871f8d06bb6de2e28fcb50da3d5

pierreprinetti:
**TL;DR: The `Extract()` method of the return type of `CreateBulkResult` should return a slice of rules and an error.**

In other words, I would suggest that `CreateBulkResult(/* ... */)` returns `CreateBulkResult` and that `CreateBulkResult` has a method with signature `Extract() ([]rules.SecGroupRule, error)`.

---

I did a couple tests with cURL. The call looked like this:
```bash
curl -sSi \
        -X POST \
        -H "X-Auth-Token: $(openstack token issue -f value -c id)" \
        -H "content-type: application/json" \
        --data-binary @sg.json \
        "$(openstack catalog show network --format json | jq -r '.endpoints[] | select(.interface == "public") | .url')/v2.0/security-group-rules"
```

When the endpoint is called with a valid payload, it returns a JSON slice containing all the created rules.

For example, when `sg.json` looks like this (**valid payload**):

```json
{
    "security_group_rules": [
        {
            "direction": "ingress",
            "port_range_min": "443",
            "ethertype": "IPv4",
            "port_range_max": "443",
            "protocol": "tcp",
            "security_group_id": "f9202b90-a34e-4804-9b72-60a9cda7bcc1"
        },
        {
            "direction": "ingress",
            "port_range_min": "80",
            "ethertype": "IPv4",
            "port_range_max": "80",
            "protocol": "tcp",
            "security_group_id": "f9202b90-a34e-4804-9b72-60a9cda7bcc1"
        }
    ]
}
```
...then the response looks like this:

```
HTTP/1.1 201 Created
Content-Length: 996
X-Openstack-Request-Id: req-bb55ef49-93e7-4ac3-9b9f-9fd424d50212
Date: Mon, 09 Dec 2024 14:22:03 GMT

{"security_group_rules": [{"id": "cf2ce71c-4c36-40d9-ab03-13afde7cb4ea", "tenant_id": "90dce24f8e6748bfba319a9223d0a7a6", "security_group_id": "f9202b90-a34e-4804-9b72-60a9cda7bcc1", "ethertype": "IPv4", "direction": "ingress", "protocol": "tcp", "port_range_min": 443, "port_range_max": 443, "remote_ip_prefix": null, "remote_group_id": null, "description": "", "created_at": "2024-12-09T14:22:02Z", "updated_at": "2024-12-09T14:22:02Z", "revision_number": 0, "project_id": "90dce24f8e6748bfba319a9223d0a7a6"}, {"id": "4d6432c9-e1a0-4861-acf0-8be423f68f25", "tenant_id": "90dce24f8e6748bfba319a9223d0a7a6", "security_group_id": "f9202b90-a34e-4804-9b72-60a9cda7bcc1", "ethertype": "IPv4", "direction": "ingress", "protocol": "tcp", "port_range_min": 80, "port_range_max": 80, "remote_ip_prefix": null, "remote_group_id": null, "description": "", "created_at": "2024-12-09T14:22:02Z", "updated_at": "2024-12-09T14:22:02Z", "revision_number": 0, "project_id": "90dce24f8e6748bfba319a9223d0a7a6"}]}
```

However when the payload contains **one invalid security group rule**, in whichever position:

```json
{
    "security_group_rules": [
        {
            "direction": "ingress",
            "port_range_min": "443",
            "ethertype": "IPv4",
            "port_range_max": "443",
            "protocol": "tcp",
            "security_group_id": "f9202b90-a34e-4804-9b72-60a9cda7bcc1"
        },
        {
            "direction": "ingress",
            "port_range_min": "81",
            "ethertype": "IPv4",
            "port_range_max": "79",
            "protocol": "tcp",
            "security_group_id": "f9202b90-a34e-4804-9b72-60a9cda7bcc1"
        }
    ]
}
```

...then none of them is created, and the response looks like this:

```
HTTP/1.1 400 Bad Request
Content-Length: 151
X-Openstack-Request-Id: req-867397b4-56e2-4f20-842a-3f92c9aae658
Date: Mon, 09 Dec 2024 14:24:46 GMT

{"NeutronError": {"type": "SecurityGroupInvalidPortRange", "message": "For TCP/UDP protocols, port_range_min must be <= port_range_max", "detail": ""}}
```

I have also tested a payload with **two errored rules**:
1. one valid rule
2. one rule with a different secgroup ID
3. one rule with ports max<min

and only one error is returned (in my case that was the first encountered, which is the different secgroup ID).

Lastly, an **empty slice of rules**:
```json
{"security_group_rules":[]}
```
Result:
```
HTTP/1.1 400 Bad Request
Content-Length: 91
X-Openstack-Request-Id: req-e493c35c-a5da-4f36-baf1-6c40a1aac118
Date: Mon, 09 Dec 2024 14:32:02 GMT

{"NeutronError": {"type": "HTTPBadRequest", "message": "Resources required", "detail": ""}}
```
---

Based on these experiments, I think it's fair to say that as long as Neutron is healthy there are only two possible results to a call to the bulk-create endpoint:
* EITHER an HTTP code of 201 and a JSON array of the created rules;
* OR exactly one error.

pierreprinetti:
oh you have added `ExtractRules` :grimacing: that's cheating

pierreprinetti:
Can you please change the return type of `CreateBulk` so that its `Extract` method returns the slice'? Having a shared object on which you can call both `Extract` and `ExtractRules` seems a footgun.

pierreprinetti (approved):
Nice, thank you.

Tested OK. Rules correctly applied, return values populated OK.
I still wonder what happens if you reach the pagination threshold (provided there is any), but I couldn't reach that with "just" 1200 rules (which took 9m22s to apply :sweat_smile:)

pierreprinetti:
FTR tested with this.

```Go
package main

import (
	"context"
	"fmt"
	"slices"
	"strconv"

	"github.com/gophercloud/gophercloud/v2/openstack"
	"github.com/gophercloud/gophercloud/v2/openstack/config"
	"github.com/gophercloud/gophercloud/v2/openstack/config/clouds"
	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/security/rules"
)

const sgID = "5eddd6ae-bad2-4290-9361-5ae5552fc1e7"

func main() {
	ctx := context.Background()
	authOptions, endpointOptions, tlsConfig, err := clouds.Parse()
	if err != nil {
		panic(err)
	}

	providerClient, err := config.NewProviderClient(ctx, authOptions, config.WithTLSConfig(tlsConfig))
	if err != nil {
		panic(err)
	}

	client, err := openstack.NewNetworkV2(providerClient, endpointOptions)
	if err != nil {
		panic(err)
	}

	createOpts := make([]rules.CreateOpts, 1200)

	for i := range createOpts {
		createOpts[i] = rules.CreateOpts{
			Direction:    rules.DirIngress,
			Description:  "Rule number " + strconv.Itoa(i),
			EtherType:    rules.EtherType6,
			SecGroupID:   sgID,
			PortRangeMax: i,
			PortRangeMin: i,
			Protocol:     rules.ProtocolUDP,
		}
	}

	sgs, err := rules.CreateBulk(ctx, client, createOpts).Extract()
	if err != nil {
		panic(err)
	}

	if want, got := len(createOpts), len(sgs); want != got {
		fmt.Printf("ERROR: expected %d rules, got %d\n", want, got)
	}

	slices.SortFunc(sgs, func(a, b rules.SecGroupRule) int { return a.PortRangeMin - b.PortRangeMin })

	for i := range sgs {
		if want, got := createOpts[i].PortRangeMin, sgs[i].PortRangeMin; want != got {
			fmt.Printf("Expected %d, got %d\n", want, got)
		}
	}

	fmt.Printf("DONE. Successfully created %d rules and got %d returned rules.\n", len(createOpts), len(sgs))
}
```

PR #3276, opened 2024-12-20
bulk security group rule creation: Allow using CreateOptsBuilder

pierreprinetti:
Before this patch, security group rules could only be created in a batch when using slices of plain `rules.CreateOpts`. With this change, a slice of any type implementing `rules.CreateOptsBuilder` can be used to create security group rules in a batch.


mandre (dismissed):

pierreprinetti:
There may be a race condition in the tests

EmilienM (approved):

Issue #3208, opened 2024-10-10
support Trove quotas


aborrero:
Thanks for your hard work with this library, it is really appreciated.

I think, as of this writing, gophercloud is missing support for Openstack Trove quotas.

Please, add support for it.

See also:
* https://docs.openstack.org/api-ref/database/#quotas

aborrero:
The mentioned patch only introduces the GET action on the endpoint.

I believe it misses the the PUT operation: https://docs.openstack.org/api-ref/database/#update-resources-quota-for-a-specific-project thus the support is incomplete.

mandre:
Indeed, this was automatically closed as completed because the related PR included the magic `Fixes` keyword. I'll re-open because we still don't have PUT.

PR #3279, opened 2024-12-24
Database V1 quotas get

tdeni:
Fixes https://github.com/gophercloud/gophercloud/issues/3208

code in this PR:

API:
https://docs.openstack.org/api-ref/database/#quotas

Get:
https://github.com/openstack/trove/blob/1526b20826bb7bfcc3c62ec3ce7c44a90b78f46f/trove/extensions/mgmt/quota/service.py#L32


EmilienM (approved):

EmilienM:
This service is not tested so we'll rely on unit tests. For some reason they didn't run.

EmilienM (approved):

PR #3284, opened 2025-01-17
block_storage: get quotausage for blockstorage volume_types

stijzermans:
The volume_type attributes are currently only set for QuotaSet objects.

Hence, implemented per `volume_type` QuotaUsage as a nested object 
of `QuotaUsageSet`for block storage. 

See API docs for refs.

block-storage v3
https://docs.openstack.org/api-ref/block-storage/v3/#show-quota-usage-for-a-project

block-storage v2
https://docs.openstack.org/api-ref/block-storage/v2/#show-quotas

### DIFF
--- a/internal/acceptance/openstack/blockstorage/v3/quotaset_test.go
+++ b/internal/acceptance/openstack/blockstorage/v3/quotaset_test.go
@@ -46,6 +46,7 @@ func TestQuotasetGetUsage(t *testing.T) {
 	th.AssertNoErr(t, err)
 
 	tools.PrintResource(t, quotaSetUsage)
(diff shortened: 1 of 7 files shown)


stijzermans:
@mandre  - sorry for that, was missing a comma in the test fixture for the v2 block storage schema, which caused the test to fail. Should be all good now.

mandre (approved):
Thanks, LGTM.

We won't be able to backport this to branch `v2` however, due to the [backward incompatible change](https://github.com/gophercloud/gophercloud/pull/3284/checks#step:5:103):

```
   Incompatible changes:
  - QuotaUsageSet: old is comparable, new is not
```

PR #3301, opened 2025-02-13
Added support for `hypervisor_hostname`

### LINKED ISSUE #2499: Create Server `hypervisor_hostname` support
In the OpenStack Nova API documentation for server creates, you can provide a `hypervisor_hostname` (with microversion `>2.74`) in the POST body.

https://docs.openstack.org/api-ref/compute/?expanded=list-hypervisors-detail,create-server-detail#create-server

I cannot find this supported in gophercloud anywhere. I attempted to use the existing scheduler hints, but this solution did not work. Is this field planned on being supported?

I believe the OpenStack python client supports this here:
https://github.com/openstack/python-openstackclient/blob/f1cb66a476b948093986afd18a00941b7807a9c4/openstackclient/compute/v2/server.py#L1685-L1694

ngin8-beta:
Fixes #2499

code in this PR:

[openstack api ref](https://docs.openstack.org/api-ref/compute/?expanded=list-hypervisors-detail,create-server-detail#create-server)
[openstack python client](https://github.com/openstack/python-openstackclient/blob/f1cb66a476b948093986afd18a00941b7807a9c4/openstackclient/compute/v2/server.py#L1685-L1694)

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -511,6 +511,9 @@ type CreateOpts struct {
 
 	// KeyName is the name of the key pair.
 	KeyName string `json:"key_name,omitempty"`
+
+	// HypervisorHostname is the name of the hypervisor to which the server is scheduled.
+	HypervisorHostname string `json:"hypervisor_hostname,omitempty"`
 }
 
 // ToServerCreateMap assembles a request body based on the contents of a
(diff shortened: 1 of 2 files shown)


EmilienM (approved):

mandre:
Why wasn't the bot able to backport this PR? The job says [it was skipped](https://github.com/gophercloud/gophercloud/actions/runs/13306925219/job/37681855952), which doesn't really makes sense to me.

EmilienM:
> Why wasn't the bot able to backport this PR? The job says [it was skipped](https://github.com/gophercloud/gophercloud/actions/runs/13306925219/job/37681855952), which doesn't really makes sense to me.

because semver vas minor and not patch

PR #3305, opened 2025-02-20
Enhance Snapshot struct and add ListDetail function in V3 blockstorage

### LINKED ISSUE #2403: V3 blockstorage snapshot list is using the wrong endpoint

dkokkino:
- Extended Snapshot struct to include additional fields: Progress, ProjectID, GroupSnapshotID, UserID, ConsumesQuota
- Implemented ListDetail function to retrieve detailed snapshot information from the snapshots/detail endpoint
- Added new test to validate the inclusion of additional fields and the functionality of  ListDetails

Fixes #2403

code in this PR:

Documentation: https://docs.openstack.org/api-ref/block-storage/v3/#list-snapshots-and-details

### DIFF
--- a/internal/acceptance/openstack/blockstorage/v3/snapshots_test.go
+++ b/internal/acceptance/openstack/blockstorage/v3/snapshots_test.go
@@ -73,7 +73,24 @@ func TestSnapshots(t *testing.T) {
 
 		return true, nil
 	})
+	th.AssertNoErr(t, err)
 
+	err = snapshots.ListDetail(client, listOpts).EachPage(context.TODO(), func(_ context.Context, page pagination.Page) (bool, error) {
+		actual, err := snapshots.ExtractSnapshots(page)
+		th.AssertNoErr(t, err)
+		th.AssertEquals(t, 1, len(actual))
+
+		var found bool
+		for _, v := range actual {
+			if v.ID == snapshot1.ID || v.ID == snapshot2.ID {
+				found = true
+			}
+		}
+
+		th.AssertEquals(t, found, true)
+
+		return true, nil
+	})
 	th.AssertNoErr(t, err)
 }
 
--- a/openstack/blockstorage/v3/snapshots/requests.go
+++ b/openstack/blockstorage/v3/snapshots/requests.go
@@ -122,6 +122,21 @@ func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pa
 	})
 }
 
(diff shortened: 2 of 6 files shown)


internal/acceptance/openstack/blockstorage/v3/snapshots_test.go
```diff
@@ -74,6 +74,23 @@ func TestSnapshots(t *testing.T) {
 		return true, nil
 	})
```
mandre, line 76: We're missing an `th.AssertNoErr(t, err)` here, to validate that the previous call didn't return an error.

mandre (commented):
This looks great. We can merge after you address the missing assert and squash your commits.

internal/acceptance/openstack/blockstorage/v3/snapshots_test.go
dkokkino: Added the missing line of code and squashed my commits

mandre (approved):

PR #3321, opened 2025-03-06
build(v2): Set Go toolchain to 1.23.6 in go.mod

EmilienM:
- Added `go 1.23.6` toolchain directive in `go.mod`.
- Ensures compatibility with `golang.org/x/tools@v0.31.0`, which requires Go >= 1.23.0.


EmilienM:
I think we should go with https://github.com/gophercloud/gophercloud/pull/3323

PR #3322, opened 2025-03-06
ci(v2): Fallback to Go 1.22 compatible goimports version

EmilienM:
- Updated `goimports` installation to fallback to `v0.30.0` if `latest` requires Go 1.23+.
- Ensures compatibility with Go 1.22 for Gophercloud v2.
- Prevents CI failures when running in environments with older Go versions.

This change allows CI and developers using Go 1.22 to continue using `goimports` without version conflicts.


EmilienM:
https://github.com/gophercloud/gophercloud/pull/3323

PR #3323, opened 2025-03-06
ci: Fallback to Go 1.22 compatible goimports version

EmilienM:
- Updated `goimports` installation to fallback to `v0.30.0` if `latest` requires Go 1.23+.
- Ensures compatibility with Go 1.22 for Gophercloud v2.
- Prevents CI failures when running in environments with older Go versions.

This change allows CI and developers using Go 1.22 to continue using `goimports` without version conflicts.

### DIFF
--- a/.github/workflows/unit.yaml
+++ b/.github/workflows/unit.yaml
@@ -32,6 +32,9 @@ jobs:
 
           go install github.com/wadey/gocovmerge@master
 
+          # Fallback to a go1.22 compatible version for gophercloud v2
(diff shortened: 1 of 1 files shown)


EmilienM:
> Rather than this hack, should instead pin `golang.org/x/tools/cmd/goimports` to v0.30.0 in our `v2` branch if that's the version that supports go 1.22.

https://github.com/gophercloud/gophercloud/blob/v2/go.mod#L10
isn't it the case already?

mandre:
Here's my attempt: https://github.com/gophercloud/gophercloud/pull/3325

PR #3326, opened 2025-03-11
CI: skip TestLayer3RouterRevision on bobcat

mandre:
Due to https://bugs.launchpad.net/neutron/+bug/2101871, which causes neutron to return a 500 error.


mandre:
We can backport after https://github.com/gophercloud/gophercloud/pull/3319 merges.

PR #3327, opened 2025-03-11
Merge pull request #3209 from shiftstack/proper-service-discovery

kayrus:
Configure and respect service type aliases

Fixes #[PUT ISSUE NUMBER HERE]
This is a backport for #3209

code in this PR:

[PUT URLS HERE]

### DIFF
--- a/endpoint_search.go
+++ b/endpoint_search.go
@@ -1,5 +1,7 @@
 package gophercloud
 
+import "slices"
+
 // Availability indicates to whom a specific service endpoint is accessible:
 // the internet at large, internal networks only, or only to administrators.
 // Different identity services use different terminology for these. Identity v2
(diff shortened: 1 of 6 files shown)


mandre (approved):
Even if this PR is flagged as a major change (`EndpointOpts: old is comparable, new is not`), I can't imagine it breaking anyone's code, and I'm for having this fix in v2.

kayrus:
The mess was initially introduced by upstream openstack developers, who decided to rename long established service types. Without this fix most of the automatic functional tests against the devstack will be broken.
You can use a workarond proposed in https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1848

P.S. I agree that this fix looks not quite well, since we could not expect that `shared-file-system` will be an alias to `sharev1`. But this was a compromise to keep the backward compatibility on the v2 branch. Feel free to suggest a better solution to hunt with the hounds.

PR #3329, opened 2025-03-12
networking: add constants for statuses

mandre:
This avoids the need to define these constants in consumer code.

Router also has a status field, but there's no documentation for it, so
let's not guess what the valid statuses are.

### DIFF
--- a/openstack/networking/v2/apiversions/constants.go
+++ b/openstack/networking/v2/apiversions/constants.go
@@ -1,0 +1,7 @@
+package apiversions
+
+const (
+	StatusCurrent    = "CURRENT"
+	StatusDeprecated = "DEPRECATED"
+	StatusStable     = "STABLE"
+)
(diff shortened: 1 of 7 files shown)


mandre:
I'm leaning towards just providing the constants without changing the types, because that's going to be a pain for users as shown with the failing CI unit tests.

openstack/networking/v2/ports/types.go
```diff
@@ -0,0 +1,10 @@
+package ports
+
+type PortStatus string
+
+const (
+	PortStatusActive PortStatus = "ACTIVE"
+	PortStatusBuild  PortStatus = "BUILD"
+	PortStatusDown   PortStatus = "DOWN"
+	PortStatusError  PortStatus = "ERROR"
```
pierreprinetti, line 9: Given the package name "ports", the identifiers for these constants can be just "StatusActive" without fear of confusion.

pierreprinetti (commented):
My opinion is that these constants can be of type string without additional typing. If designed from scratch they could be numeric Go enums, but as a retrofit I don't see a lot of value balancing the lack of backwards compatibility.

stephenfin (approved):

PR #3360, opened 2025-04-30
manila: prioritize sharev2 service type

kayrus:
Fixes #3350

code in this PR:

[PUT URLS HERE


stephenfin:
https://github.com/gophercloud/gophercloud/pull/3350 is a more complicate but IMO more comprehensive fix. Do we want to proceed with that one instead?

Issue #3343, opened 2025-04-02
functional-baremetal tests being skipped.


Sharpz7:
https://github.com/gophercloud/gophercloud/actions/runs/14184943623/job/39738477648#step:8:924

I discovered fixed this in my work with openstack-exporter: https://github.com/Sharpz7/openstack-exporter/pull/1/files#diff-ab1a500087b42774d2491892f9aaead8dd1ccf03f42b7fdfb3553123b11a07a7R26

Basically, you need to set the Ironic Endpoint.

This issue covers fixing this small issue, as well as the potential tests-failures we get from software-decay effects. I am unsure how long this issue has been ongoing so that might be a little work, or a lot.

I will take this on, and have added it to my todo list :)

PR #3365, opened 2025-05-04
CI Baremetal Fixes

Sharpz7:

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -3,6 +3,7 @@ on:
   pull_request:
     paths:
       - '**baremetal**'
+      - '.github/workflows/functional-baremetal.yaml'
   schedule:
     - cron: '0 0 */3 * *'
 jobs:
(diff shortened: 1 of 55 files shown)


Sharpz7:
Okay, so definitely some issues to work through. But, hopefully its not too complicated.

script/acceptancetest
```diff
@@ -22,5 +22,33 @@ if [[ -z "${LOG_DIR}" ]]; then
...
+    CURRENT_TIME=$(date +%s)
+    ELAPSED_TIME=$((CURRENT_TIME - START_TIME))
+    if [ "$ELAPSED_TIME" -gt "$TIMEOUT" ]; then
+      echo "Timeout reached. Ironic endpoint is not available after 2 minutes."
+      break
+    fi
+    echo "Ironic endpoint not yet available. Waiting..."
+    sleep 5
+  done
+fi
```
jayofdoom, line 50: ./script/acceptancetest: line 49: syntax error near unexpected token `done'

jayofdoom: I think you're missing a do? Add the ;do to the end of line. Plus unsure about until and the curl being on separate lines. Just generally the bash formatting looks a little weird here. Typically I'd expect this to look more like:

cmd="that long curl command"
until $cmd | grep -q '20[0-9]'; do
  [body]
done

until $cmd | grep -q '20[0-9]`
  do
    [body]
  done
is also OK but less common

PR #3368, opened 2025-05-07
CI: Add Epoxy, drop Bobcat

stephenfin:
Per $subject. I will split this into multiple PRs if it ends up with loads of failures (like I sadly suspect it will).


pierreprinetti (approved):

pierreprinetti (changes requested):
Maybe this can too run on 24.04?

stephenfin:
> Maybe this can too run on 24.04?

We can't yet. From the commit message:

> We opt to run on Jammy for now due to to [1]. That's okay since Jammy is                                                                                                                                                                                                                                                                                                                
> a supported host for Epoxy [2]. We should migrate once [1] has been                                                                                                                                                                                                                                                                                                                     
> resolved, however.                                                                                                                                                                                                                                                                                                                                                                      
>                                                                                                                                                                                                                                                                                                                                                                                         
> [1] https://bugs.launchpad.net/ubuntu/+source/apparmor/+bug/2065685                                                                                                                                                                                                                                                                                                                     
> [2] https://governance.openstack.org/tc/reference/runtimes/2025.1.html

I've fixed this issue on `master` [here](https://review.opendev.org/c/openstack/devstack/+/949510). I've just proposed a backport to `stable/2025.1` [here](https://review.opendev.org/c/openstack/devstack/+/950289) but that could be a while getting in. I'd rather come back to this later.

EmilienM (approved):

PR #3381, opened 2025-05-14
tests: Prefer extension check over environment variable

stephenfin:


stephenfin:
I don't know if these are equivalent. We're going to find out shortly.

stephenfin:
@kayrus CC

EmilienM (approved):

PR #3395, opened 2025-05-23
Fix potential panics

kayrus:
Fixes potential panic, when `BuildQueryString` returns nil url.

### DIFF
--- a/openstack/loadbalancer/v2/monitors/requests.go
+++ b/openstack/loadbalancer/v2/monitors/requests.go
@@ -45,10 +45,7 @@ type ListOpts struct {
 // ToMonitorListQuery formats a ListOpts into a query string.
 func (opts ListOpts) ToMonitorListQuery() (string, error) {
 	q, err := gophercloud.BuildQueryString(opts)
-	if err != nil {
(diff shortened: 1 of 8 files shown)


pierreprinetti:
At first sight, this looks like a no-no. What panic is this fixing?

kayrus:
I standardized the `ListOpts` interfaces to follow a consistent pattern. During this process, I noticed that `BuildQueryString` can return a `nil` URL. While I can't envision a realistic scenario where `q.String()` would panic, this change is primarily cosmetic.

stephenfin (approved):

stephenfin:
Wait, I missed @pierreprinetti's comment and can't remove my approval. Please let's ensure Pierre's happy before we merge this, if possible.

kayrus:
> I'd rather change the other functions to gracefully handle `BuildQueryString` returning a `nil` alonside a non-nil error.

this was my first attempt, but I thought it will be better to cover this in core.

stephenfin:
This appears stalled, so I'm going to close this for now. We can always reopen if necessary.

PR #3097, opened 2024-07-10
networks: Security groups rules List to accept a Builder

kayrus:
Prior to starting a PR, please make sure you have read our
[contributor tutorial](https://github.com/gophercloud/gophercloud/tree/master/docs/contributor-tutorial).

Prior to a PR being reviewed, there needs to be a Github issue that the PR
addresses. Replace the brackets and text below with that issue number.

Fixes #[PUT ISSUE NUMBER HERE]

Links to the line numbers/files in the OpenStack source code that support the
code in this PR:

[PUT URLS HERE]

This is a foillow-up for the #3070

PR #3070, opened 2024-06-14
networks: Security groups and sg rules List to accept a Builder

pierreprinetti:
Allow clients to pass a custom ListOptsBuilder to the List functions for both Security groups and Security group rules.

PR #3375, opened 2025-05-12
[v2] core: add missing Builder interfaces

kayrus:
Manual cherry-pick of the #3363 PR

PR #3397, opened 2025-05-23
[v2] Adjust List func to accept a Builder in tenants, routers and security groups packages

kayrus:
**Manual Backport:** https://github.com/gophercloud/gophercloud/pull/3097

This is a follow-up for the #3070

code in this PR:

[PUT URLS HERE]


stephenfin (approved):
As noted previously at https://github.com/gophercloud/gophercloud/pull/3375#issuecomment-2872647332, this is flagged as a major API change but isn't actually. Tests are happy. LGTM.

PR #3406, opened 2025-06-02
[WIP] Add support for blazar service

### LINKED ISSUE #3304: RFE: Support for blazar service
We're (Chameleon) heavy users of Openstack Blazar, and I wanted to propose adding support for it here, in order to consume gophercloud downstream prometheus exporters and terraform.

How "complete" would an implementation need to be? [From the blazar APIs](https://docs.openstack.org/api-ref/reservation/v1/index.html), even supporting only `GET v1/leases` and `GET v1/os-hosts` would be useful.

Previously this came up here: https://github.com/gophercloud/gophercloud/pull/316, but it requires the user to specify a reservation UUID manually; consuming the Blazar API would allow more flexible use in terraform such as "get a reservation ID from one of my active leases".

Thanks!
-Mike

winiciusallan:
This PR is still in progress; there are some tasks to complete yet.

This PR introduces:
* Reservation API support with some operations, mainly about leases;

Fixes #3304

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -492,3 +492,9 @@ func NewWorkflowV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOp
 func NewPlacementV1(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
 	return initClientOpts(client, eo, "placement")
 }
+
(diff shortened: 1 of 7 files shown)


winiciusallan:
If anyone wants to continue this work, feel free to rebase and propose a new PR.

PR #3415, opened 2025-06-05
tests: bump devstack-action

kayrus:
Fixes apt package installation, see https://github.com/gophercloud/devstack-action/pull/32


pierreprinetti:
> a new tag in devstack-action would be great so other people in the community can benefit from a new release. thanks

do you not have write permissions there?

mandre (approved):
60ca1042045c0c9e3e001c64575d381654ffcba1 is equivalent to tag v0.19.

PR #3420, opened 2025-06-11
Refine endpoint discovery

stephenfin:
This is a follow-up for #3351. In that PR we added support for discovering API versions. To do this, we added a dependency on the `version` and `min_version` headers. These are used by services with microversion support to document the maximum and minimum API versions supported, respectively. However, not all services support API microversions: services like Glance and Designate use API versions as a signal of a new feature (or, in Glance's case, as a signal that a feature is not enabled), others like Keystone just support a single version, while Neutron does its own thing with API extensions. Given this fact, relying on these fields is a mistake.

Instead, we should attempt to infer the version from the URL and only attempt version discovery if we need more information. If we do attempt version discovery, then we should be relying on the `id` field which, per the [api-sig guidelines][1], should be the API major version [3]. We continue parsing the microversion-related headers, since it will be useful later on if/when we want to do versioned discovery.

> [!NOTE]
> This appears with a `semver:major` label, but in reality the function it's modifying has not been released/backported yet so it's not actually.

[1]: https://specs.openstack.org/openstack/api-sig/guidelines/discoverability.html


openstack/utils/testing/discovery_test.go
```diff
@@ -11,162 +10,144 @@ import (
...
+	}{
 		{
 			// v2 does not support microversions and returns error
-			Endpoint:    fakeServer.Endpoint() + "compute/v2/",
-			ExpectedMax: "",
-			ExpectedMin: "",
-			ExpectedErr: true,
+			name:        "compute legacy endpoint",
+			endpoint:    fakeServer.Endpoint() + "compute/v2/",
+			expectedErr: "not supported",
```
mandre, line 319: Nit: it's weird that we expect an error when the canned result for the `compute/v2/` call returns an endpoint [marked as `SUPPORTED`](https://github.com/gophercloud/gophercloud/pull/3420/commits/aa20c7bd33abbd41778abc98f65b2d5bf2d406ca#diff-1402a85c7b7564b4150279530fede3e64be9b1040d832c8fbe31774847af59bcR44).

openstack/utils/discovery.go
```diff
@@ -46,21 +47,84 @@ type SupportedMicroversions struct {
...
+			r.Versions = versionArr
+			return nil
+		}
+	}
+
+	// case 3: we have an doubly enveloped array of version objects
+	//
+	// this is the approach used by keystone and barbican, among others, for multi-version
+	// responses
+	if data.Versions != nil {
```
mandre, line 102: Since for this case, we expect the unmarshalling to have failed [on line 91](https://github.com/gophercloud/gophercloud/pull/3420/commits/fb9337678a54cf0d194712a248b57d15785dc441#diff-86dd577329d6ae953409ee48e3e70996dc1418550a8c6be1c6ead716672fb972R91), should we move this in the same `if data.Versions != nil` conditional [on line 95](https://github.com/gophercloud/gophercloud/pull/3420/commits/fb9337678a54cf0d194712a248b57d15785dc441#diff-86dd577329d6ae953409ee48e3e70996dc1418550a8c6be1c6ead716672fb972R95)? Not sure what reads better.

.github/workflows/functional-baremetal.yaml
```diff
@@ -2,6 +2,13 @@ name: functional-baremetal
 on:
   pull_request:
     paths:
+      - 'openstack/auth_env.go'
+      - 'openstack/client.go'
+      - 'openstack/endpoint.go'
+      - 'openstack/endpoint_location.go'
+      - 'openstack/config/provider_client.go'
+      - 'openstack/utils/choose_version.go'
+      - 'openstack/utils/discovery.go'
```
mandre, line 11: Isn't it a bit fragile to list all files individually? Should we consider using a regex? Also, how about files in `openstack/common` or `openstack/utils/base_endpoint.go`?

What do you think of the following list?
```suggestion
      - 'openstack/*.go'
      - 'openstack/common/**'
      - 'openstack/config/**'
      - 'openstack/utils/**'
```

mandre (commented):
What a can of worms! Makes you appreciate the value of consistency even more. Thanks a lot for navigating this mess and put it in order.

Thanks also for updating all the workflows to trigger CI jobs when updating common files. This will be useful to avoid breaking main for next time we'll make a global change like this one.

.github/workflows/functional-baremetal.yaml
stephenfin: It is fragile, but I had intentionally avoided doing what you've suggested because I wanted to avoid running functional tests on changes to `*_test.go` files. What do you think?

openstack/utils/discovery.go
stephenfin: Yes, I kept it this way so that the "cases" are as flat as possible. It's an extra conditional but IMO it reads better.

openstack/utils/testing/discovery_test.go
stephenfin: Note that these are tests for fetching a microversion, not for fetching API information. The test for this same endpoint above in `TestGetServiceVersions` correctly returns the list of supported APIs from this endpoint.

.github/workflows/functional-baremetal.yaml
mandre: I'm of the opinion that more CI is better than not enough. There aren't too many test files, and the probability we change them without also changing the code it tests is pretty low:
```
❯ ls -1 ./openstack/*_test.go ./openstack/(common|config|utils)/**/*_test.go 
./openstack/client_test.go
./openstack/common/extensions/testing/requests_test.go
./openstack/config/clouds/clouds_test.go
./openstack/utils/testing/base_endpoint_test.go
./openstack/utils/testing/choose_version_test.go
```

However, I'm not going to block on this.

kayrus (approved):

PR #3421, opened 2025-06-12
Fix documentation for gateway_ip in subnet update

mandre:
The neutron documentation was wrong, and not providing a gateway_ip is a no-op for subnet update.

Neutron docs being fixed at [1].

[1] https://review.opendev.org/c/openstack/neutron-lib/+/952428


stephenfin (approved):
CI failures are unrelated so this lgtm

openstack/networking/v2/subnets/requests.go
```diff
@@ -199,8 +199,7 @@ type UpdateOpts struct {
 	// AllocationPools are IP Address pools that will be available for DHCP.
 	AllocationPools []AllocationPool `json:"allocation_pools,omitempty"`
 
-	// GatewayIP sets gateway information for the subnet. Setting to nil will
-	// cause a default gateway to automatically be created. Setting to an empty
+	// GatewayIP sets gateway information for the subnet. Setting to an empty
```
pierreprinetti, line 202: >Setting to an empty will cause the subnet to be created with no gateway

Maybe since we're here we can rephrase: this is `UpdateOpts` and we are not creating

mandre: Done

PR #3424, opened 2025-06-13
neutron: add more quotas members

kayrus:
An addition to #1742


stephenfin:
I'd be personally in favour of the former, if this is possible. Any extension can define its own quota, so this could quickly get out of hand otherwise. My 2c

kayrus:
I shared my thoughts on moving this into a new extension in https://github.com/gophercloud/gophercloud/pull/3447#issue-3192330597. Using pointers for these struct members should be safe and should not introduce issues. Creating separate packages for each individual option, however, could add unnecessary complexity and negatively impact the developer experience for SDK users.

PR #3426, opened 2025-06-14
neutron: fix BGP peers and speakers struct member types

kayrus:
During the work on the the https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1892 PR I discovered that speakers create opts:
* don't support networks
* local AS must have an int type
* support tenant id
* AdvertiseFloatingIPHostRoutes and AdvertiseTenantNetworks must be a pointer to allow setting `false` values

peers create opts:
* support tenant id

Also update opts must contain pointers in order to allow updates to an empty string or false boolean value.


mandre (commented):
Same comment as for https://github.com/gophercloud/gophercloud/pull/3427. I think this should a v3 addition.

PR #3427, opened 2025-06-14
neutron: support security group empty name update

kayrus:
It was not possible to update the security group name

### DIFF
--- a/internal/acceptance/openstack/networking/v2/extensions/security_test.go
+++ b/internal/acceptance/openstack/networking/v2/extensions/security_test.go
@@ -39,7 +39,7 @@ func TestSecurityGroupsCreateUpdateDelete(t *testing.T) {
 	var name = "Update group"
 	var description = ""
 	updateOpts := groups.UpdateOpts{
-		Name:        name,
+		Name:        &name,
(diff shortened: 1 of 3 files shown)


mandre (commented):
Since I had to look it up, neutron effectively allows empty security group nammes. The only validation is that the name is different from `default`:

https://github.com/openstack/neutron/blob/39b183645e071f9c2926f0bd870723fb6e300cb3/neutron/extensions/securitygroup.py#L223-L227

I'm a little less sure about the backport, it changes the API and will require everyone to change their code. What's the use case for empty names?

kayrus:
@mandre it's the matter of consistency. if it's possible to update the name to an empty string using API, it should be possible to do this in gophercloud.

mandre:
> @mandre it's the matter of consistency. if it's possible to update the name to an empty string using API, it should be possible to do this in gophercloud.

I'm all for consistency. My point is that backporting this change to v2 is going to break our API contract and affect many users. I wonder if it's not time to cut a new v3 from main very soon.

kayrus:
Then let's not backport this.

mandre (approved):

PR #3108, opened 2024-07-16
Randomize test order for unit tests

stephenfin:
I don't know why this isn't the default behavior, but relying on test order is a bad idea (TM) that we should do our best to avoid.

This is currently only done for unit tests so that we can actually get something merged. We can extend to acceptance tests in a subsequent PR.

PR #3434, opened 2025-06-19
[v2] Randomize test order for unit tests

stephenfin:
This is manual backport of #3108, with trivial conflicts on the penultimate patch addressed.

This is not critical, but backporting it will ease future backports from the `main` branch.


stephenfin:
idk if we want to include https://github.com/gophercloud/gophercloud/pull/3434/commits/fcf44f40913455ddab65c25c43a942dde1fa91c3 here. It's technically a breaking change but it's also a bugfix. Input welcome.

kayrus:
I've never seen that someone used direct access to `HypervisorPage`. Usually it is used like `HypervisorPage.AllPages(ctx)`.

stephenfin:
> I've never seen that someone used direct access to `HypervisorPage`. Usually it is used like `HypervisorPage.AllPages(ctx)`.

In that case, I think we're safe to merge this.

kayrus (approved):

PR #3435, opened 2025-06-19
[v2] Add versioned endpoint discovery

stephenfin:
This is a manual backport of #3351 and #3420, with #3108 also included to avoid tricky merge conflicts. A backport of #3108 is proposed separately at #3434 to ease reviews. We should probably merge that first and then come back to rebase and merge this.

### DIFF
--- a/openstack/client.go
+++ b/openstack/client.go
@@ -163,7 +163,7 @@ func v2auth(ctx context.Context, client *gophercloud.ProviderClient, endpoint st
 		}
 	}
 	client.EndpointLocator = func(opts gophercloud.EndpointOpts) (string, error) {
-		return V2Endpoint(ctx, client, catalog, opts)
+		return V2Endpoint(context.TODO(), client, catalog, opts)
 	}
 
 	return nil
@@ -284,7 +284,7 @@ func v3auth(ctx context.Context, client *gophercloud.ProviderClient, endpoint st
 		}
 	}
 	client.EndpointLocator = func(opts gophercloud.EndpointOpts) (string, error) {
-		return V3Endpoint(ctx, client, catalog, opts)
+		return V3Endpoint(context.TODO(), client, catalog, opts)
 	}
 
 	return nil


kayrus:
we hit this issue in https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1929
any objections on merging this? this will also require #3442

stephenfin:
> we hit this issue in [terraform-provider-openstack/terraform-provider-openstack#1929](https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1929) any objections on merging this? this will also require #3442 backport

Not from me. I'm happy to merge this now. It's been stable on `main`.

kayrus:
> This is already what we do

we do this in master branch and I have concerns for having this in the final release or in v2.

kayrus:
@stephenfin can you rebase this PR on v2 to ease the review?

kayrus:
@stephenfin #3465 and #3466 are merged, please rebase

stephenfin:
> @stephenfin can you add #3442 into this PR and make a rebase one more time?

I rebased, but can we add #3442 in here? That looks like a breaking change?

> regarding the redundant API calls, this is what I mean, see a terraform example:
> 
> ```hcl
> data "openstack_networking_port_v2" "port_1" {
>   count= 100
>   name = "port_1"
> }
> ```
> 
> dev provider compiled with gophercloud with version discovery:
> 
> ```
> $ TF_LOG=DEBUG terraform apply 2>&1 | grep 'GET ' | wc -l
> 200
> ```
> 
> recent release provider:
> 
> ```
> $ TF_LOG=DEBUG terraform apply 2>&1 | grep 'GET ' | wc -l
> 100
> ```
> 
> UPD: if you still insist on discovery, then we need to reopen a #1987 to have a way to create a client only once and clone it for each terraform resource (to properly support microversion)

I don't see any way to avoid discovery: it's what OpenStack insists on. However, yes, we should be caching the discovery documents. I can work on this.

stephenfin:
> Could we hardcode `context.TODO()` instead of using a provided `ctx` argument (only on v2 of course)? It's not ideal, but I assume any timeouts will be handled elsewhere

I've done this.

PR #3437, opened 2025-06-19
Prepare for merge queues

stephenfin:
We'd like to enable [merge queues](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue) so that we can ensure all jobs are run on a PR prior to merging that PR. To do this, we need to mark most of our jobs as *required*. However, we are currently only running those jobs on changes to specific files e.g.:

```yaml
name: functional-baremetal
on:
  pull_request:
    paths:
      - 'openstack/auth_env.go'
      - 'openstack/client.go'
      - 'openstack/endpoint.go'
      - 'openstack/endpoint_location.go'
      - 'openstack/config/provider_client.go'
      - 'openstack/utils/choose_version.go'
      - 'openstack/utils/discovery.go'
      - '**baremetal**'
      - '.github/workflows/functional-baremetal.yaml'
jobs:
    # ... rest of job definition
```

However, this causes jobs to never run rather than be "skipped". If the job is required for e.g. a merge queue, then the PR will never merge because it will wait forever for the job result.

This PR adds a local plugin to allow us to filter on files and skip steps when none of the affected files are checked. This only happens on pull requests, since during scheduled jobs we expect all jobs to run. Once this merges, we can mark all but the ironic and octavia master jobs as required.

### DIFF
--- a/.github/actions/file-filter/Dockerfile
+++ b/.github/actions/file-filter/Dockerfile
@@ -1,0 +1,5 @@
+FROM python:3.11-slim
+WORKDIR /app
+RUN pip install 'PyGithub~=2.7.0' 'requests~=2.32'
+COPY file_filter_action.py ./
+ENTRYPOINT ["python", "/app/file_filter_action.py"]
--- a/.github/actions/file-filter/action.yml
+++ b/.github/actions/file-filter/action.yml
@@ -1,0 +1,24 @@
+name: 'File Filter'
+description: 'Filter PR files by glob patterns and return true/false if any files match'
+inputs:
+  patterns:
+    description: 'String containing one or more glob patterns separated by spaces'
+    required: true
+  exclude:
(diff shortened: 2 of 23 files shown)


mandre (approved):
This commit doesn't enable merge queues, but prepares for enabling merge queues. We'll still have to configure it in the branch protection as [explained here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue#managing-a-merge-queue).

As I understand it, we also need to start making more jobs required for it to have any meaningful effect.

mandre (changes requested):
Actually, we'll also need to trigger the `hold` workflow on `merge_group` events, since this job is required.

stephenfin:
This will be a good deal easier to do once https://github.com/gophercloud/gophercloud/pull/3490 is landed.

PR #3442, opened 2025-06-26
core: pass context to the provider.EndpointLocator func

### LINKED ISSUE #3441: core: endpoint locator doesn't use active context to resolve microversions
The following func uses the context, provided during the initialization, which causes `context canceled` when the original context is done.

https://github.com/gophercloud/gophercloud/blob/c3046e033334224f7732c220b552ecb56f040f86/openstack/client.go#L286-L288

Caught by https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1899

proper func signature should be:

```go
        client.EndpointLocator = func(ctx context.Context, opts gophercloud.EndpointOpts) (string, error) {
                return V3Endpoint(ctx, client, catalog, opts)
        }
```

cc @stephenfin

kayrus:
Fixes #3441


kayrus:
@stephenfin @pierreprinetti @mandre kindly ping

stephenfin (approved):

kayrus:
Since not all tests are green, I'd like to receive one more approval before the merge

stephenfin:
The messaging failures are expected since we are not handling paths as we should. I'm looking into this. The ironic failures are also known: @Sharpz7 is looking into it. The Octavia failures are new though. I'll look at this too.

PR #3447, opened 2025-07-01
neutron: add new router attributes

kayrus:
Add new extension fields. Not sure if to put them as separate extensions, since other extension fields like `distributed` or `availability_zone_hints` were already implemented directly inside the routers package.

Likewise in #3424, if user doesn't specify these parameters, they won't be sent to API. Therefore from my perspective it is safe to keep them here, otherwise the consumer code complexity will be increased.

### DIFF
--- a/openstack/networking/v2/extensions/layer3/routers/requests.go
+++ b/openstack/networking/v2/extensions/layer3/routers/requests.go
@@ -77,14 +77,20 @@ type CreateOptsBuilder interface {
 // CreateOpts contains all the values needed to create a new router. There are
 // no required values.
 type CreateOpts struct {
-	Name                  string       `json:"name,omitempty"`
-	Description           string       `json:"description,omitempty"`
-	AdminStateUp          *bool        `json:"admin_state_up,omitempty"`
-	Distributed           *bool        `json:"distributed,omitempty"`
-	TenantID              string       `json:"tenant_id,omitempty"`
-	ProjectID             string       `json:"project_id,omitempty"`
-	GatewayInfo           *GatewayInfo `json:"external_gateway_info,omitempty"`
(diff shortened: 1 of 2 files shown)


openstack/networking/v2/extensions/layer3/routers/requests.go
```diff
@@ -77,14 +77,20 @@ type CreateOptsBuilder interface {
...
+	Distributed            *bool        `json:"distributed,omitempty"`
+	TenantID               string       `json:"tenant_id,omitempty"`
+	ProjectID              string       `json:"project_id,omitempty"`
+	HA                     *bool        `json:"ha,omitempty"`
+	ServiceTypeID          string       `json:"service_type_id,omitempty"`
+	FlavorID               string       `json:"flavor_id,omitempty"`
+	EnableNDPProxy         *bool        `json:"enable_ndp_proxy,omitempty"`
+	EnableDefaultRouteBFD  *bool        `json:"enable_default_route_bfd,omitempty"`
+	EnableDefaultRouteECMP *bool        `json:"enable_default_route_ecmp,omitempty"`
+	GatewayInfo            *GatewayInfo `json:"external_gateway_info,omitempty"`
```
stephenfin, line 92: Could you separate these and call out the extensions that add them? Alternatively, could we add a new struct tag that would allow us to e.g. check if an extension is enabled before sending/attemtpting to retrieve a field?

stephenfin:
> Likewise in #3424, if user doesn't specify these parameters, they won't be sent to API. Therefore from my perspective it is safe to keep them here, otherwise the consumer code complexity will be increased.

This is a fair point. I've said this inline, but I wonder if we could streamline things via a custom struct tag?

PR #3448, opened 2025-07-02
neutron: allow omission of subnet_id for IP address

kayrus:
I could not find any docs about `external_fixed_ip` struct to support omitted `subnet_id`, but I created test case in terraform provider and it works as expected, see https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1915

As for the port's `subnet_id`, the omission behavior is documented [here](https://docs.openstack.org/api-ref/network/v2/#create-port):

> The IP addresses for the port. If you would like to assign multiple IP addresses for the port, specify multiple entries in this field. Each entry consists of IP address (ip_address) and the subnet ID from which the IP address is assigned (subnet_id).
> If you specify both a subnet ID and an IP address, OpenStack Networking tries to allocate the IP address on that subnet to the port.
> If you specify only a subnet ID, OpenStack Networking allocates an available IP from that subnet to the port.
> If you specify only an IP address, OpenStack Networking tries to allocate the IP address if the address is a valid IP for any of the subnets on the specified network.

### DIFF
--- a/openstack/networking/v2/extensions/layer3/routers/results.go
+++ b/openstack/networking/v2/extensions/layer3/routers/results.go
@@ -21,7 +21,7 @@ type GatewayInfo struct {
 // router.
 type ExternalFixedIP struct {
 	IPAddress string `json:"ip_address,omitempty"`
-	SubnetID  string `json:"subnet_id"`
+	SubnetID  string `json:"subnet_id,omitempty"`
 }
 
 // Route is a possible route in a router.
--- a/openstack/networking/v2/ports/results.go
+++ b/openstack/networking/v2/ports/results.go
@@ -49,7 +49,7 @@ type DeleteResult struct {
 
 // IP is a sub-struct that represents an individual IP.
 type IP struct {
-	SubnetID  string `json:"subnet_id"`
+	SubnetID  string `json:"subnet_id,omitempty"`
 	IPAddress string `json:"ip_address,omitempty"`
 }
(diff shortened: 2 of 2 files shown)


mandre (commented):
I don't think this has any effect. You're adding an omitempty for fields in structs that are returned to the user, but [according to the docs](https://docs.openstack.org/api-ref/network/v2/#id69), the subnet ID is always populated:
> The IP addresses for the port. If the port has multiple IP addresses, this field has multiple entries. Each entry consists of IP address (ip_address) and the subnet ID from which the IP address is assigned (subnet_id).

That said, the change is not harmful, and it doesn't really makes sense to me why `IPAddress` had an `omitempty` tag, so... I guess we can add one for `SubnetID` for consistency?

kayrus:
> You're adding an omitempty for fields in structs that are returned to the user

not really, this struct is used in `CreateOpts` and `UpdateOpts` as well and these tags are taken into account in json marshaller.

mandre (approved):
The neutron docs for the router creation states that:

> The external gateway information of the router. If the router has an external gateway, this would be a dict with network_id, enable_snat, external_fixed_ips and qos_policy_id. Otherwise, this would be null.

However, it doesn't say what format `external_fixed_ips` must have. I think it's fair to add an omitempty for the subnet ID. And you have test cases to show that `SubnetID` is indeed optional, so all good.

PR #3449, opened 2025-07-03
Fixing Baremetal CI

Sharpz7:

Continuation from https://github.com/gophercloud/gophercloud/pull/3383.

Somepoint between May 14th and July 2nd, something changed where baremetal tests started to (mostly) work.
- Master is broken, not sure why
- Noauth + httpbasic tests all get skipped: My guess is there was a behavior change.

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -52,14 +52,11 @@ jobs:
             enable_plugin ironic https://github.com/openstack/ironic ${{ matrix.openstack_version }}
             LIBS_FROM_GIT=pyghmi,virtualbmc
             FORCE_CONFIG_DRIVE=True
-            DEFAULT_INSTANCE_TYPE=baremetal
             Q_AGENT=openvswitch
             Q_ML2_TENANT_NETWORK_TYPE=vxlan
             Q_ML2_PLUGIN_MECHANISM_DRIVERS=openvswitch
             DEFAULT_INSTANCE_TYPE=baremetal
             OVERRIDE_PUBLIC_BRIDGE_MTU=1400
-            IPV4_ADDRS_SAFE_TO_USE=10.0.0.0/22
-            FLOATING_RANGE=172.24.4.0/24
             VIRT_DRIVER=ironic
             BUILD_TIMEOUT=1800
             SERVICE_TIMEOUT=90
@@ -75,7 +72,7 @@ jobs:
             IRONIC_RAMDISK_TYPE=tinyipa
             IRONIC_TEMPEST_BUILD_TIMEOUT=720
             IRONIC_TEMPEST_WHOLE_DISK_IMAGE=False
-            IRONIC_VM_COUNT=0
+            IRONIC_VM_COUNT=1
             IRONIC_VM_EPHEMERAL_DISK=1
             IRONIC_VM_LOG_DIR=/opt/stack/new/ironic-bm-logs
             IRONIC_VM_SPECS_RAM=1024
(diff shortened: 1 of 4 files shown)


Sharpz7:
> > @stephenfin - I think this code is really old. I could totally search through now and find exactly where noauth/httpbasic was broken, but since v1 is functional and correctly uses auth (, I think this is okay to merge? (When master is fixed?)
> 
> Sorry, I don't understand the ask here. The CI is still failing: we hardly want to merge this as-is?

Sorry, this PR currently has a typo... I'm going to fix it today and it'll pass.

stephenfin:
I don't think we can deploy `master` on Ubuntu 22.04: it's not [a supported runtime for 2025.2 (Flamingo)](https://governance.openstack.org/tc/reference/runtimes/2025.2.html). While it gets things passing right now, there's a good chance this will break at some point in the near future as DevStack removes/inadvertently breaks support for Focal.

TBH, I suspect the issue here is the Ironic's devstack-plugin. We must be going off the beaten track wrt what Ironic does in their CI. My guess is that it's something to do with the networking configuration (perhaps use of OVN in one place and OVS in another). Would you be able to compare the `local.conf` we use versus those used in the upstream Ironic gates to see if there are any obvious differences? I will do so at some point myself if not, but it's just a case of finding the time.

stephenfin:
Thanks.

Just a case in point RE: Focal vs Noble: https://review.opendev.org/c/openstack/devstack/+/952839

Sharpz7:
Assuming the devstack plugin is the folder named "devstack" in the main ironic repo: https://github.com/openstack/ironic/compare/master...stable/2025.1

I do not see any changes that would suggest a fail. Likely, I need to go further back.

Sharpz7:
https://github.com/openstack/ironic/blob/stable/2025.1/devstack/lib/ironic#L1680
https://github.com/openstack/ironic/blob/master/devstack/lib/ironic#L1689

https://github.com/openstack/ironic/compare/master...stable/2025.1#diff-9b38d7edf61674b2f99d7772ea8d10a9183418d74899c8d12a31a3446089c920L337

Diff doesn't seem to show a change that does exist?

Sharpz7:
https://github.com/openstack/ironic/compare/stable/2025.1..master#diff-9b38d7edf61674b2f99d7772ea8d10a9183418d74899c8d12a31a3446089c920R2583

This seems to be the culprit, investigating now

stephenfin:
@Sharpz7 Apologies if I've missed something, but you seem to be flapping on this, so I'm offering my 2c. If *I* were to approach this, I would take the following approach:

# Approach A

First, deploy `stable/dalmatian` DevStack with the `stable/dalmatian` Ironic DevStack plugin and Ironic, using a `local.conf` from a recent successful job. Run at least one of the failing-on-master Gophercloud jobs. It should pass. This is to level set and ensure you can get tests running. Skip this if you've already achieved this.

Once done, deploy `stable/epoxy` DevStack with the `stable/dalmatian` Ironic plugin but `stable/epoxy` Ironic. This is an untested configuration but *should* work. Run at least one of the failing-on-master Gophercloud jobs. It should pass. This is to prove that the Ironic DevStack plugin is the culprit.

After this, I would likely rely on Ansible or (more realistically) a terminal running on a separate screen to effectively do a git bisect across all changes to the DevStack plugin, deploying DevStack multiple times until you identity the commit that breaks things. At this point, it's likely time to get the rest of the Ironic folks installed.

# Approach B

Do what I did in https://github.com/gophercloud/gophercloud/pull/3449#issuecomment-3057688290 and pull configuration from a basic tempest job in Ironic upstream. Deploy that and see if our tests pass. If so, start iterating on that configuration until the tests start failing. That should give us our smoking gun.

PR #3451, opened 2025-07-08
blockstorage: add isPublic query option for volume types

### LINKED ISSUE #3450: blockstorage: IsPublic list option is missing
The `IsPublic` query parameter to list private and public volume types is currently missing.
The support for this option is listed in the documentation: https://docs.openstack.org/api-ref/block-storage/v3/index.html#volume-types-types

Only public volume types are listed by default. To query private volume types, the IsPublic parameter must be provided in the OpenStack query.

VoigtS:
Fixes #3450

code in this PR:
https://docs.openstack.org/api-ref/block-storage/v3/index.html#volume-types-types

### DIFF
--- a/openstack/blockstorage/v3/volumetypes/requests.go
+++ b/openstack/blockstorage/v3/volumetypes/requests.go
@@ -73,9 +73,9 @@ type ListOptsBuilder interface {
 // ListOpts holds options for listing Volume Types. It is passed to the volumetypes.List
 // function.
 type ListOpts struct {
-	// Specifies whether to request public or private types.
-	// Set to "none" to retrieve both types. Defaults to public only.
-	IsPublic string `q:"is_public"`
+	// Specifies whether the query should include public or private Volume Types.
+	// By default, it queries both types.
+	IsPublic visibility `q:"is_public"`
 	// Comma-separated list of sort keys and optional sort directions in the
 	// form of <key>[:<direction>].
 	Sort string `q:"sort"`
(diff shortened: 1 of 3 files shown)


kayrus (approved):

openstack/blockstorage/v3/volumetypes/requests.go
```diff
@@ -73,6 +73,9 @@ type ListOptsBuilder interface {
 // ListOpts holds options for listing Volume Types. It is passed to the volumetypes.List
 // function.
 type ListOpts struct {
+	// Specifies whether to request public or private types.
+	// Set to "none" to retrieve both types. Defaults to public only.
+	IsPublic string `q:"is_public"`
```
kayrus: this must be bool. if it defaults to `true` by default, then it must be a bool pointer. would be nice to have a unit test, see https://github.com/gophercloud/gophercloud/blob/f98c2f934acfdb810509ec0af47ba777d8537940/openstack/sharedfilesystems/v2/shares/requests.go#L130 and https://github.com/gophercloud/gophercloud/blob/f98c2f934acfdb810509ec0af47ba777d8537940/openstack/blockstorage/v2/transfers/testing/fixtures_test.go#L200 for example

VoigtS: Thank you for the hint. Making this parameter a bool pointer is a good idea. However, when its value is `nil`, the BuildQueryString function ignores this parameter. This query is used for every OpenStack API implementation.

Without modifying this function and potentially causing side effects, translating this to the Python type `None` would require manually modifying the query string itself, which is an ugly hack. Therefore, I opted to create a string enum that can be used at the caller side to limit the query results. I'm open to discussing the naming of the variables (I don't think they are quite optimal yet; for example, if IsPublic has the value VisibilityPublic).

Another change is that now, by default, both public and private volume types are queried (when no ListOpts struct is provided). This aligns with the OpenStack client, which uses the parameter `?is_public=None` by default. This can be verified by enabling the debug functionality of the client for the command: `volume type list`.

kayrus (approved):
LGTM, thanks!

PR #3457, opened 2025-07-15
script: Improve getenvvar helper

stephenfin:
Make this script a little more useful, based on my personal experiences using it.

* Handle multiple `clouds.yaml` files
* Use `OS_CLOUDS` if set
* Add uv dependencies metadata header

### DIFF
--- a/script/getenvvar
+++ b/script/getenvvar
@@ -1,4 +1,10 @@
 #!/usr/bin/env python3
+# /// script
+# requires-python = ">=3.13"
+# dependencies = [
+#     "pyyaml>=6",
+# ]
+# ///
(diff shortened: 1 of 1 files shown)


stephenfin:
> @stephenfin what does this script do? it's not used anywhere in gophercloud

I use it locally so that I can run integration tests (remember how I said I only have `clouds.yaml` files?). The file docstring says as much:

https://github.com/gophercloud/gophercloud/blob/20cc6619274a578bb0ff963d17d9f31a39cc0140/script/getenvvar#L3-L14

kayrus (approved):

PR #3467, opened 2025-07-17
Add issue type to issue template

pierreprinetti:
It seems supported now


pierreprinetti:
I don't think there is a definitive way to test this before merging... I'm afraid we'll have to revert if it breaks the issue configuration.

PR #3474, opened 2025-07-22
Implement update & delete traits on resource provider

### LINKED ISSUE #2420: Add support for placement resource provider trait in placement
Adding support for updating/deleting the placement resource provider traits  in Gophercloud.

There is no support for following APIs related to update resource provider traits.
https://docs.openstack.org/api-ref/placement/?expanded=update-resource-provider-traits-detail
https://docs.openstack.org/api-ref/placement/?expanded=update-resource-provider-traits-detail,delete-resource-provider-traits-detail

fwiesel:
This implements #2420.
The relevant server side code can be found here:
https://github.com/openstack/placement/blob/stable/2025.1/placement/handlers/trait.py#L216-L274

Fixes #2420.

code in this PR:

https://github.com/openstack/placement/blob/stable/2025.1/placement/handlers/trait.py#L216-L274

### DIFF
--- a/openstack/placement/v1/resourceproviders/requests.go
+++ b/openstack/placement/v1/resourceproviders/requests.go
@@ -175,3 +175,38 @@ func GetTraits(ctx context.Context, client *gophercloud.ServiceClient, resourceP
 	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
 	return
 }
+
+// UpdateTraitsOptsBuilder allows extensions to add additional parameters to the
+// UpdateTraits request.
+type UpdateTraitsOptsBuilder interface {
+	ToResourceProviderUpdateTraitsMap() (map[string]any, error)
+}
+
+// UpdateTraitsOpts represents options used to update traits of a resource provider.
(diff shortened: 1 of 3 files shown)


openstack/placement/v1/resourceproviders/requests.go
```diff
@@ -175,3 +175,43 @@ func GetTraits(ctx context.Context, client *gophercloud.ServiceClient, resourceP
...
+type UpdateTraitsOpts = ResourceProviderTraits
+
+// ToResourceProviderUpdateTraitsMap constructs a request body from UpdateTraitsOpts.
+func (opts UpdateTraitsOpts) ToResourceProviderUpdateTraitsMap() (map[string]any, error) {
+	b, err := gophercloud.BuildRequestBody(opts, "")
+	if err != nil {
+		return nil, err
+	}
+
+	return b, nil
```
kayrus: it'd be better to avoid extra vars
```suggestion
	return gophercloud.BuildRequestBody(opts, "")
```

PR #3480, opened 2025-07-31
Implement hypervisors.GetExt: Get with Query parameter

### LINKED ISSUE #3479: Compute: Get Hypervisor with server list
### What is missing?

The Show Hypervisor Details api call introduce a query parameter `with_servers` in microversion 2.53.
See: https://docs.openstack.org/api-ref/compute/#show-hypervisor-details

It is supported in List but not the `Get` request.

fwiesel:
This introduces a new function GetExt in order to avoid breaking backwards compatibility.

Fixes #3479

code in this PR:

https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/hypervisors.py#L320-L334

### DIFF
--- a/openstack/compute/v2/hypervisors/testing/fixtures_test.go
+++ b/openstack/compute/v2/hypervisors/testing/fixtures_test.go
@@ -329,6 +329,62 @@ const HypervisorGetBody = `
 }
 `
 
+// HypervisorGetPost253Body represents a raw hypervisor GET result with Pike+
+// release with optional server list
+const HypervisorGetPost253Body = `
+{
+    "hypervisor":{
+        "cpu_info":{
+            "arch":"x86_64",
+            "model":"Nehalem",
+            "vendor":"Intel",
+            "features":[
+                "pge",
+                "clflush"
+            ],
+            "topology":{
+                "cores":1,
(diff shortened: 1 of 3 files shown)


openstack/compute/v2/hypervisors/testing/fixtures_test.go
```diff
@@ -676,6 +782,16 @@ func HandleHypervisorGetSuccessfully(t *testing.T, fakeServer th.FakeServer) {
 	})
 }
 
+func HandleHypervisorGetPost253Successfully(t *testing.T, fakeServer th.FakeServer) {
+	fakeServer.Mux.HandleFunc("/os-hypervisors/"+HypervisorFake.ID, func(w http.ResponseWriter, r *http.Request) {
+		th.TestMethod(t, r, "GET")
+		th.TestHeader(t, r, "X-Auth-Token", client.TokenID)
```
kayrus, line 788: @fwiesel can you add a check whether the `with_servers=true` URI argument is passed?

fwiesel: Sure thing. I have modified the handler to respond depending on the query parameter.

PR #3485, opened 2025-08-07
Add ProviderClient.DeepCopy, ServiceClient.DeepCopy

### LINKED ISSUE #1987: Core: DeepCopy method for the provider client

stephenfin:

Fixes #1987

This is a first pass. I need to better understand what we should be copying and what we shouldn't.

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -139,6 +139,38 @@ func (f *reauthFuture) Get() error {
 	return f.err
 }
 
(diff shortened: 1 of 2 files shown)


stephenfin:
CC @kayrus

Issue #3492, opened 2025-08-13
gophercloud >= 2.7 breaks on parsing security group time


tmmorin:
## Issue

We recently ran into a situation where the following error was raised by CAPO (Cluster API Openstack provider) which uses gophercloud:

```
E0807 12:12:22.404583       1 controller.go:324] "Reconciler error" 
   err="failed to reconcile security groups: parsing time \"2024-11-26T11:03:35.521964\" 
            as \"2006-01-02T15:04:05Z07:00\": cannot parse \"\" as \"Z07:00\"" 
...
```

## Observations

This error started to appear when upgrading CAPO from 0.11.3 to 0.11.6, which among other changes upgrades gophercloud from 2.2.0 and 2.7.0.

I have observed this:
* gophercloud has code to let it accept two timeformat for security groups `updated_at`/`created_at` fields [here](https://github.com/gophercloud/gophercloud/blob/5126c5bbd89e6e1a9aa9b00a6f862433c7b9db24/openstack/networking/v2/extensions/security/groups/results.go#L52): the standard ISO8601 format and the "older neutron time format" (called RFC3339NoZ in gophercloud code)
* in this Openstack tenant the SG groups used by CAPO indeed have such timestamp that match the RFC3339NoZ format (both for the SG updated_at/created_at fields, **and for the fields of the SG rules that it includes**):
    ```
    +------------+----------------------------+
    | Field      | Value                      |
    +------------+----------------------------+
    | created_at | 2023-03-09T14:51:27.533307 |
    | updated_at | 2025-08-06T14:48:16.792728 |
    +------------+----------------------------+
    ```
* between gophercloud 2.2.0 and 2.7.0, there was one addition (in https://github.com/gophercloud/gophercloud/commit/99b0a51a254c4e3332a0117a6f2ed4461bdc99fc) that added support for `updated_at`/`created_at` for Security Group **Rules**, but without anything to accept the "older neutron time format"

## Analysis

Based on the observations above, it seems that the explanation is the following:
* gophercloud 2.2.0 parses our SGs successfully because gophercloud accepts the "older neutron time format" for Security Groups `updated_at`/`created_at`, and does not attempt to read the `updated_at`/`created_at` fields of the SG **Rules** that are embedded in Security Group resources
[...]

PR #3493, opened 2025-08-13
Networking v2: Support two time formats for subnet, router, SG rule (#3492)

tmmorin:
For the Networking v2 API, Gophercloud has support for two time formats for the `created_at`/`updated_at` fields, for **most** of the resources for which it supports such fields.  This was introduced by https://github.com/gophercloud/gophercloud/pull/1671.  Such support however is lacking for some resources (the resources for which the support for  the `created_at`/`updated_at` fields was added later, in https://github.com/gophercloud/gophercloud/commit/99b0a51a254c4e3332a0117a6f2ed4461bdc99fc)

This PR adds support for two time formats for the subnet, router, and SG rule resources.

It does it in the same way as in https://github.com/gophercloud/gophercloud/pull/1671, and adds the corresponding unit tests.


openstack/networking/v2/extensions/layer3/routers/results.go
```diff
@@ -82,10 +82,50 @@ type Router struct {
...
+	err := json.Unmarshal(b, &s1)
+	if err == nil {
+		*r = Router(s1.tmp)
+		r.CreatedAt = time.Time(s1.CreatedAt)
+		r.UpdatedAt = time.Time(s1.UpdatedAt)
+
+		return nil
+	}
+
+	// return errors.New("test error")
```
mandre: Leftover debug?

mandre (commented):
Yes! Thanks for the patch, this is exactly how we should handle different possible formats returned by OpenStack. Just a minor nit about leftover debug in each of the added `UnmarshalJSON()` functions.

The linter is also failing due to improperly formatted code. This needs fixing before we can merge.

openstack/networking/v2/extensions/layer3/routers/results.go
tmmorin: err... yes :)

tmmorin: (should be all clean now)

PR #3500, opened 2025-09-03
Fixing Baremetal CI

Sharpz7:

Continuation from https://github.com/gophercloud/gophercloud/pull/3383 and https://github.com/gophercloud/gophercloud/pull/3449.

Somepoint between May 14th and July 2nd, something changed where baremetal tests started to (mostly) work.

image

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -25,7 +25,7 @@ jobs:
             additional_services: "openstack-cli-server"
           - name: "epoxy"
             openstack_version: "stable/2025.1"
-            ubuntu_version: "22.04"
+            ubuntu_version: "24.04"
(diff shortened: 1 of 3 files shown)


Sharpz7:
We need to decide if `IRONIC_BAREMETAL_BASIC_OPS=False` is acceptable. Apparently, we don't do anything requiring it... the question is will that hold in the long term / not cost us problems later. 

cc @jayofdoom

stephenfin:
As a follow-up, it would be great to add a comment - ideally pointing to an ironic bug - explaining why we're changed `IRONIC_BAREMETAL_BASIC_OPS` to `False` (from discussions with @jayofdoom on IRC today, it seems there's a neutron-ironic race in the DevStack plugin that this option triggers).

PR #3510, opened 2025-09-17
Make functional jobs dependent on unit tests

stephenfin:
There is no point running the (expensive) functional test suite if unit tests fail. Based on [[1][1]] and [[2][2]]

[1]: https://stackoverflow.com/a/64733705
[2]: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -4,6 +4,9 @@ on:
   pull_request:
   schedule:
     - cron: '0 0 */3 * *'
+  workflow_run:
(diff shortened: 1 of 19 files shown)


stephenfin:
Man, GitHub Actions is so unintuitive :disappointed: Will have to come back to this...

stephenfin:
So apparently the `workflow_run` trigger doesn't work on PRs. There doesn't appear to be any way to do what I want without combining things into a single workflow and using `needs` :disappointed:

PR #3514, opened 2025-09-17
Added comment addressing IRONIC_BAREMETAL_BASIC_OPS=false

Sharpz7:
Addressing https://github.com/gophercloud/gophercloud/pull/3500#issuecomment-3271229855

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -57,6 +57,9 @@ jobs:
         if: ${{ fromJSON(steps.changed-files.outputs.matches) }}
         run: sudo apt-get purge -y dnsmasq-base
 
+      # NOTE(sharpz7) IRONIC_BAREMETAL_BASIC_OPS was originally set to false as it
+      # was failing (https://review.opendev.org/c/openstack/ironic/+/960299)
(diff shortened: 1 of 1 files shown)


.github/workflows/functional-baremetal.yaml
```diff
@@ -57,6 +57,9 @@ jobs:
         if: ${{ fromJSON(steps.changed-files.outputs.matches) }}
         run: sudo apt-get purge -y dnsmasq-base
 
+      # NOTE(sharpz7) IRONIC_BAREMETAL_BASIC_OPS was originally set to false as it
+      # was failing (https://review.opendev.org/c/openstack/ironic/+/960299)
+      # See https://github.com/gophercloud/gophercloud/pull/3500
```
jayofdoom, line 62: Perhaps add something to the effect of: this disables Ironic configuration of fake baremetal nodes, which should not be necessary for SDK testing.

jayofdoom: I would've preferred we be explicit about what the option does technically (avoids creating fake BM VMs) rather than putting the value judgement in the comment, but this satisfies my request well enough for a +1

stephenfin (approved):

PR #3528, opened 2025-10-02
Compute V2: implement Get console connection information.

### LINKED ISSUE #3527: Compute V2: Implement Get console connection information
Currently, Gophercloud only implements [CreateRemoteConsole](https://github.com/gophercloud/gophercloud/pull/1159). OpenStack Compute service also provides a GET endpoint to retrieve console connection information using a console token: `/os-console-auth-tokens/{console_token}`.
Both Create and Get operations are available in the Nova API, so Gophercloud should support both.

Documentation:
[API References - Server Consoles](https://docs.openstack.org/api-ref/compute/#server-consoles)

dlaw4608:
Added remoteconsoles Get method for retrieving console connection info. Includes unit and acceptance tests, and updates to documentation.

Fixes #3527

code in this PR:

API create request handler:
https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/console_auth_tokens.py#L43-L83

API definitions:
https://github.com/openstack/nova/blob/master/nova/api/openstack/compute/schemas/console_auth_tokens.py


openstack/compute/v2/remoteconsoles/results.go
```diff
@@ -28,6 +34,27 @@ type RemoteConsole struct {
...
 }
 
+// Console represents the Compute service console object.
+type Console struct {
+	// InstanceUUID contains the UUID of the server.
+	InstanceUUID string `json:"instance_uuid"`
+
+	// Host contains the name or ID of the host.
+	// (Optional)
+	Host string `json:"host"`
```
winiciusallan, line 44: Since this field is optional, we should make it a pointer, so it can be unset when unmarshalling.

openstack/compute/v2/remoteconsoles/results.go
```diff
@@ -28,6 +34,27 @@ type RemoteConsole struct {
...
+	// Port contains the port number
+	Port int `json:"port"`
+
+	// TLSPort contains the port number of a port requiring the TLS connection.
+	// (Optional)
+	TLSPort int `json:"tls_port"`
+
+	// InternalAccessPath contains the id representing the internal access path.
+	//(Optional)
+	InternalAccessPath string `json:"internal_access_path"`
```
winiciusallan, line 55: Same for those optional fields.

internal/acceptance/openstack/compute/v2/compute.go
```diff
@@ -1023,6 +1023,20 @@ func CreateRemoteConsole(t *testing.T, client *gophercloud.ServiceClient, server
 	return remoteConsole, nil
 }
 
+// GetConsole retrieves connecton information for a remote console session.
```
winiciusallan, line 1026:
```suggestion
// GetConsole retrieves connection information for a remote console session.
```

winiciusallan (commented):
Hi, Daniel. I made a few comments. Thanks for your work!

If you agree about the optional fields, I think we also should have tests for those cases.

Also, what do you think about explicitly returning an error when microversion is not compatible (or properly set)?  IIRC, the API returns 404 for endpoints that it does not support.

openstack/compute/v2/remoteconsoles/results.go
kayrus: `omitempty` tag is enough

openstack/compute/v2/remoteconsoles/results.go
kayrus: `omitempty` tag is enough

PR #3532, opened 2025-10-10
Identity: Add description field to roles 

dlaw4608:


openstack/identity/v3/roles/requests.go
```diff
@@ -86,6 +86,9 @@ type CreateOpts struct {
 	// DomainID is the ID of the domain the role belongs to.
 	DomainID string `json:"domain_id,omitempty"`
 
+	// Description is the description of the project
+	Description string `json:"description,omitempty"`
```
kayrus, line 98: Can you also add an ability to update the role description? Make it a pointer in order to be able to unset it.

dlaw4608: Done

openstack/identity/v3/roles/requests.go
```diff
@@ -133,6 +136,9 @@ type UpdateOpts struct {
 	// Name is the name of the new role.
 	Name string `json:"name,omitempty"`
 
+	// Description is the description of the new role.
```
stephenfin:
```suggestion
	// Description is an updated description for the role.
```

openstack/identity/v3/roles/requests.go
```diff
@@ -133,6 +136,9 @@ type UpdateOpts struct {
 	// Name is the name of the new role.
```
stephenfin: not really related, but seeing as you're here

```suggestion
	// Name is an updated name for the role.
```

openstack/identity/v3/roles/testing/requests_test.go
```diff
@@ -92,8 +92,9 @@ func TestCreateRole(t *testing.T) {
...
 
 	createOpts := roles.CreateOpts{
-		Name:     "support",
-		DomainID: "1789d1",
+		Name:        "support",
+		DomainID:    "1789d1",
+		Description: "read-only support role",
 		Extra: map[string]any{
 			"description": "read-only support role",
 		},
```
stephenfin, line 100: Oh, this is interesting. So it seems this allows you to inject fields in a free-form manner, including `description`, so we don't **have** to add a `description` field because we could do things exactly like how this test is doing. However, I think we *should* add the description field, and we probably want to mark this `Extra` field as deprecated in this PR (which we will likely backport) and actually remove it in a future PR (which we won't backport). We probably want to add support for the `options` field also (which is an object that supports a single key, `immutable`, which is a boolean ([source](https://docs.openstack.org/api-ref/identity/v3/index.html#create-role)))

dlaw4608: Made a new issue (#3533) to add `option` field to role, will work on this and then look to remove the `Extra` field in the future. At that point gophercloud will support all fields available from the Keystone API.

kayrus: There are extra fields in keystone, which can not be unset, see https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1771#issuecomment-2939784423
Please consider this for new PRs.

dlaw4608: Ah ok, my plan is to leave the `Extra` field in place for now, and remove the comment marking the `Extra` field as deprecated,  and get this PR merged. I will be adding the last missing field `options` from the Keystone API in a separate PR. Once Gophercloud supports all available Keystone fields for the Role resource, it would make sense to either mark the `Extra` Field as deprecated or remove the `Extra `field entirely  in the future no?,  since it won’t serve much purpose anymore.

internal/acceptance/openstack/identity/v3/roles_test.go
```diff
@@ -94,19 +94,16 @@ func TestRolesCRUD(t *testing.T) {
 
 	th.AssertEquals(t, found, true)
 
+	description := ""
 	updateOpts := roles.UpdateOpts{
-		Extra: map[string]any{
```
mandre, line 101: We should not remove test for the `Extra` field (perhaps just change the test to use another free-form name than "description").

internal/acceptance/openstack/identity/v3/roles_test.go
```diff
@@ -94,19 +94,16 @@ func TestRolesCRUD(t *testing.T) {
 
 	th.AssertEquals(t, found, true)
 
+	description := ""
```
mandre: Looks like we try to update description from an empty string to... an empty string. If we want to test that we can unset the description (if that was your intention) we should set the original value to something else.

internal/acceptance/openstack/identity/v3/roles_test.go
dlaw4608: Done. Added back in the `Extra` field using `email` instead of `description`.

internal/acceptance/openstack/identity/v3/roles_test.go
dlaw4608: Thanks, Done.

openstack/identity/v3/roles/testing/requests_test.go
stephenfin: > There are extra fields in keystone, which can not be unset, see [terraform-provider-openstack/terraform-provider-openstack#1771 (comment)](https://github.com/terraform-provider-openstack/terraform-provider-openstack/pull/1771#issuecomment-2939784423) Please consider this for new PRs.

@kayrus I think you're mixing up two things here. Yes, there are `extra` fields in the keystone API, however, the `Extra` field here is actually flattened into the body, as seen [here](https://github.com/gophercloud/gophercloud/blob/709c32278551d28c04e65e047cae9e1c403da43f/openstack/identity/v3/roles/requests.go#L100-L106). Put another way, to get the `extra` field you're talking about, you'd need 

```
	createOpts := roles.CreateOpts{
		Name:     "support",
		DomainID: "1789d1",
		Extra: map[string]any{
			"extra": map[string]any{
                "foo": "bar",
            },
		},
	}
```

idk if we ever want to support that, especially seeing as it's deprecated in Keystone itself. We can probably still remove this (in a separate PR) once we add support for the outstanding top-level keys: `description` (this PR) and `options` (a future PR).

PR #3374, opened 2025-05-12
Move master CI jobs to Ubuntu 24.04

stephenfin:
Per $subject.

PR #3535, opened 2025-10-15
Migrate epoxy jobs to Ubuntu 24.04 (Noble), drop caracal jobs

stephenfin:
This is a follow-up to #3368 and #3374. Per $subject, we migrate Epoxy jobs to Ubuntu 24.04 (Noble) (which is now possible since the necessary fixes have long since landed in DevStack) and drop ~Dalmatian~ Caracal jobs to reduce our test matrix somewhat.

*EDIT:* Updated to drop Caracal instead of Dalmatian since Caracal is now EOL.

### DIFF
--- a/.github/workflows/functional-baremetal.yaml
+++ b/.github/workflows/functional-baremetal.yaml
@@ -13,15 +13,12 @@ jobs:
           - name: "master"
             openstack_version: "master"
             ubuntu_version: "24.04"
-            additional_services: "openstack-cli-server"
           - name: "epoxy"
             openstack_version: "stable/2025.1"
             ubuntu_version: "24.04"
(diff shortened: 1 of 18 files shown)


stephenfin:
The failed messaging job should be fixed now. I need to remove the dalmatian jobs from the set of required jobs then we can merge this.

SeanMooney:
that a long way to say you shoudl probably remove the caracal 2024.1 job as well as it is not unmtianeed with the release of falmingo 2025.2

current supported or stable reelase are 2025.2(flamingo), 2025.1(epoxy) and 2024.2(dalmatian)

mandre (approved):
Nice. Let's get this in.

PR #3536, opened 2025-10-16
Add config_drive to server struct

### LINKED ISSUE #3534: Compute V2: Add ConfigDrive to server struct
### What is missing?

Currently gophercloud allows the usage of ConfigDrive within [createOpts.](https://github.com/gophercloud/gophercloud/blob/main/openstack/compute/v2/servers/requests.go#L478) But its not part of the regular Server struct and therefore the response is never parsed. (despite the nova api including it).

Therefore one can create a server including the config_drive option but one can not verify the option for a given server. 

Nova api ref
https://docs.openstack.org/api-ref/compute/#id15

xshyve:
Adds config_drive to the regular server struct used for responses. (currently only part of createOpts) which leads to situation that one can create a server with a config_drive. But one can not verify / see that information in any kind of response.

This fixes #3534.

Fixes #3524

code in this PR:

https://docs.openstack.org/api-ref/compute/#servers-servers

### DIFF
--- a/openstack/compute/v2/servers/results.go
+++ b/openstack/compute/v2/servers/results.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/url"
 	"path"
+	"strconv"
 	"time"
 
 	"github.com/gophercloud/gophercloud/v2"
@@ -283,6 +284,9 @@ type Server struct {
 	// Locked indicates the lock status of the server
 	// This requires microversion 2.9 or later
 	Locked *bool `json:"locked"`
+
+	// ConfigDrive enables metadata injection through a configuration drive.
+	ConfigDrive bool `json:"-"`
 }
 
 type AttachedVolume struct {
@@ -343,6 +347,7 @@ func (r *Server) UnmarshalJSON(b []byte) error {
 		Image        any                             `json:"image"`
 		LaunchedAt   gophercloud.JSONRFC3339MilliNoZ `json:"OS-SRV-USG:launched_at"`
 		TerminatedAt gophercloud.JSONRFC3339MilliNoZ `json:"OS-SRV-USG:terminated_at"`
+		ConfigDrive  any                             `json:"config_drive"`
 	}
 	err := json.Unmarshal(b, &s)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


openstack/compute/v2/servers/results.go
```diff
@@ -283,6 +283,9 @@ type Server struct {
 	// Locked indicates the lock status of the server
 	// This requires microversion 2.9 or later
 	Locked *bool `json:"locked"`
+
+	// ConfigDrive enables metadata injection through a configuration drive.
+	ConfigDrive string `json:"config_drive"`
```
mandre: Wouldn't it make more sense to use a boolean type? I know that nova returns a string that is either "True" or empty (meaning false) but we don't have to adopt the same API quirks.

@stephenfin do you happen to know why nova is not returning a bool for the config_drive field?

xshyve: Yes, I 100% agree that this should be a boolean. But as long as nova returns it as a string this would mean to marshal it to a "temp struct" and then remarshall it by asserting it correctly. Which, as far as I saw, gophercloud is not doing anywhere else.

stephenfin: Wish I'd a better answer, but it's just tech debt that has never been cleaned up in the API.

mandre: Actually, there are quite a few examples in the gophercloud code base, mostly for accommodating old time format (grep for `UnmarshalJSON` or `json:"-"` to see by yourself).

Have a look at how we convert from string to bool in keystone [for the User.Enabled field](https://github.com/gophercloud/gophercloud/blob/1b200e12c43a3bd42bb98fa746f237dac6fa83fe/openstack/identity/v3/users/results.go#L62-L74), or in swift [for the GetHeader.VersionsEnabled one](https://github.com/gophercloud/gophercloud/blob/1b200e12c43a3bd42bb98fa746f237dac6fa83fe/openstack/objectstorage/v1/containers/results.go#L140-L144).

If there's no reason for it to be a string, I think we should use a boolean.

mandre (commented):
Thanks for making the requested change and making the field a boolean. Let's just wait for CI to run.

openstack/compute/v2/servers/results.go
xshyve: Ah sorry, did not catch the approach. Thanks for the hints!
I have modified the UnmarshalJSON method of the server struct. It does not match the examples 1:1 as nova can also return an empty string ("") which should be treated as boolean false. 

I have also added this case in the handling.

PR #3538, opened 2025-10-24
Identity V3: Add Options field to roles.

### LINKED ISSUE #3533: KeystoneV3: Adding support for option field in Roles
Added description field in previous PR. The role resource in keystone API also offers option as a field. Good idea to add support for this field as well in gophercloud.

reference: [click me](https://docs.openstack.org/api-ref/identity/v3/index.html#create-role:~:text=about%20the%20role.-,options%20(Optional),-body)

dlaw4608:
Fixes #3533

### DIFF
--- a/openstack/identity/v3/roles/results.go
+++ b/openstack/identity/v3/roles/results.go
@@ -23,6 +23,9 @@ type Role struct {
 
 	// Extra is a collection of miscellaneous key/values.
 	Extra map[string]any `json:"-"`
+
+	// Options are a set of defined options that allow certain features for a role
+	Options map[Option]any `json:"options"`
 }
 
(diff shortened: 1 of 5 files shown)


openstack/identity/v3/roles/results.go
```diff
@@ -23,6 +23,9 @@ type Role struct {
 
 	// Extra is a collection of miscellaneous key/values.
 	Extra map[string]any `json:"-"`
+
+	// Options are a set of defined options that allow certain features for a role
+	Options map[string]any `json:"options"`
```
mandre: Should probably be:
```suggestion
	Options map[Option]any `json:"options"`
```

Issue #3542, opened 2025-10-28
CreateImage returns empty image ID for Nova microversion 2.45+


ycalansy:
### What happened?

Currently, the [CreateImage](https://pkg.go.dev/github.com/gophercloud/gophercloud/v2@v2.8.0/openstack/compute/v2/servers#CreateImage) function in the Nova client reads the image ID from the Location header in the API response. Starting with Nova microversion 2.45, the Location header is removed. Instead, the image ID is returned in the response body as a small JSON object.

As a result, when using microversion 2.45 or higher, CreateImage returns an empty image ID, causing failures for clients relying on this value.

### Version

v2 (current)

PR #3543, opened 2025-10-28
fix: handle Nova create image response for microversion 2.45 and above

ycalansy:
## Description

This change addresses an issue where the Nova [CreateImage](https://pkg.go.dev/github.com/gophercloud/gophercloud/v2@v2.8.0/openstack/compute/v2/servers#CreateImage) API response was not handled correctly for microversion 2.45 and above.

Prior to microversion 2.45, the image ID was extracted from the Location header. Starting with 2.45, Nova no longer includes the Location header. Instead, the image ID is returned in the response body as a small JSON object. (see [API reference](https://docs.openstack.org/api-ref/compute/#create-image-createimage-action))

The `CreateImage` function has been updated to:
- Read and store the response body in `CreateImageResult`.
- Extract the image ID based on the `X-OpenStack-Nova-API-Version` header:
  - For microversions prior to 2.45, extract from the `Location` header.
  - For microversion 2.45 and above, extract from the response body.

This fix resolves the issue where the image ID was previously empty for Nova 2.45+, while maintaining compatibility with older microversions.

Fixes https://github.com/gophercloud/gophercloud/issues/3542

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -1,10 +1,12 @@
 package servers
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"io"
 	"maps"
 	"net"
 	"regexp"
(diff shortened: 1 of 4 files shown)


openstack/compute/v2/servers/requests.go
```diff
@@ -1053,10 +1055,35 @@ func CreateImage(ctx context.Context, client *gophercloud.ServiceClient, id stri
...
+	// If the caller doesn't read from the buffer, it can still be safely garbage collected.
+
+	var buf bytes.Buffer
+
+	_, r.Err = io.Copy(&buf, resp.Body)
+	if r.Err != nil {
+		return
+	}
+
+	r.Body = &buf
```
mandre, line 1085: This feels overly complex, considering that `gophercloud.ParseResponse(resp, err)` returns the body as the first return value. Any reason why we're not using it?

ycalansy: `gophercloud.ParseResponse(resp, err)` hands back the `resp.Body`, which means the body/connection stays open until the caller consumes and closes it. In this case the payload is small, so we buffer it and close the response immediately to ensure the connection can be reused even if the caller never calls the extract helper.

stephenfin (approved):

PR #3554, opened 2025-11-14
identity/services: add omitempty to the `type` field

### LINKED ISSUE #3553: service.UpdateOpts always generates `Type` as an empty value
### What happened?

While working on the Service Controller implementation on ORC, I faced an error when updating one of the Service fields, which is not the `type` field.

Type field does not have an omitempty tag, so it creates a struct with this field as an empty value every time, and this makes the API return error 400

https://github.com/gophercloud/gophercloud/blob/main/openstack/identity/v3/services/requests.go#L112

```bash
> curl -X PATCH -H 'Content-Type: application/json' -H "X-Auth-Token: $TOKEN" -d '{"name": "service-create-minimal", "type": ""}' http://13.5.0.250:5000/v3/services/2897c1bcb5214b2cb9fbd1a8ed2dcca7
{
  "error": {
    "code": 400,
    "message": "None is not of type 'object'\n\nFailed validating 'type' in schema:\n    {'additionalProperties': True,\n     'minProperties': 1,\n     'properties': {'enabled': {'enum': [True, False, None],\n                                'type': 'boolean'},\n                    'name': {'maxLength': 255,\n                             'minLength': 1,\n                             'pattern': '[\\\\S]+',\n                             'type': 'string'},\n                    'type': {'maxLength': 255,\n                             'minLength': 1,\n                             'type': 'string'}},\n     'type': 'object'}\n\nOn instance:\n    None",
    "title": "Bad Request"
  }
}
```

Expected behavior: 
The `Type` field should not be defined with a default value when UpdateOpts is created.

### Version

v2 (current)

winiciusallan:
Fixes #3553 

code in this PR:

https://docs.openstack.org/api-ref/identity/v3/index.html#update-service

### DIFF
--- a/openstack/identity/v3/services/requests.go
+++ b/openstack/identity/v3/services/requests.go
@@ -109,7 +109,7 @@ type UpdateOptsBuilder interface {
 // UpdateOpts provides options for updating a service.
 type UpdateOpts struct {
 	// Type is the type of the service.
-	Type string `json:"type"`
+	Type string `json:"type,omitempty"`
 
 	// Enabled is whether or not the service is enabled.
 	Enabled *bool `json:"enabled,omitempty"`


mandre (approved):
We'll just need https://github.com/gophercloud/gophercloud/pull/3556 first to unblock the gate.

winiciusallan:
> We'll just need #3556 first to unblock the gate.

Could you please rerun the jobs?

PR #3565, opened 2025-11-25
ImageV2: fix image properties nesting bug on JSON marshal/unmarshal r…

### LINKED ISSUE #3564: ImageV2: image properties becomes nested after JSON marshal/unmarshal round-trip
### What happened?

When an `Image` struct is marshalled to JSON and then unmarshalled back, the `Properties` map becomes incorrectly nested inside itself. Each subsequent marshal/unmarshal cycle causes deeper nesting.

  ## Steps to Reproduce

  ```go
  originalImage := images.Image{
      ID:   "da3b75d9-3f4a-40e7-8a2c-bfab23927dea",
      Name: "Test Image",
      Properties: map[string]any{
          "hw_disk_bus": "scsi",
          "custom_prop": "value",
      },
  }

  // Marshal to JSON
  jsonData, _ := json.Marshal(originalImage)

  // Unmarshal back
  var unmarshaledImage images.Image
  json.Unmarshal(jsonData, &unmarshaledImage)

  // Expected: Properties = {"hw_disk_bus": "scsi", "custom_prop": "value"}
  // Actual:   Properties = {"Properties": {"hw_disk_bus": "scsi", "custom_prop": "value"}}
```

## Cause

The Properties field in the Image struct has no json tag, so Go marshals it as "Properties" (capital P). During unmarshalling, RemainingKeys only deletes lowercase "properties" from the extras map, so "Properties" is not excluded and gets added back into the Properties field.

## Impact

This affects any use case where an Image struct is serialized to JSON and later deserialized.

### Version

v2 (current)

vjrasane:
…ound-trip

Fixes #3564

### DIFF
--- a/openstack/image/v2/images/results.go
+++ b/openstack/image/v2/images/results.go
@@ -3,6 +3,7 @@ package images
 import (
 	"encoding/json"
 	"fmt"
+	"maps"
 	"reflect"
 	"strings"
 	"time"
@@ -70,7 +71,7 @@ type Image struct {
 
 	// Properties is a set of key-value pairs, if any, that are associated with
 	// the image.
-	Properties map[string]any
+	Properties map[string]any `json:"properties"`
 
 	// CreatedAt is the date when the image has been created.
 	CreatedAt time.Time `json:"created_at"`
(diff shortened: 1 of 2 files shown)


vjrasane:
I think it would be better if the Unmarshaller interface was not used here. the root cause of the bug was due to the fact that the unmarshalling from the glance API response to Image struct is different from unmarshalling from Image struct as JSON. hijacking the JSON unrmashalling mechanism causes unpredictable behavior like this. a better approach would be to have an entirely separate function for handling the Glance API -> Image conversion as they are different data structures.

PR #3566, opened 2025-11-26
identityv3/service: add `name` and `description` fields

### LINKED ISSUE #3561: identity.v3.Services is missing fields from Keystone API
### What happened?

Currently, identity.v3.Service has missing fields from the API such as `name` and `description` which were encapsulated within `Extra` fields.

We should:
* Add support for those fields
* Deprecate Extra field in future releases

[API Reference](https://docs.openstack.org/api-ref/identity/v3/index.html)

### Version

v2 (current)

winiciusallan:
Fixes #3561

https://docs.openstack.org/api-ref/identity/v3/#show-service-details
https://github.com/openstack/keystone/blob/master/keystone/catalog/schema.py#L153

### DIFF
--- a/openstack/identity/v3/services/requests.go
+++ b/openstack/identity/v3/services/requests.go
@@ -15,6 +15,12 @@ type CreateOptsBuilder interface {
 
 // CreateOpts provides options used to create a service.
 type CreateOpts struct {
+	// Name is the name of the service.
+	Name string `json:"name,omitempty"`
+
+	// Description is the description of the service.
+	Description string `json:"description,omitempty"`
+
 	// Type is the type of the service.
 	Type string `json:"type"`
 
@@ -108,6 +114,12 @@ type UpdateOptsBuilder interface {
 
 // UpdateOpts provides options for updating a service.
 type UpdateOpts struct {
+	// Name is an updated name for the service.
+	Name *string `json:"name,omitempty"`
+
+	// Description is an update description for the service.
+	Description *string `json:"description,omitempty"`
+
 	// Type is the type of the service.
 	Type string `json:"type,omitempty"`
 
--- a/openstack/identity/v3/services/results.go
+++ b/openstack/identity/v3/services/results.go
@@ -50,6 +50,12 @@ type Service struct {
 	// ID is the unique ID of the service.
 	ID string `json:"id"`
 
+	// Name is the name of the service.
+	Name string `json:"name"`
+
+	// Description is the description of the service.
+	Description string `json:"description"`
+
 	// Type is the type of the service.
 	Type string `json:"type"`
 
@@ -75,8 +81,6 @@ func (r *Service) UnmarshalJSON(b []byte) error {
 	}
 	*r = Service(s.tmp)
 
-	// Collect other fields and bundle them into Extra
-	// but only if a field titled "extra" wasn't sent.
 	if s.Extra != nil {
 		r.Extra = s.Extra
 	} else {
@@ -85,8 +89,19 @@ func (r *Service) UnmarshalJSON(b []byte) error {
 		if err != nil {
 			return err
 		}
+
 		if resultMap, ok := result.(map[string]any); ok {
 			r.Extra = gophercloud.RemainingKeys(Service{}, resultMap)
+
+			// the following code is required for backward compatibility with the
+			// old behavior, when both description and name were in extra
+			if description, ok := resultMap["description"]; ok {
+				r.Extra["description"] = description
+			}
+
+			if name, ok := resultMap["name"]; ok {
+				r.Extra["name"] = name
+			}
 		}
 	}
 
--- a/openstack/identity/v3/services/testing/fixtures_test.go
+++ b/openstack/identity/v3/services/testing/fixtures_test.go
@@ -23,11 +23,11 @@ const ListOutput = `
             "links": {
                 "self": "https://example.com/identity/v3/services/1234"
             },
+            "name": "service-one",
             "type": "identity",
             "enabled": false,
             "extra": {
-                "name": "service-one",
-                "description": "Service One"
+                "email": "service-one@example.com"
             }
         },
         {
@@ -35,11 +35,11 @@ const ListOutput = `
             "links": {
                 "self": "https://example.com/identity/v3/services/9876"
             },
+            "name": "service-two",
+            "description": "Service Two",
             "type": "compute",
             "enabled": false,
             "extra": {
-                "name": "service-two",
-                "description": "Service Two",
                 "email": "service@example.com"
             }
         }
@@ -55,11 +55,11 @@ const GetOutput = `
         "links": {
             "self": "https://example.com/identity/v3/services/9876"
         },
+        "name": "service-two",
+        "description": "Service Two",
         "type": "compute",
         "enabled": false,
         "extra": {
-            "name": "service-two",
-            "description": "Service Two",
             "email": "service@example.com"
         }
     }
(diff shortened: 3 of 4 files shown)


openstack/identity/v3/services/requests.go
```diff
@@ -108,6 +114,12 @@ type UpdateOptsBuilder interface {
 
 // UpdateOpts provides options for updating a service.
 type UpdateOpts struct {
+	// Name is an updated name for the service.
+	Name string `json:"name,omitempty"`
+
+	// Description is an update description for the service.
+	Description string `json:"description,omitempty"`
```
kayrus: please use a pointer. this allows to unset the name/description on update
```suggestion
	// Name is an updated name for the service.
	Name *string `json:"name,omitempty"`

	// Description is an update description for the service.
	Description *string `json:"description,omitempty"`
```

winiciusallan: Done!

openstack/identity/v3/services/results.go
```diff
@@ -50,6 +50,12 @@ type Service struct {
 	// ID is the unique ID of the service.
 	ID string `json:"id"`
 
+	// Name is the name of the service.
+	Name string `json:"name"`
+
+	// Description is the description of the service.
+	Description string `json:"description,omitempty"`
```
kayrus: `omitempty` tag is not necessary in results
```suggestion
	Description string `json:"description"`
```

openstack/identity/v3/services/results.go
```diff
@@ -85,8 +91,16 @@ func (r *Service) UnmarshalJSON(b []byte) error {
 		if err != nil {
 			return err
 		}
+
+		r.Extra = make(map[string]any)
 		if resultMap, ok := result.(map[string]any); ok {
-			r.Extra = gophercloud.RemainingKeys(Service{}, resultMap)
```
kayrus, line 89: I have a gut feeling that this needs to be kept.

openstack/identity/v3/services/requests.go
```diff
@@ -15,6 +15,12 @@ type CreateOptsBuilder interface {
 
 // CreateOpts provides options used to create a service.
 type CreateOpts struct {
+	// Name is the name of the service.
+	Name string `json:"name"`
```
kayrus:
```suggestion
	Name string `json:"name,omitempty"`
```

openstack/identity/v3/services/results.go
winiciusallan: At first, I added Name and Description and left this piece of code. After modifyng the tests to accept those fields I ran them and they were failing [here](https://github.com/gophercloud/gophercloud/blob/main/openstack/identity/v3/services/testing/requests_test.go#L45), because the `expected` created service had Extra as nil, and the actual had an empty map

```bash
convenience.go:273: Failure in requests_test.go, line 45: at [0].Extra expected map[string]interface {}(nil), but got map[string]interface {}{}
```

I ended up in this function which assign a map for the Extra field. The solution that I've thought of for this was to populate only Name and Description if they exists.

openstack/identity/v3/services/results.go
```diff
@@ -85,8 +91,16 @@ func (r *Service) UnmarshalJSON(b []byte) error {
...
+			if description, ok := resultMap["description"]; ok {
```
winiciusallan, line 98: @kayrus if populating description and name makes sense, do you think tests will be necessary to cover those cases?

openstack/identity/v3/services/results.go
kayrus: > populate only Name and Description if they exists

it could be not only the name or description, it could be anything as far as I understand. I'd keep the code as is and fix the tests.

winiciusallan: > backing into your comment, I realized that this line is what will help us add other free-form keys within Extra map .e.g email . I'll revert this change.

yeah, you're right. I thought that I sent this message above yesterday, but it got stuck in pending because I replied as a review.

openstack/identity/v3/services/results.go
winiciusallan: Done.

openstack/identity/v3/services/requests.go
winiciusallan: Done. Thanks!

openstack/identity/v3/services/results.go
kayrus: I believe description and name will be covered by the default json unmarshaller according to json tags

winiciusallan: yeah, you're right, but when running the acceptance tests, I could see that was expecting the name field in Extra[1], so not filling it in might break code that expects this field to be in Extra instead of the added field. wdyt?

```bash
identity.go:226: Failure in identity.go, line 226: expected <nil> but got "ACPTTESTDT9q2QXt"
```
[1] https://github.com/gophercloud/gophercloud/blob/main/internal/acceptance/openstack/identity/v3/identity.go#L226

kayrus: you're right. please apply the following patch for the final change

```diff
diff --git a/openstack/identity/v3/services/results.go b/openstack/identity/v3/services/results.go
index b3be743a1..df599658b 100644
--- a/openstack/identity/v3/services/results.go
+++ b/openstack/identity/v3/services/results.go
@@ -93,10 +93,11 @@ func (r *Service) UnmarshalJSON(b []byte) error {
                        return err
                }
 
-               r.Extra = make(map[string]any)
                if resultMap, ok := result.(map[string]any); ok {
                        r.Extra = gophercloud.RemainingKeys(Service{}, resultMap)
-
+                       // the following code is required for backward
+                       // compatibility with the old behavior, when both
+                       // description and name were in extra
                        if description, ok := resultMap["description"]; ok {
                                r.Extra["description"] = description
                        }
diff --git a/openstack/identity/v3/services/testing/fixtures_test.go b/openstack/identity/v3/services/testing/fixtures_test.go
index 945f97a04..42811a09d 100644
--- a/openstack/identity/v3/services/testing/fixtures_test.go
+++ b/openstack/identity/v3/services/testing/fixtures_test.go
@@ -23,11 +23,11 @@ const ListOutput = `
             "links": {
                 "self": "https://example.com/identity/v3/services/1234"
             },
-                       "name": "service-one",
+            "name": "service-one",
             "type": "identity",
             "enabled": false,
             "extra": {
-                               "email": "service-one@example.com"
+                "email": "service-one@example.com"
             }
         },
         {
@@ -35,8 +35,8 @@ const ListOutput = `
             "links": {
                 "self": "https://example.com/identity/v3/services/9876"
             },
-                       "name": "service-two",
-                       "description": "Service Two",
+            "name": "service-two",
+            "description": "Service Two",
             "type": "compute",
             "enabled": false,
             "extra": {
@@ -55,8 +55,8 @@ const GetOutput = `
         "links": {
             "self": "https://example.com/identity/v3/services/9876"
         },
-               "name": "service-two",
-               "description": "Service Two",
+        "name": "service-two",
+        "description": "Service Two",
         "type": "compute",
         "enabled": false,
         "extra": {
@@ -96,8 +96,8 @@ const UpdateOutput = `
         "links": {
             "self": "https://example.com/identity/v3/services/9876"
         },
-               "name": "service-two",
-               "description": "Service Two Updated",
+        "name": "service-two",
+        "description": "Service Two Updated",
         "type": "compute2",
         "enabled": false,
         "extra": {
```

winiciusallan: Done!

mandre: Now I realize that we broke backward compatibility with https://github.com/gophercloud/gophercloud/pull/3532 because we didn't retain the `description` in the returned Extra stuct. We should probably apply the same change we're doing here, until the next major release.

mandre (approved):
Similar in principle to https://github.com/gophercloud/gophercloud/pull/3532, except that we've been more careful not breaking backward compatibility here. Good job!

LGTM

openstack/identity/v3/services/results.go
mandre: Backward compatibility fix for role at ~~https://github.com/gophercloud/gophercloud/pull/3568~~ https://github.com/gophercloud/gophercloud/pull/3569

Issue #3573, opened 2025-12-10
Application credentials doesn't work when using access_rules


eshulman2:
### What happened?

Application credentials doesn't work when using access_rules. This is due to a (poorly documented) missing header called OpenStack-Identity-Access-Rules which enables access_rules when using application credentials. To allow using access_rules + application credentials the keystone token get and validate functions need to send this header in their request.

python implementation reference: https://opendev.org/openstack/python-keystoneclient/src/branch/master/keystoneclient/v3/tokens.py#L60

### Version

v2 (current)

eshulman2:
Hi @kayrus sure,

This is related to this ORC [bug](https://github.com/k-orc/openstack-resource-controller/issues/596) in the [feature spec](https://specs.openstack.org/openstack/keystone-specs/specs/keystone/train/capabilities-app-creds.html) it is mentioned that:

> 1. When requesting token validation, keystonemiddleware (or any 3rd party application that supports access rule enforcement) sets an Openstack-Identity-Access-Rules header with a version string as its value. Token validation for an application credential with a access rule list will only succeed if this header is present. The version string will allow us to safely extend this feature by invalidating tokens using the extended version in situations where keystonemiddleware only supports an older version of this feature.
> 
> 2. If there is no Openstack-Identity-Access-Rules header in the token validation request, token validation fails.

here is an example usage with and without this flag:
without:
```bash
$ curl -g -i -X GET -H "Accept: application/json" -H "X-Auth-Token: ${TOKEN}" -H "X-Subject-Token: ${TOKEN}" "https://overcloud.enothen-kellerlab.tamlab.brq2.redhat.com:13000/v3/auth/tokens"
HTTP/1.1 404 NOT FOUND
date: Wed, 10 Dec 2025 16:49:55 GMT
server: Apache
content-length: 80
vary: X-Auth-Token
x-openstack-request-id: req-77b5fd64-8267-4038-88b7-23c77ea92c44
content-type: application/json

{"error":{"code":404,"message":"Failed to validate token","title":"Not Found"}}
```
with:
```bash
$ source app-cred-test-appcred-keystone2-openrc.sh
$ TOKEN=$(openstack token issue -c id -f value)
[...]

PR #3576, opened 2025-12-11
Tokens: Add access rules support for application credentials

eshulman2:
This change adds support for the OpenStack-Identity-Access-Rules header
when getting or validating tokens created with application credentials
that have access rules configured.

Changes:
- Add GetOptsBuilder interface and GetOpts struct for Get() function
- Add ValidateOptsBuilder interface and ValidateOpts struct for Validate()
- Both structs contain AccessRulesVersion (float64) field to specify the
  header version. Pass nil to skip access rules validation, or provide
  opts with a version >= 1 to enable it.
- Add float64 support to BuildHeaders() in params.go
- Add AccessRule and ApplicationCredential types to represent the
  application credential information in token responses
- Add ExtractApplicationCredential() method to extract application
  credential details including access rules from token responses
- Update all usages across tests and client code

The decision to use an opts struct pattern instead of passing the
parameter directly was made to align with the rest of the gophercloud
codebase conventions, although it may feel like overkill for a single
optional parameter.

This fixes the issue where tokens created with application credentials
that have access rules would return 404 Not Found when fetched without
the OpenStack-Identity-Access-Rules header.

See: https://docs.openstack.org/keystone/latest/user/application_credentials.html

[feature spec](https://specs.openstack.org/openstack/keystone-specs/specs/keystone/train/capabilities-app-creds.html) it is mentioned that:

> 1. When requesting token validation, keystonemiddleware (or any 3rd party application that supports access rule enforcement) sets an Openstack-Identity-Access-Rules header with a version string as its value. Token validation for an application credential with a access rule list will only succeed if this header is present. The version string will allow us to safely extend this feature by invalidating tokens using the extended version in situations where keystonemiddleware only supports an older version of this feature.
> 
> 2. If there is no Openstack-Identity-Access-Rules header in the token validation request, token validation fails.

Closes: #3573 

Assisted by: claude

### DIFF
--- a/openstack/identity/v3/tokens/requests.go
+++ b/openstack/identity/v3/tokens/requests.go
@@ -2,9 +2,12 @@ package tokens
 
 import (
 	"context"
+	"maps"
 
 	"github.com/gophercloud/gophercloud/v2"
 )
+
+const xSubjectTokenHeader = "X-Subject-Token"
 
 // Scope allows a created token to be limited to a specific domain or project.
 type Scope struct {
@@ -118,12 +121,6 @@ func (opts *AuthOptions) ToTokenV3HeadersMap(map[string]any) (map[string]string,
 	return nil, nil
 }
 
-func subjectTokenHeaders(subjectToken string) map[string]string {
-	return map[string]string{
-		"X-Subject-Token": subjectToken,
-	}
-}
-
 // Create authenticates and either generates a new token, or changes the Scope
 // of an existing token.
 func Create(ctx context.Context, c *gophercloud.ServiceClient, opts AuthOptionsBuilder) (r CreateResult) {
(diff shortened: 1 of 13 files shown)


openstack/identity/v3/tokens/requests.go
```diff
@@ -6,6 +6,8 @@ import (
 	"github.com/gophercloud/gophercloud/v2"
 )
 
+const defaultAccessRulesVersion = "1"
```
kayrus: it's better to set it to `1.0`, see https://github.com/openstack/keystone/blob/bd178ee91f92f50398cf940b1db3eb2157ed7fd3/keystone/models/token_model.py#L33

openstack/identity/v3/tokens/requests.go
```diff
@@ -124,6 +126,14 @@ func subjectTokenHeaders(subjectToken string) map[string]string {
 	}
 }
 
+// headersWithAccessRules adds the OpenStack-Identity-Access-Rules header
+// to allow application credentials to use access rules
+func headersWithAccessRules(subjectToken, accessRulesVersion string) map[string]string {
```
kayrus: it makes sense to adjust the original `subjectTokenHeaders`

openstack/identity/v3/tokens/requests.go
eshulman2: fair, changed (although when the python conversion into float is done the result would be similar)

openstack/identity/v3/tokens/requests.go
eshulman2: I prefer not doing it like that as it would force me to use unnecessary flags in functions that doesn't require that.

stephenfin: **I** would rather we got rid of both of these helpers and instead defined the maps in-place with constants for `OpenStack-Identity-Access-Rules` and `X-Subject-Token` 😅

openstack/identity/v3/tokens/requests.go
```diff
@@ -146,20 +143,79 @@ func Create(ctx context.Context, c *gophercloud.ServiceClient, opts AuthOptionsB
...
+	ToTokenGetParams() (map[string]string, error)
+}
+
+// GetOpts provides options for the Get request.
+type GetOpts struct {
+	// AccessRulesVersion specifies the OpenStack-Identity-Access-Rules header
+	// version. This is required when getting tokens that were created using
+	// application credentials with access rules. Versions less than 1 will cause
+	// Keystone to ignore access rules validation.
+	AccessRulesVersion float64 `h:"OpenStack-Identity-Access-Rules"`
```
stephenfin: Should this be a string? Headers are strings so I don't know if it makes sense to encode something different here.

eshulman2: @stephenfin it is converted to a string [here](https://github.com/gophercloud/gophercloud/pull/3576/changes#diff-9e53fe9bf7266754a18a5710b5e83f44a905cce9a94dd4a042969020c7653992R506) as this is a float in keystone I thought it would make sense having it as a float here and getting the benefit of the type enforcement

kayrus: It's better to stick to a string format, like in other [cases](https://github.com/gophercloud/gophercloud/blob/487e09eca1b1ec1684055c03c84fefa62e573f86/service_client.go#L30-L31), since this is an HTTP header. In the Keystone source code, this header is also handled as a [string](https://github.com/search?q=repo%3Aopenstack%2Fkeystone%20access_rules_support&type=code).

eshulman2: @kayrus is is converted/casted into a float [here](https://github.com/openstack/keystone/blob/bd178ee91f92f50398cf940b1db3eb2157ed7fd3/keystone/token/provider.py#L256) and in the go code converted to string [here](https://github.com/gophercloud/gophercloud/pull/3576/changes#diff-9e53fe9bf7266754a18a5710b5e83f44a905cce9a94dd4a042969020c7653992R506) to me it seems more type safe this way, I would say it might be nice to convert other float values to use this as well to keep type safety in opts structs

kayrus: Treating versions as strings is simpler, safer, and more idiomatic in this case. Values like `1.0` are normalized to `1` when stored as `float64`, which may cause confusion when debugging API request or response headers. Parsing the value into a `float64` and then formatting it back with `strconv.FormatFloat` provides no real benefit if the value is never used for numeric computation.

Additionally, the rest of the gophercloud codebase already represents microversions such as `"1.32"` as strings, so introducing `float64` here would be inconsistent and potentially confusing. If the code can be simpler, it's better to keep it simple.

eshulman2: switched to a string

PR #3581, opened 2025-12-12
identity.v3.Endpoint: make name as optional on creation

winiciusallan:
The `name` field is optional on Endpoint creation following the [API reference](https://docs.openstack.org/api-ref/identity/v3/index.html#create-endpoint), but Gophercloud defines this field as required.

code in this PR:

https://docs.openstack.org/api-ref/identity/v3/index.html#create-endpoint

### DIFF
--- a/openstack/identity/v3/endpoints/requests.go
+++ b/openstack/identity/v3/endpoints/requests.go
@@ -26,7 +26,7 @@ type CreateOpts struct {
 	Enabled *bool `json:"enabled,omitempty"`
 
 	// Name is the name of the Endpoint.
-	Name string `json:"name" required:"true"`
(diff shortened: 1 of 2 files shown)


mandre (approved):

mandre:
> Should we want to drop the `Name` field on the next release since it is deprecated (according to docs)?
> 
> If so, I can change this pull request to do it instead of just making it optional.

Let's make it optional now (and backport to v2) and have another PR that drops it from main (not backported to v2).

winiciusallan:
> @winiciusallan could you resolve the conflict (due to #3584)?

Done! =)

mandre (approved):

PR #3587, opened 2025-12-17
Add support for name and description filters when listing Cinder volume types

### LINKED ISSUE #3586: BlockStorage V3: Add server-side filters support for Cinder volume types
### What is missing?

It would be useful to add support for additional server-side filters when listing Cinder volume types in gophercloud in order to avoid redundant client-side filtering.

Using the Cinder API directly (via curl), it is possible to filter volume types by name. This suggests that name-based filtering is already supported internally by Cinder, but currently not exposed through gophercloud.

Additionally, Cinder allows filtering volume types by extra specs. For example, volume types can be filtered by storage_protocol using the following request:
`GET /types?extra_specs={'storage_protocol': 'nfs'}`

Koodt:
This change adds support for additional server-side filters when listing Cinder volume types

The following filters are now supported:
	•	Filtering volume types by name
	•	Filtering volume types by description

Fixes #3586

### DIFF
--- a/internal/acceptance/openstack/blockstorage/v3/volumetypes_test.go
+++ b/internal/acceptance/openstack/blockstorage/v3/volumetypes_test.go
@@ -56,6 +56,7 @@ func TestVolumeTypesCRUD(t *testing.T) {
 	th.AssertEquals(t, name, newVT.Name)
 	th.AssertEquals(t, description, newVT.Description)
 	th.AssertEquals(t, isPublic, newVT.IsPublic)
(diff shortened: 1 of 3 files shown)


kayrus:
@Koodt thanks for your effort. Could you please add functional tests as well?

kayrus:
Unfortunately due to added `ExtraSpecs` map, the PR can not be backported to v2:
```
  Incompatible changes:
  - ListOpts: old is comparable, new is not
```

@mandre @stephenfin do you think we can make an exception?

Issue #3520, opened 2025-09-26
Some resources can't unset values in UpdateOpts


mandre:
### What happened?

While working on resource mutability for ORC, we've experienced multiple cases where gophercloud allows updating a property, but not unsetting it. This issue is meant to keep track of the ones we noticed so that we can make the necessary changes for the upcoming major gophercloud version. 

* [Volume metadata](https://github.com/gophercloud/gophercloud/blob/99dfe5597440a118a1a702ec299ee0bc65ff36c6/openstack/blockstorage/v3/volumes/requests.go#L288)
* [Floating IP fixed IP](https://github.com/gophercloud/gophercloud/blob/99dfe5597440a118a1a702ec299ee0bc65ff36c6/openstack/networking/v2/extensions/layer3/floatingips/requests.go#L148)

They're all caused by a missing pointer. We should conduct a full review of all the `UpdateOpts` and make the necessary code changes before releasing v3.

### Version

v2 (current)

PR #3599, opened 2026-01-07
endpoint: add missing pointer to description on update

winiciusallan:
This PR adds a missing pointer in description when updating that field. We weren't able to set the description as an empty string.


mandre:
Related to https://github.com/gophercloud/gophercloud/issues/3520. There are probably a few more we'll discover along the way.

kayrus (approved):

kayrus (approved):
LGTM, thanks!

Issue #3597, opened 2026-01-05
networking.v2.extensions: Add support for trusted field


winiciusallan:
### What is missing?

Recently, was added to the Neutron API a new ML2 extension[1] that permits admin users to set a VIF as trusted through a new field, instead of changing it directly via binding:profile which is meant only for machine-machine communication.

It would be good to support this so we can implement this field on other projects such as CAPO and ORC.

There was a small discussion [here](https://github.com/k-orc/openstack-resource-controller/pull/622) on ORC.

[1] https://review.opendev.org/c/openstack/neutron/+/926068

Issue #3605, opened 2026-01-15
networking.v2.Port: Missing pointer on PropagateUplinkStatus


winiciusallan:
### What happened?

The [PropagateUplinkStatus](https://docs.openstack.org/api-ref/network/v2/index.html#uplink-status-propagation) is a field that is activated when `uplink_status_propagation` extension is enabled on Neutron's side. That field has a missing pointer in the Port result struct, which may cause a inconsistent behavior: if the extension is not enabled, Gophercloud will return `false` as the field value anyway.

We can not fix this without introducing a breaking change, so my suggestion is to move the PropagateUplinkStatus to a separated extension instead of just adding the missing pointer.

### Version

v2 (current)

kayrus:
I guess we need to introduce a new extension like it's done with portsecurity. As a workaround, existing v2 client can implement something like:

```go
type MyPortExt struct {
        PropagateUplinkStatusPtr *bool `json:"propagate_uplink_status"`
}

var s struct {
        ports.Port
        MyPortExt
}

err := ports.Get(ctx, client, portID).ExtractInto(&s)
... ... ...

if s.PropagateUplinkStatusPtr != nil {
... ... ...
```

PR #3608, opened 2026-01-22
         Networking V2: Added support for ML2 extension port_trusted_vif

dlaw4608:
Fixes: #3597 

Added Trusted VIF, The port-trusted-vif extension adds `trusted` attribute to the port resource.
Was implemented as an extension similar to how the `port_security_enabled` extension is currently implemented in gophercloud.

API Reference: [here](https://docs.openstack.org/api-ref/network/v2/index.html#ports:~:text=its%20connected%20to.-,Trusted%20VIF%C2%B6,-The%20port%2Dtrusted)

code in this PR: 
https://opendev.org/openstack/neutron/src/commit/762694d1bd6092060ce08c29d12575e303ae2973/neutron/plugins/ml2/extensions/port_trusted.py

### DIFF
--- a/openstack/networking/v2/extensions/portstrustedvif/requests.go
+++ b/openstack/networking/v2/extensions/portstrustedvif/requests.go
@@ -1,0 +1,53 @@
+package portstrustedvif
+
+import (
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/ports"
+)
+
+// PortCreateOptsExt adds port trusted VIF options to the base ports.CreateOpts.
+type PortCreateOptsExt struct {
+	ports.CreateOptsBuilder
+
+	// PortTrustedVIF toggles the port's trusted VIF status.
+	PortTrustedVIF *bool `json:"trusted,omitempty"`
+}
+
+// To PortCreateMap casts a CreateOpts struct to a map
+func (opts PortCreateOptsExt) ToPortCreateMap() (map[string]any, error) {
+	base, err := opts.CreateOptsBuilder.ToPortCreateMap()
+	if err != nil {
+		return nil, err
+	}
+
+	port := base["port"].(map[string]any)
+
+	if opts.PortTrustedVIF != nil {
+		port["trusted"] = *opts.PortTrustedVIF
+	}
+
+	return base, nil
+}
+
+// PortUpdateOptsExt adds port trusted VIF options to the base ports.UpdateOpts.
+type PortUpdateOptsExt struct {
+	ports.UpdateOptsBuilder
+
+	// PortTrustedVIF updates the port's trusted VIF status.
+	PortTrustedVIF *bool `json:"trusted,omitempty"`
+}
+
+// ToPortUpdateMap casts a UpdateOpts struct to a map.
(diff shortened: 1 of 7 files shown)


winiciusallan:
Jobs are failing on master branch due to eventlet module removal from neutron. This is a known problem.

Here is the upstream patch to fix it https://review.opendev.org/c/openstack/neutron-dynamic-routing/+/956747

On dalmation and epoxy, I believe it is because the extension is not enabled, so you may need to change the CI file to include it.

openstack/networking/v2/extensions/portstrustedvif/requests.go
```diff
@@ -0,0 +1,53 @@
...
+func (opts PortUpdateOptsExt) ToPortUpdateMap() (map[string]any, error) {
+	base, err := opts.UpdateOptsBuilder.ToPortUpdateMap()
+	if err != nil {
+		return nil, err
+	}
+
+	port := base["port"].(map[string]any)
+
+	if opts.PortTrustedVIF != nil {
+		port["trusted"] = &opts.PortTrustedVIF
```
kayrus:
```suggestion
		port["trusted"] = *opts.PortTrustedVIF
```

openstack/networking/v2/extensions/portstrustedvif/results.go
```diff
@@ -0,0 +1,6 @@
+package portstrustedvif
+
+type PortTrustedVIFExt struct {
+	// PortTrustedVIF stores information about whether a SR-IOV port should be trusted
+	PortTrustedVIF bool `json:"trusted"`
```
kayrus: It makes sense to convert it to a pointer, see #3605 for details
```suggestion
	PortTrustedVIF *bool `json:"trusted"`
```

openstack/networking/v2/extensions/portstrustedvif/requests.go
dlaw4608: Done, Thanks!

openstack/networking/v2/extensions/portstrustedvif/results.go
dlaw4608: Done!

openstack/networking/v2/ports/testing/requests_test.go
```diff
@@ -158,13 +159,15 @@ func TestGetWithExtensions(t *testing.T) {
 	var portWithExtensions struct {
 		ports.Port
 		portsecurity.PortSecurityExt
+		portstrustedvif.PortTrustedVIFExt
```
kayrus: semver thinks that this is an incompatible change. it make sense to move all testing under the new `portstrustedvif` package

kayrus: @mandre @stephenfin is there a config in semver checker to ignore test packages?

UPD: This was already discussed [here](https://github.com/joelanford/go-apidiff/issues/35)

dlaw4608: Done, thanks for the help @kayrus

PR #3609, opened 2026-01-22
networking: extract port uplink status propagation into extension

### LINKED ISSUE #3605: networking.v2.Port: Missing pointer on PropagateUplinkStatus

kayrus:
Fixes #3605

### DIFF
--- a/internal/acceptance/openstack/networking/v2/extensions/uplinkstatuspropagation/uplinkstatuspropagation_test.go
+++ b/internal/acceptance/openstack/networking/v2/extensions/uplinkstatuspropagation/uplinkstatuspropagation_test.go
@@ -1,0 +1,122 @@
+//go:build acceptance || networking || uplinkstatuspropagation
+
+package uplinkstatuspropagation
+
+import (
+	"context"
+	"testing"
+
+	"github.com/gophercloud/gophercloud/v2/internal/acceptance/clients"
+	networking "github.com/gophercloud/gophercloud/v2/internal/acceptance/openstack/networking/v2"
+	"github.com/gophercloud/gophercloud/v2/internal/acceptance/tools"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/uplinkstatuspropagation"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/ports"
+	th "github.com/gophercloud/gophercloud/v2/testhelper"
+)
+
+func TestUplinkStatusPropagationCRUD(t *testing.T) {
+	client, err := clients.NewNetworkV2Client()
+	th.AssertNoErr(t, err)
+
+	// Skip these tests if we don't have the required extension
+	networking.RequireNeutronExtension(t, client, "uplink-status-propagation")
+
+	// Create Network
+	network, err := networking.CreateNetwork(t, client)
+	th.AssertNoErr(t, err)
+	defer networking.DeleteNetwork(t, client, network.ID)
+
+	// Create Subnet
+	subnet, err := networking.CreateSubnet(t, client, network.ID)
+	th.AssertNoErr(t, err)
+	defer networking.DeleteSubnet(t, client, subnet.ID)
+
+	// Create port with uplink status propagation enabled
+	port, err := CreatePortWithUplinkStatusPropagation(t, client, network.ID, subnet.ID, true)
+	th.AssertNoErr(t, err)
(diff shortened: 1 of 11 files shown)


kayrus:
@mandre @stephenfin @EmilienM please review

mandre (commented):
The code itself looks good.

A couple of small things:
- we should probably add a `doc.go` file for good measure.
- I wished we had acceptance test covering `PropagateUplinkStatus`. It would have clearly highlighted the breaking change.

How easy would it be to add those?

winiciusallan:
hi @kayrus! Are you intending to continue working on this? If so, I'll be happy to review and help to merge it. If not, it'll be a pleasure to address the comments myself :).

kayrus:
according to the `internal/acceptance/openstack/networking/v2.TestExtensionsList` test results, the `uplink-status-propagation` extension is not appearing in the list of available neutron extensions, despite being enabled in the devstack config :/

winiciusallan:
@kayrus you need to enable the neutron plugin in the CI so devstack can run the function to activate the extension


https://github.com/openstack/neutron/blob/master/devstack/lib/uplink_status_propagation

kayrus:
@winiciusallan it is enabled: https://github.com/gophercloud/gophercloud/blob/148e8c345f60bf6577128c36e86790916460473d/.github/workflows/functional-networking.yaml#L78

winiciusallan:
@kayrus indeed. but I meant the devstack neutron plugin, like we did here

https://github.com/k-orc/openstack-resource-controller/blob/main/.github/workflows/e2e.yaml#L43

kayrus:
> but I meant the devstack neutron plugin, like we did here
> 
> https://github.com/k-orc/openstack-resource-controller/blob/main/.github/workflows/e2e.yaml#L43

the `uplink-status-propagation` extension is not a separate plugin requiring github repo enablement

winiciusallan:
> > but I meant the devstack neutron plugin, like we did here
> > https://github.com/k-orc/openstack-resource-controller/blob/main/.github/workflows/e2e.yaml#L43
> 
> the `uplink-status-propagation` extension is not a separate plugin requiring github repo enablement

Yeah, you're right (the tests speak for themselves). The way that I suggested is how was working for me. Thanks for the insight.

Regarding the failing test, it looks like `propagate_uplink_status` is [not a valid filter](https://github.com/openstack/neutron-lib/blob/master/neutron_lib/api/definitions/uplink_status_propagation.py#L29-L33) field, wondering why the other tests are green.

internal/acceptance/openstack/networking/v2/extensions/uplinkstatuspropagation/uplinkstatuspropagation_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/ports"
+	th "github.com/gophercloud/gophercloud/v2/testhelper"
+)
+
+func TestUplinkStatusPropagationCRUD(t *testing.T) {
+	client, err := clients.NewNetworkV2Client()
+	th.AssertNoErr(t, err)
+
+	// Skip these tests if we don't have the required extension
+	networking.RequireNeutronExtension(t, client, "uplink-status-propagation")
```
winiciusallan, line 22: nit: fwiw, we can't also test if we don't have the `uplink-status-propagation-updatable` extension

winiciusallan (approved):
Just a nit comment. Despite that, all changes look pretty good to me. Thanks @kayrus!

internal/acceptance/openstack/networking/v2/extensions/uplinkstatuspropagation/uplinkstatuspropagation_test.go
kayrus: devstack activates both `uplink-status-propagation` and `uplink-status-propagation-updatable`. for the test suite this is acceptable

PR #3614, opened 2026-01-27
feat: add network segment range extension to networking v2

### LINKED ISSUE #3613: neutron: support network segment ranges extension
### What is missing?

Add CRUD support for neutron's networking segment ranges extension.

API: https://docs.openstack.org/api-ref/network/v2/index.html#network-segment-ranges

cardoe:
Add support for Neutron's network segment range extension to the networking/v2 module.

Fixes #3613

code in this PR:

https://docs.openstack.org/neutron/2025.2//admin/config-network-segment-ranges.html
https://github.com/openstack/neutron/blob/master/neutron/extensions/network_segment_range.py
https://docs.openstack.org/api-ref/network/v2/index.html#network-segment-ranges

### DIFF
--- a/openstack/networking/v2/extensions/networksegmentranges/requests.go
+++ b/openstack/networking/v2/extensions/networksegmentranges/requests.go
@@ -1,0 +1,127 @@
+package networksegmentranges
+
+import (
+	"context"
+
+	"github.com/gophercloud/gophercloud/v2"
+	"github.com/gophercloud/gophercloud/v2/pagination"
+)
+
+// ListOpts allows filtering when listing network segment ranges.
+type ListOpts struct {
+	ID              string `q:"id"`
+	Name            string `q:"name"`
+	Default         *bool  `q:"default"`
+	Shared          *bool  `q:"shared"`
+	ProjectID       string `q:"project_id"`
+	NetworkType     string `q:"network_type"`
+	PhysicalNetwork string `q:"physical_network"`
+	Minimum         int    `q:"minimum"`
+	Maximum         int    `q:"maximum"`
+	SortDir         string `q:"sort_dir"`
+	SortKey         string `q:"sort_key"`
+}
+
+type ListOptsBuilder interface {
+	ToNetworkSegmentRangeListQuery() (string, error)
+}
+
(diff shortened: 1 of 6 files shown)


openstack/networking/v2/extensions/networksegmentranges/results.go
```diff
@@ -0,0 +1,74 @@
...
+	commonResult
+}
+
+// DeleteResult represents the result of a delete operation.
+type DeleteResult struct {
+	gophercloud.ErrResult
+}
+
+// NetworkSegmentRange represents a network segment range.
+type NetworkSegmentRange struct {
```
mandre, line 42: Should we add all fields? We're currently missing:
- Description
- TenantID (not sure this one is relevant anymore)
- CreatedAt
- UpdatedAt
- RevisionNumber
- Tags

openstack/networking/v2/extensions/networksegmentranges/requests.go
```diff
@@ -0,0 +1,127 @@
...
+	})
+}
+
+type CreateOptsBuilder interface {
+	ToNetworkSegmentRangeCreateMap() (map[string]any, error)
+}
+
+// CreateOpts represents options for creating a network segment range.
+type CreateOpts struct {
+	Name            string `json:"name,omitempty"`
```
mandre, line 58: Ideally we'd add godoc comments for all public fields.

https://github.com/gophercloud/gophercloud/blob/main/docs/contributor-tutorial/step-05-pull-requests.md#documentation

mandre (commented):
This looks great!

Would it be possible to add acceptance tests as well, to ensure the feature works as expected against a live environment?

We already [configure devstack with the `neutron-network-segment-range` extension](https://github.com/gophercloud/gophercloud/blob/01208c97119fe4bb9d3c591aac6bb1b996ed2f2e/.github/workflows/functional-networking.yaml#L74), so the environment should be able to handle it. That said, we should add a guard similar to what's done here:

https://github.com/gophercloud/gophercloud/blob/01208c97119fe4bb9d3c591aac6bb1b996ed2f2e/internal/acceptance/openstack/networking/v2/extensions/mtu/mtu_test.go#L25-L26

Issue #3618, opened 2026-01-28
Add support to health field recently added in Ironic


jacob-anders:
### What is missing?

Recently, Ironic Community implemented basic hardware health check functionality for Redfish driver type:

https://review.opendev.org/c/openstack/ironic/+/966946

It would be desirable to bring this functionality into gophercloud so that this functionality can be used in projects such as metal3. I am intending to create a PR proposing addition of this field.

jacob-anders:
@winiciusallan correct, closing. Thank you.

PR #3619, opened 2026-01-28
Add a new Ironic field representing node health to Gophercloud

jacob-anders:
Add support for the node health field which reports hardware health status from the BMC. Requires Ironic API microversion 1.109.

Signed-off-by: Jacob Anders <jacob-anders-dev@proton.me>
Assisted-by:  Claude-4.5-opus-high

Fixes #[[3618]](https://github.com/gophercloud/gophercloud/issues/3618)

code in this PR:

https://review.opendev.org/c/openstack/ironic/+/966946


openstack/baremetal/v1/nodes/requests.go
```diff
@@ -113,6 +113,10 @@ type ListOpts struct {
 	// Filter the list with the specified fault.
 	Fault string `q:"fault"`
 
+	// Filter the list with the specified health status.
+	// Requires microversion 1.109 or later.
+	Health string `q:"health"`
```
mandre, line 118: Will older versions of ironic silently ignore this query parameter? It looks like we don't have `omitempty` marker for query strings.

openstack/baremetal/v1/nodes/results.go
```diff
@@ -139,6 +139,11 @@ type Node struct {
 	// node. There are other possible types, e.g., “clean failure” and “rescue abort failure”.
 	Fault string `json:"fault"`
 
+	// Health indicates the hardware health status reported by the BMC.
+	// Possible values are "OK", "Warning", "Critical", or empty if not available.
+	// This field is read-only and requires microversion 1.109 or later.
+	Health string `json:"health"`
```
mandre: The gophercloud [dev docs](https://github.com/gophercloud/gophercloud/blob/main/docs/MICROVERSIONS.md#new-response-fields) recommend setting this new fields added in microversions to a pointer to allow differentiating between unset and empty values. I suppose we can just assume here that empty means unset?

openstack/baremetal/v1/nodes/requests.go
jacob-anders: Thank you for your review, Martin.

To my best knowledge, older versions of Ironic will indeed ignore this.

openstack/baremetal/v1/nodes/results.go
jacob-anders: Thank you for your suggestion, pushed an updated version including this.

Let me know if the updated code looks good enough. Thank you for your time.

openstack/baremetal/v1/nodes/results.go
```diff
@@ -139,6 +139,11 @@ type Node struct {
 	// node. There are other possible types, e.g., “clean failure” and “rescue abort failure”.
 	Fault string `json:"fault"`
 
+	// Health indicates the hardware health status reported by the BMC.
+	// Possible values are "OK", "Warning", "Critical", or nil if not available.
+	// This field is read-only and requires microversion 1.109 or later.
+	Health *string `json:"health"`
```
dtantsur: Honestly, I think using a pointer here is inconsistent with the rest (e.g. the Fault above is also new). An empty string is not a valid value for health.

openstack/baremetal/v1/nodes/requests.go
dtantsur: I don't think Ironic ignores unknown fields, also I'm not 100% sure. Yet, it already works somehow with Fault, so I guess we do have something remotely similar to omitempty?

openstack/baremetal/v1/nodes/results.go
jacob-anders: I rolled back this change.

@mandre @dtantsur when you have time please let me know if this looks good, or if any more edits are needed. Thank you!

PR #3643, opened 2026-03-06
Patch issue 3620

### LINKED ISSUE #3620: [Object Store]: Creating Object with slash in name fails
### What happened?

Hello,

I encountered the behavior where creating object that contains slashes (`/`) in `swift` fails with error. I was using gophercloud version v1 previously and there it worked.

Current error message from my test:
```
Failure in requests_test.go, line 342: unexpected error "Expected HTTP response code [201 202] when accessing [PUT http://127.0.0.1:42347/testContainer/testObject%2Ffile.txt], but got 404 instead: 404 page not found"
```

> Empty container names, container names containing a slash (/), and empty object names are now rejected in Gophercloud before any call to Swift.

Based on [migration document](https://github.com/gophercloud/gophercloud/blob/main/docs/MIGRATING.md#object-storage) the empty container names and container names conatining a slash (`/`) are forbidden, but the object names should allow having the `/` in them. Then they mimick the directory structure. Same can be done with `swift` CLI successfully.

```bash
$ swift upload velero-backup-ic-hubudb3 file.txt --object-name /my-prefix/dir/file.txt
my-prefix/dir/file.txt
$ swift list velero-backup-ic-hubudb3 --prefix my-prefix
my-prefix/dir/file.txt
```

The error reproduction is simple and demonstrated in PR https://github.com/gophercloud/gophercloud/pull/3621 ([job](https://github.com/gophercloud/gophercloud/actions/runs/21475312476/job/61857754163?pr=3621#step:5:2714))

### Version

v2 (current)

danchild:
Fixes #3620

Although the object API was working as intended, Horizon doesn't properly render objects that have been
created with a leading "/". It is recommended that the same approach used by the python-swiftclient
is used,  namely a leading forward slash is trimmed from the object name (e.g. "/foo/bar.txt" --> "foo/bar.txt").

code in this PR:

https://github.com/openstack/python-swiftclient/blob/master/swiftclient/service.py#L359

### DIFF
--- a/openstack/objectstorage/v1/errors.go
+++ b/openstack/objectstorage/v1/errors.go
@@ -2,6 +2,7 @@ package v1
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/gophercloud/gophercloud/v2"
@@ -21,8 +22,12 @@ func CheckObjectName(s string) error {
 	if s == "" {
 		return ErrEmptyObjectName{}
 	}
-	if strings.HasPrefix(s, "/") {
-		return ErrLeadingSlashInObjectName{name: s}
+	// Check if objet name has a leading slash or two or more consecutive slashes
+	// anywhere in the string
+	re := regexp.MustCompile(`^\/|\/{2,}`)
+	match := re.MatchString(s)
+	if match {
+		return ErrInvalidObjectName{name: s}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


openstack/objectstorage/v1/errors.go
```diff
@@ -21,6 +22,13 @@ func CheckObjectName(s string) error {
 	if s == "" {
 		return ErrEmptyObjectName{}
 	}
+	// Check if objet name has a leading slash or two or more consecutive slashes
```
dlaw4608, line 25:
```suggestion
	// Check if object name has a leading slash or two or more consecutive slashes
```
NIT: Small Typo

openstack/objectstorage/v1/errors.go
```diff
@@ -21,6 +22,13 @@ func CheckObjectName(s string) error {
...
+	// anywhere in the string
+	re := regexp.MustCompile(`^\/|\/{2,}`)
```
dlaw4608, line 27: Nice!, this does work, but could we simplify it even more and use the string package like for `CheckContainerName `

> `if strings.ContainsRune(s, '/')`

 but for `CheckObjectName` it would be 

> `if strings.HasPrefix(s, "/") || strings.Contains(s, "//")`

I think it's a simple enough check we want to do, this implementation would keep things consistent and still be enough to flag any object names that have a leading "/" or "//" anywhere else in the object name.
LMK what you think?

Issue #3240, opened 2024-11-27
can't authenticate with env vars


winiciusallan:
when using `openstack.AuthOptionsFromEnv()`, the following error was thrown
```bash
> go run main.go                               
panic: You must provide exactly one of DomainID or DomainName to authenticate by Username
```
I needed to declare `OS_DOMAIN_NAME` to make it work
```bash
> export OS_DOMAIN_NAME=xxxxx; go run main.go
 expected output ...
``` 

But when I exported the `openrc`, the `OS_USER_DOMAIN_NAME` was declared, so I expected this to work properly

kayrus:
@winiciusallan I confirm that ENV parse logic in the gophercloud/gophercloud package has been obsolete for a long time. As suggested, please use https://pkg.go.dev/github.com/gophercloud/utils/openstack/clientconfig, for example:
```go
	ao, err := clientconfig.AuthOptions(nil)
	if err != nil {
		return nil, err
	}
```
This will generate auth opts based on ENV

winiciusallan:
@kayrus what do you think about removing `AuthOptionsFromEnv` in V3? If we already have better ways to do that and if it is simpler to maintain, I'm totally agree. 

Also, It would be nice to update the documentation to use the clientconfig package considering that is the recommended.

kayrus:
@winiciusallan Removing `AuthOptionsFromEnv` is only the tip of the iceberg. The very concept of `gophercloud/utils` is now questionable, especially after we have already moved part of its functionality [out](https://github.com/gophercloud/utils/pull/261). It may make sense to merge everything currently in gophercloud/utils into gophercloud/gophercloud, but this would require a separate discussion among the maintainers.
fyi @mandre @pierreprinetti @stephenfin

majewsky:
[...]

PR #3657, opened 2026-03-23
Expand list of accepted ENV vars for openstack.AuthOptionsFromEnv

### LINKED ISSUE #3440: V3 Auth not properly implemented
OpenStack has two DOMAIN configuration parameters: `OS_PROJECT_DOMAIN_NAME` and `OS_USER_DOMAIN_NAME` and they are not the same as just `OS_DOMAIN_NAME`. I wasn't able to authenticate on OpenStack system because OS_USER_DOMAIN was used in both cases.

Auth body should look like this:

```json
{
    "auth": {
        "identity" : {
            "methods": ["password"],
            "password": {            
                "user": {                
                    "name": "<OS_USERNAME>", 
                    "domain": { "name": "<OS_USER_DOMAIN_NAME>" }, 
                    "password": "<OS_PASSWORD>"
                }
            }
        },
          "scope": {
            "project": {
                "domain": {
                    "name": "<OS_PROJECT_DOMAIN_NAME>"
                },
                "name": "<OS_PROJECT_NAME>"
            }
        }
    }
}
```

This is the debug content which I got using `openstack.AuthenticatedClient`

```json
{
    "auth": {
        "identity": {
            "methods": [
                "password"
            ],
            "password": {
                "user": {
                    "domain": {
                        "name": "my_domain"
                    },
                    "name": "USER",
                    "password": "*****"
                }
            }
        },
        "scope": {
            "project": {
                "domain": {
                    "name": "my_domain"
                },
                "name": "PROJECT"
            }
        }
    }
}

```

danchild:
Fixes #3440 #3240

According to the OpenStack API documentation, OS_USER_DOMAIN_ID, OS_USER_DOMAIN_NAME, OS_PROJECT_DOMAIN_ID, AND OS_PROJECT_DOMAIN_NAME are all valid inputs for identity v3. The current
implementation of **openstack.AuthOptionsFromEnv** does not accept these environment variables as inputs.

This fix keeps the **gophercloud.AuthOptions** struct intact, keeping the semantics of the name domainID and domainName unchanged so we do not need refactor code that consumes the the struct. Instead a new implementation
of **openstack.AuthOptionsFromEnv** would check for the presence of the OS_USER_DOMAIN_X and
OS_PROJECT_DOMAIN_X environment variables and set domainID and domainName accordingly.

The difficult part is to resolve ambiguity between the env vars passed in and provide either warnings or errors
to the users depending on different  combinations. Given that the approach to not changing **gophercloud.AuthOptions** 
is the best path forward, dealing with this ambiguity needs to be iterated on, and I certainly need some help determining what users expect and what makes sense base on OpenStack identity v3 conventions.

code in this PR:

API Reference: https://docs.openstack.org/python-openstackclient/2025.2/cli/authentication.html


openstack/auth_env.go
```diff
@@ -58,6 +90,176 @@ func AuthOptionsFromEnv() (gophercloud.AuthOptions, error) {
 		tenantName = v
 	}
 
+	// Disambiguate between OS_USER_DOMAIN_ID/OS_PROJECT_DOMAIN_ID/OS_DOMAIN_ID
+	// 		and OS_USER_DOMAIN_NAME/OS_PROJECT_DOMAIN_NAME/OS_DOMAIN_NAME
```
winiciusallan:
```suggestion
	// and OS_USER_DOMAIN_NAME/OS_PROJECT_DOMAIN_NAME/OS_DOMAIN_NAME
```

openstack/auth_env.go
```diff
@@ -58,6 +90,176 @@ func AuthOptionsFromEnv() (gophercloud.AuthOptions, error) {
...
+	// Allow setting OS_USER_DOMAIN_ID and OS_PROJECT_DOMAIN_ID if they are equal
+	if v1, v2 := os.Getenv("OS_USER_DOMAIN_ID"), os.Getenv("OS_PROJECT_DOMAIN_ID"); v1 != v2 {
+		// Do nothing if one is set and the other isn't
+		if v1 != "" && v2 != "" {
+			err := gophercloud.ErrEnvironmentVarsExpectedEqual{
+				EnvironmentVariables: []string{"OS_USER_DOMAIN_ID", "OS_PROJECT_DOMAIN_ID"},
```
winiciusallan: Could you please explain why you added this check for these two variables?

I'm asking because testing in my environment, I was able to authenticate having different values here when using project scope.

openstack/auth_env.go
```diff
@@ -58,6 +90,176 @@ func AuthOptionsFromEnv() (gophercloud.AuthOptions, error) {
...
+		if userDomainID != "" {
+			conflictingVar = "OS_USER_DOMAIN_ID"
+		} else {
+			conflictingVar = "OS_PROJECT_DOMAIN_ID"
+		}
+		err := gophercloud.ErrAmbiguousEnvironmentVarsClash{
+			EnvironmentVariables: []string{"OS_DOMAIN_NAME", conflictingVar},
+		}
+		return nilOptions, err
+	}
```
winiciusallan: Thinking about these lines because if the users are providing both `OS_PROJECT_DOMAIN_ID` and `OS_DOMAIN_NAME` so we don't have an auth scope well defined here.

Looking at how keystone handles this, an exception is raised if more than one of them is passed.

https://github.com/openstack/keystoneauth/blob/master/keystoneauth1/identity/v3/base.py#L186-L199

I'm assuming that you are checking this for backward compatibility, am I right?

winiciusallan (commented):
hi @danchild, I took a look at your changes, but I will need to do another round of review. This PR is very important to improve how Gophercloud is handling authentication.

I would say that my impression is that we now have a long function which is handling a lot of validations--most of them are doing the same kind of verification--, and this may be hard to maintain in the future. So, I believe we could extract these things to separate helper functions and use them accordingly.

Besides, IMO it would be good to document how the validation process is. For example, `1) verify the scope`, `2) override domain variables for backward compatibility`, ... 

Another pair of eyes here would be good

danchild:
Hi @winiciusallan sorry for the late reply here, and thank you for your feedback. I agree with you that creating helper functions would make the intent clearer and documenting the process as well. I will work on those and message you with an update.

openstack/auth_env.go
danchild: Yessir that is exactly right. This same check is performed here [auth_options.go#L306](https://github.com/gophercloud/gophercloud/blob/7c95518b0a45a3cf5576f89006a0809c4c6570d1/auth_options.go#L306). If we were to not add these checks in **openstack/auth_env.go** the code would still work as intended (i.e. a set ID and Name variable will eventually conflict and an error thrown), however, the user wouldn't know which environment variables caused the ambiguity. Since we are adding more env vars, there is an expanded set of env var combinations that could be ambiguous. These checks in **openstack/auth_env.go** are meant to inform the user specifically which env vars were conflicting. As you mentioned, we could make a helper function to perform these checks to make the code less verbose, maintainable, and understandable.

openstack/auth_env.go
danchild: I am adding these checks because **OS_USER_DOMAIN_ID** and **OS_PROJECT_DOMAIN_ID** eventually become **DomainID** in [auth_options.go#L53](https://github.com/gophercloud/gophercloud/blob/7c95518b0a45a3cf5576f89006a0809c4c6570d1/auth_options.go#L53). I didn't want to touch this file to make backward compatibility easier - of course, the better case would be for these user/project variables to be added to **AuthOptions** but that would be riskier and more complicated. 

Because they become the same property in **AuthOptions**, there is a problem if the values are different. However, allowing both to be set and equal would be ok even though doing so might be in unintentional. 

Regarding your comment on being able to authenticate, I tried and wasn't able to. Here is my code/setup:

**main.go**
```go
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/gophercloud/gophercloud/v2/openstack"
)

func main() {
	opts, err := openstack.AuthOptionsFromEnv()
	if err != nil {
		fmt.Fprintf(os.Stderr, "failed to read auth env vars: %v\n", err)
		os.Exit(1)
	}

	ctx := context.Background()

	provider, err := openstack.AuthenticatedClient(ctx, opts)
	if err != nil {
		fmt.Fprintf(os.Stderr, "authentication failed: %v\n", err)
		os.Exit(1)
	}

     ...
     ...
}
```

**admin-openrc.sh**
```shell
#!/usr/bin/env bash
# To use an OpenStack cloud you need to authenticate against the Identity
# service named keystone, which returns a **Token** and **Service Catalog**.
# The catalog contains the endpoints for all services the user/tenant has
# access to - such as Compute, Image Service, Identity, Object Storage, Block
# Storage, and Networking (code-named nova, glance, keystone, swift,
# cinder, and neutron).
#
# *NOTE*: Using the 3 *Identity API* does not necessarily mean any other
# OpenStack API is version 3. For example, your cloud provider may implement
# Image API v1.1, Block Storage API v2, and Compute API v2.0. OS_AUTH_URL is
# only for the Identity API served through keystone.
export OS_AUTH_URL=http://{IP_ADDRESS}/identity
# With the addition of Keystone we have standardized on the term **project**
# as the entity that owns the resources.
export OS_USER_DOMAIN_ID=werlkjiowoiuoeiuoiured
export OS_PROJECT_DOMAIN_ID=kljhdflkj
```

Are you still not seeing this error on your end? If so, could you please share your setup?

winiciusallan: My setup was quite different. I was testing using the openstackclient.
```bash
❯  o --os-cloud prod role assignment list --user winicius.silva --names --project support
+--------+------------------------+-------+-------------+--------+--------+-----------+
| Role   | User                   | Group | Project     | Domain | System | Inherited |
+--------+------------------------+-------+-------------+--------+--------+-----------+
| member | winicius.silva@Default |       | support@foo |        |        | False     |
+--------+------------------------+-------+-------------+--------+--------+-----------+
```

I assigned the member role to a user (me) who belongs to the domain `Default` to the `support` project on the domain `foo` (user in a different domain than the project in which they are acting).

Then I set the variables accordingly to be able to authenticate.
```bash
❯ env | grep OS_
OS_USERNAME=winicius.silva
OS_AUTH_URL=https://example.com:5000
OS_PROJECT_NAME=support # Using project as scope
OS_USER_DOMAIN_ID=default
OS_PROJECT_DOMAIN_ID=583c2dadbbd4419bb941fec33d8bca7e

❯ openstack server list -n
<output with servers>
```

Above, both `OS_USER_DOMAIN_ID` and `OS_PROJECT_DOMAIN_ID` are used to authenticate, and it works. Hoping it helps you

danchild: Thank you for the example - it's very clear to me now, and thank you for raising this issue. As is, the current proposal would not support a case where a user is a member of one domain and they are being scoped to a project in a different domain. This is because the current [AuthOptions](https://github.com/gophercloud/gophercloud/blob/7c95518b0a45a3cf5576f89006a0809c4c6570d1/auth_options.go#L30) struct only has DomainID and DomainName - not something like UserDomainID, UserDomainName, etc. We are basically just making that struct happy by passing in OS_USER_X or OS_PROJECT_X instead of OS_DOMAIN_X - and this fine when everything is a part of the same domain but breaks when users and projects belong to different domains. We would need to revisit this struct and its methods to properly implement the cross domain logic. I had started by adding additional fields to the **AuthOptions** struct, but the logic starts getting harry down the method chain.

winiciusallan: I'm okay with not adding this behavior now; it will increase the scope, and we can add it later in a separate PR.

> This is because the current [AuthOptions](https://github.com/gophercloud/gophercloud/blob/7c95518b0a45a3cf5576f89006a0809c4c6570d1/auth_options.go#L30) struct only has DomainID and DomainName - not something like UserDomainID, UserDomainName, etc.

Oh, ok, this clarifies a lot to me. I'll revisit your code and see if something can be addressed. Thanks.

danchild: I went ahead and created the helper functions and pushed the changes to the PR - let me know what you think among the other things you may want to address

openstack/testing/auth_env_test.go
```diff
@@ -0,0 +1,265 @@
...
+			"OS_PROJECT_DOMAIN_NAME": "domain",
+			"OS_DOMAIN_NAME":         "domain",
+		},
+	}
+
+	for _, c := range cases {
+
+		for k, v := range c {
+			os.Setenv(k, v)
+			os.Setenv(k, v)
```
winiciusallan: Maybe has set the variable twice by mistake?

openstack/testing/auth_env_test.go
```diff
@@ -0,0 +1,265 @@
...
+
+		// Didn't get the correct error
+		if _, ok := err.(gophercloud.ErrEnvironmentVarsExpectedEqual); !ok {
+			t.Errorf("Expected ErrEnvironmentVarsExpectedEqual, got %v", err)
+		}
+
+		// Clear ENV Vars for next iteration
+		for k := range c {
+			os.Unsetenv(k)
+			os.Unsetenv(k)
```
winiciusallan: Should there be just one call here as well?

openstack/testing/auth_env_test.go
```diff
@@ -0,0 +1,265 @@
...
+			"OS_PROJECT_DOMAIN_NAME": "domain",
+			"OS_DOMAIN_NAME":         "domain",
+		},
+	}
+
+	for _, c := range cases {
+
+		for k, v := range c {
+			os.Setenv(k, v)
+			os.Setenv(k, v)
```
winiciusallan: Same comment about duplicate calls. There are other occurrences along the file.

openstack/testing/auth_env_test.go
```diff
@@ -0,0 +1,265 @@
...
+		// Clear ENV Vars for next iteration
+		for k := range c {
+			os.Unsetenv(k)
+			os.Unsetenv(k)
+		}
+	}
+}
+
+// Test various cases where there is ambiguity between passed environment
+// varialbes and a ErrAmbiguousEnvironmentVarsClash is expected to be returned
```
winiciusallan:
```suggestion
// variables and a ErrAmbiguousEnvironmentVarsClash is expected to be returned
```

winiciusallan (commented):
Hi, I've added a few comments, nothing that big. If you want to wait for another review to submit a new commit, I'd say it's up to you.

I would like to see how tests will behave.

openstack/auth_env.go
```diff
@@ -58,6 +72,74 @@ func AuthOptionsFromEnv() (gophercloud.AuthOptions, error) {
 		tenantName = v
 	}
 
+	// Disambiguate between OS_USER_DOMAIN_ID/OS_PROJECT_DOMAIN_ID/OS_DOMAIN_ID
+	// and OS_USER_DOMAIN_NAME/OS_PROJECT_DOMAIN_NAME/OS_DOMAIN_NAME
+	if err := requireEnvVarsMatch("OS_USER_DOMAIN_ID", "OS_PROJECT_DOMAIN_ID", "OS_DOMAIN_ID"); err != nil {
+		return nilOptions, err
+	}
+	if err := requireEnvVarsMatch("OS_USER_DOMAIN_NAME", "OS_PROJECT_DOMAIN_NAME", "OS_DOMAIN_NAME"); err != nil {
```
mandre: As discussed already, it is entirely valid (and common) for an OpenStack user to belong to one domain while their project belongs to a different domain, and we shouldn't restrict this use case. The python-openstackclient treats these as independent values.

mandre: Additionally, variables that were previously ignored silently can now cause errors, even when it's a valid configuration. This would be a breaking change.

mandre (changes requested):
I'm really not sure about this PR. This adds a lot of complexity in a critical, and already complex part of the code.

The root problem seems to be that `gophercloud.AuthOptions` has a single DomainID/DomainName field rather than separate user-domain and project-domain fields. Until we fix this properly and add more fields to `gophercloud.AuthOptions`, I think we can't make it work for all cases.

Perhaps, instead of trying to make it work for all cases with the current `gophercloud.AuthOptions` limitation, we should instead focus on making the failure case for downloaded default openrc from horizon more user friendly (if that's really the issue)?

danchild:
Hi @mandre thank you for the feedback and I apologize for the delay. Give me a few weeks and I'll push changes with the default openrc as the primary lens for acceptance

danchild:
Hi @mandre - I took a stab at applying the fix through the lens of accepting the default openrc file from horizon. It's become clear that ToTokenV3ScopeMap was too strict in its original design. For example, it doesn't allow project name and project id to be set at the same time, and yet they are both provided on the default openrc. The patch here does the following:

**1)** Removes the restrictiveness, relying on keystone to handle its own errors
**2)** Silently, prioritize OS_USER_* and OS_PROJECT_* over the legacy OS_DOMAIN_*

This works and is perhaps overly permissive, but it jives with the project's goals of prioritizing freedom over restrictions. Can you please provide your feedback on the patch?

@winiciusallan - if you have any thoughts too please chime in.

PR #3706, opened 2026-04-09
Migrate to tools

stephenfin:
Now that we have a suitable new Go version, we can take advantage of go's `tool` directive which was introduced in [go 1.24](https://go.dev/doc/go1.24#tools).

Note that we only migrate the test runner since the golangci-lint docs are [**very** clear that you shouldn't use 'go tool' for that tool](https://golangci-lint.run/welcome/install/#install-from-sources).


IlanZuckerman (approved):
Consider fixing this link from the description: [very clear that you shouldn't use 'go tool' for that tool](https://golangci-lint.run/welcome/install/#install-from-sources).
It gets redirected to `https://golangci-lint.run/welcome/install/` for some reason...

Except for that, lgtm

mandre (approved):

PR #3709, opened 2026-04-10
[v2] Implement resource provider aggregates v2

mandre:
Related: https://github.com/gophercloud/gophercloud/issues/526

API reference:
https://docs.openstack.org/api-ref/placement/#update-resource-provider-aggregates
https://docs.openstack.org/api-ref/placement/#list-resource-provider-aggregates

Code reference:
https://opendev.org/openstack/placement/src/branch/stable/2026.1/placement/handlers/aggregate.py#L102-L132
https://opendev.org/openstack/placement/src/branch/stable/2026.1/placement/handlers/aggregate.py#L81-L99

Manual backport of https://github.com/gophercloud/gophercloud/pull/3685.

### DIFF
--- a/internal/acceptance/openstack/placement/v1/resourceproviders_test.go
+++ b/internal/acceptance/openstack/placement/v1/resourceproviders_test.go
@@ -587,3 +587,72 @@ func TestResourceProviderAggregatesUpdateNegative(t *testing.T) {
 	_, err = resourceproviders.UpdateAggregates(context.TODO(), client, resourceProvider.UUID, updateOpts).Extract()
 	th.AssertEquals(t, true, gophercloud.ResponseCodeIs(err, http.StatusConflict))
 }
+
+func TestResourceProviderAggregatesUpdatePreGenerationSuccess(t *testing.T) {
+	// Before microversion 1.19, the PUT request body is just a list of aggregate UUIDs.
+	clients.SkipReleasesBelow(t, "stable/ocata")
+	clients.RequireAdmin(t)
+
+	client, err := clients.NewPlacementV1Client()
+	th.AssertNoErr(t, err)
+
+	resourceProvider, err := CreateResourceProvider(t, client)
+	th.AssertNoErr(t, err)
+	defer DeleteResourceProvider(t, client, resourceProvider.UUID)
+
+	client.Microversion = "1.1"
+
+	updateOpts := resourceproviders.UpdateAggregatesOpts{
+		Aggregates: []string{
+			"6d84f6f6-7736-40ff-84d2-7db47f18ea25",
+			"f11f14bc-6f17-4f0a-b7c2-44b3e685ccf4",
+		},
+	}
+
+	_, err = resourceproviders.UpdateAggregates(context.TODO(), client, resourceProvider.UUID, updateOpts).Extract()
+	th.AssertNoErr(t, err)
+
+	after, err := resourceproviders.GetAggregates(context.TODO(), client, resourceProvider.UUID).Extract()
+	th.AssertNoErr(t, err)
+	th.AssertEquals(t, len(updateOpts.Aggregates), len(after.Aggregates))
+
+	for _, aggregate := range updateOpts.Aggregates {
+		th.AssertEquals(t, true, slices.Contains(after.Aggregates, aggregate))
+	}
+}
+
+func TestResourceProviderAggregatesUpdatePreGenerationWithGenerationSuccess(t *testing.T) {
+	// Before microversion 1.19, ResourceProviderGeneration in opts is silently stripped from
+	// the request body, so the operation must succeed even when the caller supplies it.
+	clients.SkipReleasesBelow(t, "stable/ocata")
+	clients.RequireAdmin(t)
+
+	client, err := clients.NewPlacementV1Client()
+	th.AssertNoErr(t, err)
+
+	resourceProvider, err := CreateResourceProvider(t, client)
+	th.AssertNoErr(t, err)
+	defer DeleteResourceProvider(t, client, resourceProvider.UUID)
+
+	client.Microversion = "1.1"
+
+	gen := 1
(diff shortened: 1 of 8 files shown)


internal/acceptance/openstack/placement/v1/resourceproviders_test.go
```diff
@@ -518,3 +519,71 @@ func TestResourceProviderAggregatesGetPreGenerationNegative(t *testing.T) {
...
+
+	before, err := resourceproviders.GetAggregates(context.TODO(), client, resourceProvider.UUID).Extract()
+	th.AssertNoErr(t, err)
+	th.AssertEquals(t, true, before.ResourceProviderGeneration != nil)
+
+	updateOpts := resourceproviders.UpdateAggregatesOpts{
+		ResourceProviderGeneration: before.ResourceProviderGeneration,
+		Aggregates: []string{
+			"6d84f6f6-7736-40ff-84d2-7db47f18ea25",
+			"f11f14bc-6f17-4f0a-b7c2-44b3e685ccf4",
```
stephenfin, line 545: These are just arbitrary IDs, yeah? And that's okay: they don't need to correspond to actual resource providers? Might be worth adding a comment here if so.

internal/acceptance/openstack/placement/v1/resourceproviders_test.go
```diff
@@ -451,3 +451,70 @@ func TestResourceProviderAllocations(t *testing.T) {
 
 	tools.PrintResource(t, usage)
 }
+
+func TestResourceProviderAggregatesGetSuccess(t *testing.T) {
+	// Resource_provider_generation in the aggregates response was introduced in microversion 1.19.
```
stephenfin: This comment isn't that helpful. It has the impression of being bound to the `clients.SkipReleasesBelow` call on the next line but I believe the more important aspect is the setting of `client.Microversion` below. Can we drop it?

Ditto for the other comments below. IMO it would be more helpful to add comments explaining the "story" of the test as we're doing in the test above this one.

internal/acceptance/openstack/placement/v1/resourceproviders_test.go
```diff
@@ -451,3 +451,70 @@ func TestResourceProviderAllocations(t *testing.T) {
...
+	clients.RequireAdmin(t)
+
+	client, err := clients.NewPlacementV1Client()
+	th.AssertNoErr(t, err)
+
+	resourceProvider, err := CreateResourceProvider(t, client)
+	th.AssertNoErr(t, err)
+	defer DeleteResourceProvider(t, client, resourceProvider.UUID)
+
+	client.Microversion = "1.1"
```
stephenfin, line 486: Any reason we can't fold these two tests into one? It should be a-okay to change the microversion on the fly?

```suggestion

	client.Microversion = "1.1"
```

Ditto for the test below and other test files added here.

internal/acceptance/openstack/placement/v1/resourceproviders_test.go
```diff
@@ -451,3 +451,70 @@ func TestResourceProviderAllocations(t *testing.T) {
...
+
+func TestResourceProviderAggregatesGetPreGenerationNegative(t *testing.T) {
+	// Resource provider aggregates operations were introduced in microversion 1.1.
+	clients.SkipReleasesBelow(t, "stable/ocata")
+	clients.RequireAdmin(t)
+
+	client, err := clients.NewPlacementV1Client()
+	th.AssertNoErr(t, err)
+
+	client.Microversion = "1.1"
```
stephenfin, line 516: Probably worth a comment here indicating that you want to prove the HTTP status code is the same regardless of microversion (assuming that's your intent?)

internal/acceptance/openstack/placement/v1/resourceproviders_test.go
```diff
@@ -451,3 +452,207 @@ func TestResourceProviderAllocations(t *testing.T) {
 
 	tools.PrintResource(t, usage)
 }
+
+func TestResourceProviderAggregatesGetSuccess(t *testing.T) {
+	// Resource_provider_generation in the aggregates response was introduced in microversion 1.19.
```
stephenfin, line 457: This comment isn't that helpful. It has the impression of being bound to the `clients.SkipReleasesBelow` call on the next line but I believe the more important aspect is the setting of `client.Microversion` below. Can we just drop it? Ditto for the other comments below. IMO it would be more helpful to add comments explaining the "story" of the test as we're doing in the test above this one.

openstack/placement/v1/resourceproviders/requests.go
```diff
@@ -254,6 +254,48 @@ func GetTraits(ctx context.Context, client *gophercloud.ServiceClient, resourceP
...
+	b, err := opts.ToResourceProviderUpdateAggregatesMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+
+	var body any = b
+	if useGenerations := microversionAtLeast(client, 1, 19); !useGenerations {
+		body = b["aggregates"]
+	}
```
stephenfin, line 291: Can we just inline this, given there's only one user.

```suggestion
	_, minor, err := utils.ParseMicroversion(client.Microversion)
	if err != nil {
		r.Err = err
		return
	}
	// In microversions prior to 1.19, the body was not enveloped.
	if minor < 19 {
		body = b["aggregates"]
	}
```

openstack/placement/v1/resourceproviders/util.go
```diff
@@ -0,0 +1,14 @@
+package resourceproviders
```
stephenfin, line 1: Let's drop this file and inline the code in its one caller, as noted above.

stephenfin (changes requested):
Couple of small things, mainly around tests, but this is otherwise good from my perspective.

stephenfin:
...and I just realised this is a backport :facepalm: I'll address these myself on `main`.

stephenfin (approved):

PR #3715, opened 2026-04-15
Implement usages

Etua:
Related https://github.com/gophercloud/gophercloud/issues/526

API reference:
https://docs.openstack.org/api-ref/placement/?expanded=#list-usages

Code reference:
https://opendev.org/openstack/placement/src/branch/stable/2026.1/placement/handlers/usage.py#L83-L143


mandre (approved):
Nice.

Issue #1775, opened 2019-11-20
Networking V2: Allow specifying 0 for security group rule PortRangeMin/PortRangeMax


kayrus:
In case when user wants to set ICMP PortRangeMin/PortRangeMax, the ICMP type should be used. In case when there is ICMP type 0 Echo Reply (see https://www.iana.org/assignments/icmp-parameters/icmp-parameters.xhtml#icmp-parameters-types), gophercloud ignores these values.

PR #3716, opened 2026-04-15
secgroup: add pointer to port_range_* on CreateOpts

### LINKED ISSUE #3644: Allow specifying 0 for PortRangeMin/PortRangeMax in security group rule
### Description
Fixes the inability to explicitly set an ICMP Type of `0` in Security Group Rules.

### Root Cause
Currently, the `PortRangeMin` and `PortRangeMax` fields use the `omitempty` JSON tag. In Go, setting an integer to `0` triggers this tag, causing the JSON marshaler to completely omit the field from the API payload. 

https://github.com/gophercloud/gophercloud/blob/a83babebedeea6ced66b9b468a507e53116f55f6/openstack/networking/v2/extensions/security/rules/requests.go#L126-L135

When creating a security group rule for the `icmp` protocol, OpenStack Neutron uses `port_range_min` to define the **ICMP Type** and `port_range_max` to define the **ICMP Code**. Because `0` is a valid ICMP Type (e.g., Echo Reply), the `omitempty` tag prevents the user from successfully sending `{"port_range_min": 0}` to the API. Consequently, Neutron receives a request with an ICMP Code but a missing ICMP Type, resulting in a `400 Bad Request.

```
failed to create security group rule in openstack: Expected HTTP response code [201 202] when accessing [POST http://10.32.50.10:9696/v2.0/security-group-rules], but got 400 instead: {\"NeutronError\": {\"type\": \"SecurityGroupMissingIcmpType\", \"message\": \"ICMP code (port-range-max) 255 is provided but ICMP type (port-range-min) is missing.\", \"detail\": \"\"}}
```

### Solution

By changing the struct fields to `*int` pointers, we can distinguish 0 from nil. This allows 0 to be safely marshaled without the omitempty tag dropping it. 

I noticed this [issue](#1775) has been open since 2019, so I would love to help get it resolved.

winiciusallan:
Currently, it is not possible to create a security group rule with `PortRangeMin: 0` and `PortRangeMax: 0`, which are valid values when using ICMP as a protocol.

This PR adds a missing pointer so we can differentiate between 0 and `nil` when creating a new rule.

Fixes #3644 #1775

### DIFF
--- a/internal/acceptance/openstack/networking/v2/extensions/extensions.go
+++ b/internal/acceptance/openstack/networking/v2/extensions/extensions.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/tools"
+	"github.com/gophercloud/gophercloud/v2/internal/ptr"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/external"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/security/addressgroups"
 	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/security/groups"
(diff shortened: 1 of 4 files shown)


winiciusallan:
In the first commit, I added a new function to convert a certain type to a pointer, but it turns out that we already have a function with this same purpose in the `internal/ptr` package. Changed to use this package instead and deleted the added one.

kayrus (approved):

stephenfin (approved):

PR #3734, opened 2026-05-05
db/instances: remove hardcoded 300GB size cap

### LINKED ISSUE #3733: db/v1/instances: remove hardcoded 300GB upper limit for CreateOpts.Size
https://github.com/gophercloud/gophercloud/blob/e12de205f3d9a635ea26deeb027d9560547977a3/openstack/db/v1/instances/requests.go#L80

In openstack/db/v1/instances/requests.go, CreateOpts.ToInstanceCreateMap() rejects sizes above 300 GB.
This appears to be environment-specific (Trove/OpenStack deployment limits vary), not a universal API constraint.
Because of this client-side validation, valid requests for deployments with larger limits cannot be sent at all.
Proposal: validate only Size >= 1 in gophercloud and let the API enforce deployment-specific maximums.

DDuckSCH21:
This PR removes the hardcoded upper limit (300 GB) from `instances.CreateOpts.Size`
client-side validation in `openstack/db/v1/instances`.
Current behavior rejects `Size > 300` before request is sent.
New behavior validates only `Size >= 1` and leaves deployment-specific maximum
validation to the OpenStack/Trove API.
Tests added:
- allow `Size: 301` in `ToInstanceCreateMap()`
- reject `Size: 0` in `ToInstanceCreateMap()`

Fixes #3733

### DIFF
--- a/openstack/db/v1/instances/requests.go
+++ b/openstack/db/v1/instances/requests.go
@@ -56,8 +56,8 @@ type CreateOpts struct {
 	// Either the integer UUID (in string form) of the flavor, or its URI
 	// reference as specified in the response from the List() call. Required.
 	FlavorRef string
-	// Specifies the volume size in gigabytes (GB). The value must be between 1
-	// and 300. Required.
+	// Specifies the volume size in gigabytes (GB). The value must be > 1.
+	// Required.
 	Size int
 	// Specifies the volume type.
 	VolumeType string
@@ -77,11 +77,11 @@ type CreateOpts struct {
 
 // ToInstanceCreateMap will render a JSON map.
 func (opts CreateOpts) ToInstanceCreateMap() (map[string]any, error) {
-	if opts.Size > 300 || opts.Size < 1 {
+	if opts.Size < 1 {
 		err := gophercloud.ErrInvalidInput{}
 		err.Argument = "instances.CreateOpts.Size"
 		err.Value = opts.Size
-		err.Info = "Size (GB) must be between 1-300"
+		err.Info = "Size (GB) must be >= 1"
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


openstack/db/v1/instances/requests.go
```diff
@@ -77,11 +77,11 @@ type CreateOpts struct {
 
 // ToInstanceCreateMap will render a JSON map.
 func (opts CreateOpts) ToInstanceCreateMap() (map[string]any, error) {
-	if opts.Size > 300 || opts.Size < 1 {
+	if opts.Size < 1 {
```
winiciusallan, line 80: Why don't we leave with the API to validate this value? Wondering if some exception is raised on Trove/Cinder side.

winiciusallan (commented):
Hi @DDuckSCH21, thanks for proposing a PR for this fix. Could you please add [acceptance tests](https://github.com/gophercloud/gophercloud/blob/main/internal/acceptance/README.md) for this change?

Also, would you mind fixing the link to the issue in the PR description?

openstack/db/v1/instances/requests.go
```diff
@@ -56,8 +56,8 @@ type CreateOpts struct {
 	// Either the integer UUID (in string form) of the flavor, or its URI
 	// reference as specified in the response from the List() call. Required.
 	FlavorRef string
-	// Specifies the volume size in gigabytes (GB). The value must be between 1
-	// and 300. Required.
+	// Specifies the volume size in gigabytes (GB). The value must be > 1.
```
mandre:
```suggestion
	// Specifies the volume size in gigabytes (GB). The value must be >= 1. The upper limit is specific to each OpenStack environment.
```

openstack/db/v1/instances/requests.go
mandre: I agree that we shouldn't validate input in the general case and leave it to OpenStack to return an appropriate error. This bug fix is a good example.

That said, I won't block on that, since [the docs](https://docs.openstack.org/api-ref/database/#create-database-instance) is explicit that values must be at least 1.

DDuckSCH21: Hi. Yes, I made sure that the value was at least 1, including 1

mandre (approved):

PR #3739, opened 2026-05-07
auth: support more scope cases in clouds.yaml parser

### LINKED ISSUE #3738: [Auth] - user_domain_name and project_domain_name are silently collapsed into a single field
### What happened?

When authenticating via `clouds.yaml` with a user and project in **different domains**, gophercloud v2
ignores the distinction between `user_domain_name` and `project_domain_name` and collapses both into
a single `DomainName` field:

```go
// openstack/config/clouds/clouds.go
DomainName: coalesce(options.domainName, cloud.AuthInfo.UserDomainName, cloud.AuthInfo.ProjectDomainName, cloud.AuthInfo.DomainName),
```

This single `DomainName` value is then used for **both** the user identity lookup and the project
scope in the resulting `gophercloud.AuthOptions`. The `clouds.yaml` spec explicitly supports separate
`user_domain_name` and `project_domain_name` for exactly this scenario, but gophercloud silently
discards the distinction.

For example, with the following `clouds.yaml`:

```yaml
clouds:
  openstack:
    auth:
      auth_url: https://identity.example.com/v3
      username: myuser
      password: mypassword
      user_domain_name: Default     # user lives in Default domain
      project_name: myproject
      project_domain_name: some_domain  # project lives in some_domain
    auth_type: password
```

The coalesce picks `user_domain_name: Default` (first non-empty value) and uses it for both the user
lookup and the project scope. Keystone then looks for `myproject` in `Default`, finds nothing, and
returns a 401 Unauthorized — with no indication that the domain scoping is the cause.

All three combinations fail:
- `user_domain_name: Default` only → project scoped to `Default` → project not found → 401
- `project_domain_name: some_domain` only → user looked up in `some_domain` → user not found → 401
- both set → `coalesce` picks the first, same result as above

The only workaround is to use IDs instead of names (`user_domain_id` + `project_id`), since
`DomainID` and `TenantID` are separate fields that are not coalesced together.

**Expectation:**

`user_domain_name` and `project_domain_name` should be treated as distinct fields, mapping to
separate parts of the Keystone v3 auth request — `user_domain_name` scoping the user identity
lookup, and `project_domain_name` scoping the project. This is how the OpenStack `clouds.yaml`
spec defines them and how direct use of `tokens.AuthOptions` already handles it via the `Scope`
struct:

```go
// tokens.AuthOptions correctly separates user domain from project domain
opts := &tokens.AuthOptions{
    Username:   "myuser",
    Password:   "mypassword",
    DomainName: "Default",       // user domain
    Scope: tokens.Scope{
        ProjectName: "myproject",
        DomainName:  "some_domain",  // project domain — separate field
    },
}
```

The `clouds.yaml` parsing path should produce equivalent behaviour to constructing `tokens.AuthOptions`
directly.

### Version

v2 (current)

kayrus:
Fixes #3738

This PR adds support for custom project scope, domain only scope and system scope in `clouds.yaml` parser

### DIFF
--- a/openstack/config/clouds/clouds.go
+++ b/openstack/config/clouds/clouds.go
@@ -154,10 +154,38 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
 
 	endpointType := coalesce(options.endpointType, cloud.EndpointType, cloud.Interface)
 
+	// Build the scope with proper project domain separation
 	var scope *gophercloud.AuthScope
 	if trustID := cloud.AuthInfo.TrustID; trustID != "" {
 		scope = &gophercloud.AuthScope{
 			TrustID: trustID,
+		}
+	} else if systemScope := cloud.AuthInfo.SystemScope; systemScope == "all" {
+		// System scoping for admin operations
+		scope = &gophercloud.AuthScope{
+			System: true,
+		}
+	} else if projectID := coalesce(options.projectID, cloud.AuthInfo.ProjectID); projectID != "" {
+		// Project scoping by ID
+		scope = &gophercloud.AuthScope{
+			ProjectID: projectID,
+		}
+	} else if projectName := coalesce(options.projectName, cloud.AuthInfo.ProjectName); projectName != "" {
+		// Project scoping by name requires project domain
+		scope = &gophercloud.AuthScope{
+			ProjectName: projectName,
+			DomainID:    coalesce(cloud.AuthInfo.ProjectDomainID, cloud.AuthInfo.DomainID),
+			DomainName:  coalesce(cloud.AuthInfo.ProjectDomainName, cloud.AuthInfo.DomainName),
+		}
+	} else if domainID := coalesce(options.domainID, cloud.AuthInfo.DomainID); domainID != "" {
+		// Domain scoping by ID (when no project is specified)
+		scope = &gophercloud.AuthScope{
+			DomainID: domainID,
+		}
+	} else if domainName := coalesce(options.domainName, cloud.AuthInfo.DomainName); domainName != "" {
+		// Domain scoping by name (when no project is specified)
+		scope = &gophercloud.AuthScope{
+			DomainName: domainName,
 		}
 	}
 
@@ -166,8 +194,8 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
 			Username:                    coalesce(options.username, cloud.AuthInfo.Username),
 			UserID:                      coalesce(options.userID, cloud.AuthInfo.UserID),
 			Password:                    coalesce(options.password, cloud.AuthInfo.Password),
-			DomainID:                    coalesce(options.domainID, cloud.AuthInfo.UserDomainID, cloud.AuthInfo.ProjectDomainID, cloud.AuthInfo.DomainID),
-			DomainName:                  coalesce(options.domainName, cloud.AuthInfo.UserDomainName, cloud.AuthInfo.ProjectDomainName, cloud.AuthInfo.DomainName),
+			DomainID:                    coalesce(options.domainID, cloud.AuthInfo.UserDomainID, cloud.AuthInfo.DomainID),
+			DomainName:                  coalesce(options.domainName, cloud.AuthInfo.UserDomainName, cloud.AuthInfo.DomainName),
 			TenantID:                    coalesce(options.projectID, cloud.AuthInfo.ProjectID),
 			TenantName:                  coalesce(options.projectName, cloud.AuthInfo.ProjectName),
 			TokenID:                     coalesce(options.token, cloud.AuthInfo.Token),
--- a/openstack/config/clouds/clouds.go
+++ b/openstack/config/clouds/clouds.go
@@ -153,20 +153,57 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
 
 	endpointType := coalesce(options.endpointType, cloud.EndpointType, cloud.Interface)
 
+	// Build the scope with proper project domain separation
 	var scope *gophercloud.AuthScope
 	if trustID := cloud.AuthInfo.TrustID; trustID != "" {
 		scope = &gophercloud.AuthScope{
 			TrustID: trustID,
 		}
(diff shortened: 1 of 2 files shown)


kayrus:
cc @mandre @stephenfin

openstack/config/clouds/clouds.go
```diff
@@ -154,20 +154,48 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
 	var scope *gophercloud.AuthScope
 	if trustID := cloud.AuthInfo.TrustID; trustID != "" {
 		scope = &gophercloud.AuthScope{
 			TrustID: trustID,
 		}
+	} else if systemScope := cloud.AuthInfo.SystemScope; systemScope == "all" {
+		// System scoping for admin operations
+		scope = &gophercloud.AuthScope{
+			System: true,
+		}
```
stephenfin, line 167: Do we want to check for `all`? Any other value is currently invalid on server-side, but I'm not sure we want to fall through here. It might be confusing.

```suggestion
	} else if systemScope := cloud.AuthInfo.SystemScope; systemScope != "" {
		// System scoping for admin operations
		if systemScope != "all" {
			return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, fmt.Errorf("only system scope of all is supported")
		}
		scope = &gophercloud.AuthScope{
			System: true,
		}
```

Alternatively, we could simplify the check to `!= ""` and add a comment that `all` is the only value allowed but we don't validate it client-side.

openstack/config/clouds/clouds.go
```diff
@@ -154,20 +154,48 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
+		// Project scoping by ID
+		scope = &gophercloud.AuthScope{
+			ProjectID: projectID,
+		}
+	} else if projectName := coalesce(options.projectName, cloud.AuthInfo.ProjectName); projectName != "" {
+		// Project scoping by name requires project domain
+		scope = &gophercloud.AuthScope{
+			ProjectName: projectName,
+			DomainID:    coalesce(cloud.AuthInfo.ProjectDomainID, cloud.AuthInfo.DomainID),
+			DomainName:  coalesce(cloud.AuthInfo.ProjectDomainName, cloud.AuthInfo.DomainName),
```
stephenfin, line 178: I know that this is what we were doing before, but the fallback to `DomainID` / `DomainName` is invalid and (I believe) not what SDK does. We should only be relying on `ProjectDomainID` / `ProjectDomainName`. Maybe one to remove for v3? If so, could we get a comment here that we can address in a follow-up PR that won't be backported to `v2` (as I expect this PR will be)

openstack/config/clouds/clouds.go
```diff
@@ -154,20 +154,48 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
 
 	return gophercloud.AuthOptions{
 			IdentityEndpoint:            coalesce(options.authURL, cloud.AuthInfo.AuthURL),
 			Username:                    coalesce(options.username, cloud.AuthInfo.Username),
 			UserID:                      coalesce(options.userID, cloud.AuthInfo.UserID),
 			Password:                    coalesce(options.password, cloud.AuthInfo.Password),
-			DomainID:                    coalesce(options.domainID, cloud.AuthInfo.UserDomainID, cloud.AuthInfo.ProjectDomainID, cloud.AuthInfo.DomainID),
-			DomainName:                  coalesce(options.domainName, cloud.AuthInfo.UserDomainName, cloud.AuthInfo.ProjectDomainName, cloud.AuthInfo.DomainName),
+			DomainID:                    coalesce(options.domainID, cloud.AuthInfo.UserDomainID, cloud.AuthInfo.DomainID),
+			DomainName:                  coalesce(options.domainName, cloud.AuthInfo.UserDomainName, cloud.AuthInfo.DomainName),
```
stephenfin, line 198: Same comment as above

openstack/config/clouds/clouds_test.go
```diff
@@ -335,4 +335,438 @@ func TestParse(t *testing.T) {
...
+      password: mypassword
+      user_domain_name: Default
+      project_name: myproject
+      project_domain_name: some_domain`
+
+		ao, _, _, err := clouds.Parse(
+			clouds.WithCloudsYAML(strings.NewReader(cloudsYAML)),
+			clouds.WithCloudName("gophercloud-test"),
+		)
+		if err != nil {
```
stephenfin, line 354: I was going to ask you to use `testhelper` here, but I see now that this isn't used in the rest of the file so what you're doing here is locally consistent.

openstack/config/clouds/clouds_test.go
```diff
@@ -335,4 +335,438 @@ func TestParse(t *testing.T) {
 			t.Errorf("unexpected identity endpoint: %q", got)
 		}
 	})
+
+	t.Run("supports user in one domain and project in another domain using names", func(t *testing.T) {
```
stephenfin, line 339: nit: I assume we can't define these tests via a single table-driven test case with expected `authOptions` structs?

stephenfin (changes requested):
I have made a few attempts at reworking the `auth_options.go` and this file to address this issue and also separate Identity v2/v3 auth, but have never settled on something I'm happy with. This is a good middle ground.

Couple of comments inline but nothing huge.

openstack/config/clouds/clouds.go
kayrus: done

openstack/config/clouds/clouds.go
kayrus: done

openstack/config/clouds/clouds.go
kayrus: done

openstack/config/clouds/clouds_test.go
kayrus: There is no a single table-driven test in this file. Probably worth adding this in a separate PR.

openstack/config/clouds/clouds_test.go
kayrus: I can add a testhelper support in the followup PR.

stephenfin (approved):

PR #3743, opened 2026-05-12
Make fields in Placement inventory update optional

### LINKED ISSUE #3742: Placement: UpdateInventoryOpts should have skippable fields
### What happened?

As per Placement API reference ([1](https://docs.openstack.org/api-ref/placement/#update-resource-provider-inventories), [2](https://docs.openstack.org/api-ref/placement/#update-resource-provider-inventory)), the inventory fields can be omit (with the exception of total).
When implementing update, I erroneously aliased the `Inventory` struct that is used as a return value, which made all of these fields mandatory.

Related: #526

### Version

v2 (current)

Etua:
Previously, the inventory update opts were incorrectly aliased to the return struct. This disallowed the SDK users to omit its values which is something API permits for all arguments but total.

Fixes #3742

### DIFF
--- a/internal/acceptance/openstack/placement/v1/allocationcandidates_test.go
+++ b/internal/acceptance/openstack/placement/v1/allocationcandidates_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/clients"
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/tools"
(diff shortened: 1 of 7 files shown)


mandre (approved):
LGTM, unfortunately we may not be able to backport to v2 because this is a breaking change to v2.12.

kayrus:
please hold on with this PR, I need some time to review it

kayrus (approved):

Issue #3747, opened 2026-05-12
Non-deterministic endpoint selection issue


reenjii:
### What happened?

Hello,

I recently moved our codebase from gophercloud v1 to gophercloud v2 (yes, it's about time). I noticed a regression regarding endpoint selection in catalog.

In short, the new aliases mechanism for endpoint selection causes a non-deterministic endpoint selection from catalog. For example, if a catalog contains endpoints for both `share` and `sharev2` endpoints, using `NewSharedFileSystemV2` can arbitrary select either endpoint from catalog in a non-deterministic way.

This behavior has been introduced in several steps:

- [Adds the support of "Aliases" in EndpointOpts](https://github.com/gophercloud/gophercloud/commit/eff7d91cd2662c11fc70d37e77a92efa7d7a3ed3)
- [Use official service types](https://github.com/gophercloud/gophercloud/commit/e3947338f69b8b52bd494f09388152febc49c6aa#diff-5643779347824e0ec4258ef187be1f23c3b1eeca504fcfe48e47efe0e051b4bf)
- [Start respecting service types when searching for endpoints](https://github.com/gophercloud/gophercloud/commit/0ba521aec71039e918235a4b5bbdeedbef732a7d)
- [Extend EndpointOpts.ApplyDefaults to set Aliases](https://github.com/gophercloud/gophercloud/commit/c507e2a9bf977d2bda06e55e9d0359baeaf11aa2)

Lets consider a catalog having endpoint for both `share` and `sharev2` endpoints. When calling `NewSharedFileSystemV2` it calls `initClientOpts(client, eo, "shared-file-system", 2)` which adds the slice `{"sharev2", "share"}` in endpoint opts aliases through [endpoint_search.go](https://github.com/gophercloud/gophercloud/blob/c507e2a9bf977d2bda06e55e9d0359baeaf11aa2/endpoint_search.go#L111).
Then in [endpoint_location.go](https://github.com/gophercloud/gophercloud/blob/main/openstack/endpoint_location.go#L86) `if (slices.Contains(opts.Types(), entry.Type))` when fetching endpoint, it will go over the catalog and select any endpoint that matches any of the service within the Aliases endpoint opts. As a result, if `share` happens to be processed first, this endpoint will be selected instead of the `sharev2` one assumed by calling `NewSharedFileSystemV2`.

The same behavior applies to any service having multiple aliases, like block-storage for example.

[...]

PR #3748, opened 2026-05-13
Add deprecation warning for V2EndpointURL, V3EndpointURL

stephenfin:
These are no longer used as of #3351. Mark them as deprecated so users know to migrate to their replacements.

Related-bug: #3747

### DIFF
--- a/openstack/endpoint_location.go
+++ b/openstack/endpoint_location.go
@@ -19,6 +19,9 @@ to return. It's an error both when multiple endpoints match the provided
 criteria and when none do. The minimum that can be specified is a Type, but you
 will also often need to specify a Name and/or a Region depending on what's
 available on your OpenStack deployment.
+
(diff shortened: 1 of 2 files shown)


openstack/endpoint_location.go
```diff
@@ -19,6 +19,9 @@ to return. It's an error both when multiple endpoints match the provided
 criteria and when none do. The minimum that can be specified is a Type, but you
 will also often need to specify a Name and/or a Region depending on what's
 available on your OpenStack deployment.
+
+Deprecated: This function does not respect the [gophercloud.EndpointOpts.Version] parameter,
```
winiciusallan, line 23: ~Is there any forecast for when we want to remove this function? If so, is it worth documenting this as well?~

update: forget this comment, I've just realized that we already have 

https://github.com/gophercloud/gophercloud/blob/main/openstack/endpoint_location.go#L11

mandre (approved):

PR #3749, opened 2026-05-13
Allow clearing Placement resource provider parent_provider_uuid

### LINKED ISSUE #3746: Placement: parent_provider_uuid cannot be set to null in UpdateOpts for resource provider
### What happened?

Despite the comment explicitly [mentioning this possibility](https://github.com/gophercloud/gophercloud/blob/main/openstack/placement/v1/resourceproviders/requests.go#L129), the ParentProviderUUID [features omitempty](https://github.com/gophercloud/gophercloud/blob/main/openstack/placement/v1/resourceproviders/requests.go#L133C2-L133C20) making it impossible to promote the child resource provider to root resource provider using the API; such value, even set explicitly, would get ignored at the SDK level before being sent.

Related #526

### Version

v2 (current)

Etua:
Current implementation describes the possibility to set parent_provider_uuid for a child resource provider to detach it from its parent, but actually doesn't allow it due to the use of omitempty which would clear such nil input.
To allow for this operation, I introduce the empty value of "" replaced by nil before being sent. This pattern allows for parent_provider_uuid to remain an optional argument just as in the API. Such pattern was previously introduced in DNS recordsets.

Fixes #3746
Related #526

### DIFF
--- a/internal/acceptance/openstack/placement/v1/resourceproviders_test.go
+++ b/internal/acceptance/openstack/placement/v1/resourceproviders_test.go
@@ -627,3 +627,41 @@ func TestResourceProviderUpdateAggregatesPreGenerationWithGenerationSuccess(t *t
 		th.AssertEquals(t, true, slices.Contains(after.Aggregates, aggregate))
 	}
 }
+
+func TestResourceProviderParentDetach(t *testing.T) {
+	clients.RequireAdmin(t)
+
+	client, err := clients.NewPlacementV1Client()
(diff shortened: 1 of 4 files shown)


mandre (approved):

Etua:
@mandre Something broke at the merge stage.

mandre:
> @mandre Something broke at the merge stage.

Indeed, coverall was down. Let's give it another try.

PR #3757, opened 2026-05-18
Add Placement resourceprovider list post 1.39

Etua:
Allows for repeating required and member_of arguments. Change of the type of field, microversions.go required.

Related #526


Etua:
@mandre By the way, could you take a look at this one? It's pretty straightforward and doesn't change the existing behaviour. It silently fixes a bad choice (IMO) in the tests, but that change is easy to review too.

mandre (approved):

PR #3765, opened 2026-05-23
ci(magnum): deprecate heat and use magnum-cluster-api driver

winiciusallan:
As Magnum moved towards removing the Heat driver to use out-of-tree drivers instead[1], we cannot rely on it anymore in our CI. This pull request changes the `functional-containerinfra` job to use the `vexxhost/magnum-cluster-api`[2] plugin that's also being used upstream[3].

The Vexxhost's driver spins up a Kubernetes cluster using Kind, installs all the dependencies, such as: kubectl, clusterctl, and CAPI, with OpenStack as the infrastructure provider (so we also have ORC here 🤩 ).

Let's see how it goes.

[1] https://review.opendev.org/c/openstack/magnum/+/985348
[2] https://github.com/vexxhost/magnum-cluster-api
[3] https://review.opendev.org/c/openstack/magnum-tempest-plugin/+/987846

### DIFF
--- a/script/stackenv
+++ b/script/stackenv
@@ -71,10 +71,7 @@ EOL
 fi
 
 if _=$(openstack service list | grep container-infra); then
-    _MAGNUM_IMAGE_ID=$(openstack image list --format value -c Name -c ID | grep coreos | cut -d ' ' -f 1)
-    if [ -z "$_MAGNUM_IMAGE_ID" ]; then
-        _MAGNUM_IMAGE_ID=$(openstack image list --format value -c Name -c ID | grep -i atomic | cut -d ' ' -f 1)
-    fi
+    _MAGNUM_IMAGE_ID=$(openstack image list --format value -c Name -c ID | grep ubuntu | cut -d ' ' -f 1)
     cat >> "openrc" <<EOL
 export OS_MAGNUM_IMAGE_ID="$_MAGNUM_IMAGE_ID"
 export OS_MAGNUM_KEYPAIR=magnum
--- a/script/stackenv
+++ b/script/stackenv
@@ -71,7 +71,12 @@ EOL
 fi
 
 if _=$(openstack service list | grep container-infra); then
-    _MAGNUM_IMAGE_ID=$(openstack image list --format value -c Name -c ID | grep ubuntu | cut -d ' ' -f 1)
+    _MAGNUM_IMAGE_ID=$(openstack image list --format value -c Name -c ID | grep ubuntu | cut -d ' ' -f 1 || true)
+
+    if [[ -z "${_MAGNUM_IMAGE_ID}" ]]; then
+        # Fallback to the CoreOS image used on Epoxy release.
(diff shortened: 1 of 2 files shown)


script/stackenv
```diff
@@ -60,10 +60,7 @@ export OS_FLAVOR_ID_RESIZE="$_FLAVOR_ALT_ID"
 EOL
 
 if _=$(openstack service list | grep container-infra); then
-    _MAGNUM_IMAGE_ID=$(openstack image list --format value -c Name -c ID | grep coreos | cut -d ' ' -f 1)
-    if [ -z "$_MAGNUM_IMAGE_ID" ]; then
-        _MAGNUM_IMAGE_ID=$(openstack image list --format value -c Name -c ID | grep -i atomic | cut -d ' ' -f 1)
-    fi
+    _MAGNUM_IMAGE_ID=$(openstack image list --format value -c Name -c ID | grep ubuntu | cut -d ' ' -f 1)
```
winiciusallan: Wondering whether we want to be able to choose the Kubernetes version to test against. We could add it to this filter instead of just grep'ing for "ubuntu".

This works because devstack does not create any ubuntu image, so there is no conflict.

But the K8s version tells more about the underlying driver support, and [CAPI compatibility](https://cluster-api.sigs.k8s.io/reference/versions#kubernetes-versions-support)

winiciusallan:
The error on master branch has an upstream bug report on Launchpad. It looks like a consequence of migrating to use the openstacksdk. 

https://bugs.launchpad.net/magnum/+bug/2153128

script/stackenv
mandre: That's a fair question. I think there's no need to do it here and we can leave that to future work.

winiciusallan:
> The error on master branch has an upstream bug report on Launchpad. It looks like a consequence of migrating to use the openstacksdk.
> 
> https://bugs.launchpad.net/magnum/+bug/2153128

As a follow-up, I've proposed a patch for this issue. Let's see if we can get it merged.

https://review.opendev.org/c/openstack/magnum/+/990397

winiciusallan:
The error on Epoxy is because this release has been using the `--no-build-isolation` flag when installing dependencies, which forces build packages (as it is used by magnum-cluster-api to build things in Rust) to be installed in advance. I'm trying to find a workaround to this; suggestions are appreciated.

I filed an issue in vexxhost/magnum-cluster-api https://github.com/vexxhost/magnum-cluster-api/issues/1045

winiciusallan:
Hey @stephenfin, do you have any suggestions for how to workaround the issue above? Looking at the history, we added the build-isolation flag to work around the `pkg_resources` removal on older branches. This flag is stopping us from building the magnum-cluster-api driver properly.

I saw Gnocchi adding a validation to skip this.

https://github.com/openstack/devstack/commit/ceec372db7b82b99e476e5abdc81b9140db8a152

I'm not sure about the likelihood of doing this upstream for the vexxhost repository.

mandre (approved):

PR #3768, opened 2026-05-26
neutron: properly enable port-forwarding extension in devstack

kayrus:


.github/workflows/functional-networking.yaml
```diff
@@ -71,11 +71,13 @@ jobs:
             enable_plugin networking-bgpvpn https://github.com/openstack/networking-bgpvpn.git ${{ matrix.openstack_version }}
             Q_ML2_PLUGIN_EXT_DRIVERS=qos,port_security,dns_domain_keywords,port_trusted
             BGP_SCHEDULER_DRIVER=neutron_dynamic_routing.services.bgp.scheduler.bgp_dragent_scheduler.StaticScheduler
+            Q_SERVICE_PLUGIN_CLASSES=port_forwarding
+            L3_AGENT_EXTENSIONS=port_forwarding
```
winiciusallan: I'm ok with this solution because it works, but testing here locally, I was able to enable the extension using:

```
enable_service neutron-port-forwarding

enable_plugin neutron https://github.com/openstack/neutron master
```

although neutron-port-forwarding is not being an external extension, it looks like we **can** enable the neutron plugin so it can check and configure it

https://github.com/openstack/neutron/blob/master/devstack/plugin.sh#L107

IMHO it is cleaner and straightforward. Thoughts?

kayrus: ah, that is why I could not find anything related to port-forwarding in devstack repo. let me try with a cleaner solution then.

kayrus: nope, this doesn't work: `devstack/lib/neutron_plugins/ovn_agent:394 The q-l3/neutron-l3 service must be disabled with OVN.`

winiciusallan: I linked the wrong line, sorry. But devstack also configures this extension when using OVN. The point of my suggestion is just to enable the neutron plugin. At least, this is how it worked for me.

https://github.com/openstack/neutron/blob/master/devstack/plugin.sh#L125

kayrus: Thanks for the hint. Now we can completely disable the `Q_ML2_PLUGIN_EXT_DRIVERS`

winiciusallan (approved):
All jobs are green. Nice!

mandre (approved):
Thanks!

PR #3773, opened 2026-05-27
CI: Use pre-built amphora images for Octavia jobs

mandre:
The octavia master job started failing recently while building the amphora image with diskimage-builder. Instead of building images from scratch (which is fragile and slow), download pre-built test images from https://tarballs.opendev.org/openstack/octavia/test-images/.

The Octavia devstack plugin skips the DIB build when the image file already exists at the path specified by `OCTAVIA_AMP_IMAGE_FILE`.

### DIFF
--- a/.github/workflows/functional-loadbalancer.yaml
+++ b/.github/workflows/functional-loadbalancer.yaml
@@ -15,21 +15,21 @@ jobs:
           - name: "master"
             openstack_version: "master"
             ubuntu_version: "24.04"
+            amphora_image_url: "https://tarballs.opendev.org/openstack/octavia/test-images/test-only-amphora-x64-haproxy-ubuntu-noble.qcow2"
             devstack_conf_overrides: |
               LIBVIRT_CPU_MODE=host-passthrough
-              OCTAVIA_AMP_DISTRIBUTION_RELEASE_ID=jammy
           - name: "gazpacho"
             openstack_version: "stable/2026.1"
             ubuntu_version: "24.04"
+            amphora_image_url: "https://tarballs.opendev.org/openstack/octavia/test-images/test-only-amphora-x64-haproxy-ubuntu-noble.qcow2"
             devstack_conf_overrides: |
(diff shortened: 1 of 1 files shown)


mandre:
Not sure this will work. The alternative would be to use pre-built image served at https://tarballs.opendev.org/openstack/octavia/test-images/, which might be a better thing to do anyway.

kayrus:
> Not sure this will work. The alternative would be to use pre-built image served at https://tarballs.opendev.org/openstack/octavia/test-images/, which might be a better thing to do anyway.

this would be definitely faster

winiciusallan:
```May 27 14:37:12 runnervmg397c nova-compute[77056]: ERROR nova.virt.libvirt.guest libvirt.libvirtError: internal error: process exited while connecting to monitor: Could not access KVM kernel module: Permission denied```

Maybe this is still related to this comment in an old PR.

https://github.com/gophercloud/gophercloud/pull/3638#issuecomment-4097705981

Is it possible that the pre-built images already have a workaround for this?

File: .github/workflows/functional-loadbalancer.yaml [hunk 1]
```diff
@@ -59,6 +59,11 @@ jobs:
           echo "No relevant files changed - skipping tests for ${{ matrix.name }}"
           echo "TESTS_SKIPPED=true" >> $GITHUB_ENV
 
+      - name: Download pre-built amphora image
+        if: ${{ fromJSON(steps.changed-files.outputs.matches) }}
+        run: |
+          wget -q "${{ matrix.amphora_image_url }}" -O /tmp/amphora.qcow2
```
kayrus:
```suggestion
          wget -q "${{ matrix.amphora_image_url }}" -O /tmp/amphora-x64-haproxy.qcow2
```

mandre: jinx :)

winiciusallan (approved):

mandre:
Looks like it worked :tada:

kayrus (approved):

PR #3775, opened 2026-05-27
[v2] CI: Use pre-built amphora images for Octavia jobs

mandre:
The octavia master job started failing recently while building the amphora image with diskimage-builder. Instead of building images from scratch (which is fragile and slow), download pre-built test images from https://tarballs.opendev.org/openstack/octavia/test-images/.

The Octavia devstack plugin skips the DIB build when the image file already exists at the path specified by OCTAVIA_AMP_IMAGE_FILE.

Manual backport of https://github.com/gophercloud/gophercloud/pull/3773

### DIFF
--- a/.github/workflows/functional-loadbalancer.yaml
+++ b/.github/workflows/functional-loadbalancer.yaml
@@ -24,21 +24,21 @@ jobs:
           - name: "master"
             openstack_version: "master"
             ubuntu_version: "24.04"
+            amphora_image_url: "https://tarballs.opendev.org/openstack/octavia/test-images/test-only-amphora-x64-haproxy-ubuntu-noble.qcow2"
             devstack_conf_overrides: |
               LIBVIRT_CPU_MODE=host-passthrough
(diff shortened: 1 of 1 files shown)


.github/workflows/functional-loadbalancer.yaml
```diff
@@ -24,35 +24,41 @@ jobs:
...
     runs-on: ubuntu-${{ matrix.ubuntu_version }}
     name: Octavia on OpenStack ${{ matrix.name }}
     steps:
       - name: Checkout Gophercloud
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           persist-credentials: false
+
+      - name: Download pre-built amphora image
+        if: ${{ fromJSON(steps.changed-files.outputs.matches) }}
```
kayrus:
```suggestion
```

mandre: Oh, right. Thanks 👍

kayrus (approved):

PR #3777, opened 2026-05-28
CI: improve test assertions

mandre:
This PR does a few things:
- fix the assertion argument order, for less confusing error messages
- fix a couple of wrong assertions that silently passed
- add new test helpers (AssertTrue, AssertFalse, AssertEqualsIgnoreCase, and their Check equivalents)
- add missing assertions to validate fields set during test

### DIFF
--- a/internal/acceptance/openstack/baremetal/httpbasic/allocations_test.go
+++ b/internal/acceptance/openstack/baremetal/httpbasic/allocations_test.go
@@ -43,5 +43,5 @@ func TestAllocationsCreateDestroy(t *testing.T) {
 		return false, nil
 	})
 	th.AssertNoErr(t, err)
-	th.AssertEquals(t, true, found)
+	th.AssertTrue(t, found)
 }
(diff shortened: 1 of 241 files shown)


internal/acceptance/openstack/baremetal/httpbasic/allocations_test.go
```diff
@@ -43,5 +43,5 @@ func TestAllocationsCreateDestroy(t *testing.T) {
 		return false, nil
 	})
 	th.AssertNoErr(t, err)
-	th.AssertEquals(t, found, true)
+	th.AssertEquals(t, true, found)
```
stephenfin: Rather than doing this, I wonder if we could add `AssertTrue` and `AssertFalse` helpers and use those instead? That way there's no second arg to mix up

stephenfin (commented):
Looks like this still needs some work, but I have a suggestion inline. We can do that later also, given this is already quite large

internal/acceptance/openstack/baremetal/httpbasic/allocations_test.go
mandre: That's not a bad idea. I'll do that in a follow-up commit.

stephenfin (approved):

PR #3778, opened 2026-05-29
docs: relax one-operation-per-PR rule

mandre:
Soften the PR scope guidance across contributor documentation to allow grouping related operations in a single PR. This better reflects current practice now that we have stable branches where non-breaking features are backported.

### DIFF
--- a/docs/contributor-tutorial/step-05-pull-requests.md
+++ b/docs/contributor-tutorial/step-05-pull-requests.md
@@ -37,9 +37,6 @@ If you're adding a single field, then a single Pull Request is also fine. See
 [#662](https://github.com/gophercloud/gophercloud/pull/662) as an example of
 this.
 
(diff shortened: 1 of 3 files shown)


docs/contributor-tutorial/step-05-pull-requests.md
```diff
@@ -49,7 +46,7 @@ this.
 ### Adding a Suite of API Calls
 
 If you're adding support for a "suite" of API calls (meaning: Create, Update,
-Delete, Get), then you will need to create one Pull Request for _each_ call.
+Delete, Get), you can group related operations into a single Pull Request.
```
winiciusallan, line 49: Do you think it is worth adding when someone is willing to add more than one suite? For example:

API X and Y are missing, each one with CRUD operations. Should X and Y be on different PRs?

stephenfin (approved):

PR #3779, opened 2026-05-29
Add AGENTS.md

mandre:
Add a reference guide for AI coding agents.
Unfortunately, claude still doesn't understand standards and we need to symlink AGENTS.md to CLAUDE.md.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,215 @@
+# AGENTS.md
+
+Quick reference guide for AI coding agents working in the Gophercloud repository.
+
+**Project:** Gophercloud - Go SDK for OpenStack services
+**Module:** `github.com/gophercloud/gophercloud/v2`
+**Language:** Go (see version in [go.mod](go.mod))
+**Stable Branch:** v2 (main development on `main`)
+
+## Build, Test & Lint Commands
+
+### Running Tests
+
+**Unit tests (default):**
+```bash
+make unit
+```
+
+**Unit tests with verbose output:**
+```bash
+go test -v ./...
+```
+
+**Run single test by name:**
+```bash
+cd openstack/compute/v2/servers
+go test -run TestCreateServer ./...
+```
+
+**Coverage:**
+```bash
+make coverage
+```
+
+**Acceptance tests (requires live OpenStack - may incur charges):**
+```bash
+make acceptance              # All services
+make acceptance-compute      # Specific service
+```
+
+**Run single acceptance test:**
+```bash
+cd internal/acceptance/openstack/compute/v2
+go test -timeout 60m -tags "acceptance" -run TestServersList
+```
+
+### Linting & Formatting
+
+```bash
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+
+**Unit tests with verbose output:**
+```bash
+go test -v ./...
+```
+
+**Run single test by name:**
+```bash
+cd openstack/compute/v2/servers
+go test -run TestCreateServer
```
winiciusallan:
```suggestion
go test -run TestCreateServer ./...
```

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+    "fmt"
+
+    "github.com/gophercloud/gophercloud/v2"
+    "github.com/gophercloud/gophercloud/v2/pagination"
+)
+```
+
+### File Structure
+
+Standard package structure under `openstack/<service>/v2/<resource>/`:
```
winiciusallan: Wouldn't it be something like this? 

```suggestion
Standard package structure under `openstack/<service>/<service_version>/<resource>/`:
```

winiciusallan (commented):
Two small comments inline, otherwise it is very good to me. Thanks!

winiciusallan (approved):
Looks very good to me. I believe that having this file in the project will be a really good addition and will help further AI-assisted contributions to properly follow the project's patterns.

Thanks for proposing this, @mandre

AGENTS.md
```diff
@@ -0,0 +1,215 @@
+# AGENTS.md
+
+Quick reference guide for AI coding agents working in the Gophercloud repository.
+
+**Project:** Gophercloud - Go SDK for OpenStack services
+**Module:** `github.com/gophercloud/gophercloud/v2`
+**Language:** Go 1.25.0
```
stephenfin: Can we point to `go.mod` since this will need manual updates or get outdated

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+```bash
+make lint     # Run golangci-lint in container (Docker/Podman)
+make format   # Run gofmt with simplify flag
+```
+
+**Note:** If lint fails with SELinux errors, run:
+```bash
+chcon -Rt svirt_sandbox_file_t .
+chcon -Rt svirt_sandbox_file_t ~/.cache/golangci-lint
+```
```
stephenfin, line 58: Do we have this in human-facing docs? Can we point to those instead?

stephenfin (changes requested):
Two small requests. Not sure if the second if feasible.

AGENTS.md
mandre: Nothing at the moment, we're a bit lacking in the docs department. One would need to read the Makefile.
Possibly for a follow-up PR?

stephenfin: Yeah, a follow-up is fine.

stephenfin (approved):

PR #3789, opened 2026-06-08
CI: skip failing Magnum and Zaqar master jobs

mandre:
Also restore Aetos that is now passing.


mandre:
I've submitted https://review.opendev.org/c/openstack/zaqar/+/992201 for fixing the zaqar gate, and we have https://github.com/gophercloud/gophercloud/pull/3765 for magnum but it needs more work.

stephenfin (approved):

PR #3806, opened 2026-06-19
dns: fix SudoTenantID type in zones ListOpts

### LINKED ISSUE #3805: DNS: wrong type for X-Auth-Sudo-Tenant-ID header
### What happened?

This PR - https://github.com/gophercloud/gophercloud/pull/3801 introduced a wrong type for the header `X-Auth-Sudo-Tenant-ID`

It should be of type `string` and not `bool`

### Version

v2 (current)

abhijith-darshan:
Fixes #3805

Change `SudoTenantID` field type from `bool` to `string` in zones `ListOpts`. The field carries a tenant UUID via `X-Auth-Sudo-Tenant-ID` header, not a boolean flag.

### DIFF
--- a/openstack/dns/v2/zones/requests.go
+++ b/openstack/dns/v2/zones/requests.go
@@ -41,8 +41,8 @@ type ListOpts struct {
 	// All projects header
 	AllProjects bool `h:"X-Auth-All-Projects"`
 
-	// Sudo tenant ID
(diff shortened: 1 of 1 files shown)


kayrus (approved):

kayrus:
I also missed this typo during the review. Thanks for the fix.

Issue #3809, opened 2026-06-20
Add Trove datastore and backup API support


stias:
This tracks adding missing Gophercloud support for several OpenStack Trove v1 API areas that were originally submitted together in #3767.

The implementation is being split into smaller PRs by API area to keep review scope manageable:

- Trove datastore version management APIs
- Trove datastore version configuration parameter APIs
- Trove database backup APIs
- Trove backup strategy APIs

OpenStack source references:

- https://github.com/openstack/trove/blob/stable/2026.1/trove/extensions/routes/mgmt.py#L73-L83
- https://github.com/openstack/trove/blob/stable/2026.1/trove/common/api.py#L191-L223

PR #3767, opened 2026-05-25
openstack/db: add Trove datastore admin and backup APIs

stias:
Links to the line numbers/files in the OpenStack source code that support the
  code in this PR:

  - Trove datastore version and datastore version configuration parameter management routes:
    https://opendev.org/openstack/trove/src/branch/master/trove/extensions/routes/mgmt.py#L68-L79

  - Trove API reference: create/update/delete datastore version configuration parameters:
    https://docs.openstack.org/api-ref/database/#create-datastore-version-configuration-parameters
    https://docs.openstack.org/api-ref/database/#update-a-datastore-version-configuration-parameter
    https://docs.openstack.org/api-ref/database/#delete-a-datastore-version-configuration-parameter

  - Trove API reference: create/list/show/update/delete datastore versions:
    https://docs.openstack.org/api-ref/database/#create-datastore-version
    https://docs.openstack.org/api-ref/database/#list-datastore-versions
    https://docs.openstack.org/api-ref/database/#show-datastore-version-details
    https://docs.openstack.org/api-ref/database/#update-datastore-version-details
    https://docs.openstack.org/api-ref/database/#delete-datastore-version

  - Trove API reference: list/create/show/delete database backups:
    https://docs.openstack.org/api-ref/database/#list-database-backups
    https://docs.openstack.org/api-ref/database/#create-database-backup
    https://docs.openstack.org/api-ref/database/#show-database-backup-details
    https://docs.openstack.org/api-ref/database/#delete-database-backup

  - Trove API reference: list/create/delete backup strategies:
    https://docs.openstack.org/api-ref/database/#list-backup-strategies
    https://docs.openstack.org/api-ref/database/#create-backup-strategy
    https://docs.openstack.org/api-ref/database/#delete-database-strategy

PR #3810, opened 2026-06-20
openstack/db: add Trove backup APIs

stias:
For #3809

Split out from #3767.

code in this PR:

- Backup routes:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/common/api.py#L191-L208
- Backup controller actions and request fields:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/backup/service.py#L40-L129
- Backup response fields:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/backup/views.py#L17-L57
- Backup model fields, including storage_driver:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/backup/models.py#L422-L429

### DIFF
--- a/openstack/db/v1/backups/requests.go
+++ b/openstack/db/v1/backups/requests.go
@@ -1,0 +1,119 @@
+package backups
+
+import (
+	"context"
+
+	"github.com/gophercloud/gophercloud/v2"
+	"github.com/gophercloud/gophercloud/v2/pagination"
+)
+
+// ListOptsBuilder allows extensions to add additional parameters to the List
+// request.
+type ListOptsBuilder interface {
+	ToBackupListQuery() (string, error)
+}
+
+// ListOpts represents options for listing database backups.
+type ListOpts struct {
+	// Return the list of backups for a particular database instance.
+	InstanceID string `q:"instance_id"`
+	// Return the list of backups for all projects. Admin only.
+	AllProjects bool `q:"all_projects"`
+	// Return backups for a particular datastore.
+	Datastore string `q:"datastore"`
+	// Return backups for a particular project. Admin only.
+	ProjectID string `q:"project_id"`
+}
+
+// ToBackupListQuery formats a ListOpts into a query string.
+func (opts ListOpts) ToBackupListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
+// List will list database backups.
+func List(client *gophercloud.ServiceClient, opts ListOptsBuilder) pagination.Pager {
+	url := baseURL(client)
+	if opts != nil {
+		query, err := opts.ToBackupListQuery()
+		if err != nil {
+			return pagination.Pager{Err: err}
+		}
+		url += query
+	}
+
+	return pagination.NewPager(client, url, func(r pagination.PageResult) pagination.Page {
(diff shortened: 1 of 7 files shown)


mandre:
Hi @stias, thanks a lot for working on improving support for trove, I'll review all of your PRs soon. You might have noticed we were lacking a CI job for trove, so I worked on adding one with https://github.com/gophercloud/gophercloud/pull/3822. Once it merges, it means we'll be able to test part of your changes against a live environment. Only a part, because some tests require an instance, which we'll not have in CI.

May I ask you to rebase your PRs on the above branch so that we have proper CI? 
I think that for backups and backups strategies, we can only test List (although we should have acceptance test for everything, just gated behind the `short` flag like we do for the instance tests), and for datastore versions and config parameters we can test everything.

openstack/db/v1/backups/requests.go
```diff
@@ -0,0 +1,119 @@
...
+// ListOpts represents options for listing database backups.
+type ListOpts struct {
+	// Return the list of backups for a particular database instance.
+	InstanceID string `q:"instance_id"`
+	// Return the list of backups for all projects. Admin only.
+	AllProjects bool `q:"all_projects"`
+	// Return backups for a particular datastore.
+	Datastore string `q:"datastore"`
+	// Return backups for a particular project. Admin only.
+	ProjectID string `q:"project_id"`
```
mandre, line 25: I found it odd at first that the trove API would accept a `project_id` query parameter, when the `project_id` is also mandatory in the request path, but now I understand this is for admin use (still a little odd IMO, but this is how they designed the API so this is what we should support).

openstack/db/v1/backups/urls.go
```diff
@@ -0,0 +1,15 @@
+package backups
+
+import "github.com/gophercloud/gophercloud/v2"
+
+func baseURL(c *gophercloud.ServiceClient) string {
+	return c.ServiceURL("backups")
```
mandre, line 6: The service URL as returned by the service catalog would typically include the project_id, so this should be good 👍

openstack/db/v1/backups/requests.go
```diff
@@ -0,0 +1,119 @@
...
+
+// CreateOptsBuilder is the top-level interface for create backup options.
+type CreateOptsBuilder interface {
+	ToBackupCreateMap() (map[string]any, error)
+}
+
+// RestoreFromOpts contains information for restoring a backup from a remote
+// location.
+type RestoreFromOpts struct {
+	RemoteLocation          string  `json:"remote_location,omitempty"`
```
mandre, line 65: Are these options really all optional in RestoreFromOpts? Not a big deal in practice.

openstack/db/v1/backups/results.go
```diff
@@ -0,0 +1,98 @@
...
+	Description   string
+	ID            string
+	InstanceID    string `json:"instance_id"`
+	LocationRef   string `json:"locationRef"`
+	Name          string
+	ParentID      string `json:"parent_id"`
+	ProjectID     string `json:"project_id"`
+	Size          float64
+	Status        string
+	StorageDriver string    `json:"storage_driver"`
```
mandre, line 25: Nit: should be a pointer, because it's optionally returned.

mandre (commented):
Really nice work, the only thing that would make it perfect would be to add acceptance tests.

PR #3811, opened 2026-06-20
openstack/db: add Trove backup strategy APIs

stias:
For #3809

Split out from #3767.

code in this PR:

- Backup strategy routes:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/common/api.py#L210-L223
- Backup strategy controller actions and request fields:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/backup/service.py#L132-L182
- Backup strategy response fields:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/backup/views.py#L60-L85

### DIFF
--- a/openstack/db/v1/backupstrategies/requests.go
+++ b/openstack/db/v1/backupstrategies/requests.go
@@ -1,0 +1,113 @@
+package backupstrategies
+
+import (
+	"context"
+
+	"github.com/gophercloud/gophercloud/v2"
+	"github.com/gophercloud/gophercloud/v2/pagination"
+)
+
+// ListOptsBuilder allows extensions to add additional parameters to the List
+// request.
+type ListOptsBuilder interface {
+	ToBackupStrategyListQuery() (string, error)
+}
+
+// ListOpts represents options for listing backup strategies.
+type ListOpts struct {
+	// Return the strategy for a particular database instance.
+	InstanceID string `q:"instance_id"`
+	// Return strategies for a particular project. Admin only.
+	ProjectID string `q:"project_id"`
+}
+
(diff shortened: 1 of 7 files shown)


openstack/db/v1/backupstrategies/requests.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Create creates a backup strategy.
+func Create(ctx context.Context, client *gophercloud.ServiceClient, opts CreateOptsBuilder) (r CreateResult) {
+	b, err := opts.ToBackupStrategyCreateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	resp, err := client.Post(ctx, baseURL(client), &b, &r.Body, &gophercloud.RequestOpts{OkCodes: []int{200}})
```
mandre, line 73: According to [the docs](https://docs.openstack.org/api-ref/database/#create-backup-strategy), we should expect a 202 on success.

openstack/db/v1/backupstrategies/requests.go
```diff
@@ -0,0 +1,113 @@
...
+
+// DeleteOptsBuilder allows extensions to add additional parameters to the
+// Delete request.
+type DeleteOptsBuilder interface {
+	ToBackupStrategyDeleteQuery() (string, error)
+}
+
+// DeleteOpts represents options for deleting backup strategies.
+type DeleteOpts struct {
+	// Delete the strategy for a particular database instance.
```
mandre, line 86: Perhaps we should call out that by default, it deletes _all_ strategies within a project if you don't specify an instance id?

PR #3812, opened 2026-06-20
openstack/db: add Trove datastore version APIs

stias:
For #3809

Split out from #3767.

code in this PR:

- Datastore version management route:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/extensions/routes/mgmt.py#L79-L83
- Datastore version controller actions and request fields:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/extensions/mgmt/datastores/service.py#L33-L207
- Datastore version response fields:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/extensions/mgmt/datastores/views.py#L16-L53

### DIFF
--- a/openstack/db/v1/datastores/requests.go
+++ b/openstack/db/v1/datastores/requests.go
@@ -7,6 +7,118 @@ import (
 	"github.com/gophercloud/gophercloud/v2/pagination"
 )
 
+// CreateVersionOptsBuilder is the top-level interface for create version
+// options.
+type CreateVersionOptsBuilder interface {
+	ToVersionCreateMap() (map[string]any, error)
+}
+
+// CreateVersionOpts represents options for registering a datastore version.
+//
+// This is an admin-only API. When the datastore specified by DatastoreName does
+// not exist, Trove creates it automatically.
+type CreateVersionOpts struct {
+	// The name of the datastore version.
+	Name string `json:"name" required:"true"`
+	// The name of the datastore.
+	DatastoreName string `json:"datastore_name" required:"true"`
+	// The datastore manager type.
+	DatastoreManager string `json:"datastore_manager" required:"true"`
+	// The ID of an image. Either Image or ImageTags must be provided.
+	Image string `json:"image,omitempty" or:"ImageTags"`
+	// Image tags used to find the latest matching image. Either Image or
+	// ImageTags must be provided.
+	ImageTags []string `json:"image_tags,omitempty"`
+	// Whether the datastore version is enabled.
+	Active gophercloud.EnabledState `json:"active" required:"true"`
+	// Whether this datastore version is the default for the datastore.
+	Default gophercloud.EnabledState `json:"default,omitempty"`
+	// Packages associated with the datastore version.
+	Packages []string `json:"packages,omitempty"`
+	// The database engine version number.
+	Version string `json:"version,omitempty"`
+}
+
+// ToVersionCreateMap converts a CreateVersionOpts struct into a request body.
+func (opts CreateVersionOpts) ToVersionCreateMap() (map[string]any, error) {
+	return gophercloud.BuildRequestBody(opts, "version")
+}
+
+// CreateVersion registers a new datastore version. This is an admin-only API.
+// Trove automatically creates the datastore when DatastoreName does not already
+// exist.
+func CreateVersion(ctx context.Context, client *gophercloud.ServiceClient, opts CreateVersionOptsBuilder) (r CreateVersionResult) {
+	b, err := opts.ToVersionCreateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	resp, err := client.Post(ctx, createVersionURL(client), &b, &r.Body, &gophercloud.RequestOpts{OkCodes: []int{202}})
+	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
+	return
+}
+
+// ListAllVersions lists all datastore versions. This is an admin-only API.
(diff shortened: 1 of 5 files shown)


openstack/db/v1/datastores/results.go
```diff
@@ -1,15 +1,54 @@
...
+	"encoding/json"
+
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/pagination"
 )
 
+// PackageList represents packages associated with a datastore version.
+type PackageList []string
+
+// UnmarshalJSON supports Trove responses that may return packages as either a
```
winiciusallan, line 13: Does not `package` return only a list? I'm assuming from these lines

https://github.com/openstack/trove/blob/stable/2026.1/trove/extensions/mgmt/datastores/views.py#L32-L33

openstack/db/v1/datastores/requests.go
```diff
@@ -7,6 +7,118 @@ import (
...
+	Name string `json:"name" required:"true"`
+	// The name of the datastore.
+	DatastoreName string `json:"datastore_name" required:"true"`
+	// The datastore manager type.
+	DatastoreManager string `json:"datastore_manager" required:"true"`
+	// The ID of an image. Either Image or ImageTags must be provided.
+	Image string `json:"image,omitempty" or:"ImageTags"`
+	// Image tags used to find the latest matching image. Either Image or
+	// ImageTags must be provided.
+	ImageTags []string `json:"image_tags,omitempty"`
```
mandre, line 31: Should we also add `or:"Image"` ?

openstack/db/v1/datastores/results.go
```diff
@@ -40,6 +79,21 @@ type GetVersionResult struct {
 	gophercloud.Result
 }
 
+// CreateVersionResult represents the result of creating a version.
+type CreateVersionResult struct {
+	gophercloud.Result
```
mandre, line 84: Since the API doesn't return a body, this should be `gophercloud.ErrResult`

openstack/db/v1/datastores/results.go
```diff
@@ -40,6 +79,21 @@ type GetVersionResult struct {
...
+}
+
+// UpdateVersionResult represents the result of updating a version.
+type UpdateVersionResult struct {
+	gophercloud.Result
```
mandre, line 89: ditto

openstack/db/v1/datastores/requests.go
```diff
@@ -7,6 +7,118 @@ import (
...
+// CreateVersion registers a new datastore version. This is an admin-only API.
+// Trove automatically creates the datastore when DatastoreName does not already
+// exist.
+func CreateVersion(ctx context.Context, client *gophercloud.ServiceClient, opts CreateVersionOptsBuilder) (r CreateVersionResult) {
+	b, err := opts.ToVersionCreateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	resp, err := client.Post(ctx, createVersionURL(client), &b, &r.Body, &gophercloud.RequestOpts{OkCodes: []int{202}})
```
mandre, line 56: It seems like Trove doesn't return a body for create requests: https://github.com/openstack/trove/blob/e9489ce73e2415ed277eb51ab45394f9247acc30/trove/extensions/mgmt/datastores/service.py#L94

openstack/db/v1/datastores/requests.go
```diff
@@ -7,6 +7,118 @@ import (
...
+}
+
+// UpdateVersion updates a datastore version. This is an admin-only API.
+func UpdateVersion(ctx context.Context, client *gophercloud.ServiceClient, versionID string, opts UpdateVersionOptsBuilder) (r UpdateVersionResult) {
+	b, err := opts.ToVersionUpdateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+	resp, err := client.Patch(ctx, mgmtVersionURL(client, versionID), &b, &r.Body, &gophercloud.RequestOpts{OkCodes: []int{202}})
```
mandre, line 110: Same here, Trove doesn't return a body: https://github.com/openstack/trove/blob/e9489ce73e2415ed277eb51ab45394f9247acc30/trove/extensions/mgmt/datastores/service.py#L169

openstack/db/v1/datastores/results.go
```diff
@@ -106,3 +160,21 @@ func (r GetVersionResult) Extract() (*Version, error) {
 	err := r.ExtractInto(&s)
 	return s.Version, err
 }
+
+// Extract retrieves a single Version struct from a create result.
```
mandre, line 164: If trove doesn't return a body, the two methods below become dead code.

openstack/db/v1/datastores/results.go
```diff
@@ -1,15 +1,54 @@
...
+	return nil
+}
+
 // Version represents a version API resource. Multiple versions belong to a Datastore.
 type Version struct {
-	ID    string
-	Links []gophercloud.Link
-	Name  string
+	Active           bool
+	Datastore        string
```
mandre, line 38: It looks like we're using the same type for the return of both [`/v1.0/{project_id}/datastores/{datastore}/versions/{version}`](https://docs.openstack.org/api-ref/database/#show-datastore-version) and [`/v1.0/{project_id}/mgmt/datastore-versions/{datastore_version_id}`](https://docs.openstack.org/api-ref/database/#show-datastore-version-details), but they populate different fields?

mandre (commented):
I think in this particular case, having acceptance test is a must because, IIUC, we're expecting a body when trove doesn't return one. We need to verify against a live environment and prove it works.

PR #3813, opened 2026-06-20
openstack/db: add Trove configuration parameter APIs

stias:
For #3809

Split out from #3767.

code in this PR:

- Datastore version configuration parameter management route:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/extensions/routes/mgmt.py#L73-L77
- Datastore version configuration parameter controller actions and request fields:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/extensions/mgmt/configuration/service.py#L32-L135
- Datastore version configuration parameter response fields:
  https://github.com/openstack/trove/blob/stable/2026.1/trove/extensions/mgmt/configuration/views.py#L21-L53

### DIFF
--- a/openstack/db/v1/configurations/requests.go
+++ b/openstack/db/v1/configurations/requests.go
@@ -120,6 +120,145 @@ func Replace(ctx context.Context, client *gophercloud.ServiceClient, configID st
 	return
 }
 
+// CreateParamOptsBuilder is the top-level interface for create parameter
+// options.
+type CreateParamOptsBuilder interface {
+	ToParamCreateMap() (map[string]any, error)
+}
+
+type createParamRequest struct {
+	Name            string `json:"name" required:"true"`
+	DataType        string `json:"data_type" required:"true"`
+	MinSize         *int   `json:"min_size,omitempty"`
+	MaxSize         *int   `json:"max_size,omitempty"`
+	RestartRequired *int   `json:"restart_required" required:"true"`
+}
+
+// CreateParamOpts represents options for registering a configuration parameter
+// for a datastore version.
+type CreateParamOpts struct {
+	// The name of the configuration parameter.
+	Name string `json:"name" required:"true"`
+	// The configuration parameter data type, for example "integer", "string",
+	// "float", or "boolean".
+	DataType string `json:"data_type" required:"true"`
+	// The minimum value for integer parameters.
+	MinSize *int `json:"min_size,omitempty"`
+	// The maximum value for integer parameters.
+	MaxSize *int `json:"max_size,omitempty"`
+	// Whether the database service needs to restart after this parameter
+	// changes.
+	RestartRequired bool `json:"restart_required"`
+}
+
+// ToParamCreateMap converts a CreateParamOpts struct into a request body.
+func (opts CreateParamOpts) ToParamCreateMap() (map[string]any, error) {
+	paramOpts := createParamRequest{
+		Name:            opts.Name,
+		DataType:        opts.DataType,
+		MinSize:         opts.MinSize,
+		MaxSize:         opts.MaxSize,
+		RestartRequired: restartRequiredToInt(opts.RestartRequired),
+	}
+
+	return gophercloud.BuildRequestBody(paramOpts, "configuration-parameter")
+}
+
+// CreateDatastoreVersionParam registers a configuration parameter for a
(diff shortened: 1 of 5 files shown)


openstack/db/v1/configurations/requests.go
```diff
@@ -120,6 +120,145 @@ func Replace(ctx context.Context, client *gophercloud.ServiceClient, configID st
...
+	MinSize *int `json:"min_size,omitempty"`
+	// The maximum value for integer parameters.
+	MaxSize *int `json:"max_size,omitempty"`
+	// Whether the database service needs to restart after this parameter
+	// changes.
+	RestartRequired bool `json:"restart_required"`
+}
+
+// ToParamCreateMap converts a CreateParamOpts struct into a request body.
+func (opts CreateParamOpts) ToParamCreateMap() (map[string]any, error) {
```
winiciusallan, line 155: You can use the existing `CreateParamOpts` struct here instead of creating a new one to build the request body. We avoid having an almost-duplicated struct. For example:

```go
func (opts CreateParamOpts) ToParamCreateMap() (map[string]any, error) {
    b, err := gophercloud.BuildRequestBody(opts, "configuration-parameter")
    if err != nil {
        return nil, err
    }

    if opts.RestartRequired != nil {
        b["restart_required"] = restartRequiredToInt(opts.RestartRequired)
    }

    return b, nil
}
```

Does it make sense?

openstack/db/v1/configurations/requests.go
```diff
@@ -120,6 +120,145 @@ func Replace(ctx context.Context, client *gophercloud.ServiceClient, configID st
...
+	// The configuration parameter data type, for example "integer", "string",
+	// "float", or "boolean".
+	DataType string `json:"data_type" required:"true"`
+	// The minimum value for integer parameters.
+	MinSize *int `json:"min_size,omitempty"`
+	// The maximum value for integer parameters.
+	MaxSize *int `json:"max_size,omitempty"`
+	// Whether the database service needs to restart after this parameter
+	// changes.
+	RestartRequired bool `json:"restart_required"`
```
winiciusallan, line 151: I would say that we need this field to be a *bool because we need to differentiate between the zero-value (false) and a not set value.

If a user does not set the RestartRequired, the marshaler will infer this field as false and pass it to Trove API. Since the field is not required, I believe it would be better to keep it as nil and let the Trove set the default value. Wdyt?

openstack/db/v1/configurations/requests.go
```diff
@@ -120,6 +120,145 @@ func Replace(ctx context.Context, client *gophercloud.ServiceClient, configID st
...
+	Name            string `json:"name,omitempty"`
+	DataType        string `json:"data_type,omitempty"`
+	MinSize         *int   `json:"min_size,omitempty"`
+	MaxSize         *int   `json:"max_size,omitempty"`
+	RestartRequired *int   `json:"restart_required,omitempty"`
+}
+
+// UpdateParamOpts represents options for updating a configuration parameter
+// for a datastore version.
+type UpdateParamOpts struct {
```
winiciusallan, line 196: Ditto for ToParamCreateMap

winiciusallan (commented):
Hi @stias! I brought some comments that I have left at the old PR. Will take another time to review it thoroughly.

openstack/db/v1/configurations/results.go
```diff
@@ -147,3 +188,19 @@ func (r ParamResult) Extract() (*Param, error) {
 	err := r.ExtractInto(&s)
 	return s, err
 }
+
+// Extract will retrieve created params from an operation result.
+func (r CreateParamResult) Extract() ([]Param, error) {
+	var s struct {
+		Params []Param `json:"configuration-parameters"`
```
mandre, line 195: I know the docs at https://docs.openstack.org/api-ref/database/#id23 show the result wrapped in `{"configuration-parameters": [...]}` but it appears the doc is wrong if you look at the source code.

https://github.com/openstack/trove/blob/e9489ce73e2415ed277eb51ab45394f9247acc30/trove/extensions/mgmt/configuration/service.py#L96

and

https://github.com/openstack/trove/blob/e9489ce73e2415ed277eb51ab45394f9247acc30/trove/extensions/mgmt/configuration/views.py#L26

Having acceptance test would prove this works as expected (or not).

PR #3843, opened 2026-07-10
Get and Update methods for Manila sharetype

### LINKED ISSUE #3842: Manila: add resources for share types and share type access
### What is missing?

It is often possible to for multi-tenant access to shared file storage to be implemented by using private share types that are then shared to the appropriate tenancy.

Gophercloud provides some support for creating, listing and deleting share types, but it needs methods for getting data on a given share type and updating the properties in-place.  This would enable the definition of appropriate terraform resources.

oneswig:
Fixes #3842 

Implement Manila share type get and update API methods.

API reference for share types: https://docs.openstack.org/api-ref/shared-file-system/#share-types

Further unit tests have been added to improve parameter coverage and some minor issues resolved with the existing unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -229,6 +229,7 @@ The above code sample creates a new server with the parameters, and returns a
 |         Messaging        |       Zaqar      |        `openstack/messaging`       |    ✔    |    ✔    |
 |        Networking        |      Neutron     |       `openstack/networking`       |    ✔    |    ✔    |
 |      Object Storage      |       Swift      |      `openstack/objectstorage`     |    ✔    |    ✔    |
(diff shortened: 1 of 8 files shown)


oneswig:
Should we make this PR on `main` first and follow with a backport to `v2`?

kayrus:
@oneswig yes, you need to make a PR against main branch first. Changing the `DriverHandlesShareServers` to a pointer can be shipped only to main branch, this is an incompatible change. If you still want to change struct member type, I'd recommend to do this in a separate PR.

stephenfin:
I'm going to close this per @kayrus's comments above. If you propose a change against `main` (without the incompatible change: you can do that in a follow-up/stacked PR) then a backport PR will be automatically proposed for you.

PR #3845, opened 2026-07-12
compute: Add shareattach package for Nova microversion 2.97

IlanZuckerman:
Add the `shareattach` package under `compute/v2` to support attaching and
detaching Manila shares on servers via Nova microversion 2.97.
Includes List, Create, Get, and Delete operations, unit tests, and an
acceptance test (`TestShareAttach`) with helpers in compute acceptance code.
The instance must be in `SHUTOFF` status to attach or detach a share.
Acceptance testing was validated against DevStack with Manila enabled and
in CI via `functional-sharedfilesystems` (all three matrix legs passed).
## CI
- `TestShareAttach` runs in `functional-sharedfilesystems` (not
  `functional-compute`) with DevStack `file_backed_memory` config for Nova
  share attach.
- Build tag `sharedfilesystems || shareattach` excludes the test from the
  compute job.
## OpenStack references
- https://docs.openstack.org/nova/2025.1/reference/api-microversion-history.html#id97 (microversion 2.97)
- https://github.com/openstack/nova/blob/stable/2026.1/nova/api/openstack/compute/server_shares.py
- https://docs.openstack.org/api-ref/compute/#servers-with-shares-attachments-servers-shares

### DIFF
--- a/.github/workflows/functional-sharedfilesystems.yaml
+++ b/.github/workflows/functional-sharedfilesystems.yaml
@@ -42,6 +42,8 @@ jobs:
             openstack/utils/choose_version.go
             openstack/utils/discovery.go
             **sharedfilesystems**
+            openstack/compute/v2/shareattach/**
+            internal/acceptance/openstack/compute/v2/compute.go
             .github/workflows/functional-sharedfilesystems.yaml
 
       - name: Skip tests for unrelated changed-files
@@ -49,6 +51,10 @@ jobs:
         run: |
           echo "No relevant files changed - skipping tests for ${{ matrix.name }}"
           echo "TESTS_SKIPPED=true" >> $GITHUB_ENV
+
+      - name: Install virtiofsd
+        if: ${{ fromJSON(steps.changed-files.outputs.matches) }}
+        run: sudo apt-get update && sudo apt-get install -y virtiofsd
 
       - name: Deploy devstack
         if: ${{ fromJSON(steps.changed-files.outputs.matches) }}
@@ -71,6 +77,15 @@ jobs:
             MANILA_DEFAULT_SHARE_TYPE_EXTRA_SPECS='snapshot_support=True create_share_from_snapshot_support=True revert_to_snapshot_support=True mount_snapshot_support=True'
             MANILA_CONFIGURE_DEFAULT_TYPES=True
             MANILA_INSTALL_TEMPEST_PLUGIN_SYSTEMWIDE=false
+
+            [[post-config|\$NOVA_CONF]]
+            [DEFAULT]
+            reserved_host_memory_mb = 0
+            ram_allocation_ratio = 1.0
+
+            [libvirt]
+            file_backed_memory = 4096
+            mem_stats_period_s = 0
 
             ${{ matrix.devstack_conf_overrides }}
           enabled_services: "openstack-cli-server"
--- a/.github/workflows/functional-sharedfilesystems.yaml
+++ b/.github/workflows/functional-sharedfilesystems.yaml
@@ -80,7 +80,6 @@ jobs:
 
             [[post-config|\$NOVA_CONF]]
             [DEFAULT]
-            reserved_host_memory_mb = 0
             ram_allocation_ratio = 1.0
 
             [libvirt]
--- a/internal/acceptance/openstack/compute/v2/shareattach_test.go
+++ b/internal/acceptance/openstack/compute/v2/shareattach_test.go
@@ -1,0 +1,49 @@
+//go:build acceptance || compute || shareattach
+
+package v2
+
+import (
+	"context"
+	"testing"
+
+	"github.com/gophercloud/gophercloud/v2/internal/acceptance/clients"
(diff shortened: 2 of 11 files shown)


.github/workflows/functional-sharedfilesystems.yaml
```diff
@@ -72,6 +78,15 @@ jobs:
             MANILA_CONFIGURE_DEFAULT_TYPES=True
             MANILA_INSTALL_TEMPEST_PLUGIN_SYSTEMWIDE=false
 
+            [[post-config|\$NOVA_CONF]]
+            [DEFAULT]
+            reserved_host_memory_mb = 0
+            ram_allocation_ratio = 1.0
```
winiciusallan, line 83: This configuration is 1.0 by default, I believe we don't need to explicitly set here.

.github/workflows/functional-sharedfilesystems.yaml
```diff
@@ -72,6 +78,15 @@ jobs:
             MANILA_CONFIGURE_DEFAULT_TYPES=True
             MANILA_INSTALL_TEMPEST_PLUGIN_SYSTEMWIDE=false
 
+            [[post-config|\$NOVA_CONF]]
+            [DEFAULT]
+            reserved_host_memory_mb = 0
```
winiciusallan: Any special reason for setting this?

.github/workflows/functional-sharedfilesystems.yaml
```diff
@@ -42,6 +42,8 @@ jobs:
             openstack/utils/choose_version.go
             openstack/utils/discovery.go
             **sharedfilesystems**
+            openstack/compute/v2/shareattach/**
+            internal/acceptance/openstack/compute/v2/compute.go
```
winiciusallan, line 46: I think this is fragile in a sense that we will run the Manila functional jobs for changes in compute.go even if they're unrelated.

What if we move this code to the `internal/acceptance/openstack/sharedfilesystems/v2`?

winiciusallan (commented):
Really great work @IlanZuckerman! It is very good overall, just minor comments inline.

.github/workflows/functional-sharedfilesystems.yaml
IlanZuckerman: Done. Removed `reserved_host_memory_mb = 0` .  
share attach acceptance + manual tests still pass on DevStack with only `ram_allocation_ratio = 1.0` + `file_backed_memory`

.github/workflows/functional-sharedfilesystems.yaml
IlanZuckerman: Tested on DevStack: Nova fails unless `ram_allocation_ratio = 1.0`  `(InternalError: Running Nova with file_backed_memory requires ram_allocation_ratio configured to 1.0).`.
DevStack default isn’t 1.0 here, so we should keep it explicit.

.github/workflows/functional-sharedfilesystems.yaml
IlanZuckerman: We're keeping the test and helpers in `compute/v2` because share attach is a Nova API call - same idea as `volumeattach`, with Manila only used to create the share fixture.
We're also keeping `compute.go` in the CI file filter so changes to the share-attach helpers still trigger the Manila job, not just edits under `shareattach/`. 
We looked at dropping that pattern or moving everything to `sharedfilesystems/v2`, but that would break the existing upstream pattern without much gain, so we're leaving it as-is.

IlanZuckerman:
> [Coverage Status](https://coveralls.io/builds/80594397)
> 
> coverage: 63.497% (-0.05%) from 63.545% — **IlanZuckerman:add-compute-shareattach** into **gophercloud:main**

Coverage dipped ~0.05% because we added new code plus acceptance-only helpers. The `shareattach` package itself has unit tests. The acceptance path isn’t included in Coveralls’ unit coverage metric IMO.

winiciusallan (approved):
Thanks @IlanZuckerman! Looks very good.

internal/acceptance/openstack/compute/v2/shareattach_test.go
```diff
@@ -0,0 +1,49 @@
...
+	err = servers.Stop(context.TODO(), client, server.ID).ExtractErr()
+	th.AssertNoErr(t, err)
+
+	err = WaitForComputeStatus(client, server, "SHUTOFF")
+	th.AssertNoErr(t, err)
+
+	shareAttachment, err := CreateShareAttachment(t, client, server, share)
+	th.AssertNoErr(t, err)
+	defer DeleteShareAttachment(t, client, server, shareAttachment)
+
```
eshulman2, line 46: Could we also exercise `shareattach.List` here and assert that the newly created attachment is present? The acceptance test currently covers Create, Get, and Delete, but without List the live endpoint and its response schema are only covered by the mocked unit test.

IlanZuckerman: Good catch! I added a List after create and assert the new attachment is present (with Tag/Status checks).

eshulman2 (approved):
lgtm

kayrus (approved):

PR #3852, opened 2026-07-14
Fix sharetype creation failing when DriverHandlesShareServers is false 

### LINKED ISSUE #3851: Creating sharetype with DriverHandlesShareServers: false fails
### What happened?

Creating a sharetype with DriverHandlesShareServers set to false fails with `Missing input for argument [DriverHandlesShareServers]`

It should accept false as a valid input for DriverHandlesShareServers

### Version

v2 (current)

TBExtent:
Fixes #3851 

Change DriverHandlesShareServers to a pointer so that false values aren't detected as a missing input.

Unit tests:
Adds a second sharetype creation test to test passing in false values.

### DIFF
--- a/openstack/sharedfilesystems/v2/sharetypes/requests.go
+++ b/openstack/sharedfilesystems/v2/sharetypes/requests.go
@@ -28,7 +28,7 @@ type CreateOpts struct {
 // ExtraSpecsOpts represent the extra specifications that can be selected for a share type
 type ExtraSpecsOpts struct {
 	// An extra specification that defines the driver mode for share server, or storage, life cycle management
-	DriverHandlesShareServers bool `json:"driver_handles_share_servers" required:"true"`
+	DriverHandlesShareServers *bool `json:"driver_handles_share_servers" required:"true"`
 	// An extra specification that filters back ends by whether they do or do not support share snapshots
 	SnapshotSupport *bool `json:"snapshot_support,omitempty"`
 }
--- a/openstack/sharedfilesystems/v2/sharetypes/testing/requests_test.go
+++ b/openstack/sharedfilesystems/v2/sharetypes/testing/requests_test.go
@@ -61,7 +61,7 @@ func TestCreateFalse(t *testing.T) {
 	th.AssertNoErr(t, err)
 
 	th.AssertEquals(t, st.Name, "my_new_share_type")
-	th.AssertEquals(t, st.IsPublic, false)
+	th.AssertFalse(t, st.IsPublic)
 }
 
 // Verifies that a share type can't be created if the required parameters are missing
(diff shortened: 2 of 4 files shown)


openstack/sharedfilesystems/v2/sharetypes/requests.go
```diff
@@ -28,7 +28,7 @@ type CreateOpts struct {
 // ExtraSpecsOpts represent the extra specifications that can be selected for a share type
 type ExtraSpecsOpts struct {
 	// An extra specification that defines the driver mode for share server, or storage, life cycle management
-	DriverHandlesShareServers bool `json:"driver_handles_share_servers" required:"true"`
+	DriverHandlesShareServers *bool `json:"driver_handles_share_servers" required:"true"`
```
kayrus, line 31: FYI: This change won't be be backported. If you still use v2, I recommend you to define a custom struct that satisfies the `CreateOptsBuilder` interface and use it until gophercloud v3 is released.

TBExtent: ok

openstack/sharedfilesystems/v2/sharetypes/testing/requests_test.go
```diff
@@ -36,6 +37,33 @@ func TestCreate(t *testing.T) {
...
+		Name:       "my_new_share_type",
+		IsPublic:   false,
+		ExtraSpecs: extraSpecs,
+	}
+
+	st, err := sharetypes.Create(context.TODO(), client.ServiceClient(fakeServer), options).Extract()
+	th.AssertNoErr(t, err)
+
+	th.AssertEquals(t, st.Name, "my_new_share_type")
+	th.AssertEquals(t, st.IsPublic, false)
```
kayrus:
```suggestion
	th.AssertFalse(t, st.IsPublic)
```

openstack/sharedfilesystems/v2/sharetypes/testing/requests_test.go
```diff
@@ -36,6 +37,33 @@ func TestCreate(t *testing.T) {
...
+	options := &sharetypes.CreateOpts{
+		Name:       "my_new_share_type",
+		IsPublic:   false,
+		ExtraSpecs: extraSpecs,
+	}
+
+	st, err := sharetypes.Create(context.TODO(), client.ServiceClient(fakeServer), options).Extract()
+	th.AssertNoErr(t, err)
+
+	th.AssertEquals(t, st.Name, "my_new_share_type")
```
winiciusallan: Just for the semantic of the assert, it should be expected then actual

```suggestion
	th.AssertEquals(t, "my_new_share_type", st.Name)
```

winiciusallan: Since we're here, I would also add some asserts verifying the dhss value.

winiciusallan (commented):
Just a nit comment, otherwise the change looks good.

openstack/sharedfilesystems/v2/sharetypes/testing/requests_test.go
TBExtent: Done!

winiciusallan (approved):
Thanks for fixing this. LGTM

kayrus (approved):

PR #3853, opened 2026-07-14
Add methods for Manila share type get and update

### LINKED ISSUE #3842: Manila: add resources for share types and share type access

TBExtent:
Fixes https://github.com/gophercloud/gophercloud/issues/3842

Implement Manila share type get and update API methods.

API reference for share types: https://docs.openstack.org/api-ref/shared-file-system/#share-types

Unit tests:
Adds unit tests for getting and updating sharetypes

### DIFF
--- a/internal/acceptance/openstack/sharedfilesystems/v2/sharetypes_test.go
+++ b/internal/acceptance/openstack/sharedfilesystems/v2/sharetypes_test.go
@@ -50,9 +50,7 @@ func TestShareTypeUpdateGet(t *testing.T) {
 		IsPublic:    &isPublic,
 	}
 
-	t.Logf("Share type ID: %s", shareType.ID)
 	_, err = sharetypes.Update(context.TODO(), client, shareType.ID, options).Extract()
-	t.Logf("Update error: %s", err)
 	if err != nil {
 		t.Fatalf("Unable to update share type: %s", shareType.Name)
 	}
--- a/internal/acceptance/openstack/sharedfilesystems/v2/sharetypes_test.go
+++ b/internal/acceptance/openstack/sharedfilesystems/v2/sharetypes_test.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/clients"
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/tools"
 	"github.com/gophercloud/gophercloud/v2/openstack/sharedfilesystems/v2/sharetypes"
+	th "github.com/gophercloud/gophercloud/v2/testhelper"
 )
 
 func TestShareTypeCreateDestroy(t *testing.T) {
@@ -29,15 +30,11 @@ func TestShareTypeCreateDestroy(t *testing.T) {
 
 func TestShareTypeUpdateGet(t *testing.T) {
 	client, err := clients.NewSharedFileSystemV2Client()
-	if err != nil {
-		t.Fatalf("Unable to create shared file system client: %v", err)
-	}
+	th.AssertNoErr(t, err)
 	client.Microversion = "2.50"
 
 	shareType, err := CreateShareType(t, client)
-	if err != nil {
-		t.Fatalf("Unable to create share type: %v", err)
-	}
+	th.AssertNoErr(t, err)
 	defer DeleteShareType(t, client, shareType)
 
 	name := "ACPTTEST_new_share_type_name"
@@ -51,26 +48,14 @@ func TestShareTypeUpdateGet(t *testing.T) {
 	}
 
 	_, err = sharetypes.Update(context.TODO(), client, shareType.ID, options).Extract()
-	if err != nil {
-		t.Fatalf("Unable to update share type: %s", shareType.Name)
-	}
+	th.AssertNoErr(t, err)
 
 	st, err := sharetypes.Get(context.TODO(), client, shareType.ID).Extract()
-	if err != nil {
-		t.Fatalf("Unable to retrieve share type: %s", shareType.Name)
-	}
+	th.AssertNoErr(t, err)
 
-	if st.Name != "ACPTTEST_new_share_type_name" {
-		t.Fatal("Share type name was expected to be ACPTTEST_new_share_type_name")
-	}
-
-	if *st.Description != "my new share type description" {
-		t.Fatal("Share type description was expected to be 'my new share type description'")
-	}
-
-	if st.IsPublic != true {
-		t.Fatal("Share type was expected to be public")
-	}
+	th.AssertEquals(t, "ACPTTEST_new_share_type_name", st.Name)
+	th.AssertEquals(t, "my new share type description", *st.Description)
+	th.AssertTrue(t, st.IsPublic)
 
 	tools.PrintResource(t, shareType)
 }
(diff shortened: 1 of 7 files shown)


openstack/sharedfilesystems/v2/sharetypes/results.go
```diff
@@ -18,6 +18,10 @@ type ShareType struct {
 	RequiredExtraSpecs map[string]any `json:"required_extra_specs"`
 	// The extra specifications for the share type
 	ExtraSpecs map[string]any `json:"extra_specs"`
+	// The human-readable description for the share type
+	Description *string `json:"description"`
+	// Indicates whether the share type is the default one
+	IsDefault bool `json:"is_default"`
```
kayrus: I wonder whether it's possible to update this field?

TBExtent: What do you mean by that?

kayrus: Is there an API that allows to set default sharetypes?

winiciusallan: Yup, it is possible, but via conf file.

https://docs.openstack.org/manila/latest/admin/shared-file-systems-share-types.html#default-share-type

I'm wondering what is returned if the a default share type is not defined. should we have a pointer here?

TBExtent: I'm happy to change it to a pointer.

winiciusallan: It looks like `is_default` is always returned indicating whether the share type is the default one or not.

```
x curl -s -H "X-Openstack-Manila-API-Version: 2.46" -H "X-Auth-Token: $TOKEN" http://devstack.example.com/share/v2/types/249b0d86-469a-4244-a40f-270f2838ffd6 | jq '.share_type'
{
  "id": "249b0d86-469a-4244-a40f-270f2838ffd6",
  "name": "dhss_false",
  "extra_specs": {
    "snapshot_support": "True",
    "create_share_from_snapshot_support": "True",
    "revert_to_snapshot_support": "True",
    "mount_snapshot_support": "True",
    "driver_handles_share_servers": "False"
  },
  "required_extra_specs": {
    "driver_handles_share_servers": "False"
  },
  "share_type_access:is_public": true,
  "description": null,
  "is_default": false
}
```

In this case I believe we don't need the pointer (we don't need to compare the zero-value with unset)

winiciusallan (commented):
Hey @TBExtent, would you mind adding some acceptance tests?

TBExtent:
Working on some, having fun with microversions

openstack/sharedfilesystems/v2/sharetypes/results.go
TBExtent: It probably should be, looking into it further. `is_default` was added in version 2.46, and I think it would be safer if it was unset if the client is using an earlier microversion

internal/acceptance/openstack/sharedfilesystems/v2/sharetypes_test.go
```diff
@@ -27,6 +27,56 @@ func TestShareTypeCreateDestroy(t *testing.T) {
...
+	client, err := clients.NewSharedFileSystemV2Client()
+	if err != nil {
+		t.Fatalf("Unable to create shared file system client: %v", err)
+	}
+	client.Microversion = "2.50"
+
+	shareType, err := CreateShareType(t, client)
+	if err != nil {
+		t.Fatalf("Unable to create share type: %v", err)
+	}
```
winiciusallan: Let's use our test helper throughout all error handling statement.


```suggestion
th.AssertNoErr(t, err)
```

internal/acceptance/openstack/sharedfilesystems/v2/sharetypes_test.go
```diff
@@ -27,6 +27,56 @@ func TestShareTypeCreateDestroy(t *testing.T) {
...
+		t.Fatal("Share type name was expected to be ACPTTEST_new_share_type_name")
+	}
+
+	if *st.Description != "my new share type description" {
+		t.Fatal("Share type description was expected to be 'my new share type description'")
+	}
+
+	if st.IsPublic != true {
+		t.Fatal("Share type was expected to be public")
+	}
```
winiciusallan: Let's also use our test helper here. You can do instead


```suggestion
th.AssertEquals(t, "ACPTTEST_new_share_type_name", st.Name)
th.AssertEquals(t, "my new share type description", st.Description)
th.AssertTrue(t, *st.IsPublic)
```

openstack/sharedfilesystems/v2/sharetypes/results.go
```diff
@@ -18,6 +18,17 @@ type ShareType struct {
...
+	Description *string `json:"description"`
+	// Indicates whether the share type is the default one
+	IsDefault *bool `json:"is_default"`
+}
+
+// shareTypeV6Compat provides compatibility for os-share-type-access:is_public
+// being renamed to share_type_access:is_public at microversion v6 or greater.
+type shareTypeV6Compat struct {
+	// Indicates whether a share type is publicly accessible
+	IsPublic *bool `json:"share_type_access:is_public"`
```
winiciusallan, line 31: Since this field is the expected in the latest versions of the API, I would love to see some tests referencing it.

winiciusallan (commented):
Thanks for adding the tests, @TBExtent. Dropped a few comments inline.

openstack/sharedfilesystems/v2/sharetypes/results.go
```diff
@@ -18,6 +18,17 @@ type ShareType struct {
 	RequiredExtraSpecs map[string]any `json:"required_extra_specs"`
 	// The extra specifications for the share type
 	ExtraSpecs map[string]any `json:"extra_specs"`
+	// The human-readable description for the share type
+	Description *string `json:"description"`
+	// Indicates whether the share type is the default one
+	IsDefault *bool `json:"is_default"`
```
kayrus, line 24: I still curious whether it is possible to set this flag using API. It is not documented in API ref for requests. Also, does it work as a list filter?

openstack/sharedfilesystems/v2/sharetypes/results.go
TBExtent: The `TestShareTypeUpdateGet` acceptance test will fail without it. I've added it to the update unit test as well, and added a get test for it.

openstack/sharedfilesystems/v2/sharetypes/results.go
TBExtent: I can't see anything in the Manila source for either

winiciusallan: I also didn't find anything in the source code. It is not used as a filter when listing shares[1] and to set this value it is via configuration file per the docs[2]

[1] https://github.com/openstack/manila/blob/master/manila/api/v2/share_types.py#L120
[2] https://docs.openstack.org/manila/latest/admin/shared-file-systems-share-types.html#default-share-type

kayrus (approved):

TBExtent:
Hi @kayrus, it failed to merge due to an internal error in Coveralls, please could you add it to the merge queue again?

PR #3869, opened 2026-07-22
auth: don't send scope with application credentials

### LINKED ISSUE #3868: auth: application credentials should not send a scope in token requests
## Bug

When authenticating with application credentials, `ToTokenV3ScopeMap` still builds and returns a project scope from `AuthScope` (or from the `TenantID`/`TenantName` backwards-compatibility fallback). Keystone rejects this with HTTP 401:

```
Application credentials cannot request a scope.
```

This is because application credentials carry an **implicit** project scope baked into the credential at creation time. Sending an explicit scope alongside them is an error per the Keystone spec.

## Keystone source reference

https://opendev.org/openstack/keystone/src/branch/stable/2025.1/keystone/auth/plugins/application_credential.py

The plugin raises `ApplicationCredentialAuthError` when a scope is provided alongside application credentials.

## Expected behaviour

`ToTokenV3ScopeMap` should return an empty map when `ApplicationCredentialID` or `ApplicationCredentialName` is set, so that no scope key is included in the token request body.

## Proposed fix

Add an early-return guard at the top of `ToTokenV3ScopeMap`:

```go
if opts.ApplicationCredentialID != "" || opts.ApplicationCredentialName != "" {
    return map[string]any{}, nil
}
```

This mirrors the behaviour of the existing unit tests in `openstack/identity/v3/tokens/testing/requests_test.go` (e.g. `TestCreateApplicationCredentialIDAndSecret`) which already pass `nil` scope and expect no scope in the JSON body — but those tests use `tokens.AuthOptions`, not `gophercloud.AuthOptions`, so the bug only surfaces when the top-level `AuthOptions` is used with a populated `Scope` field or `TenantID`/`TenantName`.

fwiesel:
Fixes #3868

When authenticating with application credentials, `ToTokenV3ScopeMap` was including a project scope in the token request. Keystone rejects this with HTTP 401: `Application credentials cannot request a scope.`

Application credentials carry an implicit project scope already. The fix returns an empty scope map early when application credentials are present.

Links to OpenStack source confirming that scopes are rejected with application credentials:
https://opendev.org/openstack/keystone/src/branch/stable/2025.1/keystone/auth/plugins/application_credential.py

### DIFF
--- a/auth_options.go
+++ b/auth_options.go
@@ -401,6 +401,13 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]any) (map[string]an
 // ToTokenV3ScopeMap builds a scope from AuthOptions and satisfies interface in
 // the v3 tokens package.
 func (opts *AuthOptions) ToTokenV3ScopeMap() (map[string]any, error) {
+	// Application credentials carry an implicit project scope in Keystone.
+	// Sending an explicit scope alongside them results in a 401 error:
+	// "Application credentials cannot request a scope."
+	if opts.ApplicationCredentialID != "" || opts.ApplicationCredentialName != "" {
+		return map[string]any{}, nil
+	}
+
 	// For backwards compatibility.
 	// If AuthOptions.Scope was not set, try to determine it.
 	// This works well for common scenarios.
--- a/auth_options.go
+++ b/auth_options.go
@@ -405,7 +405,7 @@ func (opts *AuthOptions) ToTokenV3ScopeMap() (map[string]any, error) {
 	// Sending an explicit scope alongside them results in a 401 error:
 	// "Application credentials cannot request a scope."
 	if opts.ApplicationCredentialID != "" || opts.ApplicationCredentialName != "" {
-		return map[string]any{}, nil
+		return nil, nil
 	}
 
 	// For backwards compatibility.
--- a/testing/auth_options_test.go
+++ b/testing/auth_options_test.go
@@ -145,7 +145,7 @@ func TestToTokenV3ScopeMap(t *testing.T) {
 					DomainName:  "Default",
 				},
 			},
-			map[string]any{},
+			nil,
 		},
 		{
 			"Application credential by name: no scope sent",
@@ -159,7 +159,7 @@ func TestToTokenV3ScopeMap(t *testing.T) {
 					DomainName:  "Default",
 				},
 			},
-			map[string]any{},
+			nil,
 		},
 	}
 	for _, successCase := range successCases {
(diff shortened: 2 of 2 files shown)


kayrus (dismissed):

auth_options.go
```diff
@@ -401,6 +401,13 @@ func (opts *AuthOptions) ToTokenV3CreateMap(scope map[string]any) (map[string]an
 // ToTokenV3ScopeMap builds a scope from AuthOptions and satisfies interface in
 // the v3 tokens package.
 func (opts *AuthOptions) ToTokenV3ScopeMap() (map[string]any, error) {
+	// Application credentials carry an implicit project scope in Keystone.
+	// Sending an explicit scope alongside them results in a 401 error:
+	// "Application credentials cannot request a scope."
+	if opts.ApplicationCredentialID != "" || opts.ApplicationCredentialName != "" {
+		return map[string]any{}, nil
```
kayrus: does this make sense?
```suggestion
		return nil, nil
```

winiciusallan: Yeah, this makes sense to me

winiciusallan (approved):
Per what is done in Keystone source code[1], I think this is consistent and looks correct.

https://github.com/openstack/keystone/blob/30ef2ffa65a3486ef882f00538e20f2253c57d4c/keystone/auth/core.py#L280-L284

testing/auth_options_test.go
```diff
@@ -135,6 +135,32 @@ func TestToTokenV3ScopeMap(t *testing.T) {
...
+			"Application credential by ID: no scope sent",
+			gophercloud.AuthOptions{
+				ApplicationCredentialID:     "26bb287fd56a41f8a577c47f79221187",
+				ApplicationCredentialSecret: "secret",
+				Scope: &gophercloud.AuthScope{
+					ProjectName: "admin",
+					DomainName:  "Default",
+				},
+			},
+			map[string]any{},
```
kayrus:
```suggestion
			nil,
```

testing/auth_options_test.go
```diff
@@ -135,6 +135,32 @@ func TestToTokenV3ScopeMap(t *testing.T) {
...
+				ApplicationCredentialName:   "myappcred",
+				ApplicationCredentialSecret: "secret",
+				Username:                    "someuser",
+				DomainName:                  "Default",
+				Scope: &gophercloud.AuthScope{
+					ProjectName: "admin",
+					DomainName:  "Default",
+				},
+			},
+			map[string]any{},
```
kayrus:
```suggestion
			nil,
```

kayrus (approved):

stephenfin (approved):

Issue #3889, opened 2026-07-28
Support full OS-FEDERATION resource


Sebor:
### What is missing?

OS-FEDERATION resource:
  - Identity providers
  - Federation protocols
  - Service providers

PR #3890, opened 2026-07-28
Add full OS-FEDERATION resource support

Sebor:
## Summary

  Add full OS-FEDERATION resource support for:

  - Identity providers
  - Federation protocols
  - Service providers

  Includes CRUD/list operations, pagination, nullable field handling, documentation, unit tests, and acceptance tests.

  ## Testing

  - go test ./openstack/identity/v3/...
  - Acceptance tests compile with the federation tag
  - go vet

  For [3889](https://github.com/gophercloud/gophercloud/issues/3889)

  OpenStack reference:
  - [Keystone OS-FEDERATION API implementation (stable/2025.1)](https://opendev.org/openstack/keystone/src/branch/stable%2F2025.1/keystone/api/os_federation.py)
  - [OS-FEDERATION API reference](https://docs.openstack.org/api-ref/identity/v3-ext/#os-federation-api)

### DIFF
--- a/openstack/identity/v3/federation/requests.go
+++ b/openstack/identity/v3/federation/requests.go
@@ -7,9 +7,39 @@ import (
 	"github.com/gophercloud/gophercloud/v2/pagination"
 )
 
+// ListIdentityProvidersOptsBuilder allows extensions to add additional
+// parameters to the ListIdentityProviders request.
+type ListIdentityProvidersOptsBuilder interface {
+	ToIdentityProviderListQuery() (string, error)
+}
+
+// ListIdentityProvidersOpts provides options for filtering identity providers.
+type ListIdentityProvidersOpts struct {
+	// ID filters the response by identity provider ID.
+	ID string `q:"id"`
+
+	// Enabled filters the response by whether identity providers are enabled.
+	Enabled *bool `q:"enabled"`
+}
+
+// ToIdentityProviderListQuery formats ListIdentityProvidersOpts into a query
+// string.
+func (opts ListIdentityProvidersOpts) ToIdentityProviderListQuery() (string, error) {
+	q, err := gophercloud.BuildQueryString(opts)
+	return q.String(), err
+}
+
 // ListIdentityProviders enumerates the identity providers.
-func ListIdentityProviders(client *gophercloud.ServiceClient) pagination.Pager {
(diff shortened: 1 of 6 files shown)


openstack/identity/v3/federation/requests.go
```diff
@@ -7,6 +7,143 @@ import (
 	"github.com/gophercloud/gophercloud/v2/pagination"
 )
 
+// ListIdentityProviders enumerates the identity providers.
+func ListIdentityProviders(client *gophercloud.ServiceClient) pagination.Pager {
```
winiciusallan: Per the documentation, it can accept two query parameters https://docs.openstack.org/api-ref/identity/v3-ext/#list-identity-providers. If this is the case, we should include them in the request.

If you're not familiar with the code base, you can follow one of the already implemented API suites e.g. https://github.com/gophercloud/gophercloud/blob/main/openstack/compute/v2/servers/requests.go#L105

openstack/identity/v3/federation/requests.go
```diff
@@ -89,3 +226,229 @@ func DeleteMapping(ctx context.Context, client *gophercloud.ServiceClient, mappi
...
+
+// DeleteProtocol deletes a protocol from an identity provider.
+func DeleteProtocol(ctx context.Context, client *gophercloud.ServiceClient, identityProviderID, protocolID string) (r DeleteProtocolResult) {
+	resp, err := client.Delete(ctx, protocolsResourceURL(client, identityProviderID, protocolID), nil)
+	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
+	return
+}
+
+// ListServiceProviders enumerates the service providers.
+func ListServiceProviders(client *gophercloud.ServiceClient) pagination.Pager {
```
winiciusallan: It looks like service providers also accepts query parameters https://github.com/openstack/keystone/blob/master/keystone/federation/schema.py#L93-L99

winiciusallan (commented):
A quick review. I'd need another round. Thanks for adding this @Sebor!1

openstack/identity/v3/federation/requests.go
Sebor: Hi @winiciusallan, thanks for the review! Added `id` and `enabled` query options to both `ListIdentityProviders` and `ListServiceProviders`

openstack/identity/v3/federation/requests.go
Sebor: Done too

PR #3914, opened 2026-08-06
feat: Add advanced tag filters on listing l7policy

### LINKED ISSUE #3913: Support advanced tag filters in list L7policy
### What is missing?

Hi maintainers, 
I ran into this while working on my own gateway controller that discovers and reconciles octavia resources using ownership tags.

and now Octavia supports the following advanced tag filters since API version 2.5:
- tags
- tags-any
- not-tags
- not-tags-any

These filters are documented in the [Octavia API reference](https://docs.openstack.org/api-ref/load-balancer/v2/#filtering-by-tags). L7policies and L7rules are tagged resources, and the official Octavia client exposes these filters for both list operations.

and also i found current gophercloud already supports tags when creating and updating L7policies and rules (following https://github.com/gophercloud/gophercloud/pull/2492). However, on the current main branch, neither [l7policies.ListOpts](https://github.com/gophercloud/gophercloud/blob/1c295d7a5f4ba0cc0e548934823b8f3e05ce5af3/openstack/loadbalancer/v2/l7policies/requests.go#L121-L136) nor [l7policies.ListRulesOpts](https://github.com/gophercloud/gophercloud/blob/1c295d7a5f4ba0cc0e548934823b8f3e05ce5af3/openstack/loadbalancer/v2/l7policies/requests.go#L328-L341) exposes the corresponding list filters,

Would this scope be acceptable? if it is, I am happy to implement it and test it against an amphora provider OpenStack environment.

jihyun-huh:
Fixes #3913 

Add the Octavia 2.5 advanced tag filters (`tags`, `tags-any`, `not-tags`, and `not-tags-any`) to L7policy and L7rule list options.

Includes unit tests for query generation and an acceptance test covering all filters.

code in this PR:

API reference: https://docs.openstack.org/api-ref/load-balancer/v2/#filtering-by-tags

### DIFF
--- a/internal/acceptance/openstack/loadbalancer/v2/loadbalancer.go
+++ b/internal/acceptance/openstack/loadbalancer/v2/loadbalancer.go
@@ -479,7 +479,7 @@ func CreateL7Policy(t *testing.T, client *gophercloud.ServiceClient, listener *l
 	th.AssertEquals(t, policy.ListenerID, listener.ID)
 	th.AssertEquals(t, policy.Action, string(l7policies.ActionRedirectToURL))
 	th.AssertEquals(t, "http://www.example.com", policy.RedirectURL)
-	th.AssertDeepEquals(t, policy.Tags, tags)
+	th.AssertTagsEqual(t, tags, policy.Tags)
 
 	return policy, nil
 }
@@ -509,7 +509,7 @@ func CreateL7Rule(t *testing.T, client *gophercloud.ServiceClient, policyID stri
 	th.AssertEquals(t, rule.RuleType, string(l7policies.TypePath))
 	th.AssertEquals(t, rule.CompareType, string(l7policies.CompareTypeStartWith))
 	th.AssertEquals(t, "/api", rule.Value)
-	th.AssertDeepEquals(t, rule.Tags, tags)
+	th.AssertTagsEqual(t, tags, rule.Tags)
 
 	return rule, nil
 }
--- a/openstack/loadbalancer/v2/l7policies/requests.go
+++ b/openstack/loadbalancer/v2/l7policies/requests.go
@@ -133,6 +133,22 @@ type ListOpts struct {
 	Marker         string `q:"marker"`
 	SortKey        string `q:"sort_key"`
 	SortDir        string `q:"sort_dir"`
+
+	// Tags filters L7 policies that contain all of the given tags.
+	// Requires Octavia API version 2.5 or later.
+	Tags []string `q:"tags"`
+
+	// TagsAny filters L7 policies that contain at least one of the given tags.
+	// Requires Octavia API version 2.5 or later.
+	TagsAny []string `q:"tags-any"`
+
+	// TagsNot excludes L7 policies that contain all of the given tags.
+	// Requires Octavia API version 2.5 or later.
+	TagsNot []string `q:"not-tags"`
+
+	// TagsNotAny excludes L7 policies that contain any of the given tags.
+	// Requires Octavia API version 2.5 or later.
+	TagsNotAny []string `q:"not-tags-any"`
 }
 
 // ToL7PolicyListQuery formats a ListOpts into a query string.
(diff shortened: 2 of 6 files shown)


openstack/loadbalancer/v2/l7policies/requests.go
```diff
@@ -133,6 +133,22 @@ type ListOpts struct {
...
+	// Requires Octavia API version 2.5 or later.
+	TagsAny []string `q:"tags-any"`
+
+	// TagsNot excludes L7 policies that contain all of the given tags.
+	// Requires Octavia API version 2.5 or later.
+	TagsNot []string `q:"not-tags"`
+
+	// TagsNotAny excludes L7 policies that contain any of the given tags.
+	// Requires Octavia API version 2.5 or later.
+	TagsNotAny []string `q:"not-tags-any"`
```
stephenfin, line 151: Have you tested this? I looked and saw we use `string` rather than `[]string` everywhere except for the loadbalancer service in [`openstack/loadbalancer/v2/loadbalancers/requests.go`](https://github.com/gophercloud/gophercloud/blob/a44fc6d46231cc847ebcc920deb638a316dcebd1/openstack/loadbalancer/v2/loadbalancers/requests.go#L42-L45). However, I then went and investigate how we handle `[]string` and arrived at [`BuildQueryString`](https://github.com/gophercloud/gophercloud/blob/a44fc6d46231cc847ebcc920deb638a316dcebd1/params.go#L402-L418). The [Octavia API docs](https://docs.openstack.org/api-ref/load-balancer/v2/#filtering-by-tags) explicitly state:

> When supplying a list of tags, the tags should be provided in a comma separated list.

That code (specifically [these lines](https://github.com/gophercloud/gophercloud/blob/a44fc6d46231cc847ebcc920deb638a316dcebd1/params.go#L414-L418)) indicates we need a `format=comma-separated` struct tag if we want this behavior. We haven't specified that, which means we'd call Octavia like so:

```http
GET /v2/lbaas/l7policies?tags=red&tagsblue
```

rather than

```http
GET /v2/lbaas/l7policies?tags=red,blue
```

If you have access to Octavia to test this, can you tell us: do tag filters need to be specified as CSV values or can then be passed as multiple key-value pairs? If the former, then the API reference is correct and you need to add the missing struct tag here and in the other, existing place. If the latter, then this code is fine but the API ref is wrong.

stephenfin: Answering my own question: the api-ref is wrong

https://review.opendev.org/c/openstack/octavia/+/1000092

stephenfin (dismissed):

jihyun-huh:
hi @stephenfin, thanks for looking into this PR, and for the Octavia API reference update above.
I have tested this on octavia environment(amphora provider) and both forms appear to work as expected.

please let me know if you think there are any other changes needed for this feature.

also, I found the functional test is currently failing because the L7policy test helper compares the returned tags in order, while Octavia can return the same tags in a different order. so im planning to update the test helper to fix the failed functional tests. would it be ok to you to include the fixing in this PR?

internal/acceptance/openstack/loadbalancer/v2/loadbalancer.go
```diff
@@ -445,6 +446,18 @@ func CreatePoolHTTP(t *testing.T, client *gophercloud.ServiceClient, lb *loadbal
...
+	t.Helper()
+
+	expected = append([]string(nil), expected...)
+	actual = append([]string(nil), actual...)
+
+	sort.Strings(expected)
+	sort.Strings(actual)
+
+	th.AssertDeepEquals(t, expected, actual)
+}
```
stephenfin: Could you move this to `testhelper/convenience.go`? I suspect we could use this elsewhere

stephenfin (changes requested):

stephenfin:
> hi @stephenfin, thanks for looking into this PR, and for the Octavia API reference update above. I have tested this on octavia environment(amphora provider) and both forms appear to work as expected.
> 
> please let me know if you think there are any other changes needed for this feature.

Thank you for testing this :pray: 

> also, I found the functional test is currently failing because the L7policy test helper compares the returned tags in order, while Octavia can return the same tags in a different order. so im planning to update the test helper to fix the failed functional tests. would it be ok to you to include the fixing in this PR?

No issues with this. I have left one request.

internal/acceptance/openstack/loadbalancer/v2/loadbalancer.go
jihyun-huh: sure! i moved the `assertTagsEqual` into `testhelper/convenience.go` and add some test code

stephenfin (approved):

PR #3918, opened 2026-08-07
Use slices for tag qs filters

stephenfin:
This was already done for the loadbalancer service. Do it for other services.

This is a breaking change that should not be backported to v2. While here, we also fix a TODO with another filter.

### DIFF
--- a/openstack/compute/v2/servers/requests.go
+++ b/openstack/compute/v2/servers/requests.go
@@ -77,19 +77,19 @@ type ListOpts struct {
 
 	// This requires the client to be set to microversion 2.26 or later.
 	// Tags filters on specific server tags. All tags must be present for the server.
-	Tags string `q:"tags"`
+	Tags []string `q:"tags" format:"comma-separated"`
 
 	// This requires the client to be set to microversion 2.26 or later.
 	// TagsAny filters on specific server tags. At least one of the tags must be present for the server.
-	TagsAny string `q:"tags-any"`
+	TagsAny []string `q:"tags-any" format:"comma-separated"`
 
 	// This requires the client to be set to microversion 2.26 or later.
 	// NotTags filters on specific server tags. All tags must be absent for the server.
-	NotTags string `q:"not-tags"`
+	NotTags []string `q:"not-tags" format:"comma-separated"`
 
 	// This requires the client to be set to microversion 2.26 or later.
 	// NotTagsAny filters on specific server tags. At least one of the tags must be absent for the server.
-	NotTagsAny string `q:"not-tags-any"`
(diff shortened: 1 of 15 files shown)


openstack/networking/v2/extensions/trunks/requests.go
```diff
@@ -64,22 +64,21 @@ type ListOptsBuilder interface {
...
+	Status         string   `q:"status"`
+	TenantID       string   `q:"tenant_id"`
+	ProjectID      string   `q:"project_id"`
+	SortDir        string   `q:"sort_dir"`
+	SortKey        string   `q:"sort_key"`
+	Tags           []string `q:"tags" format:"comma-separated"`
+	TagsAny        []string `q:"tags-any" format:"comma-separated"`
+	NotTags        []string `q:"not-tags" format:"comma-separated"`
+	NotTagsAny     []string `q:"not-tags-any" format:"comma-separated"`
+	RevisionNumber *int     `q:"revision_number"`
```
IlanZuckerman, line 81: IMHO, this is e second braking point (besides slices). I suggest to clearly state that in the PR description.
Meaning if someone **before** wrote: `trunks.ListOpts{RevisionNumber: "3"}`,
Now he has to write:
```
rev := 3
trunks.ListOpts{RevisionNumber: &rev}
```

IlanZuckerman (commented):
I would add a few UT's that would prove the query string is built correctly.
For example assert that list opts turn into `?tags=a,b` (not `?tags=a&tags=b`).

stephenfin:
The networking failures are unrelated. https://github.com/gophercloud/gophercloud/pull/3945 will address them.

openstack/networking/v2/extensions/trunks/requests.go
stephenfin: Noted this in the commit message :ok_hand:

winiciusallan:
Now we have https://github.com/gophercloud/gophercloud/pull/3945 merged, I believe we can re-run the jobs. @stephenfin Would you mind triggering them?

stephenfin:
> Now we have #3945 merged, I believe we can re-run the jobs. @stephenfin Would you mind triggering them?

I've rebased and forced pushed

openstack/compute/v2/servers/requests.go
winiciusallan, line 1: Can we also exercise this changes on servers in the acceptance tests?

winiciusallan (commented):
I found another set of fields that could be potentially changed to use a slice, see 

https://github.com/gophercloud/gophercloud/blob/main/openstack/orchestration/v1/stacks/requests.go#L257-L267

while we're here and if it makes sense, i believe it's worth changing.

PR #3921, opened 2026-08-07
feat: support for updating and deleting share access rule metadata

simonostendorf:
- Implement UpdateMetadata and DeleteMetadatum functions in shareaccessrules.
- Add corresponding test cases for metadata update and deletion.
- Enhance GrantAccess to include metadata options.
- Update related request and result structures to handle metadata.

### DIFF
--- a/internal/acceptance/openstack/sharedfilesystems/v2/shareaccessrules_test.go
+++ b/internal/acceptance/openstack/sharedfilesystems/v2/shareaccessrules_test.go
@@ -132,7 +132,7 @@ func TestShareAccessRulesMetadata(t *testing.T) {
 		t.Fatalf("Unable to grant access to share %s: %v", share.ID, err)
 	}
 
-	th.AssertDeepEquals(t, map[string]any{"key1": "value1"}, addedAccessRight.Metadata)
+	th.AssertDeepEquals(t, map[string]string{"key1": "value1"}, addedAccessRight.Metadata)
 
 	addedShareAccess := AccessRightToShareAccess(addedAccessRight)
 
@@ -162,7 +162,7 @@ func TestShareAccessRulesMetadata(t *testing.T) {
 
 	// key1 was set at creation time and must still be present; key2 was
 	// added via UpdateMetadata, which only touches the keys it is given.
-	th.AssertDeepEquals(t, map[string]any{"key1": "value1", "key2": "value2"}, accessRule.Metadata)
+	th.AssertDeepEquals(t, map[string]string{"key1": "value1", "key2": "value2"}, accessRule.Metadata)
 
 	err = shareaccessrules.DeleteMetadatum(context.TODO(), client, addedAccessRight.ID, "key1").ExtractErr()
 	if err != nil {
@@ -174,5 +174,5 @@ func TestShareAccessRulesMetadata(t *testing.T) {
 		t.Fatalf("Unable to get share access rule for share %s: %v", share.ID, err)
 	}
 
-	th.AssertDeepEquals(t, map[string]any{"key2": "value2"}, accessRule.Metadata)
+	th.AssertDeepEquals(t, map[string]string{"key2": "value2"}, accessRule.Metadata)
 }
--- a/internal/acceptance/openstack/sharedfilesystems/v2/shareaccessrules_test.go
+++ b/internal/acceptance/openstack/sharedfilesystems/v2/shareaccessrules_test.go
@@ -3,10 +3,13 @@
 package v2
 
 import (
+	"context"
 	"testing"
 
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/clients"
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/tools"
+	"github.com/gophercloud/gophercloud/v2/openstack/sharedfilesystems/v2/shareaccessrules"
+	"github.com/gophercloud/gophercloud/v2/openstack/sharedfilesystems/v2/shares"
 	th "github.com/gophercloud/gophercloud/v2/testhelper"
 )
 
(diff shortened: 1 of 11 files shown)


winiciusallan (commented):
Looks good overall. I'd like to see how tests goes.

simonostendorf:
> Looks good overall. I'd like to see how tests goes.

@winiciusallan thanks! Currently the tests are waiting for approval

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -281,6 +281,10 @@ type AccessRight struct {
 	State string `json:"state,omitempty"`
 	// The access rule ID.
 	ID string `json:"id"`
+	// Access rule metadata.
+	//
+	// Requires microversion 2.45 or later.
+	Metadata map[string]any `json:"metadata,omitempty"`
```
Sharpz7: Is it definitely map[string]any instead of map[string]string?

Manila defines access-rule metadata as a dictionary of strings, and the request fields and metadata extractors in this PR already use that type.

winiciusallan: yes, I believe you're actually right. good catch

when fetching a share, it joins the `share_metadata` table in the response[1]. taking a zoom into the ShareMetadata table, we can see that both key and value are strings[2]. 

[1] https://github.com/openstack/manila/blob/master/manila/db/sqlalchemy/api.py#L2655-L2659
[2] https://github.com/openstack/manila/blob/master/manila/db/sqlalchemy/models.py#L559-L560

simonostendorf: Nice catch, thanks! Ive changed it to use map[string]string instead.

simonostendorf:
One question: Why is the pr marked as major version, I thought ive only added a new feature so it should be minor, right?

winiciusallan:
> One question: Why is the pr marked as major version, I thought ive only added a new feature so it should be minor, right?

Per the job logs

```
 github.com/gophercloud/gophercloud/v2/openstack/sharedfilesystems/v2/shares
    Incompatible changes:
    - AccessRight: old is comparable, new is not
    - GrantAccessOpts: old is comparable, new is not
    Compatible changes:
    - AccessRight.Metadata: added
    - GrantAccessOpts.Metadata: added
```

Code relying on deeply comparing the struct might have breaking changes because of the new fields.

internal/acceptance/openstack/sharedfilesystems/v2/shareaccessrules_test.go
```diff
@@ -100,3 +103,76 @@ func TestShareAccessRulesList(t *testing.T) {
...
+		AccessLevel: "ro",
+		Metadata: map[string]string{
+			"key1": "value1",
+		},
+	}).Extract()
+	if err != nil {
+		t.Fatalf("Unable to grant access to share %s: %v", share.ID, err)
+	}
+
+	th.AssertDeepEquals(t, map[string]any{"key1": "value1"}, addedAccessRight.Metadata)
```
mandre: You'll need to update the acceptance tests after your latest change.
```suggestion
	th.AssertDeepEquals(t, map[string]string{"key1": "value1"}, addedAccessRight.Metadata)
```

internal/acceptance/openstack/sharedfilesystems/v2/shareaccessrules_test.go
```diff
@@ -100,3 +103,76 @@ func TestShareAccessRulesList(t *testing.T) {
...
+
+	addedShareAccess := AccessRightToShareAccess(addedAccessRight)
```
mandre, line 137: You should also update the `AccessRightToShareAccess()` helper to account for the new `Metadata` field.

https://github.com/gophercloud/gophercloud/blob/2839b4f4fb1d0317aa275d440a3c08883d7894e7/internal/acceptance/openstack/sharedfilesystems/v2/shareaccessrules.go#L27-L37

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -232,6 +232,10 @@ type GrantAccessOpts struct {
 	AccessTo string `json:"access_to"`
 	// The access level to the share is either "rw" or "ro".
 	AccessLevel string `json:"access_level"`
+	// One or more access rule metadata key and value pairs as a dictionary of strings.
+	//
```
mandre: nit: it's a bit uncommon in gophercloud to have empty lines like these, remove them.

openstack/sharedfilesystems/v2/shares/results.go
```diff
@@ -281,6 +281,10 @@ type AccessRight struct {
 	State string `json:"state,omitempty"`
 	// The access rule ID.
 	ID string `json:"id"`
+	// Access rule metadata.
+	//
```
mandre: ditto

openstack/sharedfilesystems/v2/shares/requests.go
```diff
@@ -232,6 +232,10 @@ type GrantAccessOpts struct {
...
+	// Requires microversion 2.45 or later.
+	Metadata map[string]string `json:"metadata,omitempty"`
```
mandre, line 237: Since the change is already flagged as major, would it be good to change the `shareaccessrules.ShareAccess.Metadata` field to `map[string]string` as well for consistency?

https://github.com/gophercloud/gophercloud/blob/2839b4f4fb1d0317aa275d440a3c08883d7894e7/openstack/sharedfilesystems/v2/shareaccessrules/results.go#L31

If we believe it might accept more types in the future, we can leave it as `map[string]any` but in that case we would need to change it back to the same type here.

winiciusallan (approved):

PR #3930, opened 2026-08-10
[v2] CI: improve test assertions

stephenfin:
This is a partial backport of #3777, with the last commit skipped due to the large number of merge conflicts. 2 of the 3 remaining commits also have merge conflicts but these are relatively trivial and documented in the commit message.

We want this change mainly due to the addition of the new test assertion helpers, which are relying on in `main`.

- **Fix assertion argument order: expected first, actual second**
- **Fix incorrect field assertions in acceptance tests**
- **testhelper: add new test assertion helpers**

### DIFF
--- a/docs/contributor-tutorial/.template/testing/requests_test.go
+++ b/docs/contributor-tutorial/.template/testing/requests_test.go
@@ -27,7 +27,7 @@ func TestListResources(t *testing.T) {
 		return true, nil
 	})
 	th.AssertNoErr(t, err)
-	th.AssertEquals(t, count, 1)
+	th.AssertEquals(t, 1, count)
 }
 
 func TestListResourcesAllPages(t *testing.T) {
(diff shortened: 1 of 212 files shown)


stephenfin:
I think needs a backport of https://github.com/gophercloud/gophercloud/pull/3945

winiciusallan:
I think we need to backport https://github.com/gophercloud/gophercloud/pull/3765 to V2 so we can run Magnum jobs. Created a manual backport for this https://github.com/gophercloud/gophercloud/pull/3954

`functional-baremetal` failure looks unrelated :thinking:. I asked on #openstack-ironic if the missing image should be available in the tarball repository, and it looks like tinyipa ram disk type isn't supported anymore. we may need to change to something else like `dib`. Did this on https://github.com/gophercloud/gophercloud/pull/3955

mandre (approved):

PR #3932, opened 2026-08-10
feat: add loadbalancer provider capabilities

### LINKED ISSUE #3931: loadbalancer: add support for provider flavor and availability zone capabilities
### What is missing?

Hi maintainers,

The [openstack/loadbalancer/v2/providers](https://github.com/gophercloud/gophercloud/tree/main/openstack/loadbalancer/v2/providers) package is currently supporting listing enabled provider drivers below,

```bash
GET /v2/lbaas/providers
```

and current Octavia exposes two additional read only provider APIs that are not currently available in Gophercloud yet

```bash
GET /v2/lbaas/providers/{provider}/flavor_capabilities
GET /v2/lbaas/providers/{provider}/availability_zone_capabilities
```

The flavor capabilities API was introduced in octavia API version 2.6: [Octavia API reference](https://docs.openstack.org/api-ref/load-balancer/v2/index.html#show-provider-flavor-capabilities)

and the availability zone capabilities API was also introduced in version 2.14: [Octavia API reference](https://docs.openstack.org/api-ref/load-balancer/v2/index.html#show-provider-availability-zone-capabilities)

I encountered this while adding Octavia provider validation(using amphora provider) and diagnostics to an openstack based kubernetes Gateway API controller. so without SDK support, callers currently need to send HTTP requests on these URLs manually.

If this feature proposal approach looks reasonable, i would be happy to work on this implementation also with acceptance tests.

jihyun-huh:
Fixes #3931 

This PR is to support for listing flavor and availability zone capabilities in a provider specific way.
- add `ListFlavorCapabilities`
- add `ListAvailabilityZoneCapabilities`
- add unit and acceptance tests
- update `doc.go`

code in this PR:

the requested codes are references here:

- https://docs.openstack.org/api-ref/load-balancer/v2/index.html#show-provider-flavor-capabilities
- https://docs.openstack.org/api-ref/load-balancer/v2/index.html#show-provider-availability-zone-capabilities
- https://opendev.org/openstack/octavia/src/branch/master/octavia/api/v2/types/provider.py
- https://opendev.org/openstack/octavia/src/branch/master/octavia/api/v2/controllers/provider.py

### DIFF
--- a/internal/acceptance/openstack/loadbalancer/v2/providers_test.go
+++ b/internal/acceptance/openstack/loadbalancer/v2/providers_test.go
@@ -4,6 +4,7 @@ package v2
 
 import (
 	"context"
+	"strings"
 	"testing"
 
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/clients"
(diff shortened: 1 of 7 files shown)


internal/acceptance/openstack/loadbalancer/v2/providers_test.go
```diff
@@ -31,3 +31,47 @@ func TestProvidersList(t *testing.T) {
...
+			provider = candidate.Name
+			break
+		}
+	}
+
+	flavorCapabilities, err := providers.ListFlavorCapabilities(context.TODO(), client, provider, nil).Extract()
+	if err != nil {
+		t.Fatalf("Unable to list flavor capabilities for provider %q: %v", provider, err)
+	}
+	for _, capability := range flavorCapabilities {
```
winiciusallan, line 69: Is there default values that we can check that are being returned here? So we can test the returning value from the API call. Same for the `ListAvailabilityZoneCapabilities`.

winiciusallan (commented):
Looks very good to me, thanks for adding these calls @jihyun-huh! I left one question about something that I personally would like to see in the acceptance tests.

internal/acceptance/openstack/loadbalancer/v2/providers_test.go
jihyun-huh: @winiciusallan thank you very much for your review! this makes the provider test code much more robust.

The name and description in `expectedFlavorCapability` are taken from the default values of [Octavia Tempest Plugin](https://opendev.org/openstack/octavia-tempest-plugin/src/branch/master/octavia_tempest_plugin/config.py#L143-L146), please let me know if you would prefer a different approach.

internal/acceptance/openstack/loadbalancer/v2/providers_test.go
```diff
@@ -31,3 +32,67 @@ func TestProvidersList(t *testing.T) {
...
+	}
+	expectedFlavorCapability := providers.Capability{
+		Name: "loadbalancer_topology",
+		Description: "The load balancer topology. One of: SINGLE - One amphora per load balancer. " +
+			"ACTIVE_STANDBY - Two amphora per load balancer.",
+	}
+	found := false
+	for _, capability := range flavorCapabilities {
+		tools.PrintResource(t, capability)
+		if capability == expectedFlavorCapability {
```
IlanZuckerman: Optional suggestion (not required): it might be less fragile to assert only part of the description, e.g. with `strings.Contains`, instead of comparing the whole Capability object. If Octavia (or the tempest-plugin defaults the expected text was taken from) changes the description wording, the test could fail even though the API and SDK still work correctly.

internal/acceptance/openstack/loadbalancer/v2/providers_test.go
```diff
@@ -31,3 +32,67 @@ func TestProvidersList(t *testing.T) {
...
+			found = true
+			break
+		}
+	}
+	th.AssertTrue(t, found)
```
IlanZuckerman: If found is false, failure message won’t say which capability was missing.
Suggestion:
```
th.AssertTrue(t, found, "expected flavor capability %v in %v", expectedFlavorCapability, flavorCapabilities)
```

internal/acceptance/openstack/loadbalancer/v2/providers_test.go
```diff
@@ -31,3 +32,67 @@ func TestProvidersList(t *testing.T) {
...
+	}
+	found = false
+	for _, capability := range availabilityZoneCapabilities {
+		tools.PrintResource(t, capability)
+		if capability == expectedAvailabilityZoneCapability {
+			found = true
+			break
+		}
+	}
+	th.AssertTrue(t, found)
```
IlanZuckerman: Same comment as in L79

IlanZuckerman (approved):
LGTM, except for few minor comments (for your consideration).

internal/acceptance/openstack/loadbalancer/v2/providers_test.go
jihyun-huh: hi @IlanZuckerman thanks for your review! this is very reasonable suggestion, and i understand your concern. 
I have updated the description comparison logic, so making the acceptance logic less fragile.

internal/acceptance/openstack/loadbalancer/v2/providers_test.go
jihyun-huh: thanks again! i have also improved the failure messages following your suggestion.

Sharpz7 (approved):

PR #3939, opened 2026-08-12
core: honor context while waiting for reauthentication

### LINKED ISSUE #3922: core: reauthentication waits ignore context cancellation
### What happened?

When reauthentication is already in progress, another request waits in `reauthFuture.Get()` without observing its context. A canceled request can therefore remain blocked until the active reauthentication completes.

After successful reauthentication, a failed retry is returned as `ErrErrorAfterReauthentication`. The error does not unwrap `ErrOriginal`, so `ResponseCodeIs` cannot detect status codes such as 404.

Expected behavior: a waiting request returns `ctx.Err()` when canceled without interrupting the active reauthentication, and the retried request error remains available through standard error unwrapping.

### Version

v2 (current)

rm-you:
Fixes #3922

A request waiting for another goroutine's reauthentication ignores its context and can remain blocked past its deadline. This change lets the waiter return `ctx.Err()` without interrupting the active reauthentication.

`ErrErrorAfterReauthentication` now unwraps the retried request error, allowing `ResponseCodeIs` to detect status codes such as 404.

N/A. This changes client-side synchronization and error handling.

### DIFF
--- a/provider_client.go
+++ b/provider_client.go
@@ -135,16 +135,25 @@ func (f *reauthFuture) Set(err error) {
 	close(f.done)
 }
 
-func (f *reauthFuture) Get() error {
-	<-f.done
-	return f.err
+func (f *reauthFuture) Get(ctx context.Context) error {
+	select {
+	case <-f.done:
+		return f.err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
 }
 
 // AuthenticatedHeaders returns a map of HTTP headers that are common for all
 // authenticated service requests. Blocks if Reauthenticate is in progress.
-func (client *ProviderClient) AuthenticatedHeaders() (m map[string]string) {
+func (client *ProviderClient) AuthenticatedHeaders() map[string]string {
+	headers, _ := client.authenticatedHeaders(context.Background())
+	return headers
+}
+
+func (client *ProviderClient) authenticatedHeaders(ctx context.Context) (map[string]string, error) {
 	if client.IsThrowaway() {
-		return
+		return nil, nil
 	}
 	if client.reauthmut != nil {
 		// If a Reauthenticate is in progress, wait for it to complete.
@@ -152,14 +161,16 @@ func (client *ProviderClient) AuthenticatedHeaders() (m map[string]string) {
 		ongoing := client.reauthmut.ongoing
 		client.reauthmut.Unlock()
 		if ongoing != nil {
-			_ = ongoing.Get()
+			if err := ongoing.Get(ctx); err != nil && ctx.Err() != nil {
+				return nil, ctx.Err()
+			}
 		}
 	}
 	t := client.Token()
 	if t == "" {
-		return
+		return nil, nil
 	}
-	return map[string]string{"X-Auth-Token": t}
+	return map[string]string{"X-Auth-Token": t}, nil
 }
 
 // UseTokenLock creates a mutex that is used to allow safe concurrent access to the auth token.
@@ -287,7 +298,7 @@ func (client *ProviderClient) Reauthenticate(ctx context.Context, previousToken 
 
 	// If Reauthenticate is running elsewhere, wait for its result.
 	if ongoing != nil {
-		return ongoing.Get()
+		return ongoing.Get(ctx)
 	}
 
 	// Perform the actual reauthentication.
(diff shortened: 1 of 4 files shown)


provider_client.go
```diff
@@ -287,12 +298,12 @@ func (client *ProviderClient) Reauthenticate(ctx context.Context, previousToken
...
 	// If Reauthenticate is running elsewhere, wait for its result.
 	if ongoing != nil {
-		return ongoing.Get()
+		return ongoing.Get(ctx)
 	}
 
 	// Perform the actual reauthentication.
 	var err error
-	if previousToken == "" || client.TokenID == previousToken {
+	if previousToken == "" || client.Token() == previousToken {
```
stephenfin: What's the rationale for this change? I assume it's unrelated to the main change here? If so, could you split it out to a separate change. If it's not, could you call out the reason for it in the commit message?

provider_client.go
```diff
@@ -135,31 +135,42 @@ func (f *reauthFuture) Set(err error) {
 	close(f.done)
 }
 
-func (f *reauthFuture) Get() error {
-	<-f.done
-	return f.err
+func (f *reauthFuture) Get(ctx context.Context) error {
```
stephenfin, line 138: ~jfyi this is an API change. While I don't know how many people are calling this directly, we likely can't backport this to `v2`. If you want to backport this, you might be better of introducing a new `GetWithContext` helper for v2 only, which can be immediately dropped on `main` via a subsequent/stacked PR.~

Actually, this is probably okay since the struct itself is private. Leaving this comment here in case the linter says otherwise.

provider_client.go
```diff
@@ -135,31 +135,42 @@ func (f *reauthFuture) Set(err error) {
...
+		return f.err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
 }
 
 // AuthenticatedHeaders returns a map of HTTP headers that are common for all
 // authenticated service requests. Blocks if Reauthenticate is in progress.
-func (client *ProviderClient) AuthenticatedHeaders() (m map[string]string) {
+func (client *ProviderClient) AuthenticatedHeaders() map[string]string {
```
stephenfin, line 149: Do we want to accept a ctx argument here at some point? I assume no, but if we do can we add a TODO to fix this in a later PR for `main` only?

stephenfin (changes requested):
This looks pretty good. Couple of comments/questions inline before I sign off though

provider_client.go
rm-you: Reverted. This was unrelated lock-safe cleanup, so keeping the PR focused makes sense.

provider_client.go
rm-you: I don't think we should change this public method. Accepting a context would break existing callers. `doRequest` now uses the private context-aware helper, while direct `AuthenticatedHeaders` callers retain the existing blocking behavior.

provider_client.go
rm-you: Isn't v2 == main? Am I confused? Is main for v3 now?

winiciusallan: We introduce new changes to the `main` branch, and if it's not a breaking change, we backport to the `v2` branch (usually with our backport-bot), then cut in a form of minor or patch release.

In case the PR introduces an API breaking change, which is not the case since the struct is private, we can't backport and it would be made only on v3.

stephenfin: Yeah, what @winiciusallan said

provider_client.go
stephenfin: Sorry, my suggestion was to change the behavior on `main` via a separate PR that we do not backport. `main` is what will become the future `v3`, as @winiciusallan noted [here](https://github.com/gophercloud/gophercloud/pull/3939#discussion_r3779018839).

In any case, this would have to be a follow-up so it's not something to block on here. If you are happy to do this in a `main`-only follow-up, please feel free to assign/ping me for reviews.

stephenfin (approved):

provider_client.go
rm-you: Went ahead and got it staged: https://github.com/rm-you/gophercloud/commit/2898633d8ef2ee857e8e8b8eb8919553b0332514

It makes AuthenticatedHeaders accept a context and return an error, with cancellation coverage. It is stacked on this PR for now, will make another PR for that once this merges (though it's just an AI first draft still, need to give it a bit more review and real-life testing)

rm-you:
So now we need to backport to v2?

PR #3616, opened 2026-01-27
feat: add support for the websso authentication method

cardoe:
The websso authentication method involves getting a token via the Skyline or Horizon websso endpoint for SSO authentication. Fixes #3615.

PR #3940, opened 2026-08-12
identity: add OAuth2 mTLS authentication

rm-you:
Fixes #3915

Adds Identity v3 authentication using OAuth 2.0 mutual TLS client credentials. It obtains an access token from `OS-OAUTH2/token`, validates it with Keystone to populate the service catalog, and sends it as `Authorization: Bearer` on service requests. The same flow is used for reauthentication.

`ProviderClient` now supports authentication headers other than `X-Auth-Token`; existing authentication methods are unchanged.

Includes unit and acceptance tests.

Links to the line numbers/files in the OpenStack source code that support the code in this PR:

- [keystoneauth OAuth2 mTLS client-credential implementation](https://opendev.org/openstack/keystoneauth/src/commit/7472cfd32d58f673aaaedfaf1f60df219fda46fe/keystoneauth1/identity/v3/oauth2_mtls_client_credential.py#L92-L155)
- [keystonemiddleware certificate-bound Bearer-token validation](https://opendev.org/openstack/keystonemiddleware/src/commit/9401c513219f86008d1df380a10d57464bb20b2d/keystonemiddleware/oauth2_mtls_token.py#L70-L110)

PR #1942, opened 2020-04-18
Core: set response headers for all requests

kayrus:
Resolves #1170

PR #3941, opened 2026-08-12
identity: add WebSSO authentication

### LINKED ISSUE #3615: Support the websso keystone authentication method
### What is missing?

The websso keystone authentication method is a way to get a token via the websso method in Skyline/Horizon to allow the user to auth to their SSO via a web browser. See a Python implementation at https://github.com/IFCA-Advanced-Computing/keystoneauth-oidc as well as an effort to upstream it at https://review.opendev.org/c/openstack/keystoneauth/+/970328 It's possible to install this plugin via `pip` today. There's no way to currently create an out of tree plugin for gophercloud which would support this functionality. Related to #3283.

rm-you:
Fixes #3615

Adds Identity v3 authentication through Keystone's browser-based WebSSO flow. Gophercloud opens the identity-provider and protocol-specific WebSSO endpoint with a loopback origin, receives the unscoped token posted by Keystone, and validates or scopes it with the existing token API.

The callback listener is context-aware and validates the request method, content type, body size, and token. Browser opening is injectable. Reauthentication and optional caching of validated unscoped tokens are supported, allowing tokens to be rescoped without repeating browser authentication.

- [Horizon WebSSO redirect flow](https://opendev.org/openstack/horizon/src/commit/3c5006efb4533e1cf0815fb80ede86b7a38a40bf/openstack_auth/views.py#L126-L151)
- [Horizon WebSSO token callback](https://opendev.org/openstack/horizon/src/commit/3c5006efb4533e1cf0815fb80ede86b7a38a40bf/openstack_auth/views.py#L248-L273)

Supersedes #3616, which its author closed in favor of this implementation.

### DIFF
--- a/openstack/identity/v3/tokencache/cache.go
+++ b/openstack/identity/v3/tokencache/cache.go
@@ -6,6 +6,8 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
+	"net/http"
 	"strings"
 	"time"
 
@@ -64,26 +66,27 @@ func (ct *cachedToken) valid(endpoint string) bool {
 	return time.Now().Add(tokenExpiryMargin).Before(ct.ExpiresAt)
 }
 
-// Load retrieves and validates a cached token against Keystone.
-func Load(ctx context.Context, client *gophercloud.ProviderClient, cache Cache, key, identityEndpoint string) (tokens.CreateResult, bool) {
+// Load retrieves and validates a cached token against Keystone. Invalid tokens
+// are cache misses; other validation errors are returned.
+func Load(ctx context.Context, client *gophercloud.ProviderClient, cache Cache, key, identityEndpoint string) (tokens.CreateResult, bool, error) {
 	var zero tokens.CreateResult
 	if client == nil || cache == nil {
-		return zero, false
+		return zero, false, nil
 	}
 
 	data, err := cache.Get(key)
 	if err != nil || data == "" {
-		return zero, false
+		return zero, false, nil
 	}
 
 	var cached cachedToken
 	if err := json.Unmarshal([]byte(data), &cached); err != nil {
 		_ = cache.Delete(key)
-		return zero, false
+		return zero, false, nil
 	}
 	if !cached.valid(identityEndpoint) {
 		_ = cache.Delete(key)
-		return zero, false
+		return zero, false, nil
 	}
 
 	endpoint := gophercloud.NormalizeURL(identityEndpoint)
@@ -91,29 +94,44 @@ func Load(ctx context.Context, client *gophercloud.ProviderClient, cache Cache, 
 		endpoint = strings.TrimRight(endpoint, "/") + "/v3/"
 	}
 	identityClient := &gophercloud.ServiceClient{
-		ProviderClient: &gophercloud.ProviderClient{
-			HTTPClient:        client.HTTPClient,
-			UserAgent:         client.UserAgent,
-			RetryBackoffFunc:  client.RetryBackoffFunc,
-			MaxBackoffRetries: client.MaxBackoffRetries,
-			RetryFunc:         client.RetryFunc,
-		},
-		Endpoint: endpoint,
-		Type:     "identity",
+		ProviderClient: client,
+		Endpoint:       endpoint,
+		Type:           "identity",
 	}
-
-	identityClient.SetToken(cached.TokenID)
-	result := tokens.Get(ctx, identityClient, cached.TokenID, nil)
+	result := Validate(ctx, identityClient, cached.TokenID)
 	if result.Err != nil {
-		_ = cache.Delete(key)
-		return zero, false
+		invalid := gophercloud.ResponseCodeIs(result.Err, http.StatusUnauthorized) ||
+			gophercloud.ResponseCodeIs(result.Err, http.StatusNotFound)
+		if invalid {
+			_ = cache.Delete(key)
+			return zero, false, nil
+		}
+		return zero, false, result.Err
+	}
+	return result, true, nil
+}
+
+// Validate retrieves token details from Keystone without changing client.
+func Validate(ctx context.Context, client *gophercloud.ServiceClient, tokenID string) (r tokens.CreateResult) {
+	if client == nil || client.ProviderClient == nil {
+		r.Err = fmt.Errorf("service client or provider client is nil")
+		return
 	}
 
-	var created tokens.CreateResult
-	created.Body = result.Body
-	created.Header = result.Header
-	created.Err = result.Err
-	return created, true
+	validationClient := *client
+	validationClient.ProviderClient = &gophercloud.ProviderClient{
+		TokenID:           tokenID,
+		HTTPClient:        client.HTTPClient,
+		UserAgent:         client.UserAgent,
+		RetryBackoffFunc:  client.RetryBackoffFunc,
+		MaxBackoffRetries: client.MaxBackoffRetries,
+		RetryFunc:         client.RetryFunc,
+	}
+	result := tokens.Get(ctx, &validationClient, tokenID, nil)
+	r.Body = result.Body
+	r.Header = result.Header
+	r.Err = result.Err
+	return
 }
 
 // Persist stores a successful token result. Cache errors are ignored.
--- a/openstack/identity/v3/tokencache/cache.go
+++ b/openstack/identity/v3/tokencache/cache.go
@@ -11,7 +11,8 @@ import (
 
 const tokenExpiryMargin = 5 * time.Minute
 
-// Cache stores authentication results. Implementations must be safe for concurrent use.
+// Cache stores authentication results. Values contain bearer tokens and must be
+// protected accordingly. Implementations must be safe for concurrent use.
 type Cache interface {
 	// Get returns ("", nil) for a cache miss.
 	Get(key string) (string, error)
--- a/openstack/identity/v3/tokencache/cache.go
+++ b/openstack/identity/v3/tokencache/cache.go
@@ -2,17 +2,11 @@
 package tokencache
 
 import (
-	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
-	"fmt"
-	"net/http"
 	"strings"
 	"time"
-
-	"github.com/gophercloud/gophercloud/v2"
-	"github.com/gophercloud/gophercloud/v2/openstack/identity/v3/tokens"
 )
 
 const tokenExpiryMargin = 5 * time.Minute
@@ -39,10 +33,6 @@ type KeyOptions struct {
 	IdentityProvider string
 	// Protocol is the Keystone federation protocol.
 	Protocol string
-	// AuthenticationEndpoint is the external identity endpoint.
-	AuthenticationEndpoint string
-	// Scope is the requested Keystone scope.
-	Scope tokens.Scope
 }
 
 // Key returns a deterministic cache identifier.
@@ -66,91 +56,39 @@ func (ct *cachedToken) valid(endpoint string) bool {
 	return time.Now().Add(tokenExpiryMargin).Before(ct.ExpiresAt)
 }
 
-// Load retrieves and validates a cached token against Keystone. Invalid tokens
-// are cache misses; other validation errors are returned.
-func Load(ctx context.Context, client *gophercloud.ProviderClient, cache Cache, key, identityEndpoint string) (tokens.CreateResult, bool, error) {
-	var zero tokens.CreateResult
-	if client == nil || cache == nil {
-		return zero, false, nil
+// Load retrieves a cached token ID. Missing, malformed, and expired entries are
+// cache misses.
+func Load(cache Cache, key, identityEndpoint string) (string, bool) {
+	if cache == nil {
+		return "", false
 	}
 
 	data, err := cache.Get(key)
 	if err != nil || data == "" {
-		return zero, false, nil
+		return "", false
 	}
 
 	var cached cachedToken
 	if err := json.Unmarshal([]byte(data), &cached); err != nil {
 		_ = cache.Delete(key)
-		return zero, false, nil
+		return "", false
 	}
-	if !cached.valid(identityEndpoint) {
+	if cached.TokenID == "" || !cached.valid(identityEndpoint) {
 		_ = cache.Delete(key)
-		return zero, false, nil
+		return "", false
 	}
-
-	endpoint := gophercloud.NormalizeURL(identityEndpoint)
-	if !strings.HasSuffix(strings.TrimRight(endpoint, "/"), "/v3") {
-		endpoint = strings.TrimRight(endpoint, "/") + "/v3/"
-	}
-	identityClient := &gophercloud.ServiceClient{
-		ProviderClient: client,
-		Endpoint:       endpoint,
-		Type:           "identity",
-	}
-	result := Validate(ctx, identityClient, cached.TokenID)
-	if result.Err != nil {
-		invalid := gophercloud.ResponseCodeIs(result.Err, http.StatusUnauthorized) ||
-			gophercloud.ResponseCodeIs(result.Err, http.StatusNotFound)
-		if invalid {
-			_ = cache.Delete(key)
-			return zero, false, nil
-		}
-		return zero, false, result.Err
-	}
-	return result, true, nil
+	return cached.TokenID, true
 }
 
-// Validate retrieves token details from Keystone without changing client.
-func Validate(ctx context.Context, client *gophercloud.ServiceClient, tokenID string) (r tokens.CreateResult) {
-	if client == nil || client.ProviderClient == nil {
-		r.Err = fmt.Errorf("service client or provider client is nil")
-		return
-	}
-
-	validationClient := *client
-	validationClient.ProviderClient = &gophercloud.ProviderClient{
-		TokenID:           tokenID,
-		HTTPClient:        client.HTTPClient,
-		UserAgent:         client.UserAgent,
-		RetryBackoffFunc:  client.RetryBackoffFunc,
-		MaxBackoffRetries: client.MaxBackoffRetries,
-		RetryFunc:         client.RetryFunc,
-	}
-	result := tokens.Get(ctx, &validationClient, tokenID, nil)
-	r.Body = result.Body
-	r.Header = result.Header
-	r.Err = result.Err
-	return
-}
-
-// Persist stores a successful token result. Cache errors are ignored.
-func Persist(cache Cache, key, identityEndpoint string, result tokens.CreateResult) {
-	if cache == nil {
-		return
-	}
-	tokenID, err := result.ExtractTokenID()
-	if err != nil || tokenID == "" {
-		return
-	}
-	token, err := result.ExtractToken()
-	if err != nil || token == nil {
+// Persist stores a token ID and expiration. Cache errors are ignored.
+func Persist(cache Cache, key, identityEndpoint, tokenID string, expiresAt time.Time) {
+	if cache == nil || tokenID == "" {
 		return
 	}
 
 	data, err := json.Marshal(cachedToken{
 		TokenID:   tokenID,
-		ExpiresAt: token.ExpiresAt,
+		ExpiresAt: expiresAt,
 		Endpoint:  identityEndpoint,
 	})
 	if err != nil {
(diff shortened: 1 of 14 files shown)


cardoe:
Thanks for picking this up — I closed #3616 in favor of this and it's in much better shape than what I had. Three things from reading through it, one of which is a workflow preference rather than a defect.

**1. The unscoped path will reject a `203` from Keystone.** The validation GET is hand-rolled, so it gets `defaultOkCodes("GET")` = `{200}`:

https://github.com/gophercloud/gophercloud/blob/ace4e9a6693b1cf2c2b64ffcd7413bf701e2f952/openstack/identity/v3/websso/requests.go#L133-L139

`tokens.Get` has carried `OkCodes: []int{200, 203}` for token introspection since `46d913f1d` and kept it through every later refactor of that call (#1570, #1942). On a deployment that answers `203`, `ParseResponse` sets `r.Err`, `SetTokenAndAuthResult` bails, and the whole login fails with `unexpected response code: 203`. The TokenID-passthrough branch added in #1752 already handles the same "validate a token I already hold" case the right way:

https://github.com/gophercloud/gophercloud/blob/ace4e9a6693b1cf2c2b64ffcd7413bf701e2f952/openstack/client.go#L205-L222

So I think this branch just wants `tokens.Get(ctx, client, unscopedToken, nil)` — which is also what `tokencache.Load` in this same PR does.

**2. The scope is only validated after the browser flow completes.** `captureToken` runs first, and `ToTokenV3ScopeMap()` afterward:

https://github.com/gophercloud/gophercloud/blob/ace4e9a6693b1cf2c2b64ffcd7413bf701e2f952/openstack/identity/v3/websso/requests.go#L118-L128

A `Scope{ProjectName: "x"}` with no domain makes the user sit through the full IdP login and then fails with `ProjectName requires DomainID or DomainName`, throwing away a perfectly good token. Building the scope map before `captureToken` (or checking it in `validate()`) makes that a fast failure.

**3. Caching the scoped token instead of the unscoped one costs a browser round trip per project.** `Scope` is part of the cache key:

https://github.com/gophercloud/gophercloud/blob/ace4e9a6693b1cf2c2b64ffcd7413bf701e2f952/openstack/identity/v3/websso/requests.go#L74-L84

This is a preference rather than a bug. Day to day I move across a lot of projects, and the flow I rely on today caches the *unscoped* token and rescopes locally per project — one browser login in the morning and none after it. Keying the cache on scope means a new browser round trip on every project switch. Would you consider caching the unscoped token and doing the rescope against that cache instead? It also makes item 2 cheaper, since an invalid scope would never cost a login.

cardoe:
Here's a peer change to keystoneauth: https://review.opendev.org/c/openstack/keystoneauth/+/1000759 It's still very much a WIP as I've just squashed together a number of commits in another repo and moved them over to keystoneauth and moved the tests over. Claude did that lifting so it needs to be tested.

rm-you:
Thanks—these are addressed in `0302eca7`. Token validation now accepts Keystone 203 responses, scope validation happens before opening the browser, and the cache stores the unscoped token for reuse across project scopes. Regression tests cover all three cases.

cardoe:
Ran this against a live Keystone deployment with an OIDC federated identity provider, using a small file-backed `tokencache.Cache`. Items 1-3 from my earlier comment all behave correctly outside the unit tests:

- **Unscoped-token reuse works as intended.** One browser login, then rescoping to a second project ran with no browser flow at all. That's exactly the workflow I was after in item 3 — confirmed end to end against a real deployment, not just `TestWebSSOCacheReusesUnscopedTokenAcrossProjects`.
- **The capture → validate → rescope path works against live Keystone.** `tokencache.Validate` going through `tokens.Get` returns the token with `X-Subject-Token` present, so the rescope receives a usable token ID.
- **The defaults match a real `trusted_dashboard` config.** Callback origin `http://localhost:9990/auth/websso/` worked with no server-side change on my side.

Two small notes on the new code, neither blocking:

1. `tokencache.Validate` doesn't touch the cache, and `Authenticate` calls it on the non-cached path, so callers who never enable caching still take a dependency on the `tokencache` package. It might sit better unexported, or next to `tokens` where the rest of the token operations live.

2. Both paths already know the unscoped token ID — `captureToken` returns it, and `cachedToken.TokenID` stores it — but then re-derive it from the validate response's `X-Subject-Token` header:

https://github.com/gophercloud/gophercloud/blob/784868f7297f9c58cbf9daecf4c0b5a5cc50a0ba/openstack/identity/v3/websso/requests.go#L145-L153

That works against the deployments I can test, but it's an avoidable dependency on Keystone echoing the header back; carrying the known ID forward would be more robust.

Separately, something I hit while writing the harness that isn't this PR's fault: `AuthenticateV3` with a non-empty `gophercloud.EndpointOpts` segfaults, because `v3auth` hands them to `NewIdentityV3`, which calls the still-nil `client.EndpointLocator`. `openstack.Authenticate` always passes an empty value so nothing in-tree hits it, but it will bite anyone wiring an alternate auth flow up by hand. I'll file that on its own.

Nothing blocking from my side — thanks for turning items 1-3 around so quickly.

rm-you:
Thanks for testing this. Addressed both notes in 90a9b9dd3:

- Token validation is now an unexported WebSSO helper; tokencache only reads and writes cached token metadata.
- Both browser and cache paths carry the known unscoped token ID through validation and use it directly for rescoping. Regression tests cover validation responses without X-Subject-Token on both paths.

cardoe:
Re-tested against a live Keystone deployment with an OIDC federated IdP. Both notes from my last comment are resolved, and the caching behavior holds up well in practice:

- One browser login, then rescoping to a second project — and to a project in a *different domain* — both served entirely from the cached unscoped token with no browser flow at all.
- `TestWebSSOKeepsKnownTokenIDAfterValidation` is the right test for note 2: mocking a validate response that omits `X-Subject-Token` and still asserting the correct token ID is exactly the dependency I was worried about.
- Moving `tokencache` off any gophercloud imports is a cleaner outcome than what I suggested. The package is now purely a string cache.

One thing I'd still like to see:

**Close the browser window on success.** The callback handler in `openstack/identity/v3/websso/requests.go` currently returns a page telling the user to close the window themselves ("Authentication successful ... You may close this window"). The Python WebSSO plugin I use adds a `<script>window.close()</script>` to the success page so the tab disappears on its own, which is a noticeably nicer CLI experience when you do this many times a day. It's best-effort — browsers refuse `window.close()` on tabs that script didn't open — but it costs nothing and works often enough to be worth it.

rm-you:
Added the best-effort window.close() call in aef252e7d, with the existing success message retained as the fallback when the browser refuses to close the tab.

cardoe:
This version looks great to me. The `window.close()` change is the last thing I was after, and everything I raised earlier is resolved.

Nothing outstanding from my side, and I don't want to hold this up. To be clear about my earlier comment: the acceptance-test coverage is still worth adding, but I don't think it needs to block landing — it can follow in a separate change.

Thanks for being so responsive through all of this.

cardoe (approved):

cardoe:
@stephenfin the Zun failure is unrelated so this can run again.

winiciusallan:
> @stephenfin the Zun failure is unrelated so this can run again.

An upstream [fix](https://review.opendev.org/c/openstack/zun/+/995022) has landed, so I'm rerunning the failing job. I'm also taking a look at the code right now.

openstack/identity/v3/tokencache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+func (ct *cachedToken) valid(endpoint string) bool {
+	if strings.TrimRight(ct.Endpoint, "/") != strings.TrimRight(endpoint, "/") {
+		return false
+	}
+	return time.Now().Add(tokenExpiryMargin).Before(ct.ExpiresAt)
+}
+
+// Load retrieves a cached token ID. Missing, malformed, and expired entries are
```
winiciusallan, line 60: should not malformed entries return an error or at least a warning? I think it's reasonable to consider missing and/or expired as a cache miss, but a malformed entries could be a result of something wrong somewhere else in the code.

openstack/identity/v3/tokencache/cache.go
```diff
@@ -0,0 +1,98 @@
...
+	"encoding/hex"
+	"encoding/json"
+	"strings"
+	"time"
+)
+
+const tokenExpiryMargin = 5 * time.Minute
+
+// Cache stores authentication results. Implementations must be safe for concurrent use.
+type Cache interface {
```
winiciusallan, line 16: As I can see, we don't have any concrete implementation of this interface throughout the PR (regardless the tests).

Is this being addressed in a follow-up or is it intended to be optional and implemented by the client?

openstack/identity/v3/websso/requests.go
```diff
@@ -0,0 +1,315 @@
...
+	if opts.RedirectHost != "" && opts.RedirectHost != "localhost" {
+		ip := net.ParseIP(opts.RedirectHost)
+		if ip == nil || !ip.IsLoopback() {
+			return fmt.Errorf("websso: RedirectHost must be a loopback address")
+		}
+	}
+	return nil
+}
+
+func captureToken(ctx context.Context, client *gophercloud.ServiceClient, opts *AuthOptions, port int, timeout time.Duration) (string, error) {
```
winiciusallan, line 211: Looking at the [upstream](https://review.opendev.org/c/openstack/keystoneauth/+/1000759) implementation, the callback is opened to CSRF and @cardoe tried to mitigate this with some validation in a helper function.

For security reasons, wdyt of having something similar here? We're only checking method, content-type, and body size.

winiciusallan (commented):
hey @rm-you, thank you very much for working on this and updating the PR promptly! I've left a few comments that I judged important, let me know what you think.

@stephenfin since you were involved in the upstream review, it would be nice to have eyes on this.

openstack/identity/v3/tokencache/cache.go
cardoe: So the python client uses the keyring package to store the cache in the system key chain. It's ultimately implemented by the end user client. But it would require each client to implement this. So I'm not sure how to best address this. There could be a default implementation in gophercloud? I'd be happy to do it as a follow up.

openstack/identity/v3/websso/requests.go
cardoe: It's worth noting that the CSRF is happening locally so it's not really a big danger. But I am trying to mitigate it going forward.

openstack/identity/v3/tokencache/cache.go
winiciusallan: As a SDK point of view, we want to make the end users' life easier, so IMO it's worth having at least one basic implementation while delivering the cache benefits (it improves overall performance and avoid round tripping to keystone).

If the other people agree, I think we could do this in a follow-up PR yes :slightly_smiling_face:.

rm-you:
Yeah I'm a bit distracted this week but I'll try to look at this soon -- but at a high level, yes this is in the middle of a chain of a number of PRs both here and in GARM, so there's a good chance that things unused here are in fact for use somewhere else. I'll look at the specifics and give you some links later this week, thanks for taking a look!

Edit: I had some time earlier than expected, so see in-line comments.

openstack/identity/v3/tokencache/cache.go
rm-you: I intentionally made the cache best-effort here. A malformed or stale local entry should not make authentication fail when we can delete it and continue through the normal browser flow; the next successful login repairs the entry. The cache package also has no logging facility, and returning a decoding error would make an optional cache less reliable than having no cache.

I agree that malformed data is distinct from an ordinary miss, though, and I can clarify that behavior in the docs. If the preferred contract is to expose backend or decoding errors, I can change Load to return an error, but my inclination is still delete-and-miss for malformed data and to reserve errors for failures that prevent recovery.

openstack/identity/v3/tokencache/cache.go
rm-you: The interface is already consumed by WebSSO in this PR: before opening the browser, it loads and validates a cached unscoped token; if Keystone rejects that token, it deletes the entry; after a successful browser flow, it persists the newly captured token. What is not included here is a concrete persistent-storage backend.

The original version of this integration put the system-keyring implementation in gophercloud/utils, alongside the higher-level clientconfig wiring. As the rest of this work evolved, the core clouds.ParseV3 follow-up made that Utils layer unnecessary. I did not move the keyring backend into base Gophercloud because doing so would add go-keyring and its platform-specific transitive dependencies to the core module for every user. Instead, the provider integration currently supplies the implementation in [internal/keyringcache](https://github.com/rm-you/garm-provider-openstack/blob/integration/federated-auth-reauth/internal/keyringcache/keyringcache.go) and passes it through [clouds.WithTokenCache](https://github.com/rm-you/garm-provider-openstack/blob/integration/federated-auth-reauth/client/client.go#L82).

I am not opposed to putting a concrete implementation here. If the maintainers are comfortable with those dependencies in the base Gophercloud module, I would be happy to move the keyring implementation into this PR or a small follow-up. I mainly wanted to avoid making that dependency decision implicitly.

For context, the complete dependency plan is:

- [#3939](https://github.com/gophercloud/gophercloud/pull/3939), context-aware reauthentication, is merged and is the common prerequisite.
- [#3984](https://github.com/gophercloud/gophercloud/pull/3984) adds the Identity v3 provider-construction path used by the higher-level parser.
- This PR adds WebSSO and the generic cache contract. WebSSO is its first consumer.
- [#3942](https://github.com/gophercloud/gophercloud/pull/3942) adds OIDC client-credentials auth. Assuming the PR we're commenting on now merges first, I will update this OIDC PR to make it use the same cache contract.
- [#3940](https://github.com/gophercloud/gophercloud/pull/3940) adds OAuth2 mTLS support.
- The unposted [feat/federated-auth-clouds](https://github.com/rm-you/gophercloud/tree/feat/federated-auth-clouds) follow-up adds clouds.ParseV3, registers those auth methods, and replaces the remaining gophercloud/utils clientconfig role. This might be a little confusing to look at since it is currently my own kind of ... integration branch that I use for deploying, but once a couple more of these merge I will have some follow-ups. I wish GitHub would let me do stacked PRs from my fork. :(
- [garm-provider-openstack#30](https://github.com/cloudbase/garm-provider-openstack/pull/30) is the merged Gophercloud v2 baseline. Its integration follow-up consumes clouds.ParseV3 and supplies the OS-keyring backend shown above.

Keeping the cache contract generic lets WebSSO now and OIDC later share the key format and token envelope, while leaving storage selection outside the core module unless the maintainers prefer otherwise.

Again, I am happy to add an implementation here, I was actually not happy to leave it without one either, but it felt like ya'll might balk at the added dependencies, so say the word and I'll put it in here.

openstack/identity/v3/websso/requests.go
rm-you: Good point. The listener is restricted to localhost or a loopback address, accepts only a form POST on the callback path, limits the body size, and validates the submitted token with Keystone before using it. That keeps it from being a remotely exposed arbitrary-token endpoint, but it does not bind the callback to the browser flow that opened it. A malicious local process, or a web page able to submit a form to the loopback listener, could race the legitimate callback and cause a denial of service or, with a valid token, an authentication-confusion result.

I agree that a per-flow unpredictable nonce is worthwhile defense in depth here. I can add the equivalent binding and focused tests in this PR. To be honest, I thought I remembered already doing that, but it might have been in one of my experimental branches. I'll just go ahead and handle that now.

rm-you: One correction after checking this while implementing it: a per-flow nonce cannot be added to the `origin`, because Keystone requires that value to match a `trusted_dashboard` entry verbatim. The current [keystoneauth change](https://review.opendev.org/c/openstack/keystoneauth/+/1000759) documents the same limitation.

I added the equivalent request-shape hardening in [43bc756](https://github.com/gophercloud/gophercloud/pull/3941/commits/43bc756114a126d6494233dd119199d4981c75f1): the callback now requires top-level navigation Fetch Metadata, rejects direct navigation, and rejects mismatched `Origin` or `Referer` when those headers are available. Focused tests cover each case. This narrows what can reach token handling, but it cannot cryptographically bind the callback to the flow, so the remaining login-CSRF limitation is documented.

openstack/identity/v3/tokencache/cache.go
winiciusallan: > [...] A malformed or stale local entry should not make authentication fail when we can delete it and continue through the normal browser flow; the next successful login repairs the entry. [...]

This has convinced me, so I'm okay keeping as-is. But let's keep the thread open in case anyone has objections.

openstack/identity/v3/tokencache/cache.go
winiciusallan: Although the cache is a "should have" considering the performance improvements, it isn't a hard dependency and we can do it in a follow-up PR if we want, so I'll not push this too long.

Can we at least document that users need to implement this and provide one reliable source?

I will try to thoroughly review the PRs in the correct order. This seems a perfect use case for [stacked PRs ](https://docs.github.com/en/pull-requests/get-started/about-stacked-prs):slightly_smiling_face:

rm-you: Yeah I wish they worked from a fork. If I had write access to this repo I could push them all to branches here and then created stacked PRs, but they don't work from a fork to the source repo (or fork to fork). :(

rm-you: Again, I would PREFER to add the cache implementation here NOW, the blocker for me was that I thought it would not be ideal to add client dependencies here on the keyring deps... but I am totally happy to just DO IT, if you give me a green light on that (and would rather just do it than put in temporary docs).

cardoe: Let's land this PR and let me pull it down in some local work and then stack some more PRs.

rm-you: Yeah, I mean the real issue is that I also am waiting for about three of these to land so that I can make some follow ups that tie them all together better, I literally have some branches just sitting waiting to be able to sanely make them a PR without the ability to stack :(

rm-you:
Applied the same shared-validation cleanup as #3942 in d0537360c, retaining the WebSSO-specific safety checks and excluding browser/cache hooks from JSON serialization. Added regression coverage for both entry points and hook handling; full race tests and lint pass on Go 1.25.0.

openstack/identity/v3/tokencache/cache.go
```diff
@@ -0,0 +1,99 @@
+// Package tokencache provides token caching for Identity v3 authentication.
```
stephenfin, line 1: So this is where our previously discussed lack of a `keystoneauth` equivalent comes back to bite us. Does this file belong here? Typically we only modules corresponding to APIs in `openstack/<service>/<version>` packages.

Would it be possible to have this (and it's test) in e.g. `openstack/auth` instead.

openstack/identity/v3/websso/requests.go
```diff
@@ -0,0 +1,368 @@
...
+		Flow:             "websso",
+		Principal:        opts.CacheNamespace,
+		IdentityEndpoint: identityEndpoint,
+		IdentityProvider: opts.IdentityProviderName,
+		Protocol:         opts.Protocol,
+	})
+}
+
+// Authenticate performs browser-based WebSSO authentication.
+func Authenticate(ctx context.Context, client *gophercloud.ServiceClient, builder tokens.AuthOptionsBuilder) (r tokens.CreateResult) {
```
stephenfin, line 95: I definitely feel like this doesn't belong here. I had a look for some of our more unusual existing auth mechanisms and found `openstack/identity/v3/oauth1/doc.go`. I note that that uses the more standard `AuthenticateV3` function from `openstack/client.go`. I'm wondering if this PR ought to hang off of the [pluggable auth PR](https://github.com/gophercloud/gophercloud/pull/4008) (which I'm yet to review also). I need to think about this one...

Later: Maybe if we called this "Create" we could deal with this for now?

```suggestion
// Create performs browser-based WebSSO authentication.
func Create(ctx context.Context, client *gophercloud.ServiceClient, opts tokens.AuthOptionsBuilder) (r tokens.CreateResult) {
```

It's layering on more c***, but at least it's consistent c***?

stephenfin (commented):
Partial review. I need to pull this down locally and take a second swing at it.

openstack/identity/v3/tokencache/cache.go
rm-you: Yep, or just auth/ which #4008 introduced (not sure whether that needs to move or not)

openstack/identity/v3/websso/requests.go
rm-you: I'm just going to refactor completely on #4008

rm-you:
I think you are correct that this and the other methods should just be on top of #4008 so let's get that moving 😇

PR #3942, opened 2026-08-12
identity: add OIDC client credentials authentication

### LINKED ISSUE #3283: Ability to use v3oidcclientcredentials for OpenStack keystone authentication (auth_type)
Hi and thanks for working on gophercloud :)

My understanding is that there is no support in gophercloud for OIDC keystone authentication at this time.

There are some issues around this topic but are mostly dead ends:
- https://github.com/gophercloud/gophercloud/issues/966
- https://github.com/gophercloud/gophercloud/issues/1929
- https://github.com/gophercloud/utils/issues/197

I've created an equivalent issue about this in terraform-provider-openstack (https://github.com/terraform-provider-openstack/terraform-provider-openstack/issues/1827) which is where I would use this feature if it was enabled by gophercloud and integrated into the terraform provider.

Before going further, are there any plans to develop something like this ?

As someone who isn't personally familiar with gophercloud's code, what would it look like and what kind of help or contribution would you like to see to make this happen ?

Thanks !

rm-you:
Fixes #3283

Adds Identity v3 authentication using the OpenID Connect client-credentials grant. It obtains an access token from an explicitly configured or discovered token endpoint, exchanges it for an unscoped Keystone token through the federation API, and optionally scopes the result with the existing token API.

The same flow supports reauthentication.

- [keystoneauth OIDC endpoint discovery and access-token request](https://github.com/openstack/keystoneauth/blob/83d4e0e73d50eaaa5a27f0cd2b2b4102f11b53a0/keystoneauth1/identity/v3/oidc.py#L192-L275)
- [keystoneauth federation token exchange](https://github.com/openstack/keystoneauth/blob/83d4e0e73d50eaaa5a27f0cd2b2b4102f11b53a0/keystoneauth1/identity/v3/oidc.py#L277-L316)
- [keystoneauth OIDC client-credentials plugin](https://github.com/openstack/keystoneauth/blob/83d4e0e73d50eaaa5a27f0cd2b2b4102f11b53a0/keystoneauth1/identity/v3/oidc.py#L493-L540)

### DIFF
--- a/openstack/identity/v3/oidc/requests.go
+++ b/openstack/identity/v3/oidc/requests.go
@@ -224,7 +224,7 @@ func fetchAccessToken(ctx context.Context, c *gophercloud.ServiceClient, opts *A
 	req.Header.Set("Accept", "application/json")
 
 	if opts.ClientSecret != "" {
-		req.SetBasicAuth(opts.ClientID, opts.ClientSecret)
+		req.SetBasicAuth(url.QueryEscape(opts.ClientID), url.QueryEscape(opts.ClientSecret))
 	}
 
 	httpClient := c.HTTPClient
--- a/openstack/identity/v3/oidc/requests.go
+++ b/openstack/identity/v3/oidc/requests.go
@@ -1,0 +1,299 @@
+package oidc
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"slices"
+	"strings"
+
+	"github.com/gophercloud/gophercloud/v2"
+	"github.com/gophercloud/gophercloud/v2/openstack/identity/v3/tokens"
+)
+
+type discoveryDocument struct {
+	TokenEndpoint       string   `json:"token_endpoint"`
+	GrantTypesSupported []string `json:"grant_types_supported"`
+}
+
+// AuthOptions contains OIDC client-credentials authentication options.
+type AuthOptions struct {
+	// IdentityProviderName is the Keystone federation identity provider.
+	IdentityProviderName string
+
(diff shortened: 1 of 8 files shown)


openstack/identity/v3/oidc/requests.go
```diff
@@ -0,0 +1,299 @@
...
+		return fmt.Errorf("missing required field: Protocol")
+	}
+	if opts.ClientID == "" {
+		return fmt.Errorf("missing required field: ClientID")
+	}
+	if opts.AccessTokenEndpoint == "" && opts.DiscoveryEndpoint == "" {
+		return fmt.Errorf("at least one of AccessTokenEndpoint or DiscoveryEndpoint must be provided")
+	}
+	return nil
+}
```
winiciusallan: I think you can use the `required:"true"` json tag in `AuthOptions` to use our existing validation. Then call `gophercloud.BuildRequestBody(any, parent)` in `ToTokenV3CreateMap` which you'd also need to call it at creation.

winiciusallan (commented):
Thanks @rm-you for proposing all these identity-related things. I'm doing my best to give useful insights while avoiding making wrong assumptions (I'm not familiar with keystone flow).

openstack/identity/v3/oidc/requests.go
rm-you: Done in 308362272: validation now uses `required`/`or` tags via `BuildRequestBody`, called through `ToTokenV3CreateMap` during creation. Either endpoint (or both) is still accepted, explicit endpoint precedence is unchanged, and regression tests cover both entry points.

Good call, missed this. :)
Also adjusted a couple other PRs to match.

PR #3945, opened 2026-08-14
pagination: Fix AllPages returning wrong key for empty linked pages

stephenfin:
When `AllPages()` processes a linked page where all results are empty, `EachPage` returns early without invoking the callback (because `IsEmpty` returns `true`). This left the resource key unset (empty string `""`), producing a body like `{"": nil}` instead of `{"networks": []}`.

This was harmless until b80f3715a added validation in `extractIntoPtr` that correctly rejects responses missing the expected key, causing `ExtractNetworks` (and similar) to fail with "expected response key not found in response" on empty list results.

Fix by discovering the resource key from the first page body before calling `EachPage`, rather than inside the callback.

### DIFF
--- a/pagination/pager.go
+++ b/pagination/pager.go
@@ -165,8 +165,19 @@ func (p Pager) AllPages(ctx context.Context) (Page, error) {
 	// `[]byte`, and `[]any`.
 	switch pb := firstPage.GetBody().(type) {
 	case map[string]any:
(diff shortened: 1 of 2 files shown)


winiciusallan (approved):

stephenfin:
I think this needs to be backported to `v2`. We're seeing similar issues in e.g. #3930

mandre (approved):
LGTM

PR #3947, opened 2026-08-14
pagination: Add KeyedPage interface

stephenfin:
Currently, pagination of list resources works by scanning resource bodies, looking for the first non-"links" resource key. This mostly works but is crude. It also doesn't work for resources that have multiple non-"links" top-level keys, such as placement's list resource provider inventories API, which has [both an `inventories` envelope key and a separate `resource_provider_generation` key](https://docs.openstack.org/api-ref/placement/#list-resource-provider-inventories).

Improve this by adding a new `KeyedPage` interface, with a `ResourceKey()` string value. When a `Page` struct implements this interface, the static key will be used rather than us scanning the body for the key. When this interface is *not* implemented, we continue to fallback to the old heuristic. The latter will allow us to implement the interface incrementally, and we only do it here for `NetworkPage` in `openstack/networking/v2/networks/results.go`.

### DIFF
--- a/pagination/pager.go
+++ b/pagination/pager.go
@@ -181,15 +181,8 @@ func (p Pager) AllPages(ctx context.Context) (Page, error) {
 		// Iterate over the pages to concatenate the bodies.
 		err = p.EachPage(ctx, func(_ context.Context, page Page) (bool, error) {
 			b := page.GetBody().(map[string]any)
-			for k, v := range b {
-				// If it's a linked page, we don't want the `links`, we want the other one.
-				if !strings.HasSuffix(k, "links") {
-					// check the field's type. we only want []any (which is really []map[string]any)
-					switch vt := v.(type) {
-					case []any:
-						pagesSlice = append(pagesSlice, vt...)
-					}
-				}
+			if v, ok := b[key].([]any); ok {
+				pagesSlice = append(pagesSlice, v...)
 			}
 			return true, nil
 		})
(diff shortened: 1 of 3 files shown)


winiciusallan:
@stephenfin once the conflicts are solved, I believe this will be good to go. :ok_hand:

stephenfin:
> @stephenfin once the conflicts are solved, I believe this will be good to go. 👌

Done. I also split the change into separate commits to make it clearer what I'm doing here.

winiciusallan (approved):

pagination/pager.go
```diff
@@ -165,31 +175,28 @@ func (p Pager) AllPages(ctx context.Context) (Page, error) {
...
-				if !strings.HasSuffix(k, "links") {
-					// check the field's type. we only want []any (which is really []map[string]any)
-					switch vt := v.(type) {
-					case []any:
-						pagesSlice = append(pagesSlice, vt...)
-					}
-				}
+			if v, ok := b[key].([]any); ok {
+				pagesSlice = append(pagesSlice, v...)
 			}
```
mandre, line 200: Should we return an error when the implementor has the wrong ResourceKey or it contains a typo? Right now, this silently returns empty results.

stephenfin: I didn't add this initially because it's a change in behavior under the dynamic case: previously, it would have happily pulled the first non-`link` key it found and used that. However, you're probably right: it would be wiser to explode here rather than silently do the wrong thing. I'll add a follow-up shortly.

stephenfin: Done in https://github.com/gophercloud/gophercloud/pull/3975

PR #3949, opened 2026-08-15
fix(config/clouds): resolve reference to clouds-public.yaml

### LINKED ISSUE #3948: clouds.Parse does not read clouds-public.yaml
### What happened?

Openstack configuration defines a clouds-public.yaml (https://docs.openstack.org/python-openstackclient/2026.1/configuration/index.html#clouds-public-yaml) file containing public informations that can be shared across clouds and users. 

clouds.Parse does not take this file into account leading to erronous authOptions when using the file. 

For example with the config given in the aftermentionned docs and the following program

```go
package main

import (
	"fmt"

	"github.com/gophercloud/gophercloud/v2/openstack/config/clouds"
)

func main() {
	authOptions, _, _, err := clouds.Parse()
	if err != nil {
		panic(err)
	}

	fmt.Printf("%+v\n", authOptions)
}
```

We have

```
$ OS_CLOUD=infra go run .
{IdentityEndpoint: Username:openstack UserID: Password:xyzpdq!lazydog Passcode: DomainID: DomainName: TenantID:275610 TenantName: AllowReauth:false TokenID: Scope:<nil> ApplicationCredentialID: ApplicationCredentialName: ApplicationCredentialSecret:}
```
``IdentifyEndpoint``is not set while it should be ``https://identity.api.rackspacecloud.com/v2.0/`` as shown by

```
$ os configuration show | grep auth.auth_url                                                                                                                                                                     
| auth.auth_url                               | https://identity.api.rackspacecloud.com/v2.0/ | 
```

### Version

v2 (current)

AntoninRuan:
Fixes #3948

code in this PR:

https://docs.openstack.org/python-openstackclient/2026.1/configuration/index.html#clouds-public-yaml

### DIFF
--- a/openstack/config/clouds/clouds.go
+++ b/openstack/config/clouds/clouds.go
@@ -81,15 +81,9 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
 			if err != nil {
 				return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, fmt.Errorf("failed to get the current working directory: %w", err)
 			}
-			// Use XDG_CONFIG_HOME or fall back to ~/.config, matching the
-			// OpenStack convention for clouds.yaml location on all platforms.
-			userConfig := os.Getenv("XDG_CONFIG_HOME")
-			if userConfig == "" {
-				homeDir, err := os.UserHomeDir()
-				if err != nil {
-					return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, fmt.Errorf("failed to get the user home directory: %w", err)
-				}
-				userConfig = path.Join(homeDir, ".config")
+			userConfig, err := getUserConfig()
+			if err != nil {
+				return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, err
 			}
 			options.locations = []string{path.Join(cwd, "clouds.yaml"), path.Join(userConfig, "openstack", "clouds.yaml"), path.Join("/etc", "openstack", "clouds.yaml")}
 		}
@@ -141,10 +135,15 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
 				var err error
 				cloud, err = mergeClouds(secureCloud, cloud)
 				if err != nil {
-					return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, fmt.Errorf("unable to merge information from clouds.yaml and secure.yaml")
+					return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, fmt.Errorf("unable to merge information from clouds.yaml and secure.yaml: %w", err)
 				}
 			}
 		}
+	}
+
+	cloud, err := mergeWithPublicClouds(cloud, &options)
+	if err != nil {
+		return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, err
 	}
 
 	tlsConfig, err := computeTLSConfig(cloud, options)
@@ -181,6 +180,81 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
 		},
 		tlsConfig,
 		nil
+}
+
+func getUserConfig() (string, error) {
+	// Use XDG_CONFIG_HOME or fall back to ~/.config, matching the
+	// OpenStack convention for clouds.yaml location on all platforms.
+	userConfig := os.Getenv("XDG_CONFIG_HOME")
+	if userConfig != "" {
+		return userConfig, nil
+	}
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get the user home directory: %w", err)
+	}
+	userConfig = path.Join(homeDir, ".config")
+	return userConfig, nil
+}
+
+func mergeWithPublicClouds(cloud Cloud, options *cloudOpts) (Cloud, error) {
+	var mergeWith string
+	if cloud.Profile != "" {
+		mergeWith = cloud.Profile
+	} else if cloud.Cloud != "" {
+		mergeWith = cloud.Cloud
+	} else {
+		return cloud, nil
+	}
+
+	// Code is executed only if cloud needs to be merged with a public
+	// profile, error are returned only loading clouds-public.yaml was
+	// needed
+	if options.cloudsPublicyamlReader == nil {
+		if len(options.publicLocations) < 1 {
+			cwd, err := os.Getwd()
+			if err != nil {
+				return cloud, fmt.Errorf("failed to get the current directory: %w", err)
+			}
+			userConfig, err := getUserConfig()
+			if err != nil {
+				return cloud, err
+			}
+			options.publicLocations = []string{path.Join(cwd, "clouds-public.yaml"), path.Join(userConfig, "openstack", "clouds-public.yaml"), path.Join("/etc", "openstack", "clouds-public.yaml")}
+		}
+		for _, publicPath := range options.publicLocations {
+			f, err := os.Open(publicPath)
+			if err != nil {
+				continue
+			}
+			defer f.Close()
+			options.cloudsPublicyamlReader = f
+			break
+		}
+		if options.cloudsPublicyamlReader == nil {
+			return cloud, fmt.Errorf("clouds file not found. Search locations were: %v", options.publicLocations)
+		}
+	}
+
+	var publicClouds PublicClouds
+	if err := yaml.NewDecoder(options.cloudsPublicyamlReader).Decode(&publicClouds); err != nil {
+		return cloud, err
+	}
+
+	pCloud, ok := publicClouds.Clouds[mergeWith]
+	if !ok {
+		return cloud, nil
+	}
+
+	var err error
+	cloud, err = mergeClouds(cloud, pCloud)
+	if err != nil {
+		return cloud, fmt.Errorf("unable to merge information from clouds-public.yaml: %w", err)
+	}
+
+	return cloud, nil
+
 }
 
 // computeAvailability is a helper method to determine the endpoint type
--- a/openstack/config/clouds/clouds_test.go
+++ b/openstack/config/clouds/clouds_test.go
@@ -8,6 +8,7 @@ import (
 	"testing"
 
 	"github.com/gophercloud/gophercloud/v2/openstack/config/clouds"
+	th "github.com/gophercloud/gophercloud/v2/testhelper"
 )
 
 func ExampleWithCloudName() {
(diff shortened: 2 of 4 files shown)


openstack/config/clouds/clouds.go
```diff
@@ -183,6 +184,68 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
 
+func getUserConfig() (string, error) {
+	userConfig := os.Getenv("XDG_CONFIG_HOME")
+	if userConfig != "" {
+		return userConfig, nil
+	}
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get the user hom directory: %w", err)
```
winiciusallan:
```suggestion
		return "", fmt.Errorf("failed to get the user home directory: %w", err)
```

openstack/config/clouds/clouds.go
```diff
@@ -83,13 +83,9 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
 			}
 			// Use XDG_CONFIG_HOME or fall back to ~/.config, matching the
 			// OpenStack convention for clouds.yaml location on all platforms.
```
winiciusallan: Please, move this comment to the `getUserConfig` helper function.

stephenfin: Can you move this note into the new helper function

openstack/config/clouds/clouds.go
stephenfin:
```suggestion
		return "", fmt.Errorf("failed to get the user home directory: %w", err)
```

openstack/config/clouds/clouds.go
```diff
@@ -183,6 +184,68 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return "", fmt.Errorf("failed to get the user hom directory: %w", err)
+	}
+	userConfig = path.Join(homeDir, ".config")
+	return userConfig, nil
+}
+
+func mergeWithPublicClouds(cloud Cloud, options *cloudOpts) (Cloud, error) {
+	if cloud.Cloud == "" {
```
stephenfin: [`cloud` is the deprecated alias for `profile`](https://github.com/openstack/openstacksdk/blob/6946e39adfc0c26a4535b64481256ea6619353f0/openstack/config/loader.py#L664-L666). We need to check both `cloud.Cloud` and `cloud.Profile`, with a preference for the latter (higher priority).

As a follow-up, we should likely also remove the `Cloud` field from the `Cloud` struct and instead modify the `UnmarshalJSON` to scan for the deprecated `cloud` option and populate `Profile` if found. I say this should be a follow-up since it would be a major version change, unlike the rest of this PR.

openstack/config/clouds/clouds.go
```diff
@@ -183,6 +184,68 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
+	var publicClouds PublicClouds
+	if err := yaml.NewDecoder(options.cloudsPublicyamlReader).Decode(&publicClouds); err != nil {
+		return cloud, err
+	}
+
+	pCloud, ok := publicClouds.Clouds[cloud.Cloud]
+	if !ok {
+		return cloud, nil
+	}
+
```
stephenfin: We likely want to emit some kind of warning here, else things could get very confusing if a user requests a profile that doesn't exist.

openstack/config/clouds/clouds.go
```diff
@@ -183,6 +184,68 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
+	var err error
+	cloud, err = mergeClouds(cloud, pCloud)
+	if err != nil {
+		return cloud, fmt.Errorf("unable to merge informations from clouds-public.yaml")
```
stephenfin: Grammar nit but `information` doesn't have a plural (Google says it's an "uncountable mass noun"). Also, we probably want to log the error here (`%w`)?


```suggestion
		return cloud, fmt.Errorf("unable to clouds-public.yaml: %w", err)
```

openstack/config/clouds/options.go
```diff
@@ -53,6 +55,15 @@ func WithLocations(locations ...string) ParseOption {
...
 }
 
+// WithPublicLocations is a functional option that sets the search location for
+// the clouds-public.yaml. Each location is a file path pointing to a possible
+// `clouds-public.yaml`
+func WithPublicLocations(locations ...string) ParseOption {
+	return func(co *cloudOpts) {
+		co.publicLocations = locations
+	}
+}
```
stephenfin, line 65: openstacksdk has independent searches for the three files (`clouds.yaml`, `secure.yaml`, and `clouds-public.yaml`). It's possible for `clouds.yaml` to be retrieved from `/etc` while `secure.yaml` is retrieved from `~/.config/openstack`. We don't currently have this distinction here for `secure.yaml`: it will only ever be read from the directory `clouds.yaml` is read from.

This being the case, I think we should either (a) fix `secure.yaml` loading, to make it independent from `clouds.yaml` loading, or (b) make `clouds-public.yaml` dependent on `clouds.yaml`. (a) is probably the more correct answer, but I'm happy with either. A `TODO` would be okay for now.

openstack/config/clouds/clouds.go
```diff
@@ -183,6 +184,68 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
+			defer f.Close()
+			options.cloudsPublicyamlReader = f
+			break
+		}
+	}
+
+	var publicClouds PublicClouds
+	if err := yaml.NewDecoder(options.cloudsPublicyamlReader).Decode(&publicClouds); err != nil {
+		return cloud, err
+	}
```
stephenfin, line 243: There's a missing `nil` check here: `options.cloudsPublicyamlReader` can be `nil` if none of the files in `options.publicLocations` were found

openstack/config/clouds/types.go
```diff
@@ -9,6 +9,10 @@ type Clouds struct {
 	Clouds map[string]Cloud `yaml:"clouds" json:"clouds"`
 }
 
+type PublicClouds struct {
+	Clouds map[string]Cloud `yaml:"public-clouds" json:"public-clouds"`
+}
+
 // Cloud represents an entry in a clouds.yaml/public-clouds.yaml/secure.yaml file.
 type Cloud struct {
 	Cloud      string    `yaml:"cloud,omitempty" json:"cloud,omitempty"`
```
stephenfin, line 18: As I noted elsewhere, this should probably be removed in favour of `Profile` and `cloud` handled via the unmarshaller. That's a separate change though

stephenfin (changes requested):

openstack/config/clouds/clouds.go
AntoninRuan: I've tried to make anything as close as python openstack cli, and from my experiments (I did not have a look at the code), a reference to a non declared public-cloud is silently ignored. But yes maybe a warning would be a good idea.

openstack/config/clouds/clouds.go
AntoninRuan: I've almost copied the error message from a failed merge with secure.yaml (adding a 's' to information while doing it). If log the error here maybe we could also do that at line 138

openstack/config/clouds/options.go
AntoninRuan: Hmm okay, I assumed reading ``secure.yaml`` from the same place as ``clouds.yaml`` was intended openstack behavior. 
At one point while doing this I tried to abstract the loading file/reader mechanism (line 78-112) into an helper function, but that failed as moving the ``defer f.Close()`` into helper function broke the yaml decode. 
Maybe moving a bit more into the helper could help fix it, I'll have a look and see if it can be a quick fix that fit into this PR

openstack/config/clouds/clouds.go
AntoninRuan: Hmm that's true. But what do we want to do in this case, do we silently ignore it, and like you proposed for reference to non existent public-cloud and throw a warning

AntoninRuan:
And a more overall question that I don't remenber seeing addresses in the contributing guideline, what is the preffered way to edit a PR: amending the commit to keep only one commit in the PR, or adding new commit for each round of corrections?

openstack/config/clouds/clouds.go
stephenfin: You're probably not seeing the warning because OSC silences most/all warnings from SDK, but SDK does log this. See https://github.com/openstack/openstacksdk/blob/a01433391d49c88b1033c990befefca48b1836a2/openstack/config/loader.py#L705-L711

openstack/config/clouds/clouds.go
stephenfin: > If log the error here maybe we could also do that at line 138

Yes, good idea. The only reason I can think to *not* log the error is if that meant we'd log the contents of the file (including passwords) but that isn't the case.

openstack/config/clouds/options.go
stephenfin: > Hmm okay, I assumed reading `secure.yaml` from the same place as `clouds.yaml` was intended openstack behavior.

So did I, but I looked through SDK and sure enough, it loads them all individually. To be fair, this allows for scenarios like e.g. share a common `clouds.yaml` at `/etc` while users can have `secure.yaml` in their own home directory.

> At one point while doing this I tried to abstract the loading file/reader mechanism (line 78-112) into an helper function, but that failed as moving the `defer f.Close()` into helper function broke the yaml decode. Maybe moving a bit more into the helper could help fix it, I'll have a look and see if it can be a quick fix that fit into this PR

Great. Just to be clear though, I am also happy with a TODO for loading `secure.yaml` independently of `clouds.yaml`. We don't have to block your work while we bikeshed on that

openstack/config/clouds/clouds.go
stephenfin: I suspect we just ignore it: the absence of any `clouds-public.yaml` file isn't an issue unless a user requested a `profile`.

stephenfin:
> And a more overall question that I don't remenber seeing addresses in the contributing guideline, what is the preffered way to edit a PR: amending the commit to keep only one commit in the PR, or adding new commit for each round of corrections?

We don't have one. I'd typically prefer amending commits and force pushing, but GitHub's UI works better with follow-up commits. Do whatever you're used to/prefer.

openstack/config/clouds/clouds.go
AntoninRuan: Do you have any pattern to emit such a warning? I've grepped the project and did not found anything with "warn". Would a simple printf to stderr be sufficient?

AntoninRuan:
@stephenfin I've just added in a different commit my attempt at loading every file independently of each others. I am not fully satisfied notably the way I handled errors I think it could be better. (maybe it would be a good idea to separate the commit in a different pr, I'll let you decide what you think is best).
I think I've also fixed almost all comments in an amend of the first commit, expect the warning one as I did not know how you wanted to handle warning emmision

stephenfin:
> @stephenfin I've just added in a different commit my attempt at loading every file independently of each others. I am not fully satisfied notably the way I handled errors I think it could be better. (maybe it would be a good idea to separate the commit in a different pr, I'll let you decide what you think is best). I think I've also fixed almost all comments in an amend of the first commit, expect the warning one as I did not know how you wanted to handle warning emmision

Yes, could you keep this locally but drag it out of this PR (or consider creating one of the newfangled [stacked PRs](https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests)?) I'll hold off reviewing until that's done to avoid losing review comments.

AntoninRuan:
Ok I'll do that

AntoninRuan:
Done. I did not find a way to create stacked PRs as both branch are on my fork rather than this repo. I'll just keep the branch and open a new later on

stephenfin (dismissed):

openstack/config/clouds/clouds_test.go
```diff
@@ -335,4 +335,175 @@ func TestParse(t *testing.T) {
...
+  gophercloud-test-1:
+    auth:
+      auth_url: https://example.com/gophercloud-test-1:13000
+      user_domain_name: Default`
+
+		tmpDir, err := os.MkdirTemp(os.TempDir(), tempDirPrefix)
+		if err != nil {
+			t.Fatalf("unable to create a temporary directory: %v", err)
+		}
+		defer rmTmpDirOrPanic(tmpDir)
```
stephenfin:
```suggestion
		th.AssertNoErr(t, err)
		defer rmTmpDirOrPanic(tmpDir)
```

openstack/config/clouds/clouds_test.go
```diff
@@ -335,4 +335,175 @@ func TestParse(t *testing.T) {
...
+
+		cwd, err := os.Getwd()
+		if err != nil {
+			t.Fatalf("unable to determine the current working directory: %v", err)
+		}
+		if err := os.Chdir(tmpDir); err != nil {
+			t.Fatalf("unable to move to a temporary directory: %v", err)
+		}
```
stephenfin:
```suggestion
		th.AssertNoErr(t, err)
		err = os.Chdir(tmpDir)
		th.AssertNoErr(t, err)
```

etc. etc.

stephenfin (commented):
Just noting that we have test helpers to simplify much of the error checking in tests. If you have time to fix, I'm happy to re-review.

openstack/config/clouds/clouds_test.go
AntoninRuan: Oh yeah seems nicer. I'll update the tests

stephenfin (approved):

PR #3962, opened 2026-08-20
dependabot: group gomod and github-actions updates into single PR

mandre:
Dependabot currently opens one PR per dependency, which causes interdependent bumps to land as separate, independently-mergeable PRs. For example, https://github.com/gophercloud/gophercloud/pull/3937, https://github.com/gophercloud/gophercloud/pull/3935, and https://github.com/gophercloud/gophercloud/pull/3933 each bump one of github/codeql-action/{analyze,autobuild,init} within the same codeql-analysis.yaml workflow, instead of landing together.

Add a groups entry to each gomod and github-actions update block (for both the main and v2 target branches) so that all updates for a given ecosystem/branch are grouped into a single pull request.

Also reduce frequency to reduce noise.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -4,7 +4,8 @@ updates:
   directory: "/"
   target-branch: "main"
   schedule:
-    interval: daily
+    interval: "weekly"
+    day: "monday"
   cooldown:
     default-days: 7
   open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


IlanZuckerman (approved):
lgtm

winiciusallan (approved):

PR #3975, opened 2026-08-20
pagination: Return error on missing resource key

stephenfin:
A follow-up to #3947 that returns an error if the requested key was not found on a page, rather than silently ignoring the key.


mandre (approved):
Nice, thanks for the quick fix.

PR #3957, opened 2026-08-18
feat: add loadbalancer advanced tag filters

jihyun-huh:
Fixes #3956 

This PR is to add the remaining Octavia advanced tag filters to listener, pool, health monitor, and member list operations below.

- `tags-any`
- `not-tags`
- `not-tags-any`

and member list options now support all four tag filters, including `tags`.

This follows the existing naming and query encoding used by load balancers, L7 policies, and L7 rules. Unit tests cover query generation, while acceptance coverage verifies matching and non-matching behavior for every filter.

References:

- https://docs.openstack.org/api-ref/load-balancer/v2/#filtering-by-tags
- https://opendev.org/openstack/octavia/src/branch/master/octavia/api/common/pagination.py

related to https://github.com/gophercloud/gophercloud/pull/3914

PR #3977, opened 2026-08-20
[v2] add tags on octavia loadbalancers API 

winiciusallan:
Manual backport of https://github.com/gophercloud/gophercloud/pull/3957 and https://github.com/gophercloud/gophercloud/pull/3914

The apidiff returned

```
    Incompatible changes:
    - ListOpts: old is comparable, new is not
    - ListRulesOpts: old is comparable, new is not
```

but this might be safe to backport

### DIFF
--- a/internal/acceptance/openstack/loadbalancer/v2/l7policies_test.go
+++ b/internal/acceptance/openstack/loadbalancer/v2/l7policies_test.go
@@ -7,8 +7,10 @@ import (
 	"testing"
 
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/clients"
+	networking "github.com/gophercloud/gophercloud/v2/internal/acceptance/openstack/networking/v2"
 	"github.com/gophercloud/gophercloud/v2/internal/acceptance/tools"
 	"github.com/gophercloud/gophercloud/v2/openstack/loadbalancer/v2/l7policies"
+	th "github.com/gophercloud/gophercloud/v2/testhelper"
 )
 
 func TestL7PoliciesList(t *testing.T) {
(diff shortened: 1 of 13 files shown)


mandre:
Even if the backward incompatible change is limited to structs that were previously comparable and that no longer are, I don't feel totally comfortable backporting a change flagged as `major`. If we start making exceptions, it's going to be unmanageable very quickly.

winiciusallan:
Yeah, I agree. This is a practice that we usually don't do and there aren't enough reasons to start doing considering the we'll need to manage them afterwards. Closing then.

stephenfin:
> Even if the backward incompatible change is limited to structs that were previously comparable and that no longer are, I don't feel totally comfortable backporting a change flagged as `major`. If we start making exceptions, it's going to be unmanageable very quickly.

Could we document the list of exceptions we allow to prevent this "slippery slope" occurring?

If not, the best alternative I can think of is to create new structs that embed the old ones (`ListOptsExt`?) and add these fields to that, like we used to do in v1 when we still maintained the idea of (non-networking) extensions. But that would mean users would need to use a different struct and then switch back to `ListOpts` in v3. The other alternative is to use a pointer instead of a slice, but that's worse UX and, again, presents an API difference to manage come v3.

winiciusallan:
> Could we document the list of exceptions we allow to prevent this "slippery slope" occurring?

I agree this could be a good one, but is there any exception? Desconsidering the "prepare" commits, we haven't merged too major changes. To be more exact, since the v2.0.0 tag:

```
> jq 'map(select(contains({labels: [{name: "semver:major"}]}) ))' ../prs2.json | jq '.[] | select(.title | startswith("Prepare") | not) | {number, title}'
{
  "number": 3327,
  "title": "Merge pull request #3209 from shiftstack/proper-service-discovery"
}
{
  "number": 3375,
  "title": "[v2] core: add missing Builder interfaces"
}
{
  "number": 3393,
  "title": "[v2] dns: implement shared zones list"
}
{
  "number": 3397,
  "title": "[v2] Adjust List func to accept a Builder in tenants, routers and security groups packages"
}
{
  "number": 3434,
  "title": "[v2] Randomize test order for unit tests"
}
{
  "number": 3632,
  "title": "[v2] CI: Fix fwaas jobs"
}
{
  "number": 3648,
  "title": "[v2] Do not specify go patch version"
}
{
  "number": 3808,
  "title": "dns: fix SudoTenantID type in zones ListOpts"
```

> If not, the best alternative I can think of is to create new structs that embed the old ones (`ListOptsExt`?) and add these fields to that, like we used to do in v1 when we still maintained the idea of (non-networking) extensions. But that would mean users would need to use a different struct and then switch back to `ListOpts` in v3. The other alternative is to use a pointer instead of a slice, but that's worse UX and, again, presents an API difference to manage come v3.

While this solves the problem, we will need to maintain various new structs and then take care of get rid of all this when branching V3. Our code is extensible and people can, for example, implement their own `ListOptsBuilder`.

stephenfin:
> > Could we document the list of exceptions we allow to prevent this "slippery slope" occurring?
> 
> I agree this could be a good one, but is there any exception? Desconsidering the "prepare" commits, we haven't merged too major changes. To be more exact, since the v2.0.0 tag:

The things I can think of are:

* Bumps to minimum go version
* Unavoidable fixes for security issues
* Things like this where we decide the positives to users outweigh the negatives

winiciusallan:
fyi, another similar case

https://github.com/gophercloud/gophercloud/pull/3921#issuecomment-5512578384

PR #3978, opened 2026-08-20
Add support for default security group rules

### LINKED ISSUE #3175: support neutron security group default rules
The neutron v2 API has an endpoint to operate on the template of security group default rules.

Interacting with the endpoint should be rather easy, and very similar to the standard security group rules one.

See the following pages to learn more about this:
* https://docs.openstack.org/api-ref/network/v2/index.html#security-group-default-rules-security-group-default-rules
* https://docs.openstack.org/python-openstackclient/latest/cli/command-objects/default-security-group-rule.html
* https://github.com/openstack/neutron/blob/08fff4087dc342be40db179fca0cd9bbded91053/neutron/db/securitygroups_db.py#L1124
* https://github.com/openstack/neutron/blob/08fff4087dc342be40db179fca0cd9bbded91053/neutron/db/securitygroups_db.py#L1133

pslestang:
Add a new openstack/networking/v2/extensions/security/defaultrules package supporting the List, Create, Get and Delete operations of the Neutron /v2.0/default-security-group-rules API, provided by the security-groups-default-rules extension (Neutron 2023.2). The API does not support update nor bulk create for this resource.

The package mirrors the existing security/rules package and reuses its RuleDirection, RuleProtocol and RuleEtherType constants. Unit tests and an acceptance test are included.

Fixes #3175

List security group default rules (GET /v2.0/default-security-group-rules):
https://docs.openstack.org/api-ref/network/v2/index.html#list-security-group-default-rules

Create security group default rule (POST /v2.0/default-security-group-rules):
https://docs.openstack.org/api-ref/network/v2/index.html#create-security-group-default-rule

Show security group default rule (GET /v2.0/default-security-group-rules/{id}):
https://docs.openstack.org/api-ref/network/v2/index.html#show-security-group-default-rule

Delete security group default rule (DELETE /v2.0/default-security-group-rules/{id}):
https://docs.openstack.org/api-ref/network/v2/index.html#delete-security-group-default-rule

### DIFF
--- a/internal/acceptance/openstack/networking/v2/extensions/security_default_rules_test.go
+++ b/internal/acceptance/openstack/networking/v2/extensions/security_default_rules_test.go
@@ -1,0 +1,83 @@
+//go:build acceptance || networking || security
+
+package extensions
+
+import (
+	"context"
+	"testing"
+
+	"github.com/gophercloud/gophercloud/v2/internal/acceptance/clients"
+	networking "github.com/gophercloud/gophercloud/v2/internal/acceptance/openstack/networking/v2"
+	"github.com/gophercloud/gophercloud/v2/internal/acceptance/tools"
+	"github.com/gophercloud/gophercloud/v2/internal/ptr"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/security/defaultrules"
+	"github.com/gophercloud/gophercloud/v2/openstack/networking/v2/extensions/security/rules"
+	th "github.com/gophercloud/gophercloud/v2/testhelper"
+)
+
+func TestDefaultSecurityGroupRulesCreateGetListDelete(t *testing.T) {
+	clients.RequireAdmin(t)
+
+	client, err := clients.NewNetworkV2Client()
+	th.AssertNoErr(t, err)
+
+	networking.RequireNeutronExtension(t, client, "security-groups-default-rules")
+
+	description := "Default rule description"
+	fromPort := tools.RandomInt(10080, 10089)
+	toPort := tools.RandomInt(10090, 10099)
+
+	createOpts := defaultrules.CreateOpts{
+		Description:        description,
+		Direction:          rules.DirIngress,
+		EtherType:          rules.EtherType4,
+		PortRangeMin:       ptr.To(fromPort),
+		PortRangeMax:       ptr.To(toPort),
+		Protocol:           rules.ProtocolTCP,
+		RemoteGroupID:      "PARENT",
+		UsedInDefaultSG:    ptr.To(true),
+		UsedInNonDefaultSG: ptr.To(false),
+	}
+
+	rule, err := defaultrules.Create(context.TODO(), client, createOpts).Extract()
+	th.AssertNoErr(t, err)
+	defer func() {
+		err := defaultrules.Delete(context.TODO(), client, rule.ID).ExtractErr()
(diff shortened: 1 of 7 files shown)


openstack/networking/v2/extensions/security/defaultrules/requests.go
```diff
@@ -0,0 +1,164 @@
...
+	Direction            string `q:"direction"`
+	EtherType            string `q:"ethertype"`
+	PortRangeMax         int    `q:"port_range_max"`
+	PortRangeMin         int    `q:"port_range_min"`
+	Protocol             string `q:"protocol"`
+	RemoteAddressGroupID string `q:"remote_address_group_id"`
+	RemoteGroupID        string `q:"remote_group_id"`
+	RemoteIPPrefix       string `q:"remote_ip_prefix"`
+	UsedInDefaultSG      *bool  `q:"used_in_default_sg"`
+	UsedInNonDefaultSG   *bool  `q:"used_in_non_default_sg"`
```
winiciusallan, line 35: The api-ref says that these fields should go into the [body](https://docs.openstack.org/api-ref/network/v2/index.html#list-security-group-default-rules) request, but I believe this is a bug on the documentation. Pushed a [patch](https://review.opendev.org/c/openstack/neutron-lib/+/1001816) to fix this

openstack/networking/v2/extensions/security/defaultrules/requests.go
```diff
@@ -0,0 +1,164 @@
...
+
+// ListOptsBuilder allows extensions to add additional parameters to the List
+// request.
+type ListOptsBuilder interface {
+	ToDefaultSecGroupRuleListQuery() (string, error)
+}
+
+// ListOpts allows the filtering and sorting of paginated collections through
+// the API. Filtering is achieved by passing in struct field values that map to
+// the default security group rule attributes you want to see returned. SortKey
```
winiciusallan, line 19: This sentence about filtering confused me a little. Wouldn't be the idea of this something like: we pass field values for the attributes that we would like to use as filters.?

internal/acceptance/openstack/networking/v2/extensions/security_default_rules_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	th.AssertEquals(t, description, rule.Description)
+	th.AssertEquals(t, "ingress", rule.Direction)
+	th.AssertEquals(t, "IPv4", rule.EtherType)
+	th.AssertEquals(t, fromPort, rule.PortRangeMin)
+	th.AssertEquals(t, toPort, rule.PortRangeMax)
+	th.AssertEquals(t, string(rules.ProtocolTCP), rule.Protocol)
+	th.AssertEquals(t, "PARENT", rule.RemoteGroupID)
+	th.AssertEquals(t, true, rule.UsedInDefaultSG)
+	th.AssertEquals(t, false, rule.UsedInNonDefaultSG)
```
winiciusallan, line 59: We usually do a GET for the created resource and then compare the returned value with the expected, instead of comparing with the result of a creation operation. Some APIs return different values for these operations and comparing with the GET result shows that in a N + 1 operation the resource state is as expected.

So IMO, you might want to move these checks after the `defaultrules.Get` and then compare with the `getRule` SG.

winiciusallan (commented):
Thanks @pslestang for working on this. Left a few comments, let me know what you think.

PR #3979, opened 2026-08-21
ci(db): add slashes to file pattern

winiciusallan:
Adding slashes between the word "db" so we can match only packages related with Trove. This avoids running other unrelated jobs e.g. loadbalancer, because the filter was matching the "db" substring from the word.


IlanZuckerman (approved):
LGTM

kayrus (approved):

stephenfin (approved):

PR #3981, opened 2026-08-21
Add apicheck tool to validate resources against Gophercloud

stephenfin:
The https://github.com/gtema/openstack repo provides OpenAPI schema for many of the same OpenStack services that Gophercloud supports. These can be found [here](https://github.com/gtema/openstack/tree/main/openstack_types/data). These OpenAPI schemas are built on a combination of JSON Schema schemas provided in-tree by some projects (Nova and Keystone primarily, though Ironic and Cinder also have partial coverage) along with parsed versions of the api-ref and some custom glue. Long-term, the hope is that these schemas will become "official" and replace the api-ref, but in the interim they still provide great machine-readable "docs" for the OpenStack APIs.

This PR is an early attempt to use these in Gophercloud. Long-term, we would like to take advantage of Go's code generation capabilities to start generating most of our structs and functions from the OpenAPI schemas. This should make the addition of a new API or APIs as simple as pulling in a new version of the OpenAPI schema. That's a lot of work though and might result in breaking changes as we massage Gophercloud's API into something better suited to code generation. So instead of doing that, we add a tool that will allow us to verify what we have against the schemas. This allows us to highlight gaps, which we can either fix manually or call out and wait for auto-generation capabilities.

This PR is currently almost entirely one-shotted by Claude and should mostly be ignored until it's in a better state. I mainly have it up here to show what could be possible and ensure this is a direction we want to go in.


winiciusallan:
@stephenfin do you mind giving me some context about this? there is `+23,447`, so it's a lot  and I'd like to be in the same page.

stephenfin:
> @stephenfin do you mind giving me some context about this? there is `+23,447`, so it's a lot and I'd like to be in the same page.

This is highly WIP and probably should be ignored for now. I'll update comment 0 now to describe what I'm trying to achieve here though.

gtema:
we may want to address few other issues:
- a better delivery of specs to github (or publishing them in openinfra side). Now it is a semi automated process where every merged codegenerator change pushed commit to the 'openstack' repo, but doesn't open a PR.  Also a change in the service repo doesn't trigger on it's own new schema change/pr
- you need to find a strategy dealing with deviating response (not only request) schemas per MV - they are not additive as you perfectly know. One of the worst examples is with server flavor where server.flavor.id is present only till 2.46 and server.flavor.original_name exists only from 2.47 (just had fun with it in cli issue report)
- sometimes responses have _a, _b schemas within a single MV. I only try them serially in cli/tui - I am not aware of any other way.
- you are most likely not that interested in query parameters to the list endpoints where param types are also sometimes changing with MV

And don't forget, codegenerator already pushed generated rust code to the 'openstack' repository, and so it can also push generated go code to gophercloud repo so that we unify all worlds to the single source of truth.

PR #3983, opened 2026-08-21
fix(config/clouds): find the clouds config file independently of each…

AntoninRuan:
Opening the PR as a follow up of this comment: https://github.com/gophercloud/gophercloud/pull/3949#issuecomment-5345636690. (@stephenfin)

Change should make config files loading independent of each others as openstacksdk does

### DIFF
--- a/openstack/config/clouds/clouds.go
+++ b/openstack/config/clouds/clouds.go
@@ -23,6 +23,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path"
 	"reflect"
@@ -57,12 +58,14 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
 		cloudName:    os.Getenv("OS_CLOUD"),
 		region:       os.Getenv("OS_REGION_NAME"),
 		endpointType: os.Getenv("OS_INTERFACE"),
-		locations: func() []string {
+		locations: func() map[CloudsType][]string {
+			out := make(map[CloudsType][]string)
 			if path := os.Getenv("OS_CLIENT_CONFIG_FILE"); path != "" {
-				return []string{path}
+				out[Default] = []string{path}
 			}
-			return nil
+			return out
 		}(),
+		readers: make(map[CloudsType]io.Reader),
 	}
 
 	for _, apply := range opts {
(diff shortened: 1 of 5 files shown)


openstack/config/clouds/clouds.go
```diff
@@ -73,77 +76,54 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
 
-	cloud, ok := clouds.Clouds[options.cloudName]
+	cloud, ok := clouds[options.cloudName]
 	if !ok {
 		return gophercloud.AuthOptions{}, gophercloud.EndpointOpts{}, nil, fmt.Errorf("cloud %q not found in clouds.yaml", options.cloudName)
 	}
 
-	if options.secureyamlReader != nil {
-		var secureClouds Clouds
-		if err := yaml.NewDecoder(options.secureyamlReader).Decode(&secureClouds); err != nil {
```
Sharpz7, line 127: `WithCloudsYAML` only sets the default reader, so this call falls through to filesystem discovery for `secure.yaml`. That means an in-memory config can unexpectedly pick up a same-named entry from the current directory or XDG config. Can we avoid filesystem lookup unless a secure reader/location was explicitly provided?

openstack/config/clouds/clouds.go
```diff
@@ -182,79 +162,71 @@ func Parse(opts ...ParseOption) (gophercloud.AuthOptions, gophercloud.EndpointOp
...
-		return cloud, nil
-	}
-
-	// Code is executed only if cloud needs to be merged with a public
-	// profile, error are returned only loading clouds-public.yaml was
-	// needed
-	if options.cloudsPublicyamlReader == nil {
-		if len(options.publicLocations) < 1 {
+func readClouds(options *cloudOpts, ctype CloudsType) (map[string]Cloud, error) {
+	// Set the defaults and open the files for reading. This code only runs
```
Sharpz7, line 166: `WithLocations` only sets `locations[Default]`, so the secure lookup does not check beside the selected `clouds.yaml`. Existing callers using `/custom/clouds.yaml` and `/custom/secure.yaml` will silently lose the secure values. Can we preserve the companion lookup, or add a separate secure-file location option and update the docs?

Sharpz7 (commented):
I found two issues with the new independent file lookup behavior.

Assisted-By: Codex

openstack/config/clouds/clouds.go
AntoninRuan: I am adding a test to detect it and I'll see if I have time to change it in the week

openstack/config/clouds/clouds.go
AntoninRuan: Maybe we could make ``WithLocation`` also update the ``secure.yaml`` location file to keep old behavior. And have independent find only when no ``WithLocation is set``.  

That would mean we match openstack-cli when no options are passed. When only ``WithLocation`` is given we preserve old behavior and when both ``WithLocation`` and ``WithSecureLocation`` are given ``WithSecureLocation`` overrides the secure location.

Would you be ok with that?

AntoninRuan:
@Sharpz7 Fixed the two issues you mentionned in your previous comment.

openstack/config/clouds/options.go
```diff
@@ -51,7 +51,17 @@ func WithCloudName(osCloud string) ParseOption {
...
 func WithLocations(locations ...string) ParseOption {
 	return func(co *cloudOpts) {
-		co.locations = locations
+		co.locations[Default] = locations
+		co.locationsOverriden[Default] = true
+	}
+}
+
+// WithSecureLocations is a functional option that sets the search locations for
+// secure.yaml file, Each location is a file path to a possible `secure.yaml`
```
winiciusallan, line 60: typo

```suggestion
// secure.yaml file. Each location is a file path to a possible `secure.yaml`
```

openstack/config/clouds/clouds_test.go
winiciusallan, line 219: I would like to see a test case when setting `OS_CLIENT_CONFIG_FILE`. The `Parse` function docs says

```
//   - if the environment variable `OS_CLIENT_CONFIG_FILE` is set and points to a
//     clouds.yaml, use that location as the only search location for `clouds.yaml` and `secure.yaml`;
```

and I'm afraid that this might no be true anymore, since both files are read separately and when OS_CLIENT_CONFIG_FILE is set, the out[Default] with the path is returned.

https://github.com/gophercloud/gophercloud/pull/3983/changes#diff-eab218efe948a7873c1abd24c3ad5a42cba607599f3f8eb0b1ce4d4e4160fb41R61-R66

winiciusallan (commented):
Overall, it looks good to me. I left a comment that might be a regression or just a change in the function docs. Please see in line

openstack/config/clouds/options.go
```diff
@@ -7,12 +7,12 @@ import (
...
-	publicLocations        []string
-	cloudsyamlReader       io.Reader
-	secureyamlReader       io.Reader
-	cloudsPublicyamlReader io.Reader
+	cloudName           string
+	locations           map[CloudsType][]string
+	locationsOverriden  map[CloudsType]bool
+	secureAsideLocation string
+	readers             map[CloudsType]io.Reader
+	readersOverriden    map[CloudsType]bool
```
winiciusallan, line 15: typo on `Overriden`. The same applies for `locationsOverriden`. You might want to fix all references.

```suggestion
	readersOverridden    map[CloudsType]bool
```

openstack/config/clouds/types.go
```diff
@@ -207,4 +210,8 @@ const (
 
 	// AuthV3ApplicationCredential defines version 3 of the application credential
 	AuthV3ApplicationCredential AuthType = "v3applicationcredential"
+
+	Default CloudsType = "clouds.yaml"
+	Secure  CloudsType = "secure.yaml"
+	Public  CloudsType = "clouds-public.yaml"
```
winiciusallan, line 216: If we want to make these constants public, it'd be nice to have godoc on these.

AntoninRuan: I didn't think much about it when adding them, but in retrospect I don't think they should be public. I'll update (as well as the type of these constants)

openstack/config/clouds/clouds_test.go
AntoninRuan: I've added the test case and it seems the case was already handled by the changes I previously made to load aside ``secure.yaml`` when ``WithLocation`` is passed

PR #3986, opened 2026-08-27
skills: add a review skill for Gophercloud

winiciusallan:
This commit introduces a new skill for helping reviewers and authors to thoroughly review their codes and follow the best project practices, leveraging the use of AI to achieve this.


.agents/skills/review/SKILL.md
```diff
@@ -0,0 +1,86 @@
...
+- **Positives**: always include at least one concrete item when anything is correct; if the review is all blockers, still note any solid parts
+
+If there are no blockers, say the code is not blocked and keep warnings/suggestions/positives as applicable. Residual risk (no live API, untested microversions) goes under Warnings.
+
+## Rules of the review
+
+- Follow existing package patterns over inventing new ones.
+- List `Page` types whose JSON body has more than one top-level key (besides `*links`) must implement `pagination.KeyedPage` (`ResourceKey()`), so `AllPages` does not guess the collection key. See `pagination/pager.go` and [gophercloud#3947](https://github.com/gophercloud/gophercloud/pull/3947).
+- Assertion order is **expected, actual**: `th.AssertEquals(t, expected, actual)`.
+- Cite code with `startLine:endLine:path`.
```
mandre, line 83: Shouldn't it be `path:line` as above (line 43)? Or perhaps `path:startLine-endLine` for multiline citations.

.agents/skills/review/SKILL.md
```diff
@@ -0,0 +1,86 @@
+---
+name: review-gophercloud
+description: Review Gophercloud Go SDK code for API coverage gaps, convention violations, microversion mistakes, missing tests, and bugs. Use when the user invokes this skill or asks to review Gophercloud packages, PRs, diffs, or OpenStack API implementations.
+disable-model-invocation: true
```
mandre, line 4: We may actually want to remove the `disable-model-invocation: true` setting so that the skill is visible to the agent. Otherwise it forces you to invoke the skill with a slash command, which kinda defeats the purpose IMO.

.agents/skills/review/checklist.md
```diff
@@ -0,0 +1,91 @@
+# Gophercloud review checklist
+
+Copy this list into the review notes and check items that apply. Skip items that are N/A and say so.
```
mandre, line 3: This new skill has a LOT of overlap with AGENTS.md, which is always loaded as project context. I'd suggest reworking this skill to be more about verification than explaining the project conventions.

Having overlapping content is also another potential source of drift over time.

mandre (commented):
This is most welcome, thanks! I believe we should try to rely more on AGENTS.md rather than trying to restating some of the project conventions.

.agents/skills/review/SKILL.md
winiciusallan: yup! specified that.

.agents/skills/review/SKILL.md
winiciusallan: This is tipically the review that you, as a user, would invoke, but is not harmful if your favorite clanker decide to start reviewing some code (there should be something in the context to do so).

With other changes, I removed the model invocation boolean.

.agents/skills/review/checklist.md
winiciusallan: YES! This makes a lot of sense and I learned a lot how agents work with skills. 

Now we have a much smaller skill file and now keeping the project conventions in AGENTS.md as a single source of truth for this. Now, we have general and discoverable instructions about the review process. My personal experience is doing great... if you want to try it out.

winiciusallan:
@mandre may I have another round of review? :)

AGENTS.md
```diff
@@ -207,7 +213,23 @@ err := pager.EachPage(ctx, func(ctx context.Context, page pagination.Page) (bool
 })
 ```
 
-## Key Reminders
+### 5.3 KeyedPage interface
+
+`AllPages` concatenates list bodies. By default it picks the first top-level JSON key that is not `*links` and whose value is an array.
```
mandre, line 218: nit: 
```suggestion
`AllPages` concatenates list bodies. By default it picks an arbitrary top-level JSON key that is not `*links` and whose value is an array.
```
(go map iteration order is non-deterministic).

.agents/skills/review/SKILL.md
```diff
@@ -0,0 +1,98 @@
+---
+name: review-gophercloud
+description: Review Gophercloud Go SDK code for API coverage gaps, convention violations, microversion mistakes, missing tests, and bugs. Use when the user invokes this skill or asks to review Gophercloud packages, PRs, diffs, or OpenStack API implementations.
```
mandre, line 3: That's quite a long description, and the second sentence is redundant. Keep in mind that the frontmatter always end up in the system prompt, so if we can save a few char it's nice.


```suggestion
"Gophercloud: Review packages, PRs, and diffs for convention violations, API coverage gaps, microversion mistakes, missing tests, and bugs."
```

.agents/skills/review/SKILL.md
```diff
@@ -0,0 +1,98 @@
...
+---
+
+# Review Gophercloud
+
+Review Gophercloud (`github.com/gophercloud/gophercloud/v2`) as a core contributor would. Do not implement fixes unless the user asks.
+
+## Patterns
+
+Do not restate repository conventions here. Read [AGENTS.md](../../../AGENTS.md) and cite the matching section when a finding is a pattern violation, for example `AGENTS.md § 2.4` or `see § 5.3`.
```
mandre, line 12: Pick one convention, and stick with it.
```suggestion
Do not restate repository conventions here. Read [AGENTS.md](../../../AGENTS.md) and cite the matching section when a finding is a pattern violation, for example `AGENTS.md § 2.4`.
```

.agents/skills/review/SKILL.md
```diff
@@ -0,0 +1,98 @@
...
+
+Consult AGENTS.md for: running tests and lint (§ 1); imports, file layout, naming, and OptsBuilder (§ 2.1–2.3); pointers, `omitempty`, errors, and GoDoc (§ 2.4–2.6); unit and acceptance tests (§ 2.7); microversions (§ 3, § 6); pull requests (§ 4); `context.Context`, pagination, and the KeyedPage interface (§ 5).
```
mandre, line 14: I have a suspicion the paragraph number is going to drift very quickly :D. Consider removing them.

Issue #3987, opened 2026-08-28
Add support for Blazar (reservation)


barnes-c:
There is currently no support for the Blazar (reservation service) in gophercloud.

https://docs.openstack.org/api-ref/reservation/v1/index.html

As mentioned in #3304 Im going to pick this up and continue the work already done before me

---
- [x] Service client, acceptance scaffolding and CI workflow (#3988)

---
- [ ] Leases: List
- [ ] Leases: Create
- [ ] Leases: Get
- [ ] Leases: Update
- [ ] Leases: Delete

---- 
- [ ] Hosts: List (#3998)
- [ ] Hosts: Create
- [ ] Hosts: Get
- [ ] Hosts: Update
- [ ] Hosts: Delete

---
- [ ] Floating IPs: List
- [ ] Floating IPs: Create
- [ ] Floating IPs: Get
- [ ] Floating IPs: Delete

---
- [ ] Allocations: List
- [ ] Allocations: Get

---
- [ ] Resource properties: List
- [ ] Resource properties: Update

PR #3988, opened 2026-08-28
reservation: Add Blazar service client and CI job

barnes-c:
Hi!

For #3987

This is a initial PR to add reservation (blazar) support. This PR brings in a service client, acceptance test scaffolding and a ci workflow so future additions can be tested immediately.

I was mostly orientating myself along the code already present for other resources

Related:
https://github.com/gophercloud/gophercloud/issues/3304
https://github.com/gophercloud/gophercloud/pull/3406

### DIFF
--- a/.github/workflows/functional-reservation.yaml
+++ b/.github/workflows/functional-reservation.yaml
@@ -58,8 +58,6 @@ jobs:
           branch: ${{ matrix.openstack_version }}
           conf_overrides: |
             enable_plugin blazar https://github.com/openstack/blazar ${{ matrix.openstack_version }}
-
-            KEYSTONE_ADMIN_ENDPOINT=true
           enabled_services: "openstack-cli-server"
 
       - name: Checkout go
--- a/.github/workflows/functional-reservation.yaml
+++ b/.github/workflows/functional-reservation.yaml
@@ -1,0 +1,101 @@
+name: functional-reservation
+on:
+  merge_group:
+  pull_request:
+  schedule:
+    - cron: "0 0 */3 * *"
+jobs:
+  functional-reservation:
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - name: "master"
+            openstack_version: "master"
+            ubuntu_version: "24.04"
+          - name: "gazpacho"
+            openstack_version: "stable/2026.1"
+            ubuntu_version: "24.04"
+          - name: "epoxy"
+            openstack_version: "stable/2025.1"
+            ubuntu_version: "24.04"
+    runs-on: ubuntu-${{ matrix.ubuntu_version }}
+    name: Blazar on OpenStack ${{ matrix.name }}
+    steps:
+      - name: Checkout Gophercloud
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Check changed files
+        uses: ./.github/actions/file-filter
+        id: changed-files
+        with:
+          patterns: |
+            openstack/auth_env.go
+            openstack/client.go
+            openstack/endpoint.go
+            openstack/endpoint_location.go
+            openstack/config/provider_client.go
+            openstack/utils/choose_version.go
+            openstack/utils/discovery.go
+            **reservation**
+            .github/workflows/functional-reservation.yaml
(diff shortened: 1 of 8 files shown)


.github/workflows/functional-reservation.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+      - name: Deploy devstack
+        if: ${{ fromJSON(steps.changed-files.outputs.matches) }}
+        uses: gophercloud/devstack-action@60ca1042045c0c9e3e001c64575d381654ffcba1 # v0.19
+        with:
+          branch: ${{ matrix.openstack_version }}
+          conf_overrides: |
+            enable_plugin blazar https://github.com/openstack/blazar ${{ matrix.openstack_version }}
+
+            KEYSTONE_ADMIN_ENDPOINT=true
```
mandre: Is blazar using admin endpoints internally? Perhaps let's add this once we have tests that require the setting.

openstack/client.go
```diff
@@ -508,3 +508,12 @@ func NewWorkflowV2(ctx context.Context, client *gophercloud.ProviderClient, eo g
 func NewPlacementV1(ctx context.Context, client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
 	return initClientOpts(ctx, client, eo, "placement", 1)
 }
+
+// NewReservationV1 creates a ServiceClient that may be used with the reservation package.
+func NewReservationV1(ctx context.Context, client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
+	sc, err := initClientOpts(ctx, client, eo, "reservation", 1)
+	if !strings.HasSuffix(strings.TrimSuffix(sc.Endpoint, "/"), "v1") {
```
mandre, line 542: This can actually be quite confusing if `initClientOpts()` returned an error, we would set `ResourceBase` to `"v1/"` on error paths. It would be better to have:

```go
sc, err := initClientOpts(ctx, client, eo, "reservation", 1)
if err != nil {
	return sc, err
}
if !strings.HasSuffix(strings.TrimSuffix(sc.Endpoint, "/"), "v1") {
	sc.ResourceBase = sc.Endpoint + "v1/"
}
return sc, nil
```

I understand that you copied other functions that have the same issue, I'll submit a patch to fix them all.

mandre:
Thanks for the PR, I'm looking forward to reviewing the service implementation.

openstack/client.go
mandre: Cleanup patch in https://github.com/gophercloud/gophercloud/pull/3989.

barnes-c: Adapted the changes here too!

.github/workflows/functional-reservation.yaml
barnes-c: Removed the setting :)

barnes-c:
> Thanks for the PR, I'm looking forward to reviewing the service implementation.

Thanks for being open to review this!

.github/workflows/functional-reservation.yaml
```diff
@@ -0,0 +1,99 @@
...
+        if: ${{ fromJSON(steps.changed-files.outputs.matches) }}
+        uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7.0.0
+        with:
+          go-version-file: "go.mod"
+          cache: true
+
+      - name: Run Gophercloud acceptance tests
+        if: ${{ fromJSON(steps.changed-files.outputs.matches) }}
+        run: |
+          source ${{ github.workspace }}/script/stackenv
```
winiciusallan, line 73: is stackenv going to be used in acceptance tests? if not or if you're unsure, I'd remove it. if yes, we can keep it.

winiciusallan (commented):
Nice kickstart @barnes-c! Good job.

.github/workflows/functional-reservation.yaml
barnes-c: > Nice kickstart @barnes-c! Good job.

Thanks!

Regarding stackenv, I was following the convention of the other clients. When creating a client they use `openstack.AuthOptionsFromEnv()` which needs the env vars that stackenv brings, so yes its needed

mandre (approved):

barnes-c:
Thanks for merging @mandre!
I will start with the resources now

PR #3989, opened 2026-08-28
Return early on initClientOpts error in service client constructors

mandre:
Several New* functions access the `ServiceClient` returned by `initClientOpts` without checking the error first. On error paths, this results in setting fields like `ResourceBase` on a half-initialized client (e.g.` ResourceBase = "v2.0/"`). Add early error returns to avoid modifying a client that failed to initialize.


kayrus (approved):

winiciusallan (approved):
Thanks for the fix! Waiting for green jobs before merging; the failure looks unrelated with the changes

PR #3990, opened 2026-08-28
test(db): use datastore-appropriate config parameter in TestConfigurationsCRUD

### LINKED ISSUE #3982: Trove jobs fail on openstack main
### What happened?

Trove jobs are constantly failing since 6th of July with:

```
got 422 instead: {"unprocessableEntity": {"code": 422, "message": "The configuration parameter collation_server is not supported for this datastore: postgresql 18."}}
```

https://github.com/gophercloud/gophercloud/actions?query=workflow%3Afunctional-db+is%3Afailure

### Version

main (development)

mandre:
Trove's devstack plugin (commit 25dde62d8f0b) now registers three datastores (mysql, mariadb, postgresql) instead of just mysql. Our script/stackenv picks whichever openstack datastore list returns first, which is non-deterministic. When postgresql is selected, the test fails because collation_server is a MySQL/MariaDB-only parameter.

Use max_connections for PostgreSQL and keep collation_server for MySQL/MariaDB.

Fixes #3982

### DIFF
--- a/internal/acceptance/openstack/db/v1/configurations_test.go
+++ b/internal/acceptance/openstack/db/v1/configurations_test.go
@@ -35,7 +35,11 @@ func TestConfigurationsCRUD(t *testing.T) {
 	createOpts.Datastore = &datastore
 
 	values := make(map[string]any)
-	values["collation_server"] = "latin1_swedish_ci"
+	if choices.DBDatastoreType == "postgresql" {
+		values["max_connections"] = 200
+	} else {
+		values["collation_server"] = "latin1_swedish_ci"
+	}
 	createOpts.Values = values
 
 	cgroup, err := configurations.Create(context.TODO(), client, createOpts).Extract()


mandre:
We'll need to run the CI several times until we confirm this has worked with postgresql backend.

internal/acceptance/openstack/db/v1/configurations_test.go
```diff
@@ -35,7 +35,11 @@ func TestConfigurationsCRUD(t *testing.T) {
 	createOpts.Datastore = &datastore
 
 	values := make(map[string]any)
-	values["collation_server"] = "latin1_swedish_ci"
+	if choices.DBDatastoreType == "postgresql" {
+		values["max_connections"] = "200"
```
winiciusallan: Per CI error message, I believe this value should be an `int`.

```
The value provided for the configuration parameter max_connections is not of type integer.
```

winiciusallan (commented):
I was about to suggest to pin just one of the datastores, instead of having one of the three depending on the run. But if this works, it's better for now, so we can have a coverage for all datastores.

mandre:
The last [master job ran with postgresql](https://github.com/gophercloud/gophercloud/actions/runs/33192854596/job/98957885887?pr=3990#step:8:930) and succeeded :green_circle:

winiciusallan (approved):
Sweet!

PR #3995, opened 2026-08-31
Fix factual errors and omissions in README

mandre:
Fixes a few issues with the README:
- Fix branch description: main is the v3 development branch, not the stable v2 branch
- Fix servers.Create example: add missing hintOpts (nil) fourth argument
- Fix misleading comment about NewComputeV2 resolving microversions; the context is used for endpoint resolution, cancellation, and tracing
- Fix typo: contructor -> constructor
- Add missing services to Supported Services table: Metric (Aetos), Orchestration (Heat), Placement, and Workflow (Mistral)
- Fix Keystone identity endpoint in manual auth example: v2.0 -> v3

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,8 @@ Gophercloud is a Go SDK for OpenStack.
 Join us on kubernetes slack, on [#gophercloud](https://kubernetes.slack.com/archives/C05G4NJ6P6X). Visit [slack.k8s.io](https://slack.k8s.io) for an invitation.
 
 > **Note**
-> This branch contains the current stable branch of Gophercloud: `v2`.
+> This branch contains the current development version of Gophercloud (`v3`).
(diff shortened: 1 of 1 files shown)


README.md
winiciusallan, line 213: Do we want to have something similar with ORC regarding the "level of implementation" of the services? Like "most implemented", "basic ops"... (this is not the better naming 🙁)

ex.: the placement service only has the `resourceproviders` API on v1, however there are a lot of other resources that could have been supported.

mandre: I think that might be a little too much for this PR.

But otherwise I agree it would be nice to have a clue how complete the implementation is, and I think https://github.com/gophercloud/gophercloud/pull/3981 is a step in that direction.

winiciusallan (approved):

PR #3998, opened 2026-08-31
reservation: add hosts resources

barnes-c:
For #3987

Adds the `hosts` package to the Reservation (Blazar) v1 service client, covering the Hosts suite of the [api-ref](https://docs.openstack.org/api-ref/reservation/v1/index.html#hosts):

| Operation |           Endpoint           | Status |
| --------- | ---------------------------- | ------ |
| `List`    | `GET /os-hosts`              | 200    |
| `Create`  | `POST /os-hosts`             | 201    |
| `Get`     | `GET /os-hosts/{host_id}`    | 200    |
| `Update`  | `PUT /os-hosts/{host_id}`    | 200    |
| `Delete`  | `DELETE /os-hosts/{host_id}` | 204    |

- The blueprint is mounted at `/v1/os-hosts` ([v1_0.py#L27][1])
- and the routes are [list][2], [create][3], [get][4], [update][5] and [delete][6]
- The status codes above are the defaults of the `Rest` helper ([utils.py#L47-L59][7]).

[1]: https://opendev.org/openstack/blazar/src/branch/stable/2026.1/blazar/api/v1/oshosts/v1_0.py#L27
[2]: https://opendev.org/openstack/blazar/src/branch/stable/2026.1/blazar/api/v1/oshosts/v1_0.py#L33-L36
[3]: https://opendev.org/openstack/blazar/src/branch/stable/2026.1/blazar/api/v1/oshosts/v1_0.py#L39-L42
[4]: https://opendev.org/openstack/blazar/src/branch/stable/2026.1/blazar/api/v1/oshosts/v1_0.py#L45-L49
[5]: https://opendev.org/openstack/blazar/src/branch/stable/2026.1/blazar/api/v1/oshosts/v1_0.py#L52-L61
[6]: https://opendev.org/openstack/blazar/src/branch/stable/2026.1/blazar/api/v1/oshosts/v1_0.py#L63-L68
[7]: https://opendev.org/openstack/blazar/src/branch/stable/2026.1/blazar/api/v1/utils.py#L47-L59

### DIFF
--- a/openstack/reservation/v1/hosts/requests.go
+++ b/openstack/reservation/v1/hosts/requests.go
@@ -1,0 +1,13 @@
+package hosts
+
+import (
+	"github.com/gophercloud/gophercloud/v2"
+	"github.com/gophercloud/gophercloud/v2/pagination"
+)
+
+// List retrieves a list of compute hosts.
+func List(client *gophercloud.ServiceClient) pagination.Pager {
+	return pagination.NewPager(client, listURL(client), func(r pagination.PageResult) pagination.Page {
+		return HostPage{pagination.SinglePageBase(r)}
+	})
+}
--- a/openstack/reservation/v1/hosts/requests.go
+++ b/openstack/reservation/v1/hosts/requests.go
@@ -3,6 +3,7 @@ package hosts
 import (
 	"context"
 	"fmt"
+	"maps"
 
 	"github.com/gophercloud/gophercloud/v2"
 	"github.com/gophercloud/gophercloud/v2/pagination"
@@ -89,6 +90,45 @@ func Create(ctx context.Context, client *gophercloud.ServiceClient, opts CreateO
 // Get retrieves a specific compute host based on its unique ID.
 func Get(ctx context.Context, client *gophercloud.ServiceClient, id string) (r GetResult) {
 	resp, err := client.Get(ctx, getURL(client, id), &r.Body, &gophercloud.RequestOpts{
+		OkCodes: []int{200},
+	})
+	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
+	return
+}
+
+// UpdateOptsBuilder allows extensions to add parameters to the Update request.
+type UpdateOptsBuilder interface {
+	ToHostUpdateMap() (map[string]any, error)
+}
+
+// UpdateOpts specifies the parameters for updating a host. Only the extra
+// capabilities of a host can be changed.
+type UpdateOpts struct {
+	// ExtraCapabilities holds the operator-defined capabilities to set on the host.
+	ExtraCapabilities map[string]any
+}
+
+// ToHostUpdateMap formats an UpdateOpts into a request body.
+func (opts UpdateOpts) ToHostUpdateMap() (map[string]any, error) {
+	if len(opts.ExtraCapabilities) == 0 {
+		return nil, gophercloud.ErrMissingInput{Argument: "ExtraCapabilities"}
+	}
+
+	b := make(map[string]any, len(opts.ExtraCapabilities))
+	maps.Copy(b, opts.ExtraCapabilities)
+
+	return b, nil
+}
+
+// Update changes the extra capabilities of a compute host.
+func Update(ctx context.Context, client *gophercloud.ServiceClient, id string, opts UpdateOptsBuilder) (r UpdateResult) {
+	b, err := opts.ToHostUpdateMap()
+	if err != nil {
+		r.Err = err
+		return
+	}
+
+	resp, err := client.Put(ctx, updateURL(client, id), b, &r.Body, &gophercloud.RequestOpts{
 		OkCodes: []int{200},
 	})
 	_, r.Header, r.Err = gophercloud.ParseResponse(resp, err)
--- a/openstack/reservation/v1/hosts/requests.go
+++ b/openstack/reservation/v1/hosts/requests.go
@@ -15,8 +15,8 @@ type ListOptsBuilder interface {
 }
 
 // ListOpts allows the filtering of paginated collections through the API.
-//
-// Blazar accepts query parameters on this endpoint but ignores them. It returns every host.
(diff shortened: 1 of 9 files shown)


openstack/reservation/v1/hosts/requests.go
```diff
@@ -0,0 +1,13 @@
+package hosts
+
+import (
+	"github.com/gophercloud/gophercloud/v2"
+	"github.com/gophercloud/gophercloud/v2/pagination"
+)
+
+// List retrieves a list of compute hosts.
+func List(client *gophercloud.ServiceClient) pagination.Pager {
```
mandre: Gophercloud `List` functions would typically take a `ListOptsBuilder` arg, so that users can extend it if needed.
Here, it seems that [the GET](https://docs.openstack.org/api-ref/reservation/v1/index.html#list-hosts) doesn't support any query parameters so this is fine, however if filters are added in the future, we would need to change the function signature.

We may want to stub out an empty ListOpts, just in case, unless of course we're confident the Blazar API is unlikely to change.

mandre (commented):
Seems good, just a concern about possible API breakage, in case the blazar API evolves. I'm actually unsure about this, so looking for your feedback as well.

Also feel free to bundle more API calls in the same PR, for example all of the [Host calls](https://docs.openstack.org/api-ref/reservation/v1/index.html#hosts): this will be simpler if we want to backport the implementation to v2, and will ensure we have proper acceptance tests before merging anything.

openstack/reservation/v1/hosts/requests.go
barnes-c: > free to bundle more API calls in the same PR

Will do! I will add the other methods for hosts in this PR then!

> just a concern about possible API breakage, in case the blazar API evolves. I'm actually unsure about this

Yes, its behind other projects on this. I see that upstream already is "query-aware" and most pieces are already in place. I will try to upstream this too but I see the usual velocity of Blazar is very slow

I will add the stub here for now. Upstream already accepts [query parameters](https://opendev.org/openstack/blazar/src/branch/master/blazar/api/v1/oshosts/v1_0.py#L32-L36).

openstack/reservation/v1/hosts/requests.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/gophercloud/gophercloud/v2/pagination"
+)
+
+// ListOptsBuilder allows extensions to add parameters to the List request.
+type ListOptsBuilder interface {
+	ToHostListQuery() (string, error)
+}
+
+// ListOpts allows the filtering of paginated collections through the API.
+//
```
winiciusallan: We usually don't put empty comment lines in docstring.

openstack/reservation/v1/hosts/requests.go
```diff
@@ -0,0 +1,145 @@
...
+type UpdateOpts struct {
+	// ExtraCapabilities holds the operator-defined capabilities to set on the host.
+	ExtraCapabilities map[string]any
+}
+
+// ToHostUpdateMap formats an UpdateOpts into a request body.
+func (opts UpdateOpts) ToHostUpdateMap() (map[string]any, error) {
+	if len(opts.ExtraCapabilities) == 0 {
+		return nil, gophercloud.ErrMissingInput{Argument: "ExtraCapabilities"}
+	}
```
winiciusallan, line 115: I think we should not return an error here when no extra capability is passed. In this case, Blazar identifies that no value has been passed and just immediately returns the compute host with no exceptions. What do you think?

https://github.com/openstack/blazar/blob/d2b6e5d030b6eeb6c26ac4b7e257c289f3b4e841/blazar/plugins/oshosts/host_plugin.py#L481-L483

openstack/reservation/v1/hosts/requests.go
```diff
@@ -0,0 +1,145 @@
...
+
+	b := make(map[string]any, len(opts.ExtraCapabilities))
+	maps.Copy(b, opts.ExtraCapabilities)
```
winiciusallan, line 118: Does call `gophercloud.BuildRequestBody` work in this case?

openstack/reservation/v1/hosts/requests.go
barnes-c: Thanks for your review!

openstack/reservation/v1/hosts/requests.go
barnes-c: Hi!
I think you are looking at the wrong location, you checked the plugin. 
In the code for the API you see that this check also exists with an error too

https://opendev.org/openstack/blazar/src/branch/stable/2026.1/blazar/api/v1/oshosts/v1_0.py#L56-L58

winiciusallan: oh, yup, I think you're right. I traced the request using the v2 path: https://github.com/openstack/blazar/blob/d2b6e5d030b6eeb6c26ac4b7e257c289f3b4e841/blazar/api/v2/controllers/extensions/host.py#L156.

So I'm ok with this.

openstack/reservation/v1/hosts/requests.go
barnes-c: looking at `BuildRequestBody` it calls `json.Marshal` so the field would be in the body

Blazar wants the capabilities at the top so it needs to be like this

https://opendev.org/openstack/blazar/src/branch/stable/2026.1/doc/api_samples/hosts/host-create-req.json
https://opendev.org/openstack/blazar/src/branch/stable/2026.1/doc/api_samples/hosts/host-update-req.json

winiciusallan: yup, you're right. I just checked here with a code snippet and the way you handled seems good.

winiciusallan (approved):

mandre (approved):
Sweet, really nice work.

barnes-c:
Hi @winiciusallan & @mandre, 

I think this PR got stuck. 
Could you try to add it to the merge queue again

Cheers,
Chris

PR #4005, opened 2026-09-07
auth: do not fall back to user_domain_*/project_domain_* in clouds.yaml parser

kayrus:
The clouds.yaml parser previously coalesced the user identity domain from user_domain_id/user_domain_name and the project scope domain from domain_id/domain_name. Those fallbacks are removed:

- AuthOptions.DomainID/DomainName now come only from domain_id/domain_name.
- The project-scope domain now comes only from project_domain_id/ project_domain_name.

Callers must specify the user and project domains explicitly. Tests updated to reflect the new behavior.

Followup for the #3739 PR

### DIFF
--- a/openstack/config/clouds/clouds_test.go
+++ b/openstack/config/clouds/clouds_test.go
@@ -88,46 +88,32 @@ func TestParse(t *testing.T) {
       username: gophercloud-test-username`
 
 		tmpDir, err := os.MkdirTemp(os.TempDir(), tempDirPrefix)
-		if err != nil {
-			t.Fatalf("unable to create a temporary directory: %v", err)
-		}
+		th.AssertNoErr(t, err)
 		defer rmTmpDirOrPanic(tmpDir)
 
 		cwd, err := os.Getwd()
-		if err != nil {
-			t.Fatalf("unable to determine the current working directory: %v", err)
-		}
-		if err := os.Chdir(tmpDir); err != nil {
-			t.Fatalf("unable to move to a temporary directory: %v", err)
-		}
+		th.AssertNoErr(t, err)
+		err = os.Chdir(tmpDir)
+		th.AssertNoErr(t, err)
 		defer func() {
 			if err := os.Chdir(cwd); err != nil {
 				panic("unable to reset the current working directory: " + err.Error())
 			}
 		}()
 
-		if err := os.WriteFile("clouds.yaml", []byte(cloudsYAML), 0644); err != nil {
-			t.Fatalf("unable to create a mock clouds.yaml file: %v", err)
-		}
+		err = os.WriteFile("clouds.yaml", []byte(cloudsYAML), 0644)
+		th.AssertNoErr(t, err)
 
-		if err := os.WriteFile("secure.yaml", []byte(secureYAML), 0644); err != nil {
-			t.Fatalf("unable to create a mock secure.yaml file: %v", err)
-		}
+		err = os.WriteFile("secure.yaml", []byte(secureYAML), 0644)
+		th.AssertNoErr(t, err)
 
 		ao, _, _, err := clouds.Parse(
 			clouds.WithCloudName("gophercloud-test"),
 		)
-		if err != nil {
-			t.Fatalf("unexpected error: %v", err)
-		}
+		th.AssertNoErr(t, err)
 
-		if got := ao.IdentityEndpoint; got != "https://example.com/gophercloud-test-12345:13000" {
-			t.Errorf("unexpected identity endpoint: %q", got)
-		}
-
-		if got := ao.Username; got != "gophercloud-test-username" {
-			t.Errorf("unexpected username: %q", got)
-		}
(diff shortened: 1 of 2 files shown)


openstack/config/clouds/clouds_test.go
```diff
@@ -371,20 +309,15 @@ func TestParse(t *testing.T) {
...
+		th.AssertEquals(t, "https://example.com/gophercloud-test-1:13000", ao.IdentityEndpoint)
+		th.AssertEquals(t, "CustomDomain", ao.DomainName)
 	})
 	t.Run("parses clouds-public.yaml locations in order", func(t *testing.T) {
 		const cloudsYAML = `clouds:
   gophercloud-test-0:
     cloud: gophercloud-test-1
     auth:
-      user_domain_name: CustomDomain`
+      domain_name: CustomDomain`
```
danchild, line 320: As a small nitpick, the other `clouds.yaml` should also have `domain_name` as to not be confusing to the reader

openstack/config/clouds/clouds_test.go
```diff
@@ -419,20 +352,15 @@ func TestParse(t *testing.T) {
...
+		th.AssertEquals(t, "https://example.com/gophercloud-test-1:13000", ao.IdentityEndpoint)
+		th.AssertEquals(t, "CustomDomain", ao.DomainName)
 	})
 	t.Run("fall back to next location if clouds-public.yaml is not found", func(t *testing.T) {
 		const cloudsYAML = `clouds:
   gophercloud-test-0:
     cloud: gophercloud-test-1
     auth:
-      user_domain_name: CustomDomain`
+      domain_name: CustomDomain`
```
danchild, line 363: same comment at line `320` above

openstack/config/clouds/clouds_test.go
```diff
@@ -823,26 +648,14 @@ func TestParse(t *testing.T) {
 			clouds.WithCloudsYAML(strings.NewReader(cloudsYAML)),
 			clouds.WithCloudName("gophercloud-test"),
 		)
-		if err != nil {
-			t.Fatalf("unexpected error: %v", err)
-		}
+		th.AssertNoErr(t, err)
```
danchild, line 651: small nitpick: `user_domain_name` is stale and no longer needed in example clouds.yaml

openstack/config/clouds/clouds_test.go
```diff
@@ -861,19 +674,11 @@ func TestParse(t *testing.T) {
...
```
danchild, line 677: for clarity, I would remove `user_domain_name` and use `Default` as the value for `domain_name`

openstack/config/clouds/clouds_test.go
```diff
@@ -861,19 +674,11 @@ func TestParse(t *testing.T) {
...
-		}
-		if got := ao.Scope.DomainName; got != "ProjectDomain" {
-			t.Errorf("expected project domain in scope, got DomainName=%q", got)
-		}
+		th.AssertTrue(t, ao.Scope != nil)
+		th.AssertEquals(t, "myproject", ao.Scope.ProjectName)
+		th.AssertEquals(t, "ProjectDomain", ao.Scope.DomainName)
 	})
 
 	t.Run("domain scoping with user_domain_id and domain_id", func(t *testing.T) {
```
danchild, line 684: I would name this test `domain scoping with domain_id` and remove `user_domain_id` from `cloud.yaml` to be more clear on the intent of the test

openstack/config/clouds/clouds_test.go
```diff
@@ -574,28 +469,16 @@ func TestParse(t *testing.T) {
...
-		}
+		th.AssertEquals(t, "myuser", ao.Username)
+		th.AssertEquals(t, "shared-domain", ao.DomainName)
+		th.AssertTrue(t, ao.Scope != nil)
+		th.AssertEquals(t, "myproject", ao.Scope.ProjectName)
+		th.AssertEquals(t, "", ao.Scope.DomainName)
 	})
 
-	t.Run("user_domain_name takes precedence over domain_name for user identity", func(t *testing.T) {
+	t.Run("domain_name sets user identity and user_domain_name is not used", func(t *testing.T) {
```
danchild, line 481: since this test doesn't concern scope, we should remove `th.AssertEquals(t, "", ao.Scope.DomainName)`, and for clarity, I would change the name of the `domain_name` to `domain` or something like that (no longer a fallback)

danchild:
@kayrus the patch looks good to me. As noted in my comments, the only things I would change concern making sure that the intent of the tests are clear for future readers. All nitpicks

mandre (changes requested):
The documentation for a few fields is now inaccurate in [types.go](https://github.com/gophercloud/gophercloud/blob/main/openstack/config/clouds/types.go). We should fix it before merging the PR.
Also, it seems like UserDomainName and UserDomainID are now completely ignored, we should drop them, or at least mark them as deprecated/unused if we want to avoid breaking backward compatibility.

PR #4008, opened 2026-09-11
Pluggable Keystone v2 and v3 authentication

danchild:
For #3440

## Summary

This is a prototype `auth` package that supports both Keystone Identity **v2** and **v3** while continuing to use Gophercloud's existing `ProviderClient` machinery.

The design follows the python `keystoneauth` model as closely as is practical in Go - in particular, the concept of **auth plugins**.

`AuthResult` becomes a way for users to store the results of a successful authentication and can be used independently of `ProviderClient`. Callers can authenticate once, transport or retain the resulting token and catalog, rescope an unscoped token later, and then create a normal Gophercloud provider client from the result.

## Changes

- Adds `v2password`, `v2token`, and `v2rescopetoken` plugins.
- Adds `v3password`, `v2token`, `v3totp`, `v3applicationcredential`, and `v3multifactor` plugins
- Adds generic password and token plugins (ultimately utilizes the plugins already mentioned)
- Properly implements project, domain, system, and trust scoping for keystone v3
- Adds normalized `AuthResult` struct to store the result of a successful athentication
- Builds `AuthOptions` that implement the `Authenticator` interface from `OS_*` environment variables or `clouds.yaml`.
- Connects `AuthResult` to `ProviderClient`
- Updates acceptance helpers and identity tests to use the new authentication flow.

### Create a provider client from the environment

```go
ctx := context.Background()

opts, err := auth.AuthOptionsFromEnv()

if err != nil {
    return err
}

provider, err := openstack.AuthenticatedClient(ctx, opts)

if err != nil {
    return err
}
```

### Create a provider client from `clouds.yaml`

```go
ctx := context.Background()

cloud, _, _, err := clouds.Parse()

if err != nil {
    return err
}

opts, err := auth.AuthOptionsFromCloud(cloud)

if err != nil {
    return err
}

provider, _ := openstack.AuthenticatedClient(ctx, opts)

```

### Keep an unscoped token and scope it later

```go
ctx := context.Background()

// assume no scope was provided in opts
opts, err := auth.AuthOptionsFromEnv()

if err != nil {
    return err
}

authResult, err := opts.Authenticate(ctx, nil)

if err != nil {
    return err
}

// Rescope token
opts = auth.AuthOptionsV3{
	AuthURL: opts.GetAuthURL(),
	Auth: auth.V3RescopeTokenOpts{
		Token: authResult.TokenID,
		Scope: &auth.Scope{ProjectName: "admin", ProjectDomainID: "default"},
	},
}

// Authenticate with new scope
authResult, err = opts.Authenticate(ctx, nil)
```

## Breaking changes

This prototype intentionally changes Gophercloud's authentication API:

- `openstack.AuthenticatedClient` and `openstack.Authenticate` accept an `auth.Authenticator` rather than `gophercloud.AuthOptions`.
- `config.NewProviderClient` accepts an `auth.Authenticator`.


auth/clouds.go
```diff
@@ -0,0 +1,423 @@
...
+			IdentityBase:     gophercloud.NormalizeURL(base),
+			IdentityEndpoint: gophercloud.NormalizeURL(authURL),
+		}
+
+		versions := []*utils.Version{
+			{ID: "v2.0", Priority: 20, Suffix: "/v2.0/"},
+			{ID: "v3", Priority: 30, Suffix: "/v3/"},
+		}
+
+		chosen, _, err := utils.ChooseVersion(context.TODO(), client, versions)
```
rm-you, line 151: Could generic password/token discovery be deferred to `Authenticate(ctx, httpClient)`? This runs before `config.NewProviderClient` receives the caller's HTTP/TLS settings, using a fresh client and `context.TODO()`. With a versionless endpoint signed by a private CA, I reproduced `auth_type: password` failing with an unknown-CA error during option construction, while explicit `v3password` succeeds with the configured client. The environment loader has the same issue.

The endpoint returned by `ChooseVersion` also needs to be retained: currently only the version is used, so an advertised endpoint on a different path is lost. I have an [isolated implementation and regression test](https://github.com/rm-you/gophercloud/commit/ba08986293586dbc1b6046a56b419a7f288a6d72) covering the configured TLS client, cancellation and a discovered `/identity/v3/` path, if useful.

auth/env.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	scope := &Scope{
+		DomainID:          os.Getenv("OS_DOMAIN_ID"),
+		DomainName:        os.Getenv("OS_DOMAIN_NAME"),
+		ProjectDomainID:   os.Getenv("OS_PROJECT_DOMAIN_ID"),
+		ProjectDomainName: os.Getenv("OS_PROJECT_DOMAIN_NAME"),
+		ProjectID:         os.Getenv("OS_PROJECT_ID"),
+		ProjectName:       os.Getenv("OS_PROJECT_NAME"),
+	}
```
rm-you, line 157: The separation between user/project domains and authorization domain scope makes sense and matches keystoneauth. Could we check consistency between the loaders? `OS_SYSTEM_SCOPE=all` and `OS_TRUST_ID` are currently ignored by environment loading, although the corresponding cloud settings work.

Conversely, a `v3password` cloud entry with valid credentials, `user_domain_id: user-domain` and `domain_id: scope-domain`, with no project, loses the requested domain scope; the equivalent environment configuration preserves it. I reproduced these differences in the generated auth bodies/scopes. Tests at that level would help keep the loaders aligned without restoring the old domain-field conflation.

auth/auth.go
```diff
@@ -0,0 +1,571 @@
...
+type Endpoint struct {
+	ID, Region, RegionID, Interface, URL string
+}
+
+func (r AuthResult) Token() string {
+	return r.TokenID
+}
+
+func (r AuthResult) AuthenticatedHeaders() map[string]string {
+	return map[string]string{"X-Auth-Token": r.TokenID}
```
rm-you, line 290: Could the plugin result express the authentication header scheme for subsequent service requests, and have `ProviderClient` consume it? Our [OAuth2 mTLS flow (#3940)](https://github.com/gophercloud/gophercloud/pull/3940) needs `Authorization: Bearer`, while this method always returns `X-Auth-Token`. `Authenticator` returns a concrete `*AuthResult`, so an external implementation cannot substitute a different result type; the provider request path also independently hardcodes `X-Auth-Token` rather than calling this method.

This isn't a bug in the password/token plugins, but it seems worth agreeing on the extension point here. Our [mTLS port](https://github.com/rm-you/gophercloud/commit/6ea02a5e15af799f5d347b4d931aff7d427ad0b6) demonstrates one option; I'm not attached to that API. Whichever approach we use should honor caller header overrides/omissions and track the actual token used for 401 reauthentication independently of its header name.

auth/auth.go
```diff
@@ -0,0 +1,571 @@
...
+	if result.Err != nil {
+		return nil, result.Err
+	}
+
+	var respBody v3TokenBody
+	if err := result.ExtractIntoStructPtr(&respBody, "token"); err != nil {
+		return nil, err
+	}
+
+	return respBody.toAuthResult(result.Header.Get("X-Subject-Token"), ao.Auth.CanReauth()), nil
```
rm-you, line 214: Optional: would you be willing to extract the v3 request execution and response normalization into reusable internal helpers here? When adapting WebSSO/OIDC/mTLS, a [small extraction](https://github.com/rm-you/gophercloud/commit/b2e38c986b957e6c29b9a80797ec897b0ad8e8a9) let the flows reuse this handling while keeping their own overall reauthentication policy. It doesn't change the public `Authenticator` interface. Including it here would let the three flow PRs branch directly from this work rather than depend on a separate preparatory refactor; equally happy to carry it as a follow-up.

rm-you (commented):
I've adapted our WebSSO, OIDC and OAuth2 mTLS work to this interface in an [integration branch](https://github.com/rm-you/gophercloud/tree/integration/pluggable-auth-4008), and the overall approach works well. That exercise exposed two configuration/discovery issues and a service-header design question worth addressing here. I've also suggested a small optional helper extraction; the flow implementations and shared token cache can remain separate follow-ups.

PR #4012, opened 2026-09-14
fix(params): allow zero values for required numeric fields

### LINKED ISSUE #3866: RegisteredLimit does not allow a `defaultLimit` of `0`
### What happened?

The [`defaultLimit`](https://github.com/gophercloud/gophercloud/blob/7544537fabb8278f7b0b41f2cab1b65ae1527355/openstack/identity/v3/registeredlimits/requests.go#L69) field for RegisteredLimits in Gophercloud is an integer with `required:true`

While OpenStack [permits a value of `0`](https://docs.openstack.org/keystone/latest/admin/unified-limits.html), Gophercloud will fail validation for the field in that case.

Gophercloud runs into this: https://github.com/gophercloud/gophercloud/blob/7544537fabb8278f7b0b41f2cab1b65ae1527355/params.go#L92-L97

Expected behavior would be that Gophercloud permits requests containing a `defaultLimit` of `0` (in fact everything from `-1` to `2147483647` should be valid)

See also https://github.com/k-orc/openstack-resource-controller/pull/782#discussion_r3614334846

### Version

v2 (current)

LucasLeao18:
Fixes #3866

code in this PR:

- https://docs.openstack.org/keystone/latest/admin/unified-limits.html (Keystone permits a `default_limit` of `0`)
- https://docs.openstack.org/api-ref/identity/v3/index.html#create-registered-limit

Problem
-------

`registeredlimits.CreateOpts.DefaultLimit` is declared as `int` with
`required:"true"`, and `BuildRequestBody` rejects the zero value of any
required field. As a result the SDK fails client-side with `Missing input
for argument [DefaultLimit]` when creating a registered limit with a
`default_limit` of `0`, even though Keystone accepts `0` as a valid limit
(valid range is `-1` to `2147483647`).

Solution
--------

`BuildRequestBody`, `BuildQueryString`, and `BuildHeaders` no longer treat a numeric zero as missing input for
fields tagged `required:"true"`. Zero is legitimate input for numeric
fields, so range validation is left to the server. Other zero values
(empty strings, nil pointers, empty structs, ...) are still rejected
client-side as before, so existing validation behavior is unchanged
everywhere else.

This implements option 1 suggested by @winiciusallan in the issue,
scoped to numeric kinds so the existing `ErrMissingInput` expectations
for string/struct/pointer fields keep working.

Validation
----------

- Added `TestBuildRequestBodyRequiredNumericZero` (`testing/params_test.go`):
  required `int` zero passes and is serialized; required `string` empty
  still returns `ErrMissingInput`.
- Extended `TestBuildQueryString` and `TestBuildHeaders` to verify required
  numeric zero values are serialized while other required zero values remain
  rejected.
- Added `TestCreateRegisteredLimitWithZeroDefaultLimit` plus fixtures
  (`openstack/identity/v3/registeredlimits/testing/`): end-to-end create
  with `DefaultLimit: 0`, asserting the request body sent on the wire.
- Verified the new test fails before the fix with exactly the reported
  error (`Missing input for argument [DefaultLimit]`) and passes after.
- Full unit suite (`go test ./...`) passes, except the pre-existing
  `TestNormalizePathURL` failure which also fails on a clean checkout on
  Windows (path/URL normalization, unrelated to this change).
- `go vet` is clean on the touched packages, including the acceptance
  test package (`-tags acceptance`).
- Extended the `TestRegisteredLimitsCRUD` acceptance test with a
  `default_limit: 0` create/assert/delete block (not run here, requires a
  live OpenStack cloud).


internal/acceptance/openstack/identity/v3/registeredlimits_test.go
```diff
@@ -101,4 +101,25 @@ func TestRegisteredLimitsCRUD(t *testing.T) {
...
+			ServiceID:    serviceID,
+			ResourceName: zeroResourceName,
+			DefaultLimit: 0,
+			RegionID:     "RegionOne",
+		},
+	}
+
+	zeroLimits, err := registeredlimits.BatchCreate(context.TODO(), client, zeroCreateOpts).Extract()
+	th.AssertNoErr(t, err)
+	th.AssertIntGreaterOrEqual(t, 1, len(zeroLimits))
```
mandre, line 120: Nit: the order is reversed. It works now because there's exactly 1 registered limit but will break if we have more.

https://github.com/gophercloud/gophercloud/blob/fab772e81d1f7bfd3863363a4b07539220e7a60d/testhelper/convenience.go#L513-L520

mandre (commented):
Nice!

One question, though. Should we add the same fix for `BuildQueryString()` and `BuildHeaders()` that also reject 0 values for numeric when the value is marked as required?

internal/acceptance/openstack/identity/v3/registeredlimits_test.go
LucasLeao18: Fixed in 9efd28573 by swapping the arguments so the returned slice length is checked as greater than or equal to 1. Thanks for catching this.

LucasLeao18:
Thanks for the follow-up. I extended the same required numeric-zero handling to BuildQueryString and BuildHeaders, with regression coverage for both, in 9efd28573.

params.go
```diff
@@ -443,7 +447,7 @@ func BuildQueryString(opts any) (*url.URL, error) {
 					}
 				} else {
 					// if the field has a 'required' tag, it can't have a zero-value
```
winiciusallan: You also need to fix this comment line to reflect the new behavior.

params.go
```diff
@@ -523,7 +531,7 @@ func BuildHeaders(opts any) (map[string]string, error) {
...
```
winiciusallan: Ditto

params.go
```diff
@@ -401,8 +401,12 @@ func BuildQueryString(opts any) (*url.URL, error) {
...
 				tags := strings.Split(qTag, ",")
 
-				// if the field is set, add it to the slice of query pieces
-				if !isZero(v) {
+				required := f.Tag.Get("required") == "true"
+
+				// if the field is set, add it to the slice of query pieces.
+				// Required numeric zero values are also set explicitly and must
+				// not be mistaken for omitted input.
+				if !isZero(v) || (required && v.Kind() == reflect.Int) {
```
winiciusallan, line 410: Isn't better to use the same `isNumeric` function here instead of checking if the field is `int`?

winiciusallan (commented):
Bom trabalho, @LucasLeao18! I left a few comments inline, please see.

params.go
LucasLeao18: Fixed in bec3ee50972a68b2560248576912da8f39188301. The comment and BuildQueryString documentation now explain that required int fields are included even when zero, while other required zero values still return an error. Thanks for catching this.

params.go
LucasLeao18: Fixed in bec3ee50972a68b2560248576912da8f39188301 as well. The comment and BuildHeaders documentation now explicitly describe the zero-value exception for required int and int64 fields.

params.go
LucasLeao18: Reusing isNumeric would be more consistent, but its accepted types are broader than the serializers currently support: BuildQueryString handles only reflect.Int for numeric scalars, and BuildHeaders handles reflect.Int and reflect.Int64. isNumeric also accepts the other signed integer widths, unsigned integers, and floats.

I checked the direct replacement: a required int64(0) query field would produce an empty query with no error, and required uint(0) or float64(0) fields would be omitted without error from both queries and headers. They pass the isNumeric check but have no matching serialization case.

For that reason, I kept the checks aligned with each serializer's supported kinds and documented that constraint in bec3ee50972a68b2560248576912da8f39188301. Using isNumeric safely would also require extending serialization support, which would broaden this fix. The targeted tests for BuildQueryString, BuildHeaders, and required numeric request-body zeros, plus go vet on the affected packages, pass.

mandre (approved):

mandre:
https://github.com/gophercloud/gophercloud/pull/4013 should fix the keystone job failure.

PR #4022, opened 2026-09-17
reservation: Add Blazar service client and CI job

mandre:
Manual backport of https://github.com/gophercloud/gophercloud/pull/3988 to v2.


winiciusallan (dismissed):

openstack/client.go
```diff
@@ -508,3 +508,15 @@ func NewWorkflowV2(client *gophercloud.ProviderClient, eo gophercloud.EndpointOp
 func NewPlacementV1(client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
 	return initClientOpts(client, eo, "placement", 1)
 }
+
+// NewReservationV1 creates a ServiceClient that may be used with the reservation package.
+func NewReservationV1(ctx context.Context, client *gophercloud.ProviderClient, eo gophercloud.EndpointOpts) (*gophercloud.ServiceClient, error) {
```
winiciusallan: We need to remove the context from this function signature. This looks like failures' root cause

mandre: Indeed, I've pushed an update.

winiciusallan (approved):

PR #4028, opened 2026-09-21
dependabot: ignore minor and major golang bumpings

winiciusallan:
The commit makes dependabot ignores major and minor bumps to the required golang version. This increases the control over the used golang version and avoid unnecessary bumpings.

Looking at dependabot documentation, it's possible to ignore considering semver levels. See https://docs.github.com/en/code-security/how-tos/secure-your-supply-chain/manage-your-dependency-security/controlling-dependencies-updated#allowing-specific-semantic-versioning-levels-for-updates


stephenfin (approved):
I think this is a good idea, but I'll leave merging to another maintainer in case others think otherwise

winiciusallan:
Dependabot continues to bump go version as a side-effect of updating some dependencies. This PR doesn't have the result that we expect.

